PR #6, opened 2016-07-20
Remove duplicate definition

feihujiang:

### DIFF
--- a/gossip/gossip_core.py
+++ b/gossip/gossip_core.py
@@ -145,7 +145,6 @@ class Gossip(object, DatagramProtocol):
 
         self.PendingAckMap = {}
         self.MessageHandledMap = {}
-        self.MessageHandlerMap = {}
 
         self.SequenceNumber = 0
         self.NextCleanup = time.time() + self.CleanupInterval


feihujiang:
@rbuysse Could you please review this PR?

gossip/gossip_core.py
```diff
@@ -145,7 +145,6 @@ def __init__(self, node, **kwargs):
 
         self.PendingAckMap = {}
         self.MessageHandledMap = {}
-        self.MessageHandlerMap = {}
```
boydjohnson: MessageHandlerMap and MessageHandledMap are two different dicts. The MessageHandlerMap is a map of message types to handler functions and the MessageHandledMap is a map of handled messages where keys are message identifiers and values are message expiration times. Checkout the code comments from lines 85-118 in gossip_core.py. Thanks for looking through the code.

boydjohnson:
@feihujiang Checkout my comment on your code. This isn't a duplicate.

feihujiang:
Yes, you are right!

PR #17, opened 2016-08-25
Permit disable client-side validation

feihujiang:

Address part of moving client-side validation to server-side.
This PR permits disable client-side validation.


feihujiang:
And print error message returned from server.

danmack:

PR #18, opened 2016-08-27
Implement server-side prevalidation session

feihujiang:

Implement server-side prevalidation session


feihujiang:
@vaporos Address part of server-side prevalidation, please review.

feihujiang:
POST /prevalidate
Do server-side prevalidation.

extensions/mktplace/mktplace/mktplace_client.py
```diff
@@ -168,6 +178,9 @@ def _sendtxn(self, update):
         self.LastTransaction = txnid
         txn.apply(self.CurrentState.State)
 
+#         if not self.enable_session:
```
vaporos: Please remove commented out code.

extensions/mktplace/mktplace/mktplace_client.py
```diff
@@ -189,6 +202,8 @@ def _register(self, update):
         # the name map
         if txnid:
             fqname = self.CurrentState.i2n(txnid)
+#             fqname = update.Name
```
vaporos: Please remove commented out code.

extensions/mktplace/mktplace/mktplace_communication.py
```diff
@@ -135,12 +137,26 @@ def postmsg(self, msgtype, info):
...
+                    logger.debug('self._cookie %s', self._cookie)
+            else:
+                response = opener.open(request, timeout=10)
 
         except urllib2.HTTPError as err:
             logger.warn('operation failed with response: %s', err.code)
-            raise MessageException(
-                'operation failed with response: {0}'.format(err.code))
+            err_content = err.read()
+            if err_content.find("InvalidTransactionError"):
```
vaporos: We should potentially check for a status code of BAD_REQUEST here.

extensions/mktplace/mktplace/transactions/market_place.py
```diff
@@ -495,6 +496,10 @@ def is_valid(self, store):
 
         return True
 
+#     def check_valid(self, store):
```
vaporos: Please remove commented out code.

validator/setup.py
```diff
@@ -114,7 +114,7 @@ def version(default):
     url='http://www.intel.com',
     packages=find_packages(),
     install_requires=['sawtooth-core', 'cbor>=0.1.23', 'colorlog',
-                      'twisted', 'PyYAML', 'psutil', 'numpy'],
```
vaporos: This was perhaps accidentally added to this PR?

vaporos:
The changes to mktclient should (at least eventually) be more extensive, removing it's local state entirely in favor of getting that state from the server.  (And, essentially, removing any direct interaction between mktclient and the mktplace transaction family.)  We don't need to do it all at once, but it is good to keep in mind the goal is to remove that dependency.

vaporos:
Should we explicitly handle sessions with create/delete?  At a minimum, we need some way to clean up sessions from the validator process.

extensions/mktplace/mktplace/mktplace_communication.py
feihujiang: Done. Others Done too.

feihujiang:
@vaporos I updated this PR, and support creating/deleting sessions, please review. And yes, the goal is removing all of the interaction between mktclient and the mktplace transaction family. But if session is disable, state from the server has no pending transactions now, but some commands need [n2i] (https://github.com/hyperledger/sawtooth-core/blob/master/extensions/mktplace/mktmain/client_cli.py#L600),  so I remain [transaction.apply()](https://github.com/feihujiang/sawtooth-core/blob/897adbdac01aabb7f9924495306ff84bca0f4d8d/extensions/mktplace/mktplace/mktplace_client.py#L185) function. I'll remove it when I have a good idea to handle this case. Or could you give me some advice?

danmack:

validator/txnserver/web_pages/prevalidation_page.py
```diff
@@ -0,0 +1,131 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+
+import copy
+import logging
+import traceback
+
+from zope.interface import Interface, Attribute, implements
```
vaporos: We don't want to add a zope dependency here.

vaporos:
+ ./bin/run_lint
************* Module txnserver.web_pages.prevalidation_page
E: 33, 0: Inheriting 'Interface', which is not a class. (inherit-non-class)

validator/txnserver/web_pages/prevalidation_page.py
feihujiang: Why we don't want to add a zope dependency?  And what you suggest?

feihujiang:
@vaporos Could you give some advice if we don't want to add a zope dependency?

PR #26, opened 2016-09-08
Fix too many replicated warnings are throwed

feihujiang:

Everytime we invoke [client_cli.main()](https://github.com/hyperledger/sawtooth-core/blob/master/extensions/mktplace/tests/integration/test_all_transactions.py#L65), Logger().addHandler(clog) [here](https://github.com/hyperledger/sawtooth-core/blob/master/extensions/mktplace/mktmain/client_cli.py#L1532), too many handlers are added. So test_transactions_reg() throws two replicated warnings,  test_transactions_unr() throws three replicated warnings etc. This PR fixes the bug.

### DIFF
--- a/extensions/mktplace/mktmain/client_cli.py
+++ b/extensions/mktplace/mktmain/client_cli.py
@@ -1496,6 +1496,8 @@ def get_configuration(args, os_name=os.name, config_files_required=True):
 
     return cfg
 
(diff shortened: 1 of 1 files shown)


danmack:

feihujiang:
@boydjohnson Please review this PR, thanks.

PR #29, opened 2016-10-09
Fix sawtooth transaction list command bug

feihujiang:
Before this PR, "sawtooth_xo", or "mktplace.transactions.market_place" etc TransactionFamilies loaded, when using "sawtooth transaction list" command, it would crash. Throws "update = trans_info["Update"] KeyError: 'Update'" error. This PR fixes the bug.


feihujiang:
@boydjohnson I updated this PR, please review.

PR #30, opened 2016-11-13
Use the validator registry with a whitelist

feihujiang:
The administrator provide a validator public keys, then send a ValidatorRegistryTransaction to register the validator.

### DIFF
--- a/core/sawtooth/cli/admin.py
+++ b/core/sawtooth/cli/admin.py
@@ -21,6 +21,10 @@ from sawtooth.cli.admin_sub.poet0_genesis import add_poet0_genesis_parser
 from sawtooth.cli.admin_sub.poet0_genesis import do_poet0_genesis
 from sawtooth.cli.admin_sub.poet1_genesis import add_poet1_genesis_parser
 from sawtooth.cli.admin_sub.poet1_genesis import do_poet1_genesis
+from sawtooth.cli.admin_sub.permissioned_validator_registry\
+    import add_validator_parser
+from sawtooth.cli.admin_sub.permissioned_validator_registry\
+    import do_validator_registry
 from sawtooth.cli.admin_sub.clean import add_clean_parser
 from sawtooth.cli.admin_sub.clean import do_clean
 
@@ -34,6 +38,8 @@ def do_admin(args):
         do_poet0_genesis(args)
     elif args.admin_cmd == 'poet1-genesis':
         do_poet1_genesis(args)
+    elif args.admin_cmd == 'validator-registry':
+        do_validator_registry(args)
     elif args.admin_cmd == 'clean':
         do_clean(args)
     else:
@@ -46,4 +52,5 @@ def add_admin_parser(subparsers, parent_parser):
     add_dev_mode_genesis_parser(admin_sub, parser)
     add_poet0_genesis_parser(admin_sub, parser)
     add_poet1_genesis_parser(admin_sub, parser)
+    add_validator_parser(admin_sub, parser)
     add_clean_parser(admin_sub, parser)
--- a/core/sawtooth/cli/admin_sub/permissioned_validator_registry.py
+++ b/core/sawtooth/cli/admin_sub/permissioned_validator_registry.py
@@ -1,0 +1,145 @@
+# Copyright 2016 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+
+import json
+import logging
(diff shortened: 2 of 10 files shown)


dcmiddle:
This is a great start without having been given very prescriptive requirements.
We can catch up on slack for the right policy to start with.

validator/sawtooth_validator/consensus/poet1/validator_registry.py
```diff
@@ -293,7 +411,11 @@ def __init__(self, minfo=None):
         self.update = None
 
         if 'Update' in minfo:
-            self.update = Update(minfo=minfo['Update'])
+            if 'signup_info' in minfo['Update']:
```
dcmiddle: We need to think about whether its really ok to just switch on the message contents or whether its better to have an exclusive-or on these policies. I think a ledger that allows a mix of white-listed nodes and full poet nodes to intermix might lead to difficulties.

core/sawtooth/cli/transaction.py
```diff
@@ -155,26 +155,12 @@ def print_trans_info(args, web_client, tsctids):
         "Identifier": "TRANS",
         "InBlock": "BLOCK",
         "Status": "STATUS",
-        "TransactionType": "TXNTYPE",
-        "Update": {
```
dcmiddle: do these changes relate to the whitelist PR?

feihujiang: It's just my idea that an administrator has the role to register the permissioned validator. If we don't want to introduce an powerful node like an administrator. I'll change this file, and just get validators information like [an endpoint client](https://github.com/hyperledger/sawtooth-core/blob/master/core/sawtooth/endpoint_client.py).

docs/source/tutorial/genesis.sh
```diff
@@ -5,5 +5,5 @@ mkdir /home/vagrant/sawtooth/keys
 mkdir /home/vagrant/sawtooth/data
 mkdir /home/vagrant/sawtooth/logs
 ./bin/sawtooth keygen base000 --key-dir /home/vagrant/sawtooth/keys
-./bin/sawtooth admin poet0-genesis --keyfile /home/vagrant/sawtooth/keys/base000.wif -F mktplace.transactions.market_place -F ledger.transaction.integer_key
+./bin/sawtooth admin poet1-genesis --keyfile /home/vagrant/sawtooth/keys/base000.wif -F mktplace.transactions.market_place -F sawtooth_validator.consensus.poet1.validator_registry
```
feihujiang: OK. I changed it here is just for somebody want to test this PR.

core/sawtooth/cli/transaction.py
feihujiang: Sorry, the above comment doesn't answer your question. Not related, but we need to print different transaction information due to different transaction type, for example,  the update of ValidatorTransaction is different from the update of the other transaction type.

validator/sawtooth_validator/consensus/poet1/validator_registry.py
feihujiang: @dcmiddle So, I'll implement the ValidatorRegistryTransaction as the subclass of the [UpdateTransaction](https://github.com/hyperledger/sawtooth-core/blob/master/validator/journal/transaction.py#L228). What's your point?

feihujiang: Done. Implement PermissionedValidatorRegistryTransaction independently.

validator/sawtooth_validator/consensus/poet1/validator_registry.py
```diff
@@ -241,6 +241,124 @@ def dump(self):
         return result
 
 
+class PermissionedValidtaorUpdate(object):
```
feihujiang: Done.

docs/source/tutorial/genesis.sh
feihujiang: Done.

feihujiang:
@dcmiddle I updated this PR, please review. Thanks!

feihujiang:
Start a validator using the following command:
```./bin/txnvalidator -vvv --config ./validator/etc/permissioned-validators.json --keyfile /home/vagrant/sawtooth/keys/base000.wif```

Send a PermissionedValidatorRegistryTransaction using the following command:
```./bin/sawtooth admin validator-registry --keyfile /home/vagrant/sawtooth/keys/base000.wif --whitelist ./validator/etc/permissioned-validators.json```

core/sawtooth/cli/transaction.py
dcmiddle: The way I understand this module to work is it will print transaction info for a client. Since transaction families are diverse there are only a few general fields that make sense. For now, we expect that the validator will always install the endpoint registry. Hence, these update fields are supported in this CLI helper. 
We shouldn't cut them out though when adding a new family. If you see some error condition we should figure out how to resolve that, otherwise we should drop these changes from the commit.

feihujiang:
@dcmiddle Added a unit test, please take a look.

dcmiddle:
Looks good @feihujiang 
Just need you to add the signed-off message for licensing.
(Bonus points if you rebase when you amend the commit message).

feihujiang:
@dcmiddle Added the signed-off message for licensing using '''git commit -s''', please have a look.

feihujiang:
@dcmiddle I updated the PR, and ran ./bin/run_tests locally, it finished successfully. Please take a look.

PR #34, opened 2016-12-01
Consolidate tutorial setup, update genesis script, update tutorial

nickdrozd:
Signed-off-by: nickdrozd <drozd@bitwise.io>

### DIFF
--- a/docs/source/tutorial.rst
+++ b/docs/source/tutorial.rst
@@ -21,7 +21,7 @@ Prerequisites
 
 The following tools are required:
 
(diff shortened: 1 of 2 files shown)


setup.sh
```diff
@@ -0,0 +1,11 @@
+build_in() {
```
vaporos: Scripts go in the bin/ directory and don't contain the ".sh" extension as the script/language is really an implementation detail.  I have a different version of this script called "build_all" which I'll create a PR for - we should use that instead.

Style note - we don't use tabs in source code (bash included).

PR #36, opened 2016-12-06
Rename signingkey and SignatureKey

feihujiang:

I think [signingkey](https://github.com/hyperledger/sawtooth-core/blob/master/validator/gossip/signed_object.py#L117) and [SignatureKey](https://github.com/hyperledger/sawtooth-core/blob/master/validator/gossip/signed_object.py#L128) here isn't really a signing key, it is just the name of the key related to the signature. I renamed them to avoid confusing and fixed the wrong case in test_signed_object.

### DIFF
--- a/validator/gossip/signed_object.py
+++ b/validator/gossip/signed_object.py
@@ -108,24 +108,25 @@ class SignedObject(object):
 
     Attributes:
         Signature (str): The signature used to sign the object.
-        SignatureKey (str): The name of the key related to the signature.
-            Used to build dict return types.
+        SignatureDictKey (str): The key in the serialized dict containing the
+            signature. Used to build dict return types.
 
     """
     signature_cache = LruCache()
 
-    def __init__(self, minfo=None, signkey='Signature'):
+    def __init__(self, minfo=None, sig_dict_key='Signature'):
         """Constructor for the SignedObject class.
 
         Args:
             minfo (dict): object data
-            signkey (str): the field name for the signature within the
+            sig_dict_key (str): the field name for the signature within the
                 object data
         """
         if minfo is None:
             minfo = {}
-        self.Signature = minfo.get(signkey)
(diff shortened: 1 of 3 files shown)


validator/gossip/signed_object.py
```diff
@@ -260,30 +260,30 @@ def sign_object(self, signingkey):
         self._identifier = hashlib.sha256(self.Signature).hexdigest()
 
     def serialize(self, signable=False):
-        """Generates a CBOR serialized dict containing the a SignatureKey
+        """Generates a CBOR serialized dict containing the a SignatureName
```
ojalatodd: Delete extra article "a" in this comment: Generates a CBOR serialized dict containing the a SignatureName

validator/tests/unit/test_signed_object.py
```diff
@@ -78,8 +76,7 @@ def test_signed_node(self):
         # Verify that signed_node and sign_object does not invalidate the
         # signed object and can be returned to original
         # create initial signed object
-        signkey = SigObj.generate_signing_key()
-        temp = SignedObject({signkey: "test"}, signkey)
+        temp = SignedObject({"TestSignatureName": "test"}, "TestSignatureName")
```
ojalatodd: You simplified some of these tests. Looks good to me, but are you sure you didn't break anything by simplifying this tests and the ones below? If you are confident, I'll merge the PR after the minor change above.

feihujiang: I didn't simplify any of these tests, I just fixed the misunderstanding use of the ```signkey``` replaced by ```SignatureName``` attribute. So yes and I checked, I didn't break anything.

feihujiang:
@ojalatodd I updated this PR, please have a look.

feihujiang:
@vaporos I updated this PR according to your advice, please take a look.

PR #37, opened 2016-12-07
Print more linted files when using 'run_lint -n' command

feihujiang:

Print linted files from missed module  when using 'run_lint -n' command


feihujiang:
@vaporos Could you please have a look?

PR #40, opened 2016-12-09
Add module option in run_tests

feihujiang:

Usage:
  ./bin/run_tests [-m Module]
Options:
  -h          print usage and exit
  -m Module   only run tests of module, <core, validator, mktplace, bond, arcade>

### DIFF
--- a/bin/run_tests
+++ b/bin/run_tests
@@ -1,4 +1,28 @@
 #!/bin/bash
+#
+# Copyright 2016 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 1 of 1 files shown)


feihujiang:
@vaporos Could you please review this PR?  Running ./bin/run_tests totally costs too much time, this PR could run tests of specified module.

bin/run_tests
```diff
@@ -7,56 +31,122 @@ top_dir=$(cd $(dirname $(dirname $0)) && pwd)
...
+        MODULE=$OPTARG
+        ;;
+      \?)
+        echo "Invalid option: -$OPTARG" >&2
+        exit 2
+        ;;
+    esac
+done
+
+case $MODULE in
```
vaporos: Let's use the same indent here as above (4 spaces instead of 2).

bin/run_tests
```diff
@@ -7,56 +31,122 @@ top_dir=$(cd $(dirname $(dirname $0)) && pwd)
...
+  all)
+    All
```
vaporos: Let's rename Core, Validator, etc. functions to 'lint_core', 'lint_validator', etc.  Let's also remove 'All' and just call each of these "lint_" functions directly from the case statement.

bin/run_tests
feihujiang: Done.

bin/run_tests
feihujiang: I renamed  Core, Validator, etc. functions to ```test_core```, ```test_validator```, and removed ```All```. Done.

feihujiang:
@vaporos I udpated this PR, please take a look.

PR #41, opened 2016-12-13
Load libor publisher addr from a config file

feihujiang:

@jjason Could you please have a look?

### DIFF
--- a/extensions/bond/sawtooth_bond/txn_family.py
+++ b/extensions/bond/sawtooth_bond/txn_family.py
@@ -14,6 +14,7 @@
 # ------------------------------------------------------------------------------
 
 import logging
+import json
 import time
 import copy
 from datetime import datetime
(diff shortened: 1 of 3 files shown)


extensions/bond/sawtooth_bond/txn_family.py
```diff
@@ -49,6 +50,19 @@
...
+    try:
+        with open('./etc/sawtooth-bond.json') \
+                as bond_fd:
+            bond_config = json.load(bond_fd)
+    except IOError, ex:
+        LOGGER.exception('IOError: %s', str(ex))
+        return
+
+    libor_publisher_addr = bond_config["LiborPublisherAddr"]
+    CreateLIBORUpdate.LIBOR_PUBLISHER_ADDR(libor_publisher_addr)
```
jjason: Am I missing something?  LIBOR_PUBLISHER_ADDR has been been made a global variable in sawtooth_bond.updates.libor and here you are accessing it like a function on the CreateLIBORUpdate class.

extensions/bond/sawtooth_bond/txn_family.py
```diff
@@ -49,6 +50,19 @@
...
+def _load_config(journal):
+    try:
+        with open('./etc/sawtooth-bond.json') \
+                as bond_fd:
+            bond_config = json.load(bond_fd)
+    except IOError, ex:
+        LOGGER.exception('IOError: %s', str(ex))
+        return
+
+    libor_publisher_addr = bond_config["LiborPublisherAddr"]
```
jjason: I think you need to guard against the "LiborPublisherAddr" attribute not existing in the configuration file:
libor_publisher_addr = bond_config.get('LiborPublisherAddr')
if libor_publisher_addr is not None:
  ....

This way the code works with the default LIBOR address if one is not provided.  Either that, or flag an error if it doesn't exist.

feihujiang:
@jjason Sorry, I forget to say this PR is just an idea. The idea in this PR is accepted, isn't it? If it is, I'll update this PR tomorrow.

jjason:
@vaporos - I tend to agree.

PR #47, opened 2016-12-16
Add logger warning when signed object has no pub key and addr

feihujiang:

If signed object has no pub key and addr, verifying its signature always return true. We should add logger warning when such case occurs.

### DIFF
--- a/validator/gossip/signed_object.py
+++ b/validator/gossip/signed_object.py
@@ -229,6 +229,9 @@ class SignedObject(object):
         try:
             # force validation of the signature
             recovered_id = self.OriginatorID
+            if self.public_key is None and originatorid is None:
+                logger.warning("Identifier %s of SignedObject has no public "
+                    "key and addr used to verify signature", self.Identifier)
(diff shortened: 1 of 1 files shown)


vaporos:
If this is a case that deserves a warning, should is_valid() also return False in that case as well?

feihujiang:
Yes, as the function of SignedObject, is_valid() should return False in this case. But I found at least EndpointRegistryTransaction and Block have no pub key and addr. We need to found is there any other signedobjects that have no pub key and addr,  fix all of them firstly, then change  is_valid() to return False. So I added a warning here.

feihujiang:
@vaporos I'll open a new PR to fix the bug that EndpointRegistryTransaction has no pub key and addr.

feihujiang:
@vaporos what's your point about this PR?

vaporos:
This change has been accepted and cherry-picked to the 0-7 branch.  Changes to "legacy code" that is not part of the 0.8 validator are not being merged into master, so closing this PR against master.

PR #50, opened 2016-12-22
Fix args typo

feihujiang:


feihujiang:
poet0 was deleted, so this PR is meaningless, closed this PR.

PR #51, opened 2016-12-23
Add public key for blocks

feihujiang:

Blocks has no public key and addr, [is_valid()](https://github.com/hyperledger/sawtooth-core/blob/master/validator/gossip/signed_object.py#L199) always return true, [verify_signature()](https://github.com/hyperledger/sawtooth-core/blob/master/validator/gossip/signed_object.py#L212) is meaningless, we need to fix this.

### DIFF
--- a/validator/journal/journal_core.py
+++ b/validator/journal/journal_core.py
@@ -34,6 +34,7 @@ from journal.messages import transaction_block_message
 from journal.messages import transaction_message
 
 from sawtooth.exceptions import NotAvailableException
+from sawtooth_signing import pbct_nativerecover as signing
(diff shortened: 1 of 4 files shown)


feihujiang:
I rebased this PR.

vaporos:
Not sure if #59 is a more general solution or if they are both needed.  In any case, this PR could not be cherry-picked to 0-7 without modifications.  If the change is still relevant to 0-7 (not solved by #59), please resubmit against that branch.

Since changes to "legacy code" that is not part of the 0.8 validator are not being merged into master, I'm closing this particular PR.

PR #54, opened 2016-12-27
Fix invoking non-public member

feihujiang:

### DIFF
--- a/validator/journal/journal_core.py
+++ b/validator/journal/journal_core.py
@@ -520,6 +520,14 @@ class Journal(object):
 
         self.on_initialization_complete.fire(self)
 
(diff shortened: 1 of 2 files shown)


feihujiang:
A name prefixed with an underscore (e.g. journal._txn_lock) should be treated as a non-public part of the API, should not be invoked from outside of the object.

feihujiang:
I rebased this PR.

vaporos:
This PR could not be cherry-picked to 0-7 without modifications.  If the change is still relevant to 0-7, please resubmit against that branch.

Since changes to "legacy code" that is not part of the 0.8 validator are not being merged into master, I'm closing this particular PR.

PR #55, opened 2016-12-28
Remove duplicate functions

feihujiang:


vaporos:
The duplication is intentional, as we are attempting to separate core and validator code.  In particular, core should not rely on anything in the validator package, but the reverse is also true.  I think this is mostly the case currently with the exception of the CLI utility for creating the genesis block.

feihujiang:
@vaporos You mean  the validator package should not rely on anything in the core package either? 
I don't think the duplication is a good idea. Maybe we need to put them into an separate package. then the validator and core packages rely on the separate package.

PR #56, opened 2016-12-29
Maintain consistent for certificates of poet0 and poet1

feihujiang:

Maintain consistent for fields of certificates of poet0 and poet1. So whenever we using poet0 or poet1, ./bin/sawtooth block list has no bug.


feihujiang:
I rebased this PR.

vaporos:
We do not plan compatibility between poet0 and poet1 at this level.  For master (0.8), we don't have poet0.  For the 0-7 branch, we should fix this within the CLI code.  This will keep the 0-7 poet1 closer to the poet1 in master.

PR #58, opened 2017-01-03
set keyfile config for validator tutorial

feihujiang:

When we use poet1, if we don't use a registered validator, we get an error ```Cannot validate wait certificate because cannot retrieve PoET public key for validator with ID=1MJYJNLiN9m2zbPkK21BW4nMn5RfsDQKzs``` and ```block test failed```. We need to use the registered validator.

### DIFF
--- a/docs/source/tutorial.rst
+++ b/docs/source/tutorial.rst
@@ -115,7 +115,7 @@ and run the following commands:
 
    $ /project/sawtooth-core/docs/source/tutorial/genesis.sh
    $ cd /project/sawtooth-core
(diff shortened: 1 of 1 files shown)


ojalatodd:
@feihujiang Sorry for delay on looking at this. 
Is your change here still relevant in light of recent changes to 0-7?

I think you might want to rebase a repush this change. 
If you do that, and you think it still is necessary, I will test it myself then merge it!
thanks,
Todd

ojalatodd:
Reviewing to see if still valid.

ojalatodd:
Superseded by current tutorial, and keyfile no longer needs to be specified.

ojalatodd:
Not relevant with current tutorials and codebase.

PR #69, opened 2017-02-12
Add head parameter, update /state endpoint

delventhalz:
* Add `head` parameter to all endpoints
  - Specificies a block id, which will be treated as the chain head
  - Update response envelope to include `head` and `link`
* Update state routes to bring in line with spec
  - No more fetching or specifying merkle root
  - `head` can be used to specify a particular tree instead
  - Without `head`, routes default to most recent tree
* Update Rest Api error handling as needed to implement changes
* Refactor ClientHandlers to handle genesis changes, and make
    future refactors easier
* Update xo and inkey smoke tests to use new `/state`

### DIFF
--- a/cli/sawtooth_cli/block.py
+++ b/cli/sawtooth_cli/block.py
@@ -74,12 +74,6 @@ def add_block_parser(subparsers, parent_parser):
 
 
 def do_block(args):
-    subcommands = ['list', 'show']
-    if args.subcommand not in subcommands:
-        print('Unknown sub-command, expecting one of {0}'.format(
-            subcommands))
-        return
-
     rest_client = RestClient(args.url)
 
     def print_json(data):
@@ -114,7 +108,6 @@ def do_block(args):
 
         if args.format == 'default':
             print('{:<3}  {:88.88}  {:<7}  {:<4}  {:20.20}'.format(*headers))
-
             for block in blocks:
                 print('{:<3}  {:88.88}  {:<7}  {:<4}  {:17.17}...'.format(
                     *get_block_data(block)))
@@ -142,7 +135,7 @@ def do_block(args):
         else:
             raise CliException('unknown format: {}'.format(args.format))
 
-    elif args.subcommand == 'show':
+    if args.subcommand == 'show':
         block = rest_client.get_block(args.block_id)
 
         if args.key:
--- a/cli/sawtooth_cli/main.py
+++ b/cli/sawtooth_cli/main.py
@@ -33,6 +33,8 @@ from sawtooth_cli.config import add_config_parser
 from sawtooth_cli.config import do_config
 from sawtooth_cli.block import add_block_parser
 from sawtooth_cli.block import do_block
+from sawtooth_cli.state import add_state_parser
+from sawtooth_cli.state import do_state
 
 
 def create_console_handler(verbose_level):
@@ -91,6 +93,7 @@ def create_parser(prog_name):
     add_admin_parser(subparsers, parent_parser)
     add_config_parser(subparsers, parent_parser)
     add_block_parser(subparsers, parent_parser)
+    add_state_parser(subparsers, parent_parser)
 
     return parser
 
@@ -115,6 +118,8 @@ def main(prog_name=os.path.basename(sys.argv[0]), args=sys.argv[1:],
         do_config(args)
     elif args.command == 'block':
         do_block(args)
+    elif args.command == 'state':
+        do_state(args)
     else:
         raise AssertionError("invalid command: {}".format(args.command))
 
--- a/cli/sawtooth_cli/state.py
+++ b/cli/sawtooth_cli/state.py
@@ -1,0 +1,151 @@
+# Copyright 2016 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the 'License');
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+
+import sys
+from base64 import b64decode
+
+import csv
+import json
+import yaml
+
+from sawtooth_cli.rest_client import RestClient
+from sawtooth_cli.exceptions import CliException
+
+
+def add_state_parser(subparsers, parent_parser):
+    parser = subparsers.add_parser('state')
+
+    grand_parsers = parser.add_subparsers(title='grandchildcommands',
+                                          dest='subcommand')
+
+    epilog = '''
+    details:
+        Lists leaves on the merkle tree, storing state. List can be
+        narrowed using the address of a subtree.
+    '''
+
+    list_parser = grand_parsers.add_parser('list', epilog=epilog)
+    list_parser.add_argument(
+        'subtree',
+        type=str,
+        nargs='?',
+        default=None,
+        help='the address of a subtree to filter list by')
+    list_parser.add_argument(
+        '--url',
+        type=str,
+        help="the URL of the validator's REST API")
+    list_parser.add_argument(
+        '--head',
+        action='store',
+        default=None,
+        help='the id of the block to set as the chain head')
+    list_parser.add_argument(
+        '--format',
+        action='store',
+        default='default',
+        help='the format of the output, options: csv, json or yaml')
+
+    epilog = '''
+    details:
+        Shows the data for a single leaf on the merkle tree.
+    '''
+    show_parser = grand_parsers.add_parser('show', epilog=epilog)
+    show_parser.add_argument(
+        'address',
+        type=str,
+        help='the address of the leaf')
+    show_parser.add_argument(
+        '--url',
+        type=str,
+        help="the URL of the validator's REST API")
+    show_parser.add_argument(
+        '--head',
+        action='store',
+        default=None,
+        help='the id of the block to set as the chain head')
+
+
+def do_state(args):
+    rest_client = RestClient(args.url)
+
+    def print_json(data):
+        print(json.dumps(
+            data,
+            indent=2,
+            separators=(',', ': '),
+            sort_keys=True))
+
+    def print_yaml(data):
+        print(yaml.dump(data, default_flow_style=False)[0:-1])
+
+    if args.subcommand == 'list':
+        response = rest_client.list_state(args.subtree, args.head)
+        leaves = response['data']
+        head = response['head']
+
+        keys = ('address', 'size', 'data')
+        headers = (k.upper() for k in keys)
+        max_data_width = 15
+        format_string = '{{:{addr}.{addr}}} {{:<4}} {{!s:.{data}}}'.format(
+            addr=len(leaves[0]['address']) if len(leaves) > 0 else 30,
+            data=max_data_width)
+
+        def get_leaf_data(leaf, use_decoded=True):
+            data = leaf['data']
+            decoded_data = b64decode(data)
+            return (
+                leaf['address'],
+                len(decoded_data),
+                decoded_data if use_decoded else data)
+
+        if args.format == 'default':
+            print(format_string.format(*headers))
+            for leaf in leaves:
+                print_string = format_string.format(*get_leaf_data(leaf))
+                if len(str(leaf['data'])) > max_data_width:
+                    print_string += '...'
+                print(print_string)
+            print('HEAD BLOCK: "{}"'.format(head))
+
+        elif args.format == 'csv':
+            try:
+                writer = csv.writer(sys.stdout)
+                writer.writerow(headers)
+                for leaf in leaves:
+                    writer.writerow(get_leaf_data(leaf))
+            except csv.Error:
+                raise CliException('Error writing CSV.')
+            print('(data for head block: "{}")'.format(head))
+
+        elif args.format == 'json' or args.format == 'yaml':
+            state_data = {
+                'head': head,
+                'data': list(map(
+                    lambda b: dict(zip(keys, get_leaf_data(b, False))),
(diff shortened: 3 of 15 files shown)


vaporos:
Breaking this up into multiple commits would make it easier to review.  For example, breaking out the error handling changes and genesis changes.  The head parameter and /state could be in different commits.

delventhalz:
Added CLI command to PR.

Re: breaking up commits
It wasn't immediately obvious as I was working on it how to break things up, since everything was fairly interdependent. In retrospect, those are good suggestions, and I will try to keep my commits smaller in the future.

cli/sawtooth_cli/main.py
```diff
@@ -25,14 +25,11 @@
...
-from sawtooth_cli.keygen import do_keygen
-from sawtooth_cli.config import add_config_parser
-from sawtooth_cli.config import do_config
-from sawtooth_cli.block import add_block_parser
-from sawtooth_cli.block import do_block
+from sawtooth_cli.admin import add_admin_parser, do_admin
+from sawtooth_cli.keygen import add_keygen_parser, do_keygen
+from sawtooth_cli.config import add_config_parser, do_config
+from sawtooth_cli.block import add_block_parser, do_block
+from sawtooth_cli.state import add_state_parser, do_state
```
vaporos: Prefer the original format (one import per line) as it is more diff-friendly and thus easier to resolve merge conflicts.

cli/sawtooth_cli/state.py
```diff
@@ -0,0 +1,147 @@
...
+        help='the address of a subtree to filter list by')
+    list_parser.add_argument(
+        '--url',
+        type=str,
+        help="the URL of the validator's REST API")
+    list_parser.add_argument(
+        '--head',
+        action='store',
+        default=None,
+        help='the format of the output, options: csv, json or yaml')
```
boydjohnson: This is the wrong help text.

cli/sawtooth_cli/state.py
```diff
@@ -0,0 +1,147 @@
...
+        help='the address of the leaf')
+    show_parser.add_argument(
+        '--url',
+        type=str,
+        help="the URL of the validator's REST API")
+    show_parser.add_argument(
+        '--head',
+        action='store',
+        default=None,
+        help='the format of the output, options: csv, json or yaml')
```
boydjohnson: Wrong help text.

protos/client.proto
```diff
@@ -56,16 +56,17 @@ message ClientStateCurrentResponse {
 message ClientStateListRequest {
     string merkle_root = 1;
     string subtree = 2;
+    string head = 3;
```
vaporos: Is 'head' always a block id?  If so, should this be called 'block_id'?

These should potentially be different messages - ClientStateListByAddressRequest, ClientStateListByBlockIdRequest.

rest_api/sawtooth_rest_api/exceptions.py
```diff
@@ -0,0 +1,85 @@
...
+        super().__init__(reason=message)
+
+
+class ValidatorUnavailable(web.HTTPServiceUnavailable):
+    def __init__(self):
+        message = 'Could not reach validator, validator timed out'
+        super().__init__(reason=message)
+
+
+class _ErrorTrap(object):
```
vaporos: I don't understand the exception strategy; could probably use more documentation here describing the overall approach.

rest_api/sawtooth_rest_api/exceptions.py
```diff
@@ -0,0 +1,85 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+
+from aiohttp import web
+from sawtooth_rest_api.protobuf import client_pb2 as client
+
+
+class WrongBodyType(web.HTTPBadRequest):
```
vaporos: I assume these are derived from aiohttp.web as aiohttp cares about them when they are returned?  Is the reason for the exceptions here to hard-code the message string used, or do we also except these types aiohttp exceptions?

protos/client.proto
```diff
@@ -50,72 +50,80 @@ message ClientStateCurrentResponse {
     }
     Status status = 1;
     string merkle_root = 2;
-
 }
 
-// A request from a client for a particular address's data, for a given merkle root
-message ClientStateGetRequest {
+// A request to list every entry under a given merkle root or head block.
```
delventhalz: You can specify either a head block, or a merkle root. Or neither, and the most current will be fetched. Could be clarified.

protos/client.proto
```diff
@@ -50,72 +50,80 @@ message ClientStateCurrentResponse {
...
+message ClientStateListResponse {
     enum Status {
         OK = 0;
-        NORESOURCE = 1;
-        NONLEAF = 2;
-        ERROR = 3;
+        ERROR = 1;
+        NOROOT = 3;
+        NORESOURCE = 2;
+        NOGENESIS = 4;
```
peterschwarz: The error here is the state where the validator in question is not ready - e.g it's a peer that hasn't yet received the genesis block from the network.  Perhaps another name for this, though...

delventhalz: Yes. The client reports back that the validator has no genesis block yet. Given that it is possible to distinguish between this and a general error, I erred on the side of being more informative.

rest_api/sawtooth_rest_api/routes.py
```diff
@@ -178,56 +139,60 @@ def _try_validator_request(self, message_type, content):
...
         return MessageToDict(parsed, preserving_proto_field_name=True)
 
     @staticmethod
-    def _wrap_response(data=None, head=None, link=None):
+    def _safe_get(obj, key, default=''):
+        """
+        aiohttp very helpfully parses param strings to replace '+' with ' '
+        This is very bad when your block ids contain meaningful +'s
+        """
+        return obj.get(key, default).replace(' ', '+')
```
peterschwarz: my suggestion here would be to base64 decode it and do a `urlsafe_b64encode` on the result

cli/sawtooth_cli/block.py
```diff
@@ -160,3 +153,6 @@ def get_block_data(block):
                 print_json(block)
             else:
                 raise CliException('unknown format: {}'.format(args.format))
+
+    else:
+        raise CliException('Unknown subcommand: {}'.format(args.subcommand))
```
peterschwarz: if these are specified via parsers, this should be an AssertionError, as argparse should catch incorrect commands

cli/sawtooth_cli/state.py
```diff
@@ -0,0 +1,155 @@
...
+        else:
+            raise CliException('Unknown format: {}'.format(args.format))
+
+    elif args.subcommand == 'show':
+        leaf = rest_client.get_leaf(args.address, args.head)
+        print('DATA: "{}"'.format(b64decode(leaf['data'])))
+        print('HEAD: "{}"'.format(leaf['head']))
+
+    else:
+        raise CliException('Unknown subcommand: {}'.format(args.subcommand))
```
peterschwarz: This should be an `AssertionError` as `argparse` will catch unknown subcommands

protos/client.proto
```diff
@@ -50,53 +50,66 @@ message ClientStateCurrentResponse {
...
-// merkle trie implementation requires addresses and prefixes to be length % 2 == 0)
-// any length % 2 == 0 prefix namespace.
-message ClientStateListRequest {
+// A request from a client for a particular address's data, or for a given
+// merkle root or head block
+message ClientStateGetRequest {
     string merkle_root = 1;
-    string prefix = 2;
+    string address = 2;
+    string head = 3;
```
peterschwarz: Is `head` and `merkle_root` and either/or situtation?  If so, would it make sense to use protobuf's `oneof` here?

rest_api/sawtooth_rest_api/exceptions.py
delventhalz: Yes, aiohttp will take them and turn them into proper HTTP errors. They are grouped here to provide a central location for the hard-coded messages. I'm also planning ahead to when simple error messages are replaced with actual error bodies. Keeping the errors in one place will simplify the refactor.

rest_api/sawtooth_rest_api/routes.py
delventhalz: Well, with the query strings, I can grab them raw and parse them myself, which wouldn't be too bad. For the path parameters, I don't think there is anything to do other than perhaps write a custom middleware to parse them. Neither seems like a fantastic use of time though. Perhaps easiest of all would be switching to url-safe block ids.

delventhalz: The problem is, by the time the string gets to the route handlers, it has had `'+'` swapped out for `' '`. You can't b64decode it by then, because the +'s are already gone. Currently, there are two properties that this is happening for, and they each have their own more correct solution as I see it.

`request.url.query`:
This covers query string parameters. It is possible to get the raw query string (i.e. `'head=n40+2/3h...'`), which should solve the issue, _but_ that does mean parsing the query string manually.

`request.match_info`:
This is where parameters in the path come from, and don't really have a raw equivalent. You can get the whole url string, but unlike with query parameters, there is not an immediately obvious programatic way to no which URL segment is the parameter you are looking for. Probably what we'd have to do is write a custom middleware to build a custom `match_info` equivalent.

I think the _best_ and _easiest_ solution is just to have block ids that are url-safe. We could maybe use `urlsafe_b64encode_ somewhere in the validator to do that, but I understand from Dan, that the ids are changing soon anyway. Might be best to punt on this for now.

rest_api/sawtooth_rest_api/exceptions.py
delventhalz: Documentation expanded.

delventhalz:
I've pushed some updates. All comments not under discussion should have been addressed.

validator/sawtooth_validator/state/client_handlers.py
```diff
@@ -32,182 +32,222 @@
...
+    Args:
+        request_proto - protobuf class for the request
+        response_proto - protobuf class for the response
+        result_type - the Message enum for the response
+    """
+    def __init__(self, request_proto, response_proto, result_type):
+        self._request = request_proto()
+        self._response_proto = response_proto
+        self._result_type = result_type
+        self._result = None
```
delventhalz: A new `HandlerResult` is generated every time `handle` is called, but as far as configuration goes (the protos and result type), they are stateful now, yes.

peterschwarz: Yes, but are there situations where this can cause issues with multiple threads?  There is only one instance of any of the subclasses of `_ClientHandler`.

delventhalz: That I don't know, probably a question for Shawn/Andi/Boyd? For what it's worth, the db and blockstore were already getting saved statefully to the handlers previous to this refactor. I think the only new thing that might create an issue is the request proto class, which gets instantiated during `init`. That could get moved to `handle` if someone who knows more than me about python and multithreading thinks it could cause issues.

rest_api/sawtooth_rest_api/routes.py
vaporos: Shouldn't clients be required to encode when crafting the URL, with decode happening within the rest_api?

rest_api/sawtooth_rest_api/exceptions.py
vaporos: ok

rest_api/sawtooth_rest_api/routes.py
delventhalz: Yes, and they do. So a client has to encode `'/'` as `'%2F'`, and that will work fine. The problem is that whether they use `'+'` or `'%2B'` for pluses, they both get turned into `' '` by aiohttp.

validator/sawtooth_validator/state/client_handlers.py
vaporos: @boydjohnson - can you comment here.

I believe there is one handler for all requests of a type, and so the following implementation of handle() would break under load due to self._result being written to by multiple threads.  Looks like result could be passed into the methods as an argument instead avoiding the state on the handler itself.

rest_api/sawtooth_rest_api/routes.py
vaporos: How about using urllib.parse.parse_qs() to parse the query string?

peterschwarz: It think my suggestion above was to apply that to where it's needed on the output coming from zmq before heading to the client.

validator/sawtooth_validator/state/client_handlers.py
boydjohnson: Yeah, I think that is a problem. One Handler subclass instance can be run in many threads at the same time, so all shared data should be in threadsafe datastructures.

rest_api/sawtooth_rest_api/routes.py
delventhalz: Ah. I'm not sure I'm a fan of that. Feels like swapping one hacky work around for another. We'd be lying to the client about what the block ids are, so the block ids we give them would _only_ be usable through the Rest Api. They wouldn't match the protobuf binaries that got posted through `POST /batches` either.

Also, the query string is the least of the issue. That you _can_ use percent escaping for, for whatever reason. Also, it would be easy enough to parse those manually. But `match_info` is giving us path parameters that that we named when we were setting up the routes. The only way to manually parse those would either be _very_ hacky, or very low level, some sort of middleware I don't entirely know how to build at the moment.

validator/sawtooth_validator/state/client_handlers.py
```diff
@@ -25,189 +25,253 @@
...
 from sawtooth_validator.protobuf.block_pb2 import BlockHeader
-from sawtooth_validator.protobuf.state_context_pb2 import Entry
 from sawtooth_validator.protobuf import validator_pb2
 
 
 LOGGER = logging.getLogger(__name__)
 
 
-class StateCurrentRequestHandler(Handler):
+class StateCurrentRequest(Handler):
```
peterschwarz: These could probably use some doc string, but could be handled as a separate commit (since they weren't introduced here).

PR #73, opened 2017-02-13
Block validator

cintel:
Enhance block validator design, add block validation of completeness and signature. Enhance test framework to support this additional validation. 
Add BlockValidator test for various fork conditions to resolve. 
Clean up callback interface between BlockValidator and ChainController

### DIFF
--- a/validator/sawtooth_validator/journal/chain.py
+++ b/validator/sawtooth_validator/journal/chain.py
@@ -15,6 +15,8 @@
 import logging
 from threading import RLock
 
+from sawtooth_signing import pbct as signing
+
 from sawtooth_validator.journal.block_wrapper import BlockStatus
 from sawtooth_validator.journal.block_wrapper import NULL_BLOCK_IDENTIFIER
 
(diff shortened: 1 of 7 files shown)


validator/sawtooth_validator/journal/chain.py
```diff
@@ -62,10 +64,28 @@ def __init__(self, consensus,
         }
 
     def _is_block_complete(self, blkw):
+        """
```
agunde406: The completer will only pass block to the journal if they are deem complete. What is your reasoning for "re-checking" the completeness?

agunde406:
The "fixes" and "clean up" commits should be squashed into the corresponding commits.

validator/sawtooth_validator/journal/chain.py
```diff
@@ -43,161 +54,246 @@ def __init__(self, consensus,
...
+            'new_chain': [],
+            'cur_chain': [],
+            'committed_batches': [],
+            'uncommitted_batches': [],
+        }
 
-    @property
-    def chain_head(self):
-        return self._chain_head
+    def _is_block_complete(self, blkw):
```
agunde406: The completer will only pass block to the journal if they are deem complete. What is your reasoning for "re-checking" the completeness?

cintel: 1) The completer is not the only source of blocks. the blocks can be loaded from the BlockCache or the BlockStore. Both of risk of having corrupted data. Long term it is not clear the policy of the BlockCache will be only to contain completed blocks. 
2) During testing blocks do not go thru the completer. Best to fail early on bad data.

PR #79, opened 2017-02-14
add run_lint to Jenkinsfile

agunde406:

### DIFF
--- a/docker/sawtooth-build-ubuntu-xenial
+++ b/docker/sawtooth-build-ubuntu-xenial
@@ -116,6 +116,7 @@ RUN apt-get update && apt-get install -y -q \
     python3-nose2 \
     python3-nose2-cov \
     python3-stdeb \
+    python3-yaml \
     python3-zmq \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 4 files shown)


docker/sawtooth-build-ubuntu-xenial
```diff
@@ -123,6 +123,7 @@ RUN apt-get update && apt-get install -y -q \
 
 RUN pip3 install bitcoin && \
     pip3 install grpcio-tools && \
-    pip3 install secp256k1
+    pip3 install secp256k1 && \
+    pip3 install pylint
```
aludvik: Try adding:
`RUN pip install pylint`

To install pylint for python2. It looks like we are still linting some python2 code.

I think you can also test run_lint locally by doing something like:
```
cd sawtooth-core/docker
vagrant $ docker build . -f sawtooth-build-ubuntu-xenial -t sawtooth-build
vagrant $ docker run -it -v /project/sawtooth-core:/project/sawtooth-core sawtooth-build /bin/bash
root@[container-id]# cd /project/sawtooth-core
root@[container-id]# ./bin/run_lint
```

Message me if that doesn't work.

agunde406: pip install pylint is happening on line 61

aludvik:
Did we figure out what the `Unable to create directory /.pylint.d` message is from?

PR #80, opened 2017-02-14
Fix failing smoke tests

boydjohnson:
The depends_on portion of the transaction processor service
definition is necessary because they reference a service
with a particular container that the docker dns gives
an address.

The sleeps on integration_test service is necessary until we have a way to get the rest_api status of if the validator is ready. ~~The other sleeps are necessary for unknown reasons that hopefully will be solved in my work this sprint, but if not we should figure out why they are necessary.~~ The socket id was not being set correctly in the python sdk. It was classname-pid which was not guaranteed to be unique in docker.

### DIFF
--- a/integration/sawtooth_integration/docker/intkey-smoke.yaml
+++ b/integration/sawtooth_integration/docker/intkey-smoke.yaml
@@ -23,8 +23,9 @@ services:
       - ../../../:/project/sawtooth-core
     expose:
       - 40000
-    entrypoint: "bash -c \"\
-        ./bin/tp_config -v validator:40000 \""
+    depends_on:
+      - validator
+    entrypoint: ./bin/tp_config -vv validator:40000
     stop_signal: SIGKILL
 
   tp_intkey:
@@ -33,9 +34,9 @@ services:
       - ../../../:/project/sawtooth-core
     expose:
       - 40000
-    entrypoint: "bash -c \"\
-        sleep 1 && \
-        ./bin/tp_intkey_python -v validator:40000\""
+    depends_on:
+      - validator
+    entrypoint: ./bin/tp_intkey_python -vv validator:40000
     stop_signal: SIGKILL
 
   validator:
@@ -44,13 +45,10 @@ services:
       - ../../../:/project/sawtooth-core
     expose:
       - 40000
-    depends_on:
-      - tp_intkey
-      - tp_config
     # start the validator with an empty genesis batch
     entrypoint: "bash -c \"\
         ./bin/sawtooth-0.8 admin genesis && \
-        ./bin/validator -v\""
+        ./bin/validator -vv\""
     stop_signal: SIGKILL
 
   rest_api:
@@ -63,9 +61,7 @@ services:
     depends_on:
       - validator
     # Starting rest_api at the same time as tp_intkey breaks
-    entrypoint: "bash -c \"\
-      sleep 4 && \
-      ./bin/rest_api --stream-url validator:40000\""
+    entrypoint: ./bin/rest_api --stream-url validator:40000
     stop_signal: SIGKILL
 
   integration_test:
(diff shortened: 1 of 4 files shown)


boydjohnson:
I can increase the sleeps. I think they are consistent between xo and intkey smoke. It ran again and passed. If you have time run it locally and see how many times it passes. I have only seen 1 fail in ~20 runs locally, but also the fail on the server.

integration/sawtooth_integration/docker/intkey-smoke.yaml
```diff
@@ -64,7 +65,7 @@ services:
       - validator
     # Starting rest_api at the same time as tp_intkey breaks
     entrypoint: "bash -c \"\
-      sleep 4 && \
+      sleep 2 && \
```
vaporos: Is this needed at all?

integration/sawtooth_integration/docker/xo-smoke.yaml
```diff
@@ -74,7 +81,7 @@ services:
       - rest_api
     # Wait for rest_api
     entrypoint: "bash -c \"\
-      sleep 2 && \
+      sleep 8 && \
```
vaporos: Let's use 30 seconds here.

integration/sawtooth_integration/docker/intkey-smoke.yaml
```diff
@@ -33,6 +35,8 @@ services:
       - ../../../:/project/sawtooth-core
     expose:
       - 40000
+    depends_on:
+      - validator
     entrypoint: "bash -c \"\
         sleep 1 && \
```
vaporos: Can this be removed?

integration/sawtooth_integration/docker/xo-smoke.yaml
```diff
@@ -33,7 +35,11 @@ services:
       - ../../../:/project/sawtooth-core
     expose:
       - 40000
-    entrypoint: ./bin/tp_xo_python validator:40000
+    depends_on:
+      - validator
+    entrypoint: "bash -c \"\
+     sleep 1 && \
```
vaporos: Do we really need a sleep here?

rest_api/sawtooth_rest_api/routes.py
```diff
@@ -167,7 +167,7 @@ def _try_validator_request(self, message_type, content):
         Sends a protobuf message to the validator
         Handles a possible timeout if validator is unresponsive
         """
-        timeout = 5
+        timeout = 20
```
vaporos: This seems arbitrary, how about 5 minutes?

integration/sawtooth_integration/docker/xo-smoke.yaml
boydjohnson: From local tests it seems necessary. I'm wondering if it has to do with dns resolution since each thing that resolves a container name into an address needs to be staggered.

rest_api/sawtooth_rest_api/routes.py
peterschwarz: Since we're responding to http requests, looking around at what browsers will time out, 300 seconds is the minimum (Firefox), everything else will wait longer (for example, https://support.microsoft.com/en-us/help/181050/internet-explorer-error-connection-timed-out-when-server-does-not-respond).  So we don't have to be stingy here.

integration/sawtooth_integration/docker/xo-smoke.yaml
```diff
@@ -23,8 +23,9 @@ services:
       - ../../../:/project/sawtooth-core
     expose:
       - 40000
-    entrypoint: "bash -c \"\
-        ./bin/tp_config -v validator:40000 \""
+    depends_on:
+      - validator
+    entrypoint: ./bin/tp_config -vv validator:40000
```
peterschwarz: Do we want to leave the debug logging in?

boydjohnson: I think so. Until we have tar files of the logs, it will help debugging failed tests.

PR #82, opened 2017-02-15
Initial connectivity handling fixes

vaporos:
Prior to these changes, it was not possible to use a the default config file with the cluster command since it specified a config argument pointing at a config it wrote to disk.  Tested with sawtooth cluster and launcher.

### DIFF
--- a/core/sawtooth/manage/daemon.py
+++ b/core/sawtooth/manage/daemon.py
@@ -35,14 +35,6 @@ class DaemonNodeController(NodeController):
         self._host_name = host_name
         self._verbose = verbose
         self._base_config = get_validator_configuration([], {})
-        # Additional configuration for the genesis validator
-        self._genesis_cfg = {
-            "InitialConnectivity": 0,
-        }
-        # Additional configuration for the non-genesis validators
-        self._non_genesis_cfg = {
-            "InitialConnectivity": 1,
-        }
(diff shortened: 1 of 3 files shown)


core/sawtooth/manage/daemon.py
```diff
@@ -77,13 +69,11 @@ def _construct_start_command(self, node_args):
         if not os.path.exists(config_dir):
             os.makedirs(config_dir)
```
ojalatodd: "config_dir" doesn't seem to even be used elsewhere in the code now.  Should the creation of this directory, and references to it, just be removed altogether?

core/sawtooth/manage/subproc.py
```diff
@@ -59,13 +51,10 @@ def _construct_start_command(self, node_args):
             config_dir = os.path.join(node_args.currency_home, 'etc')
```
ojalatodd: Same as above for daemon:
"config_dir" doesn't seem to even be used elsewhere in the code now. Should the creation of this directory, and references to it, just be removed altogether?

core/sawtooth/manage/daemon.py
ojalatodd: Just wanted to make sure you saw my comment @vaporos. ;)

PR #85, opened 2017-02-15
Add State Views and Config View specialization

peterschwarz:
**Add StateView and its Factory**

StateViewFactory provides read-only views into state stored in a merkle tree database. These views can provide a window into the state which can be used throughout a block of processing without concern of the state being changed underneath the process.

**Add ConfigView and its Factory**

ConfigView is a specialized StateView which translates keys into state addresses and transforms state entries into their respective values.  This can be used by validator components to read on-chain settings at viewed from the lens of a specific point in the chain.

### DIFF
--- a/validator/sawtooth_validator/state/state_view.py
+++ b/validator/sawtooth_validator/state/state_view.py
@@ -1,0 +1,93 @@
+
+# Copyright 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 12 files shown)


validator/sawtooth_validator/state/state_view.py
```diff
@@ -0,0 +1,93 @@
...
+        """
+        Returns:
+            list of str: the list of addresses available in this view
+        """
+        return self._tree.addresses()
+
+    def leaves(self, prefix):
+        """
+        Args:
+            prefix (str): an address prefix under which to look fo leaves
```
jsmitchell: s/fo/for

cintel:
Probably a nit but shouldn't all of these files have a new line at the end?

peterschwarz:
new-lines would fail our linting rules. But, since #79 is not merged in yet, it did not catch some other linting issues that are there.  Will fix.

nit away!

peterschwarz:
Hold off on this - need to correct some tests

PR #68, opened 2017-02-10
Move sawtooth cluster new cli

ojalatodd:

PR #88, opened 2017-02-15
Remove expected column from cluster status

ojalatodd:


ojalatodd:
Waiting on this until PR #68 is merged.

PR #90, opened 2017-02-15
Add discard builds property to Jenkinsfile

rbuysse:
This will discard builds older than 31 days from Jenkins.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,6 +15,11 @@
 // limitations under the License.
 // ------------------------------------------------------------------------------
 
+// Discard old builds after 31 days
+properties([[$class: 'BuildDiscarderProperty', strategy:
+        [$class: 'LogRotator', artifactDaysToKeepStr: '',
+        artifactNumToKeepStr: '', daysToKeepStr: '31', numToKeepStr: '']]]);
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -15,6 +15,11 @@
 // limitations under the License.
 // ------------------------------------------------------------------------------
 
+// Discard old builds after 31 days
+properties([[$class: 'BuildDiscarderProperty', strategy:
+        [$class: 'LogRotator', artifactDaysToKeepStr: '',
+        artifactNumToKeepStr: '', daysToKeepStr: '31', numToKeepStr: '']]]);
+
```
aludvik: Do you need to set artifactDaysToKeepStr too? The pipeline script generator also has a schedule/build periodically if not run option. Does leaving this option out mean the cleanup will only happen when the job is triggered by a PR?

rbuysse: artifactDaysToKeepStr results in a different effect: "if not empty, artifacts from builds older than this number of days will be deleted, but the logs, history, reports, etc for the build will be kept"

We'll be triggering this job out of band, so I left out scheduled runs here.

PR #93, opened 2017-02-16
Remove legacy tests from Jenkins, build python 3 deb packages

vaporos:


vaporos:
This PR does not remove a lot of legacy code, that will come in a future PR.  This one is more about shifting to building/testing only python3 packages and updating Jenkins accordingly.

Jenkinsfile
```diff
@@ -57,49 +57,12 @@ node ('master') {
...
 
         // Run the tests
-        stage("Run 0-8 Tests"){
+        stage("Run Tests") {
             // Required docker containers are built by the tests
             sh './bin/docker_build_all -p $(printf $BUILD_TAG | sha256sum | cut -c1-64)'
             sh './bin/run_tests -p $(printf $BUILD_TAG | sha256sum | cut -c1-64)'
         }
 
-        stage("Remove Docker Images") {
```
peterschwarz: Should this have been removed as well?

sdk/python/setup.py
```diff
@@ -0,0 +1,93 @@
+# Copyright 2016 Intel Corporation
```
peterschwarz: License year

signing/setup.py
```diff
@@ -72,7 +72,9 @@ def version(default):
     author='Mic Bowman, Intel Labs',
```
peterschwarz: Is this still an accurate author list?

vaporos:
protobuf reorg is in this PR as it is required to build the deb files.

sdk/python/setup.py
vaporos: Updated to "2016, 2017" (this file is derived from other files.

signing/setup.py
vaporos: Will update this, descriptions, versioning in another PR.  This wasn't required to disable legacy tests or to build python 3 deb files, so I did not update it here.

Jenkinsfile
vaporos: I think so, as a similar stage "Remove docker images" removes both the build and docs images.

vaporos:
@aludvik - should we do that in a different move-docker-images PR, or is this related to changes here?

aludvik:
We can do it in another PR

PR #94, opened 2017-02-16
Implement CurveZMQ security for the network

jsmitchell:
This implements an authentication handshake and transport encryption using ZMQ's CurveZMQ support (ec based on edwards). The support is implemented in the shared Interconnect object, but is only enabled for the network (validator to validator) instance. In the future, we can also enable this for the intercomponent instance, but that will require changes to 'clients'.

### DIFF
--- a/validator/sawtooth_validator/networking/interconnect.py
+++ b/validator/sawtooth_validator/networking/interconnect.py
@@ -21,8 +21,11 @@ from threading import Thread
 import uuid
 
 import zmq
+import zmq.auth
+from zmq.auth.asyncio import AsyncioAuthenticator
 import zmq.asyncio
 
+from sawtooth_validator.exceptions import LocalConfigurationError
 from sawtooth_validator.protobuf import validator_pb2
 from sawtooth_validator.networking import future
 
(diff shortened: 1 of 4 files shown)


validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -39,11 +42,16 @@ def get_enum_name(enum_value):
 
 
 class _SendReceive(object):
-    def __init__(self, address, futures, identity=None, dispatcher=None):
+    def __init__(self, address, futures, identity=None,
+                 dispatcher=None, secured=False,
+                 server_public_key=None, server_private_key=None):
```
peterschwarz: Can we document these parameters?  Are they strings? File names?

validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -143,20 +181,30 @@ def __init__(self,
                  endpoint,
                  dispatcher,
                  identity=None,
-                 peer_connections=None):
+                 peer_connections=None,
+                 secured=False,
+                 server_public_key=None,
+                 server_private_key=None):
```
peterschwarz: Doc here on these fields

validator/sawtooth_validator/networking/interconnect.py
jsmitchell: Added in squashed commit 3448bb7

validator/sawtooth_validator/networking/interconnect.py
jsmitchell: Added in squashed commit 3448bb7

PR #95, opened 2017-02-16
Fix LIBOR updates

aludvik:


bin/load_bond_data
```diff
@@ -1,6 +1,17 @@
 #!/bin/bash
 
-./bin/btp load -vvv --wait \
+SAWTOOTH=$(cd $(dirname $(dirname $0)) && pwd)
```
vaporos: Using top_dir= here would match other scripts in bin/

bin/load_bond_data
```diff
@@ -1,6 +1,17 @@
...
+PYTHONPATH=$SAWTOOTH/core:$SAWTOOTH/extensions/bond
+PYTHONPATH=$PYTHONPATH:$SAWTOOTH/core/build/lib.linux-x86_64-2.7
+PYTHONPATH=$PYTHONPATH:$SAWTOOTH/signing
+PYTHONPATH=$PYTHONPATH:$SAWTOOTH/signing/build/lib.linux-x86_64-2.7
+PYTHONPATH=$PYTHONPATH:$SAWTOOTH/validator
+PYTHONPATH=$PYTHONPATH:$SAWTOOTH/extensions/mktplace
```
vaporos: mktplace, arcade probably not required for this particular script.  Same might be true of validator?

aludvik: Addressed these issues and force pushed update

PR #99, opened 2017-02-17
Sequence independent validator-transaction processor shutdown, restart

boydjohnson:
After this, python transaction processors can stop (if allowed to do so gracefully (1 keyboardinterrupt)) even during load. The tps can also recover after validator shutdown, even during load. The validator still needs at least 2 keyboard interrupts to stop it.

The protos are updated, so if seeking to test it, run protogen.

### DIFF
--- a/protos/processor.proto
+++ b/protos/processor.proto
@@ -18,7 +18,8 @@ option java_multiple_files = true;
 option java_package = "sawtooth.sdk.protobuf";
 
 
-// The registration request from the transaction processor to the validator/executor
+// The registration request from the transaction processor to the
+// validator/executor
 message TpRegisterRequest {
 
     // A settled upon name for the capabilities of the transaction processor.
@@ -40,8 +41,27 @@ message TpRegisterRequest {
     repeated string namespaces = 4;
 }
 
-// A response sent from the validator to the transaction processor acknowledging the registration
+// A response sent from the validator to the transaction processor
+// acknowledging the registration
 message TpRegisterResponse {
+    enum Status {
+        OK = 0;
+        ERROR = 1;
+    }
+
+    Status status = 1;
+}
+
+// The unregistration request from the transaction processor to the
+// validator/executor. The correct handlers are determined from the
+// zeromq identity of the tp, on the validator side.
+message TpUnregisterRequest {
+
+}
+
+// A response sent from the validator to the transaction processor
+// acknowledging the unregistration
+message TpUnregisterResponse {
     enum Status {
         OK = 0;
         ERROR = 1;
--- a/protos/validator.proto
+++ b/protos/validator.proto
@@ -31,61 +31,70 @@ message Message {
         DEFAULT = 0;
         // Registration request from the transaction processor to the validator
         TP_REGISTER_REQUEST = 1;
-        // Registration response from the validator to the transaction processor
+        // Registration response from the validator to the
+        // transaction processor
         TP_REGISTER_RESPONSE = 2;
-        // Process Request from the validator/executor to the transaction processor
-        TP_PROCESS_REQUEST = 3;
+        // Tell the validator that the transaction processor
+        // won't take any more transactions
+        TP_UNREGISTER_REQUEST = 3;
+        // Response from the validator to the tp that it won't
+        // send any more transactions
+        TP_UNREGISTER_RESPONSE = 4;
+        // Process Request from the validator/executor to the
+        // transaction processor
+        TP_PROCESS_REQUEST = 5;
         // Process response from the transaction processor to the validator/executor
-        TP_PROCESS_RESPONSE = 4;
+        TP_PROCESS_RESPONSE = 6;
         // State get request from the transaction processor to validator/context_manager
-        TP_STATE_GET_REQUEST = 5;
+        TP_STATE_GET_REQUEST = 7;
         // State get response from the validator/context_manager to the transaction processor
-        TP_STATE_GET_RESPONSE = 6;
+        TP_STATE_GET_RESPONSE = 8;
         // State set request from the transaction processor to the validator/context_manager
-        TP_STATE_SET_REQUEST = 7;
+        TP_STATE_SET_REQUEST = 9;
         // State set response from the validator/context_manager to the transaction processor
-        TP_STATE_SET_RESPONSE = 8;
+        TP_STATE_SET_RESPONSE = 10;
         // State delete request from the transaction processor to the validator/context_manager
-        TP_STATE_DEL_REQUEST = 9;
+        TP_STATE_DEL_REQUEST = 11;
         // State delete response from the validator/context_manager to the transaction processor
-        TP_STATE_DEL_RESPONSE = 10;
+        TP_STATE_DEL_RESPONSE = 12;
         // Submission of a batchlist from the web api or another client to the validator
-        CLIENT_BATCH_SUBMIT_REQUEST = 11;
+        CLIENT_BATCH_SUBMIT_REQUEST = 100;
         // Response from the validator to the web api/client that the submission was accepted
-        CLIENT_BATCH_SUBMIT_RESPONSE = 12;
+        CLIENT_BATCH_SUBMIT_RESPONSE = 101;
         // A request to list blocks from the web api/client to the validator
-        CLIENT_BLOCK_LIST_REQUEST = 13;
-        CLIENT_BLOCK_LIST_RESPONSE = 14;
-        CLIENT_BLOCK_GET_REQUEST = 15;
-        CLIENT_BLOCK_GET_RESPONSE = 16;
-        CLIENT_BATCH_LIST_REQUEST = 17;
-        CLIENT_BATCH_LIST_RESPONSE = 18;
-        CLIENT_BATCH_GET_REQUEST = 19;
-        CLIENT_BATCH_GET_RESPONSE = 20;
-        CLIENT_TRANSACTION_LIST_REQUEST = 21;
-        CLIENT_TRANSACTION_LIST_RESPONSE = 22;
-        CLIENT_TRANSACTION_GET_RESPONSE = 23;
+        CLIENT_BLOCK_LIST_REQUEST = 102;
+        CLIENT_BLOCK_LIST_RESPONSE = 103;
+        CLIENT_BLOCK_GET_REQUEST = 104;
+        CLIENT_BLOCK_GET_RESPONSE = 105;
+        CLIENT_BATCH_LIST_REQUEST = 106;
+        CLIENT_BATCH_LIST_RESPONSE = 107;
+        CLIENT_BATCH_GET_REQUEST = 108;
+        CLIENT_BATCH_GET_RESPONSE = 109;
+        CLIENT_TRANSACTION_LIST_REQUEST = 110;
+        CLIENT_TRANSACTION_LIST_RESPONSE = 111;
+        CLIENT_TRANSACTION_GET_REQUEST = 112;
+        CLIENT_TRANSACTION_GET_RESPONSE = 113;
         // Client state request of the current state hash to be retrieved from the journal
-        CLIENT_STATE_CURRENT_REQUEST = 24;
+        CLIENT_STATE_CURRENT_REQUEST = 114;
         // Response with the current state hash
-        CLIENT_STATE_CURRENT_RESPONSE = 25;
+        CLIENT_STATE_CURRENT_RESPONSE = 115;
         // A request of all the addresses under a particular prefix, for a state hash.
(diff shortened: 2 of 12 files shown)


protos/validator.proto
```diff
@@ -33,59 +33,63 @@ message Message {
         TP_REGISTER_REQUEST = 1;
         // Registration response from the validator to the transaction processor
         TP_REGISTER_RESPONSE = 2;
+        // Tell the validator that the transaction processor won't take any more transactions
```
peterschwarz: This reordering of the enums is certainly fine here, but we should probably come up with a strategy for giving ourselves room to grow.

sdk/python/sawtooth_sdk/client/future.py
```diff
@@ -37,6 +44,9 @@ def result(self, timeout=None):
             if self._result is None:
                 if not self._condition.wait(timeout):
                     raise FutureTimeoutError('Future timed out')
+        if isinstance(self._result, FutureError):
+            raise ValidatorConnectionError(
```
peterschwarz: The error could be stored in the FutureError - keeping the error as part of the operation, not an interpretation of the error.

sdk/python/sawtooth_sdk/client/stream.py
```diff
@@ -46,17 +50,24 @@ class _SendReceiveThread(Thread):
     Internal thread to Stream class that runs the asyncio event loop.
     """
 
-    def __init__(self, url, futures):
+    def __init__(self, url, futures, ready_event):
+        """constructor for background thread
+
+        :param url (str): the address to connect to the validator on
+        :param futures (FutureCollection):
+        :param ready_event (threading.Event:
```
peterschwarz: Missing closing paren

sdk/python/sawtooth_sdk/processor/core.py
```diff
@@ -33,65 +43,175 @@
...
+            response = TpUnRegisterResponse()
+            response.ParseFromString(f.result(1).content)
+            LOGGER.debug("unregister attempt: %s",
+                         TpUnRegisterResponse.Status.Name(response.status))
+
+    def start(self):
+        fut = None
+        try:
+            self._register()
+            while True:
```
peterschwarz: There probably could be a control-flag here, to be set during the shutdown phase.

sdk/python/sawtooth_sdk/client/stream.py
```diff
@@ -114,59 +124,124 @@ def _get_message(self):
...
-        asyncio.ensure_future(self._receive_message(), loop=self._event_loop)
-        self._event_loop.run_forever()
+        first_time = True
+        while True:
+            if self._event_loop is None:
+                self._event_loop = zmq.asyncio.ZMQEventLoop()
+                asyncio.set_event_loop(self._event_loop)
+            if self._context is None:
+                self._context = zmq.asyncio.Context()
+            self._sock = self._context.socket(zmq.DEALER)
```
peterschwarz: Should set the linger property

sdk/python/sawtooth_sdk/client/future.py
boydjohnson: Yeah, sounds good. I will update this.

sdk/python/sawtooth_sdk/client/stream.py
boydjohnson: Thanks, updating.

sdk/python/sawtooth_sdk/client/stream.py
boydjohnson: Thanks, updating.

sdk/python/sawtooth_sdk/processor/core.py
boydjohnson: So my thinking was that the transaction processor would only stop due to signals (sigint is handled by the KeyboardInterrupt exception handling) and not programmatically. As I see it, any exception bumps the flow of control out of the while loop, with sigint getting special handling.

sdk/python/sawtooth_sdk/processor/core.py
```diff
@@ -33,65 +44,174 @@
...
+        """Returns all of the TpRegisterRequests for handlers
 
-        while True:
-            if self._stop:
-                break
-            msg = self._stream.receive()
-            LOGGER.debug(
-                'received message of type: %s',
-                Message.MessageType.Name(msg.message_type))
+        :note: itertools.chain.from_iterable flattens the nested lists
```
vaporos: Is this note useful as API doc, or should it be a non-API doc comment within the method?

sdk/python/sawtooth_sdk/processor/core.py
```diff
@@ -33,65 +44,174 @@
...
                 self._stream.send_back(
                     message_type=Message.TP_PROCESS_RESPONSE,
                     correlation_id=msg.correlation_id,
                     content=TpProcessResponse(
                         status=TpProcessResponse.INVALID_TRANSACTION
                     ).SerializeToString())
-            except InternalError as ie:
-                LOGGER.warning("State Error! %s", ie)
+            except ValidatorConnectionError as vce:
+                LOGGER.warning("%s", vce)
```
vaporos: Can we provide more context here?  Tracking down where this was output would be challenging if we don't add any unique text.

sdk/python/sawtooth_sdk/processor/core.py
```diff
@@ -33,65 +44,174 @@
...
+        except InternalError as ie:
+            LOGGER.warning("State Error! %s", ie)
```
vaporos: This assumes all InternalErrors are state errors, but presumably there may be other reasons we have InternalErrors.

sdk/python/sawtooth_sdk/processor/core.py
vaporos: Same comment for other logging statements here.

sdk/python/sawtooth_sdk/processor/core.py
```diff
@@ -33,65 +44,174 @@
...
                 self._stream.send_back(
                     message_type=Message.TP_PROCESS_RESPONSE,
                     correlation_id=msg.correlation_id,
                     content=TpProcessResponse(
                         status=TpProcessResponse.INTERNAL_ERROR
                     ).SerializeToString())
+            except ValidatorConnectionError as vce:
+                LOGGER.warning("%s", vce)
+        except ValidatorConnectionError as vce:
+            LOGGER.warning("%s", vce)
```
vaporos: It might be good to add comments on these except blocks that log errors to explain why we are only logging.

sdk/python/sawtooth_sdk/processor/core.py
```diff
@@ -33,65 +44,174 @@
...
+                while True:
+                    if fut is not None:
+                        # process futures as long as the tp has them,
+                        # if the message doesn't come from zeromq->asyncio
+                        # in 1 second raise a TimeOutError (TOError) and
+                        # be done.
+                        self._process_future(fut, 1, sigint=True)
+                        fut = self._stream.receive()
+            except TOError:
+                pass
```
vaporos: pass statements that throw away exceptions always warrant a comment to explain.

sdk/python/sawtooth_sdk/processor/core.py
```diff
@@ -13,17 +13,28 @@
 # limitations under the License.
 # ------------------------------------------------------------------------------
 
+from concurrent.futures import CancelledError
+from concurrent.futures import TimeoutError as TOError
```
vaporos: Renaming things on import generally makes it harder to read the code.  If this conflicts with another TimeoutError, we could instead refer to it fully qualified concurrent.futures.TimeoutError.

protos/processor.proto
```diff
@@ -50,6 +50,32 @@ message TpRegisterResponse {
     Status status = 1;
 }
 
+// The unregistration request from the transaction processor to the validator/executor
+message TpUnRegisterRequest {
```
vaporos: Another possible approach would be to send less information in this request, then find the processors based on identity instead of (family, version, encoding).  Presumably that approach would be more of an unregister-all message.

protos/processor.proto
```diff
@@ -50,6 +50,32 @@ message TpRegisterResponse {
...
+    // the transaction family name for the logic of the handler
```
vaporos: Capitalize the first t

protos/processor.proto
```diff
@@ -50,6 +50,32 @@ message TpRegisterResponse {
...
+    string family = 1;
+
+    // The version supported;
```
vaporos: remove the trailing ;

protos/processor.proto
```diff
@@ -50,6 +50,32 @@ message TpRegisterResponse {
...
+    string version = 2;
+
+    //supported encoding of the handler
```
vaporos: Add space and capitalize the s

protos/processor.proto
```diff
@@ -50,6 +50,32 @@ message TpRegisterResponse {
...
+    string encoding = 3;
+
+    // The namesapces that this transaction processor handler expects
+    // to interact with.
+    repeated string namespaces = 4;
+}
+
+// a response sent from the validator to the transaction processor acknowledging the unregistration
```
vaporos: Curious if the comments wrap < 80. Hard to judge from the github interface.

protos/processor.proto
```diff
@@ -50,6 +50,32 @@ message TpRegisterResponse {
...
+    // the transaction family name for the logic of the handler
+    string family = 1;
+
+    // The version supported;
+    string version = 2;
+
+    //supported encoding of the handler
+    string encoding = 3;
+
+    // The namesapces that this transaction processor handler expects
```
vaporos: Typo 'namesapces'.

Do we really need to specify the namespaces here?  I wouldn't think so, because they don't uniquely identify the TP in any way.

protos/processor.proto
vaporos: Seems like this should be TpUnregisterRequest since unregister is a word.

sdk/python/sawtooth_sdk/client/stream.py
```diff
@@ -26,14 +27,17 @@
...
 from sawtooth_sdk.client.future import Future
 from sawtooth_sdk.client.future import FutureCollection
 from sawtooth_sdk.client.future import FutureCollectionKeyError
 from sawtooth_sdk.client.future import FutureResult
-
+from sawtooth_sdk.client.future import FutureError
 
 LOGGER = logging.getLogger(__file__)
 
+RECONNECT_EVENT = -1
```
vaporos: A comment here describing why this is set to -1 would be helpful.

sdk/python/sawtooth_sdk/client/stream.py
```diff
@@ -46,17 +50,24 @@ class _SendReceiveThread(Thread):
     Internal thread to Stream class that runs the asyncio event loop.
     """
 
-    def __init__(self, url, futures):
+    def __init__(self, url, futures, ready_event):
+        """constructor for background thread
+
+        :param url (str): the address to connect to the validator on
+        :param futures (FutureCollection):
+        :param ready_event (threading.Event):
```
vaporos: Should fill out this API doc.  In particular, some description of ready_event works would be useful.

sdk/python/sawtooth_sdk/processor/core.py
```diff
@@ -33,65 +44,174 @@
...
+                self._process_future(fut)
+        except KeyboardInterrupt:
+            try:
+                # tell the validator to not send any more messages
+                self._unregister()
+                while True:
+                    if fut is not None:
+                        # process futures as long as the tp has them,
+                        # if the message doesn't come from zeromq->asyncio
+                        # in 1 second raise a TimeOutError (TOError) and
```
vaporos: What is "the message" in this context?  Why 1s?

protos/processor.proto
boydjohnson: I like the approach of sending less and sending just one unregister message per request to disconnect even if there are multiple handlers. But now I can't think of anything that would need to be in the message since the zeromq identity is already known on the validator side.

protos/validator.proto
boydjohnson: I could add a break of 10 (20?) numbers between top level namespace message types. So TP would end at 12, but CLIENT wouldn't start until 23 (33?).

peterschwarz:
You can add the break, but start the next block on an even number: TP ends
at 12, CLIENT starts at 100, etc.  Using http error codes as an example
here of the distances between blocks of related items.

boydjohnson:
I updated logging statements, fixed typos, and added more comments. The largest difference is that one TpUnregisterRequest is sent for all handlers on a transaction processor, so the handling in processor_iterator.py and processor_handlers.py was updated.

PR #100, opened 2017-02-17
Tests for the parallel scheduler predecessor tree

nick-drozd:


validator/tests/unit3/test_scheduler/tests.py
```diff
@@ -299,104 +297,788 @@ def test_valid_batch_invalid_batch(self):
...
+        self.tree = PredecessorTree(token_size=1)
+
+    def tearDown(self):
+        self.tree = None
+
+    def test_common_initial_segments(self):
+        '''Tests basic predecessor tree functions
+        '''
+
+        # self.verbose = 1
```
vaporos: Should remove the comment-out code.

validator/tests/unit3/test_scheduler/tests.py
```diff
@@ -299,104 +297,788 @@ def test_valid_batch_invalid_batch(self):
...
+
+    def setUp(self):
+        self.verbose = 0
+        self.tree = PredecessorTree(token_size=1)
+
+    def tearDown(self):
+        self.tree = None
+
+    def test_common_initial_segments(self):
+        '''Tests basic predecessor tree functions
```
vaporos: I feel like the above class-level comment does a good job explaining the general approach, but that I'm left at the per-statement level wondering why the specific readers/writers/assertions are made.  I feel like we need some ASCII way to visualize the tree, and then we could have comments "the tree is now expected to looks like this: ...", which would then explain the result and why we are checking the addresses as we are.  Wondering if we can use the verbose output to capture the tree representations and then embed that as comments.

vaporos: Should we break the tests down into more granular tests?

validator/tests/unit3/test_scheduler/tests.py
```diff
@@ -299,104 +297,788 @@ def test_valid_batch_invalid_batch(self):
...
+
+    def set_writer(self, address, txn):
+        self.tree.set_writer(address, txn)
+
+    def get_node(self, address):
+        return self.tree.get(address)
+
+    # display
+
+    def show_tree(self):
```
vaporos: If we turned this into logger statements instead of print, the verbosity can probably be controlled by the test runner (nose2-3 -v).

vaporos: Because of the newlines here, the approach would be to split on newline and log each line separately; ideally, prefixing the line with something friendly to grep.

validator/tests/unit3/test_scheduler/tests.py
```diff
@@ -299,104 +297,788 @@ def test_valid_batch_invalid_batch(self):
...
+        reader_dict = {address: reader}
+        '''
+        for address, reader in reader_dict.items():
+            self.add_reader(address, reader)
+
+    def set_writers(self, writer_dict):
+        '''
+        writer_dict = {address: writer}
+
+        WARNING: The order in which writers are set is not specified.
```
vaporos: If the caller cared about order, they could use an OrderedDict and get predictable results while using this method.

validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -116,23 +116,25 @@ def find_write_predecessors(self, address):
...
+
+        for token in tokens:
             if token not in node.children:
-                break
+                if enclosing_writer is not None:
+                    predecessors.add(enclosing_writer)
+                return predecessors
 
             node = node.children[token]
-            if i == len(tokens) - 1:
```
vaporos: If I understand this correctly, it looks like node will equal the address node when we leave the for loop, which is guaranteed because of the return within the loop would have prevented partial addresses from continuing to be processed.  Might be worth a comment that this side-effect of the return is very important to ensuring correctness.  If we were not returning, then we would process a node higher in the tree which would be very incorrect.

validator/tests/unit3/test_scheduler/tests.py
nick-drozd: I can add what it currently prints to the comments. That would make it easier to visualize what's happening, but I don't know if it's the best way to represent it.

nick-drozd: My thought was that there should be at least one test that works through a normal case in reasonable depth (this test add readers and writer 13 times; maybe that depth is unreasonable?). It can also serve as a reference example, since personally I find it difficult to think of good examples at the whiteboard.

validator/tests/unit3/test_scheduler/tests.py
```diff
@@ -299,104 +301,1031 @@ def test_valid_batch_invalid_batch(self):
...
+
+    def tearDown(self):
+        self.tree = None
 
     def test_predecessor_tree(self):
-        """Tests basic functionality of the scheduler's radix tree.
+        '''Tests basic predecessor tree functions
+        '''
+
+        # This test is intended to show the evolution of a tree
```
peterschwarz: This comment should be the doc string

vaporos:
I wonder if we can kick the build off again. If not, we should re-open this PR.

PR #101, opened 2017-02-17
Move poet1 to new home for 0.8 in sawtooth validator.

cintel:


validator/tests/unit3/test_poet1/utils.py
```diff
@@ -0,0 +1,51 @@
...
+# limitations under the License.
+# ------------------------------------------------------------------------------
+import hashlib
+import random
+import string
+
+from sawtooth_signing import pbct as signing
+
+
+class AttrDict(dict):
```
peterschwarz: Doesn't look like it's used.

docker/sawtooth-base
```diff
@@ -21,6 +21,7 @@ RUN apt-get update && apt-get install -y -q \
     python3-pip \
     python3-colorlog \
     python3-cbor \
+    python3-requests \
```
aludvik: Is the `python3-requests` package something that will be used by all components, or just the validator? If it is just the validator, then it should be added to `docker/sawtooth-validator`. Otherwise this is fine.

PR #102, opened 2017-02-18
Add units tests for client request handlers

delventhalz:


validator/tests/unit3/test_client_request_handlers/tests.py
```diff
@@ -0,0 +1,410 @@
...
+    def _initialize(self, handler, request_proto, response_proto,
+                    store=None, roots=None):
+        self._identity = '1234567'
+        self._handler = handler
+        self._status = response_proto
+        self._request_proto = request_proto
+        self._store = store
+        self._roots = roots
+
+    def _make_request(self, **kwargs):
```
vaporos: These are used by the subclass, so probably shouldn't be prefixed with _.

validator/tests/unit3/test_client_request_handlers/tests.py
```diff
@@ -0,0 +1,410 @@
...
+
+    def _serialize(self, **kwargs):
+        request = self._request_proto(**kwargs)
+        return request.SerializeToString()
+
+    def _handle(self, request):
+        result = self._handler.handle(self._identity, request)
+        return result.message_out
+
+    def _make_bad_request(self, **kwargs):
```
vaporos: What is the significance here that differentiates a bad request from a normal request?

validator/tests/unit3/test_client_request_handlers/tests.py
```diff
@@ -0,0 +1,410 @@
...
+            roots=roots)
+
+    def test_state_list_request(self):
+        """Verifies requests for data lists without parameters work properly.
+        Checks status, head_id, and length and type of leaves list.
+        """
+        response = self._make_request()
+
+        self.assertEqual(self._status.OK, response.status)
+        self.assertEqual('2', response.head_id)
```
vaporos: Additional comments in the API doc explaining the magic values here ('2' and 3) would help (why should it be '2' and 3 at this point?).  (Comment really applies to several of the tests here that test 'magic' values.)

validator/tests/unit3/test_client_request_handlers/tests.py
```diff
@@ -0,0 +1,410 @@
+# Copyright 2016 Intel Corporation
```
vaporos: Should be 2017

validator/tests/unit3/test_client_request_handlers/tests.py
delventhalz: It truncates the protobuf.

PR #103, opened 2017-02-19
Add the Validator Registry Transaction Family

agunde406:
This includes the specification and transaction processor tests

### DIFF
--- a/docs/source/transaction_family_specifications/validator_registry_transaction_family.rst
+++ b/docs/source/transaction_family_specifications/validator_registry_transaction_family.rst
@@ -1,0 +1,244 @@
+***************************************************
+Validator Registry Transaction Family Specification
+***************************************************
+
+Overview
+=========
+
+The validator registry transaction family provides a way to add new validators
+to the network.
+
+The validator's information is used by poet_consensus to verify that when a
+validator tries to claim a block it is following the block claiming policies of
+PoET consensus. For example, the Z policy will refuse blocks from a validator
+if it has already claimed a disproportionate amount of blocks compared to the
+other validators on the network. The C policy says that a validator may not
+claim blocks, after being added to the registry or updating their information,
+until a certain number of blocks are claimed by other participants. The number
+of blocks claimed since signup can be found by checking the current block number
+against the block number that is stored in the validator information when the
+validator registry information has either been added or updated. And finally the K
+policy requires new signup information to be sent to the validator registry
+after it has claimed a maximum number of block. If the validator is not found
+in the validator registry the block will be rejected since there is no way to
+validate.
+
+Currently the number of blocks claimed by a specific validator is stored within
+the poet_consensus, not within the validator registry.
+
+
+State
+=====
+This section describes in detail how validator information, including
+identification and signup data, is stored and addressed using the validator
+registry transaction family.
+
+The following protocol buffers definition defines the validator info:
+
+.. code-block:: protobuf
+
+  message ValidatorInfo {
+    // Used to check if the validator's signup has been revoked.
+    string registered = 1;
+
+    // The name of the endpoint. It should be a human readable name.
+    string name = 2;
+
+    // The validator's public key (currently using signer_pubkey as this is
+    // stored in the transaction header)
+    string id = 3;
+
+    // This is the protocol buffer described below.
+    SignUpInfo signup_info = 4;
+
+    // The block number that this ValidatorInfo was set on.
+    // This is used as a form of time to help with the enforcement of the C test.
+    int32 block_num = 5;
+  }
+
+The above ValidatorInfo includes the following protocol buffer:
+
+.. code-block:: protobuf
+
+  message SignUpInfo {
+    // Encoded public key corresponding to private key used by PoET to sign
+    // wait certificates.
+    string poet_public_key = 1;
+
+    // Information that can be used internally to verify the validity of
+    // the signup information stored as an opaque buffer.
+    string proof_data = 2;
+
+    // A string corresponding to the anti-Sybil ID for the enclave that
+    // generated the signup information.
+    string anti_sybil_id = 3;
+  }
+
+Currently the poet_enclave needs a way to check how many validators are
+currently registered within the validator registry. It also needs a way to
+find the validator_id associated with an anti-sybil_id. A ValidatorMap, where
+the anti_sybil_id is the key and the validator_id is the value solves both of
+these problems.
+
+.. code-block:: protobuf
+
+  message ValidatorMap {
+      // Contains a validator entry where the key is an  anti_sybil_id,
+      // and the value is a validator_id
+      message Entry {
+          string key = 1;
+          string value = 2;
+      }
+
+      // List of validator entries
+      repeated Entry entries = 1;
+  }
+
+
+This can be used to check the number of validators registered. This
+information is used to decide what number of blocks a validator has to wait
+for before it can start claiming blocks after it adds new signup information
+to the validator registry. This check is necessary because if the number of
+blocks that must be waited on is greater than the number of validators minus
+one, it is possible for the network to get into a state where nobody can
+publish blocks because the validators are all waiting for more blocks to be
+committed or their signup information to be added to a block.
+
+Validator registry transaction would not be able to be done at the same time as
+any other transaction as an update to the ValidatorMap is necessary. However all
+other transaction that need to access the state set by the validator registry,
+can be done in parallel since it will only be a read and  the statics for each
+validator is stored in the poet_enclave. If this was changed so that the stats
+were stored in the validator registry this would require a write to state every
+time a block is published and would reduce the ability for parallelism.
+
+Addressing
+----------
+
+When a validator’s signup info is registered or updated it should be accessed
+using the following algorithm:
+
+Addresses for the validator registry transaction family are set by adding
+sha256 hash of the validator's id to the validator registry namespace. The
+namespace for the validator registry will be the first 6 characters of the
+sha256  hash of the string “validator_registry”, which is “6a4372” For example,
+the validator signup info of a validator with the id “validator_id” could be
+set like this:
+
+.. code-block:: pycon
+
+  >>>“6a4372” + hashlib.sha256('validator_id'.encode("utf-8")).hexdigest()
+  '6a43722aee5b550a3cbd1595f4de10049ee805bc035b5e232dfacfc31cc6275170b30d'
+
+The map of the current registered validator_id should be stored at the
+ following address:
+
+
+.. code-block:: pycon
+
+  >>>“6a4372” + hashlib.sha256('validator_map'.encode()).hexdigest()
+  '6a437247a1c12c0fb03aa6e242e6ce988d1cdc7fcc8c2a62ab3ab1202325d7d677e84c'
+
+Transaction Payload
+===================
+Validator registry transaction family payloads are defined by the following
+protocol buffers code:
+
+.. code-block:: protobuf
+
+  message ValidatorRegistryPayload {
+    // The action that the transaction processor will take. Currently this
+    // is only “register”, but could include other actions in the futures
+    // such as “revoke”
+    string verb = 1;
+
+    // The name of the endpoint. It should be a human readable name.
+    string name = 2;
+
+
+    // Validator's public key (currently using signer_pubkey as this is
+    // stored in the transaction header)
+    string id = 3;
+
+    // This is the protocol buffer described above.
+    SignUpInfo signup_info = 4;
+
+    // The block that this transaction is stored on.
+    // This is used as a form of time to help with the enforcement of the C test.
+    int32 block_num = 5;
+
+  }
+
+
+
+Transaction Header
+==================
+
+Inputs and Outputs
+------------------
+
+The inputs for validator registry family transactions must include:
+
+* the address of *validator_id*
+* the address of *validator_map*
+
+The outputs for validator registry family transactions must include:
+
+* the address of *validator_id*
+* the address of *validator_map*
+
+
+Dependencies
+------------
+
+None
+
+Family
+------
+- family_name: "sawtooth_validator"
+- family_version: "1.0"
+
+
+Encoding
+--------
+
+The encoding field must be set to "application/protobuf".
+
+
+Execution
+=========
+
+Untrusted python code, that is a part of the transaction processor, will verify
+the attestation verification report for the signup information. It is important to
+note, that the IAS report public key will need to be on the block chain and it will need to
+be set on configuration. This will allow both the simulator logic and real
+SGX logic to be the same.
+
+The validator_registry transaction will need to be injected into a block by
+poet_consensus. At this time, the block number for the block that poet_consensus
+is checking will be set into the transaction.
+
+The journal will need a new method to retrieve this information.
+
+If the validator_name does not match syntactic requirements, the transaction is
+invalid. The current requirement is that the validator_name is 64 characters
+or less.
+
+If the validator_id does not match the transaction signer, the transaction is
+invalid. The validator_id should be the same as the signer_pubkey stored in the
+transaction header.
+
+The signup info needs to be verified next. The signup info, public key hash, and
+the most recent wait_certificate_id are passed to verify the signup data.
+
+If any of the signup info checks fail verification, the validator_registry
+transaction is rejected and a invalid transaction response is returned.
+
+If the transaction is deemed to be valid, the validator_id is used to find
+the address where the validator_info should be stored. Store the serialized
+ValidatorInfo protocol buffer in state at the address as mentioned above. If
+this validator is new (not updating its SignUpInfo), the validator’s id needs
+to be added to the validator_map.
+
+For any previous entries with the anti_sybil_id, the public key is revoked before
+resetting the new information.
--- a/poet/sawtooth_poet/protos/validator_registry.proto
+++ b/poet/sawtooth_poet/protos/validator_registry.proto
@@ -1,0 +1,87 @@
+// Copyright 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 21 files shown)


docs/source/transaction_family_specifications/validator_registry_transaction_family.rst
```diff
@@ -0,0 +1,249 @@
...
+
+.. code-block:: protobuf
+
+  message ValidatorMap {
+    // anti_sybil_id, validator_id
+    map<string,string> validator_ids = 1;
+  }
+
+
+This could be used to check the number of validators registered. This
```
peterschwarz: could -> can

docs/source/transaction_family_specifications/validator_registry_transaction_family.rst
```diff
@@ -0,0 +1,249 @@
...
+Transaction Payload
+===================
+Validator registry transaction family payloads are defined by the following
+protocol buffers code:
+
+.. code-block:: protobuf
+
+  message ValidatorRegistryPayload {
+    // The action that the transaction processor will take. Currently this
+    //is only “register”, but could include other actions in the futures
```
peterschwarz: comment spacing

docs/source/transaction_family_specifications/validator_registry_transaction_family.rst
```diff
@@ -0,0 +1,249 @@
...
+
+Transaction Header
+==================
+
+Inputs and Outputs
+------------------
+
+The inputs for validator registry family transactions must include:
+
+* the address of *validator_map*
```
peterschwarz: Is the `validator_id` address should be included in the inputs.

poet/sawtooth_poet/validator_registry/processor/handler.py
```diff
@@ -0,0 +1,337 @@
...
+    def __init__(self, namespace_prefix):
+        self._report_private_key = \
+            signing.encode_privkey(
+                signing.decode_privkey(self.__REPORT_PRIVATE_KEY_WIF, 'wif'),
+                'hex')
+
+        self._report_public_key = signing.generate_pubkey(
+            self._report_private_key)
+
+        self._namespace_prefix = namespace_prefix
```
agunde406: Used now in def namespace

poet/sawtooth_poet/validator_registry/processor/handler.py
```diff
@@ -0,0 +1,334 @@
...
+from sawtooth_poet.protobuf.validator_registry_pb2 import \
+    ValidatorRegistryPayload
+
+LOGGER = logging.getLogger(__name__)
+
+STATE_TIMEOUT_SEC = 10
+
+
+def _get_address(key):
+    namespace = hashlib.sha256("validator_registry".encode()).hexdigest()[0:6]
```
peterschwarz: This can be a module-level constant - otherwise, you're computing this hash every `_get_address` call.

poet/sawtooth_poet/validator_registry/processor/handler.py
peterschwarz: If the namespace is a module level constant, you don't need this.  Really there's a disconnect between this being a parameter but never actually being used to compute addresses, which could cause a mismatch later down the road.

docs/source/transaction_family_specifications/validator_registry_transaction_family.rst
```diff
@@ -0,0 +1,257 @@
...
+    // The validator's public key (currently using signer_pubkey as this is
+    // stored in the transaction header)
+    string id = 3;
+
+    // This is the protocol buffer described below.
+    SignUpInfo signup_info = 4;
+
+    // The current chain head’s block number at the time of submission.
+    // This is used as a form to time to help with the enforcement of the C
+    // and K tests
```
jjason: I am not certain this helps with the K test as the K test uses the number of blocks a particular validator claims and that is mostly independent from the block number that contains the transaction used to add/update a validator's registry information.

docs/source/transaction_family_specifications/validator_registry_transaction_family.rst
```diff
@@ -0,0 +1,257 @@
...
+
+.. code-block:: protobuf
+
+  message SignUpInfo {
+    // Encoded public key corresponding to private key used by PoET to sign
+    // wait certificates.
+    string poet_public_key = 1;
+
+    // Information that can be used internally to verify the validity of
+    // the signup information. This is a JSON formatted string.
```
jjason: While that will probably always be true for IAS, I think that at this level it should simply be considered an opaque buffer.

docs/source/transaction_family_specifications/validator_registry_transaction_family.rst
```diff
@@ -0,0 +1,257 @@
...
+    string proof_data = 2;
+
+    // A string corresponding to the anti-Sybil ID for the enclave that
+    // generated the signup information.
+    string anti_sybil_id = 3;
+
+    // A base 64 string representing data that can be persisted and
+    // presented at a later time to restore the PoET enclave.
+    string sealed_signup_data = 4;
+  }
```
jjason: Is there a reason for the validator registry to store this?  Validators will/should not be passing around their sealed signup data.  Maybe my question will be answered later, but currently it is the poet_consensus module for PoET 1 that maintains this and upon startup will read it and give it to the poet_enclave to reset key state.

docs/source/transaction_family_specifications/validator_registry_transaction_family.rst
```diff
@@ -0,0 +1,257 @@
...
+sha256  hash of the string “validator_registry”, which is “6a4372” For example,
+the validator signup info of a validator with the id “validator_id” could be
+set like this:
+
+.. code-block:: pycon
+
+  >>>“6a4372” + hashlib.sha256('validator_id'.encode("utf-8")).hexdigest()
+  '6a43722aee5b550a3cbd1595f4de10049ee805bc035b5e232dfacfc31cc6275170b30d'
+
+The map of current the current registered validator_id should be stored at the
```
jjason: "The map of the current..."

docs/source/transaction_family_specifications/validator_registry_transaction_family.rst
```diff
@@ -0,0 +1,257 @@
...
+    // The action that the transaction processor will take. Currently this
+    // is only “register”, but could include other actions in the futures
+    // such as “revoke”
+  	string verb = 1;
+
+    // The name of the endpoint. It should be a human readable name.
+  	string name = 2;
+
+
+  	// validator's public key (currently using signer_pubkey as this is
```
jjason: Capitalize "validator's"

docs/source/transaction_family_specifications/validator_registry_transaction_family.rst
```diff
@@ -0,0 +1,257 @@
...
+Validator registry transaction family payloads are defined by the following
+protocol buffers code:
+
+.. code-block:: protobuf
+
+  message ValidatorRegistryPayload {
+    // The action that the transaction processor will take. Currently this
+    // is only “register”, but could include other actions in the futures
+    // such as “revoke”
+  	string verb = 1;
```
jjason: Inconsistent indenting below.

docs/source/transaction_family_specifications/validator_registry_transaction_family.rst
```diff
@@ -0,0 +1,257 @@
...
+  	// validator's public key (currently using signer_pubkey as this is
+    // stored in the transaction header)
+      string id = 3;
+
+    // This is the protocol buffer described above.
+    SignUpInfo signup_info = 4;
+
+    // The current chain head’s block num. This is used as a form of time
+    // to help with the enforcement of the C and K tests
+    int32 block_num = 5:
```
jjason: Same comment as above with regard to K.

Also, the transaction payload contains the block number?  Isn't this the block number when the transaction was created, not the block number that contains this transaction when the block was committed?  Technically the C policy uses the block number for when the validator information was added/updated via the transaction being included in a committed block.

docs/source/transaction_family_specifications/validator_registry_transaction_family.rst
```diff
@@ -0,0 +1,257 @@
...
+  	string name = 2;
+
+
+  	// validator's public key (currently using signer_pubkey as this is
+    // stored in the transaction header)
+      string id = 3;
+
+    // This is the protocol buffer described above.
+    SignUpInfo signup_info = 4;
+
```
jjason: Same comment as above with regard to the sealed signup data.  We don't want to be passing this around.

docs/source/transaction_family_specifications/validator_registry_transaction_family.rst
```diff
@@ -0,0 +1,257 @@
...
+The encoding field must be set to "application/protobuf".
+
+
+Execution
+=========
+
+Untrusted python code, that is a part of the transaction processor, will verify
+the attestation verification report for the signup information. It is important to
+note, that the IAS keys will need to be on the block chain and it will need to
```
jjason: "...that the IAS report public key..."

docs/source/transaction_family_specifications/validator_registry_transaction_family.rst
```diff
@@ -0,0 +1,257 @@
...
+be set on configuration. This will allow both the simulator logic and real
+SGX logic to be the same.
+
+Before the validator_registry transaction can be sent the SignUpInfo needs to be
+created by the poet_consensus and it  will need to request the current
+chain_head’s block number from the journal.
```
jjason: Same comment before with respect to the block number and the transaction.

docs/source/transaction_family_specifications/validator_registry_transaction_family.rst
```diff
@@ -0,0 +1,257 @@
...
+
+The journal will need a new method to retrieve this information.
+
+If the validator_name does not match syntactic requirements, the transaction is
+invalid. The current requirement is that the validator_name is 64 characters
+or less.
```
jjason: And presumably non-empty???  Although really that is just for human consumption, so maybe that doesn't matter.

docs/source/transaction_family_specifications/validator_registry_transaction_family.rst
```diff
@@ -0,0 +1,257 @@
...
+the most recent wait_certificate_id are passed to the enclave where the
+following checks are performed:
+
+- Verify the signature of the attestation verification report that is
+  stored in the signup_info.proof_data.
+- The verification report’s has a PSE manifest status and it must be OK.
+- PSE manifest hash exists and is value we expect.
+- The enclave quote is OK and contains a report body with the value we expect.
+- The wait_certificate_id stored in the nonce field matches the provided
+  wait_certificate_id.
```
jjason: Not certain if you need to provide this level of detail for the validator registry as I am not certain how static these tests are....

docs/source/transaction_family_specifications/validator_registry_transaction_family.rst
```diff
@@ -0,0 +1,257 @@
...
+If any of the signup info checks fail verification, the validator_registry
+transaction is rejected and a invalid transaction response is returned.
+
+If the transaction is deemed to be valid, the validator_id is used to find
+the address where the validator_info should be stored. Store the serialized
+ValidatorInfo protocol buffer in state at the address as mentioned above. If
+this validator is new (not updating its SignUpInfo), the validator’s id needs
+to be added to the validator_map.
+
+Any previous entries with the anti_sybil_id, the public key is revoked before
```
jjason: "For any ...."

poet/sawtooth_poet/protos/validator_registry.proto
```diff
@@ -0,0 +1,95 @@
...
+
+    // Validator's public key (currently using signer_pubkey as this is
+    // stored in the transaction header)
+    string id = 3;
+
+    // This is the protocol buffer described below. Need to figure out how
+    // this protobuf is created (currently uses poet_enclave_simulator)
+    SignUpInfo signup_info = 4;
+
+	  // The current chain head’s block number at the time of submission.
```
jjason: Inconsistent indent.

poet/sawtooth_poet/protos/validator_registry.proto
```diff
@@ -0,0 +1,95 @@
...
+    // This is used as a form to time to help with the enforcement of the C
+    // and K tests
+    int32 block_num = 5;
```
jjason: Same comment about K.

docs/source/transaction_family_specifications/validator_registry_transaction_family.rst
vaporos: Indenting here seems suspect - maybe there are tabs?

poet/sawtooth_poet/protos/validator_registry.proto
```diff
@@ -0,0 +1,95 @@
...
+
+
+message SignUpInfo {
+    // Encoded public key corresponding to private key used by PoET to sign
+    // wait certificates.
+    string poet_public_key = 1;
+
+    // Information that can be used internally to verify the validity of
+    // the signup information. This is a JSON formatted string.
+    string proof_data = 2;
```
jjason: Same comment about removing mention of JSON.

poet/sawtooth_poet/protos/validator_registry.proto
```diff
@@ -0,0 +1,95 @@
...
+
+    // A string corresponding to the anti-Sybil ID for the enclave that
+    // generated the signup information.
+    string anti_sybil_id = 3;
+
+    // A base 64 string representing data that can be persisted and
+    // presented at a later time to restore the PoET enclave.
+    string sealed_signup_data = 4;
+}
```
jjason: Same comment about sealed signup data.

poet/sawtooth_poet/tests/validator_registry_test/tests.py
```diff
@@ -0,0 +1,436 @@
...
+                    'wif'), 'hex')
+
+    def _expect_invalid_transaction(self):
+        self.tester.expect(
+            self.factory.create_tp_response("INVALID_TRANSACTION"))
+
+    def _expect_ok(self):
+        self.tester.expect(self.factory.create_tp_response("OK"))
+
+    def test_valid_signup_info(self):
```
vaporos: test methods should have API doc.

poet/sawtooth_poet/validator_registry/processor/handler.py
```diff
@@ -0,0 +1,333 @@
...
+    if len(entries_list) != 0:
+        validator_state.ParseFromString(entries_list[0].data)
+
+    return validator_state
+
+
+def _update_validator_state(state, validator_id, anti_sybil_id):
+    validator_map = _get_validator_state(state)
+    for entry in validator_map.entries:
+        if anti_sybil_id == entry.key:
```
jjason: Does it make any sense to have a reverse lookup for anti_sybil_id or is the size of the validator map going to be small enough that linearly searching is sufficient?

poet/sawtooth_poet/validator_registry/processor/handler.py
```diff
@@ -0,0 +1,333 @@
...
+            validator_info = _get_validator_state(state, validator_id)
+
+            validator_info.registered = "revoked"
+            validator_info = validator_info.SerializeToString()
+            _set_validator_info(state, validator_id, validator_info)
+        return
+
+    # This is a new validator that needs to be added to the map
+    validator_map.entries.add(key=anti_sybil_id, value=validator_id)
+    address = _get_address('validator_map')
```
jjason: Can the code below that sets the map entry be refactored with the _set_validator_info method so that it is not repeated?

poet/sawtooth_poet/validator_registry/processor/handler.py
```diff
@@ -0,0 +1,333 @@
...
+        raise InternalError(
+            'Unable to save validator info for {}'.format(validator_id))
+
+    LOGGER.info("Validator info was set for %s", validator_id)
+
+
+class ValidatorRegistryTransactionHandler(object):
+
+    __REPORT_PRIVATE_KEY_WIF = \
+        '5Jz5Kaiy3kCiHE537uXcQnJuiNJshf2bZZn43CrALMGoCd3zRuo'
```
jjason: This will only ever work with the simulator.  This private key is tied to simulated PoET1 enclave.  The real IAS will use different key to sign the report.

poet/sawtooth_poet/validator_registry/processor/handler.py
```diff
@@ -0,0 +1,333 @@
...
+
+    @property
+    def encodings(self):
+        return ['application/protobuf']
+
+    @property
+    def namespaces(self):
+        return [VAL_REG_NAMESPACE]
+
+    def verify_signup_info(self,
```
jjason: Why is the validator registry verifying the signup information and not simply giving it to the poet_enclave to do it?

docs/source/transaction_family_specifications/validator_registry_transaction_family.rst
agunde406: This is not the sealed signup data, It is the SignUpInfo protobuf, so if it is removed from SingUpInfo it would not be here.

docs/source/transaction_family_specifications/validator_registry_transaction_family.rst
agunde406: The legacy validator registry counted "" as valid.

poet/sawtooth_poet/tests/validator_registry_test/tests.py
agunde406: Added

poet/sawtooth_poet/validator_registry/processor/handler.py
agunde406: This is temporary until we can set the IAS information on the block chain.

poet/sawtooth_poet/validator_registry/processor/handler.py
agunde406: Currently we have no good way to have the transaction processor accesses the poet_enclave and since this code is untrusted python code, it can be done here.

PR #98, opened 2017-02-17
Consensus interface

cintel:
I still have a few more updates to the Consensus interface to support Poet1, these are mainly around injecting transactions in to the pending block. Which should not affect the consensus objects and where/how they are constructed.

PR #105, opened 2017-02-20
Generate Validator Signing Key and Use for Genesis Block

peterschwarz:
This adds the `admin keygen` subcommand, which produces a key file in the validator's `keys` directory.

The validator loads the key and passes it to the `GenesisController`, for use when signing the genesis block.

This key will most likely, be used by journal, as well, to sign blocks, after #98 is merged.

### DIFF
--- a/cli/sawtooth_cli/admin.py
+++ b/cli/sawtooth_cli/admin.py
@@ -40,7 +40,7 @@ def do_admin(args):
 
 
 def add_admin_parser(subparsers, parent_parser):
-    parser = subparsers.add_parser('admin')
+    parser = subparsers.add_parser('admin', parents=[parent_parser])
     admin_sub = parser.add_subparsers(title='admin_commands', dest='admin_cmd')
 
     add_genesis_parser(admin_sub, parser)
--- a/cli/sawtooth_cli/admin.py
+++ b/cli/sawtooth_cli/admin.py
@@ -12,35 +12,30 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ------------------------------------------------------------------------------
-import os
-import sys
 
 from sawtooth_cli.exceptions import CliException
 
 from sawtooth_cli.admin_command.genesis import add_genesis_parser
 from sawtooth_cli.admin_command.genesis import do_genesis
+from sawtooth_cli.admin_command.keygen import add_keygen_parser
+from sawtooth_cli.admin_command.keygen import do_keygen
+from sawtooth_cli.admin_command.utils import ensure_directory
 
 
 def do_admin(args):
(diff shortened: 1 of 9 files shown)


cli/sawtooth_cli/admin.py
```diff
@@ -19,13 +19,15 @@
...
+from sawtooth_cli.admin_command.keygen import add_keygen_parser
+from sawtooth_cli.admin_command.keygen import do_keygen
 
 
 def do_admin(args):
     if 'SAWTOOTH_HOME' in os.environ:
         data_dir = os.path.join(os.environ['SAWTOOTH_HOME'], 'data')
     else:
-        data_dir = '/var/lib/sawtooth'
+        data_dir = os.path.join(os.sep, 'var', 'lib', 'sawtooth')
```
vaporos: The assumption that the os separator matters here is incorrect.  "/var/lib/sawtooth" is only valid for Linux distributions, not Windows.

cli/sawtooth_cli/admin_command/keygen.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    if args.key_name is not None:
+        key_name = args.key_name
+    else:
+        key_name = 'validator'
+
+    if 'SAWTOOTH_HOME' in os.environ:
+        key_dir = os.path.join(os.environ['SAWTOOTH_HOME'], 'keys')
+    else:
+        key_dir = os.path.join(os.sep, 'etc', 'sawtooth', 'keys')
```
vaporos: This should not use os.sep for same reason as above (it isn't valid on Windows).

validator/sawtooth_validator/server/cli.py
```diff
@@ -47,27 +47,56 @@ def parse_args(args):
     return parser.parse_args(args)
 
 
+def ensure_directory(sawtooth_home_path, fallback_path):
```
vaporos: This assumes a single default path, which isn't accurate as the windows default will differ from the Linux default.

vaporos:
With respect to Windows default path support...

In 0.7, we used a fixed path in Program Files.  In 0.8, we should perhaps set the default relative to the executable -- this would work similar to setting SAWTOOTH_HOME to the directory above bin/, though the subdirectories for 0.7 where not the same as when using SAWTOOTH_HOME. (conf instead of etc for example).  For reference on the previous approach:

https://github.com/hyperledger/sawtooth-core/blob/0-7/core/sawtooth/validator_config.py#L242

validator/sawtooth_validator/server/cli.py
peterschwarz: See the update code changes for what I've done here.

PR #106, opened 2017-02-20
Pass proxy info into docker container.

aludvik:


cintel:
@aludvik Any problem if we have jjason, tbarnes or nconde(other intel folks that may be affected sign off) on this?

aludvik:
No, I will add them as reviewers.

PR #108, opened 2017-02-21
Remove legacy code, scripts, and dependencies

vaporos:


peterschwarz:
Separating the code and unit test removals means if one rolls back to just before the unit tests are removed, the build is broken.  Unit tests are dependent on the production code, and really should be removed before or at the same time as the removal of the relevant production code.

vaporos:
Re-ordered commits and fixed linting issue (pep8/pylint3 were not installed in previous attempt).  Also attempted to fix the build issue by removing more from the dockerfile.

vaporos:
Added pep8 to docker/sawtooth-build-ubuntu-xenial...

vaporos:
Updated to keep installing pylint via pip3 instead of apt-get in vagrant, since the apt-get version is broken.

peterschwarz:
It's looking like the PoET consensus code is still dependent on the legacy gossip package.  Not sure if that cleanup should be part of this PR or not.

vaporos:
The PoET code integration is still in-progress and not expected to work in the scope of this PR.  The PoET code move was done in a different PR and (some of) the tests are currently disabled pending completion of that work.

docs/README.md
```diff
@@ -1,32 +0,0 @@
...
-community and documented [here](http://bit.ly/1T6eVBH). Information on what 
-_Incubation_ entails can be found in the [Hyperledger Project Lifecycle document](https://goo.gl/4edNRc).
-
-Distributed Ledger
-==================
-
-Distributed Ledger is a highly modular platform for building, deploying and
-running distributed ledgers.
-
-Security Notice
```
TomBarnes: Security notice is important - what replaces this in the docs?

validator/etc/txnvalidator.js.example
```diff
@@ -1,67 +0,0 @@
-{
```
TomBarnes: what replaces txnvalidator.js.example?  If multiple files (which I'm wary of), how does a tester or deployer determine what files need to be modified and where to find them all?

peterschwarz: At the moment, nothing replaces it.

docs/README.md
vaporos: Duplicate of top-level README.md

PR #109, opened 2017-02-22
Break PoET1 dependencies on 0.7 gossip.

cintel:

Goals of this PR is to break any 0.7 dependences and address any other Python3 issues that were uncovered by PyLint.

### DIFF
--- a/validator/sawtooth_validator/journal/consensus/poet1/poet_consensus.py
+++ b/validator/sawtooth_validator/journal/consensus/poet1/poet_consensus.py
@@ -20,12 +20,9 @@ import hashlib
 import random
 import math
 
-from gossip import common
-from gossip import stats
 from sawtooth_signing import pbct as signing
+from sawtooth_validator.journal.block_wrapper import NULL_BLOCK_IDENTIFIER
 from sawtooth_validator.journal.consensus.poet1 import poet_transaction_block
-from sawtooth_validator.journal.consensus.poet1\
-    import validator_registry as val_reg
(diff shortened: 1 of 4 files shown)


vaporos:
Suggest we split this into two different PRs and use this PR for reference.  One immediate one that fixes pylint to not exclude the consensus directory (might be more involved than what's just in this PR, since it is excluding more than intended), and then another that introduces the new validator registry state view.

I can pull the relevant lint-related pieces together into a PR.  @agunde406 can comment on the amount of work to introduce the view.

agunde406:
The view for the Validator Registry has not been written yet.

cintel:
The validator_registry was removed in the previous move commit. All changes in this commit are fixing lint errors stemming from this. As stated in the commit messages this is a cleanup only commit not a porting commit. I will not grow the scope of this PR to include PoET1 porting to the 0.8 consensus interface.

peterschwarz:
Writing a StateView for the registry should be straight forward.  Submitting the transactions for the registry might be a bit more work, as I don't know if the facilities for sending a txn exists from within the validator.

cintel:
It is pointless to add the StateView to this commit. None of the porting activities of moving PoET1 to the new consensus activities are done in this PR nor will this commit be expanded to encompass them. 

As it seems cleanup lint errors and incorrect references are not desired. I will close this PR.

vaporos:
Sorry for any confusion, we were suggesting two PRs (one attempting to solve the larger linting issue and another to address the view).  Was originally hoping if we coordinated it right, we could avoid the commented-out-code, which isn't actually practical.

This commit is included as part of #112 .

PR #111, opened 2017-02-22
Bug fix for transaction processors maxing cpu on validator disconnect

boydjohnson:
The fix was disabling the monitor socket right after it receives a disconnect event and disconnecting but not closing the external socket on validator disconnect. After this, the validator can stop and restart with no ill effects on the transaction processor.

Also, the hard-coded "tcp://" is removed from the Stream python class as we might want to support other transport protocols. Then main modules used in bins are updated to specify tcp, as well as docker files and the docker manage library.

### DIFF
--- a/docker/sawtooth-tp_xo_python
+++ b/docker/sawtooth-tp_xo_python
@@ -21,4 +21,4 @@ ARG PROJECT_DIR="/project/sawtooth-core"
 
 WORKDIR $PROJECT_DIR
 
-ENTRYPOINT ./bin/tp_xo_python validator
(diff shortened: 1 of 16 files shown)


docker/sawtooth-tp_xo_python
```diff
@@ -21,4 +21,4 @@ ARG PROJECT_DIR="/project/sawtooth-core"
 
 WORKDIR $PROJECT_DIR
 
-ENTRYPOINT ./bin/tp_xo_python validator
+ENTRYPOINT ./bin/tp_xo_python tcp://validator
```
agunde406: This seems to be missing ":40000" that all others have.

PR #112, opened 2017-02-23
Increase lint coverage

vaporos:


vaporos:
The first commit here has been modified from the version in #109, with the two commits immediately after it resolving the a few of the linting issues in a slightly different manner.

validator/sawtooth_validator/journal/consensus/poet1/poet_transaction_block.py
```diff
@@ -158,7 +165,7 @@ def update_block_weight(self, journal):
             assert self.wait_certificate
             self.aggregate_local_mean = self.wait_certificate.local_mean
 
-            if self.PreviousBlockID != NullIdentifier:
+            if self.PreviousBlockID != NULL_BLOCK_IDENTIFIER:
```
peterschwarz: Any reason why this doesn't need to be lower-cased?

validator/sawtooth_validator/journal/consensus/poet1/wait_certificate.py
```diff
@@ -14,7 +14,7 @@
 # ------------------------------------------------------------------------------
 
 import logging
-from requests import ConnectionError
+from requests import ConnectionError as RequestsConnectionError
```
peterschwarz: Instead renaming, just reference the exception via `requests.ConnectionError`

vaporos:
Commit d07a6ed is broken.  I'll work on a replacement.

Also, cool that the smoke test caught it.

validator/sawtooth_validator/journal/consensus/poet1/poet_transaction_block.py
peterschwarz: And by "this" I mean `PreviousBlockId`

vaporos: It declared at the module level (similar to LOGGER), and pep8 specifies those follow all-caps.

peterschwarz: See my previous reply - I meant `PreviousBuildID`, not `NULL_BLOCK_IDENTIFIER`

cintel: PreviousBlockID will be converted to lower case when this module is converted to use the 0.8 protobuf block definition.

validator/sawtooth_validator/journal/consensus/poet1/wait_certificate.py
vaporos: This has been updated.

vaporos:
The super.__init__() build issue has been fixed.

PR #113, opened 2017-02-23
Make intkey batch creation insert proper txn deps

jsmitchell:
This makes intkey batch creation set dependencies in the
inc/dec transactions for the corresponding set transaction.

### DIFF
--- a/sdk/examples/intkey_python/sawtooth_intkey/client_cli/create_batch.py
+++ b/sdk/examples/intkey_python/sawtooth_intkey/client_cli/create_batch.py
@@ -59,7 +59,26 @@ class IntKeyPayload(object):
         return self._sha512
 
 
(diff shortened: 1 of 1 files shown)


sdk/examples/intkey_python/sawtooth_intkey/client_cli/create_batch.py
```diff
@@ -59,7 +59,8 @@ def sha512(self):
         return self._sha512
 
 
-def create_intkey_transaction(verb, name, value, private_key, public_key):
+def create_intkey_transaction(verb, name, value, deps,
```
peterschwarz: Since you're modifying the function signature, could you add some doc string here?

jsmitchell: Docstring added in squashed commit

PR #115, opened 2017-02-23
Implement executor support for TP list from config txn family

boydjohnson:
To require a transaction processor, sawtooth config --key <your key> 'sawtooth.validator.transaction_families=[{"family": "intkey", "version": "1.0", "encoding":"application/cbor"}]', and then either "intkey load -f config.batch" on an already running validator or give the config.batch file to the genesis util: sawtooth admin genesis config.batch, and then start the validator.

If there are no required transaction processors, the executor will wait for transaction processors for all transactions (This is useful on startup). If there are required transaction processors and a transaction comes in that is not on the list, the Executor will immediately fail it.


validator/sawtooth_validator/execution/executor.py
```diff
@@ -13,29 +13,41 @@
...
 
 class TransactionExecutorThread(threading.Thread):
-    def __init__(self, service, context_manager, scheduler, processors,
-                 require_txn_processors=False):
+    def __init__(self,
+                 service,
+                 context_manager,
+                 scheduler,
+                 processors,
+                 state_database):
```
peterschwarz: Don't pass the state database, base the config_view_factory.  We want to limit components to read access unless they really need it.

validator/sawtooth_validator/execution/executor.py
```diff
@@ -77,49 +89,121 @@ def run(self):
...
+
+            # If there are required transaction processors, sawtooth_config
+            # should be one of them, so the TransactionExecutor doesn't
+            # automatically fail config transactions.
+            if len(required_transaction_processors) > 0 and config_processor \
+                    not in required_transaction_processors:
+                required_transaction_processors += [config_processor]
+
+            if processor_type not in required_transaction_processors:
+                if len(required_transaction_processors) == 0:
```
peterschwarz: So if no required processors are configured, all are required?

validator/sawtooth_validator/execution/executor.py
peterschwarz: Also, since you're changing the signature, can you doc string this?

validator/sawtooth_validator/execution/executor.py
```diff
@@ -77,49 +89,121 @@ def run(self):
...
+            validator_pb2.Message.TP_PROCESS_REQUEST,
+            content,
+            identity=identity,
+            has_callback=True)
+        future.add_callback(self._future_done_callback)
 
 
 class TransactionExecutor(object):
-    def __init__(self, service, context_manager):
+    def __init__(self, service, context_manager, state_database):
```
peterschwarz: Again, pass the `config_view_factory`, not the database.

validator/sawtooth_validator/journal/genesis.py
```diff
@@ -135,8 +135,7 @@ def start(self, on_done):
             for batch in genesis_data.batches:
                 scheduler.add_batch(batch)
 
-            self._transaction_executor.execute(scheduler,
-                                               require_txn_processors=True)
+            self._transaction_executor.execute(scheduler)
```
peterschwarz: This change should be part of the previous commit where this signature changed.  Separately, the previous commit is broken.

boydjohnson: I squashed it with the previous commit, now.

validator/sawtooth_validator/execution/executor.py
```diff
@@ -77,47 +102,188 @@ def run(self):
...
+                               "misconfigured. Expecting a json array, found"
+                               " %s", transaction_families)
+                required_transaction_processors = []
+
+            config_processor = processor_iterator.ProcessorType(
+                name='sawtooth_config',
+                version='1.0',
+                encoding='application/protobuf')
+
+            # If there are required transaction processors, sawtooth_config
```
vaporos: The config transaction family should be explicitly in the list once it is set.  Imagine that during the lifetime of the chain we go from sawtooth_config 1.0 to sawtooth_config 1.1 and no longer want to accept 1.0 transactions.  Or, alternatively, another implementation of a config txn family is being used entirely.

validator/sawtooth_validator/execution/executor.py
```diff
@@ -28,14 +31,36 @@
 
 
 class TransactionExecutorThread(threading.Thread):
-    def __init__(self, service, context_manager, scheduler, processors,
-                 require_txn_processors=False):
+    """A thread of execution controlled by the TransactionExecutor.
+        Provides the functionality that the journal can process on several
```
vaporos: Indent is a bit odd here.

validator/sawtooth_validator/execution/executor.py
```diff
@@ -77,47 +102,188 @@ def run(self):
...
+
+            if processor_type not in required_transaction_processors:
+                # If there aren't any configured required transaction
+                # processors, wait for all transactions. This is useful
+                # for startup with the assumption being that the
+                # required transaction processors would be configured soon
+                # after startup.
+                if len(required_transaction_processors) == 0:
+                    # There are no required txn processors
+                    LOGGER.debug("there are no required transaction "
```
vaporos: We could log the processor_type we are looking for and the txn id.

validator/sawtooth_validator/execution/executor.py
```diff
@@ -77,47 +102,188 @@ def run(self):
...
+                                 " required in the configuration",
+                                 processor_type.name,
+                                 processor_type.version,
+                                 processor_type.encoding)
+                    self._context_manager.delete_context(
+                        context_id_list=[context_id])
+                    self._scheduler.set_transaction_execution_result(
+                        txn_signature=txn.header_signature,
+                        is_valid=False,
+                        context_id=context_id)
```
vaporos: Should context_id be None when is_valid is False?

validator/sawtooth_validator/execution/executor.py
```diff
@@ -77,47 +102,188 @@ def run(self):
...
+                    LOGGER.debug("there are no required transaction "
+                                 "processors")
+                    self._execute_or_wait_for_processor_type(
+                        processor_type=processor_type,
+                        content=content)
+                else:
+                    # The txn processor type is not in the required
+                    # transaction processors so
+                    # failing transaction right away
+                    LOGGER.debug("failing transaction of type (name=%s,"
```
vaporos: We could log the transaction id here as well.

validator/sawtooth_validator/execution/executor.py
```diff
@@ -77,47 +102,188 @@ def run(self):
...
+                    not in required_transaction_processors:
+                required_transaction_processors += [config_processor]
+
+            if processor_type not in required_transaction_processors:
+                # If there aren't any configured required transaction
+                # processors, wait for all transactions. This is useful
+                # for startup with the assumption being that the
+                # required transaction processors would be configured soon
+                # after startup.
+                if len(required_transaction_processors) == 0:
```
vaporos: If required_transaction_processors is of length 0, then that should be interpreted as "any".  This allows us to avoid explicit configuration of the sawtooth.validator.transaction_families setting at startup.

validator/sawtooth_validator/execution/executor.py
```diff
@@ -77,47 +102,188 @@ def run(self):
...
+                    # The txn processor type is not in the required
+                    # transaction processors so
+                    # failing transaction right away
+                    LOGGER.debug("failing transaction of type (name=%s,"
+                                 "version=%s,encoding=%s) since it isn't"
+                                 " required in the configuration",
+                                 processor_type.name,
+                                 processor_type.version,
+                                 processor_type.encoding)
+                    self._context_manager.delete_context(
```
vaporos: As an optimization, the context creation could have been skipped in the case that we aren't intending to execute  the transaction.

validator/sawtooth_validator/execution/executor.py
```diff
@@ -77,47 +102,188 @@ def run(self):
...
+            if processor_type not in self._waiters_by_type:
+                in_queue = queue.Queue()
+                in_queue.put_nowait(content)
+                waiter = _Waiter(self._send_and_process_result,
+                                 processor_type=processor_type,
+                                 processors=self._processors,
+                                 in_queue=in_queue,
+                                 waiters_by_type=self._waiters_by_type)
+                self._waiters_by_type[processor_type] = waiter
+                waiter.start()
```
vaporos: All Thread start()s must have an appropriate matching join() at the right time.  This might be a good reason to use a thread pool here.

validator/sawtooth_validator/execution/executor.py
boydjohnson: Since config should be explicitly set, it seems like there needs to be a check that there is some transaction processor with the capabilities to process config transactions. The validator is misconfigured otherwise. I agree that checking for a specific name, version, encoding isn't right, when for the lifetime of a chain there could be different config processors. Could we require that any config processor has "config" in the name? I think we are inviting bad situations where, after a mistake in changing required transaction processors, we fail every config transaction and aren't able to make another config change.

PR #117, opened 2017-02-23
Fix error handling of sawtooth cli

ojalatodd:
Sawtooth cli error handling fixed.

### DIFF
--- a/cli/sawtooth_cli/admin.py
+++ b/cli/sawtooth_cli/admin.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 # ------------------------------------------------------------------------------
 
-from sawtooth_cli.exceptions import CliException
 
 from sawtooth_cli.admin_command.genesis import add_genesis_parser
 from sawtooth_cli.admin_command.genesis import do_genesis
@@ -30,12 +29,12 @@ def do_admin(args):
     elif args.admin_cmd == 'keygen':
         do_keygen(args)
     else:
-        raise CliException("invalid command: {}".format(args.admin_cmd))
+        raise AssertionError("invalid command: {}".format(args.admin_cmd))
 
 
 def add_admin_parser(subparsers, parent_parser):
     parser = subparsers.add_parser('admin', parents=[parent_parser])
     admin_sub = parser.add_subparsers(title='admin_commands', dest='admin_cmd')
-
+    admin_sub.required = True
     add_genesis_parser(admin_sub, parent_parser)
     add_keygen_parser(admin_sub, parent_parser)
--- a/cli/sawtooth_cli/block.py
+++ b/cli/sawtooth_cli/block.py
@@ -29,7 +29,7 @@ def add_block_parser(subparsers, parent_parser):
 
     grand_parsers = parser.add_subparsers(title='grandchildcommands',
                                           dest='subcommand')
-
+    grand_parsers.required = True
     epilog = '''
     details:
         Lists committed blocks from the newest to the oldest, including
@@ -46,6 +46,7 @@ def add_block_parser(subparsers, parent_parser):
         '--format',
         action='store',
         default='default',
+        choices=['csv', 'json', 'yaml', 'default'],
         help='the format of the output, options: csv, json or yaml')
 
     epilog = '''
@@ -61,7 +62,7 @@ def add_block_parser(subparsers, parent_parser):
     show_parser.add_argument(
         '-k', '--key',
         type=str,
-        help='specficy to show a single property from the block or header')
+        help='specify to show a single property from the block or header')
     show_parser.add_argument(
         '--url',
         type=str,
@@ -70,10 +71,12 @@ def add_block_parser(subparsers, parent_parser):
         '--format',
         action='store',
         default='yaml',
+        choices=['yaml', 'json'],
         help='the format of the output, options: yaml (default), or json')
 
 
 def do_block(args):
+
     rest_client = RestClient(args.url)
 
     def print_json(data):
@@ -134,7 +137,7 @@ def do_block(args):
                 print_yaml(block_data)
 
         else:
-            raise CliException('unknown format: {}'.format(args.format))
+            raise AssertionError('unknown format: {}'.format(args.format))
 
     if args.subcommand == 'show':
         block = rest_client.get_block(args.block_id)
@@ -153,4 +156,4 @@ def do_block(args):
             elif args.format == 'json':
                 print_json(block)
             else:
-                raise CliException('unknown format: {}'.format(args.format))
+                raise AssertionError('unknown format: {}'.format(args.format))
--- a/cli/sawtooth_cli/main.py
+++ b/cli/sawtooth_cli/main.py
@@ -90,7 +90,7 @@ def create_parser(prog_name):
         formatter_class=argparse.RawDescriptionHelpFormatter)
 
     subparsers = parser.add_subparsers(title='subcommands', dest='command')
-
+    subparsers.required = True
     add_keygen_parser(subparsers, parent_parser)
     add_admin_parser(subparsers, parent_parser)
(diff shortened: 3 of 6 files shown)


ojalatodd:
Part of NML-2010

cli/sawtooth_cli/main.py
```diff
@@ -126,7 +126,7 @@ def main(prog_name=os.path.basename(sys.argv[0]), args=sys.argv[1:],
     elif args.command == 'cluster':
         do_cluster(args)
     else:
-        raise AssertionError("invalid command: {}".format(args.command))
```
peterschwarz: This is an AssertionError, because if we get to this point, `argparse` is busted.

peterschwarz: In other words, I don't think this change is necessary.

ojalatodd: Thanks. Fixing up this morning. There is still an issue with error handling, based on testing. I'll iron this out today and resubmit.

ojalatodd: I've made more changes than I expected, to clean up the user experience of the sawtooth command. Lots of commands/subcommands produced ungraceful errors. These have been fixed.

cli/sawtooth_cli/admin.py
```diff
@@ -22,15 +21,19 @@
 from sawtooth_cli.admin_command.utils import ensure_directory
 
 
-def do_admin(args):
+def do_admin(args, parser):
     data_dir = ensure_directory('data', '/var/lib/sawtooth')
 
+    if args.admin_cmd is None:
```
peterschwarz: I think this also will be handled by arg parse.

ojalatodd: I tested it. This condition does not get caught by argparse. At least that is what I determined empirically.

ojalatodd: Addendum: Argparse handles an argument of None, but doesn't provide any useful info or help. It just provides an error.

cli/sawtooth_cli/main.py
```diff
@@ -133,6 +137,9 @@ def main_wrapper():
     # pylint: disable=bare-except
     try:
         main()
+    except AssertionError as e:
```
vaporos: AssertionError is used to indicate a code bug, and thus it should fall through to the bare except which prints out the stack trace.  It should be impossible for users to cause this.

cli/sawtooth_cli/admin.py
vaporos: We should look into this a bit more, maybe we are using argparse incorrectly.

ojalatodd: Well, what I have done is to make the user experience better. Instead of getting an error about the argument "None" being not supported, it spits out help on the subcommand. 
I thought this was a good solution, and found similar examples during my research.

ojalatodd: @vaporos Should I close this PR, so that we can consider this issue more?
My changes prevent a bunch of exceptions or errors when subcommands are left out. Maybe some of the changes can be implemented, or this should be a new Jira that involves discussion and then a more  considered change.

cli/sawtooth_cli/main.py
ojalatodd: @vaporos I believe I resolved this. Main wrapper now does not catch the AssertionError.

ojalatodd:
@vaporos and @peterschwarz  Current state of code provides better user experience. We can either just close this PR and discuss in future, or merge this (if it looks OK, or after another round), and have this better user experience right away.

vaporos:
This is a python 2->3 regression due to changed behavior in argparse.  See:

http://stackoverflow.com/questions/22990977/why-does-this-argparse-code-behave-differently-between-python-2-and-3

If we took a similar approach, the fix would look something like:

```
diff --git a/cli/sawtooth_cli/admin.py b/cli/sawtooth_cli/admin.py
index 4ed48fc..23dbe16 100644
--- a/cli/sawtooth_cli/admin.py
+++ b/cli/sawtooth_cli/admin.py
@@ -36,6 +36,7 @@ def do_admin(args):
 def add_admin_parser(subparsers, parent_parser):
     parser = subparsers.add_parser('admin', parents=[parent_parser])
     admin_sub = parser.add_subparsers(title='admin_commands', dest='admin_cmd')
+    admin_sub.required = True
 
     add_genesis_parser(admin_sub, parent_parser)
     add_keygen_parser(admin_sub, parent_parser)
diff --git a/cli/sawtooth_cli/main.py b/cli/sawtooth_cli/main.py
index 60c7855..84175c2 100644
--- a/cli/sawtooth_cli/main.py
+++ b/cli/sawtooth_cli/main.py
@@ -90,6 +90,7 @@ def create_parser(prog_name):
 
     subparsers = parser.add_subparsers(title='subcommands', dest='command')
+    subparsers.required = True
 
     add_keygen_parser(subparsers, parent_parser)
     add_admin_parser(subparsers, parent_parser)
```

ojalatodd:
@vaporos  Thanks for review above. Are you saying that I should keep this PR open and apply this?

ojalatodd:
@vaporos I rewrote this PR based on the tip you gave me above. 
Tested.

ojalatodd:
@peterschwarz I changed this quite a bit based on the fact that Python3 subparsers have a "required" attribute you can set. Ready for a new review!

cli/sawtooth_cli/block.py
```diff
@@ -134,7 +135,7 @@ def get_block_data(block):
                 print_yaml(block_data)
 
         else:
-            raise CliException('unknown format: {}'.format(args.format))
+            raise AssertionError('unknown format: {}'.format(args.format))
```
peterschwarz: This actually is a `CliException`,  since we're not enforcing it via argparse.  You can either add `choices=['json', 'yaml', 'csv']` to the `--format` definition, and keep the `AssertionError`, or change this back to a `CliException`.

cli/sawtooth_cli/block.py
```diff
@@ -153,4 +154,4 @@ def get_block_data(block):
             elif args.format == 'json':
                 print_json(block)
             else:
-                raise CliException('unknown format: {}'.format(args.format))
+                raise AssertionError('unknown format: {}'.format(args.format))
```
peterschwarz: Same as above comment.

cli/sawtooth_cli/block.py
ojalatodd: @peterschwarz Done. I used the 'choices' option.

cli/sawtooth_cli/block.py
ojalatodd: @peterschwarz Done. I used the 'choices' option.

cli/sawtooth_cli/block.py
```diff
@@ -46,6 +46,7 @@ def add_block_parser(subparsers, parent_parser):
         '--format',
         action='store',
         default='default',
+        choices=['csv', 'json', 'yaml'],
```
peterschwarz: I think `'default'` also needs to be a choice here.

cli/sawtooth_cli/state.py
```diff
@@ -79,6 +79,11 @@ def add_state_parser(subparsers, parent_parser):
 
 
 def do_state(args):
+
+    # if args.subcommand is None:
```
vaporos: Let's remove the commented-out code.

cli/sawtooth_cli/block.py
ojalatodd: It worked without it, because to trigger 'default', user simply left off the flag/option altogether. However, putting it in.

cli/sawtooth_cli/state.py
ojalatodd: Missed that. Removed!

PR #118, opened 2017-02-23
Determine and use Consensus Module using On-chain Settings 

peterschwarz:
During genesis block construction, the consensus module can be loaded based on the settings supplied by the sawtooth_config transaction family.  The module is then used to correctly initialize and finalize the block before committing it to the block store.

The consensus module is also before it is used during Journal's processing of blocks.

### DIFF
--- a/validator/sawtooth_validator/journal/consensus/consensus_factory.py
+++ b/validator/sawtooth_validator/journal/consensus/consensus_factory.py
@@ -1,0 +1,47 @@
+# Copyright 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+import importlib
+
+from sawtooth_validator.exceptions import UnknownConsensusModuleError
+
+
+class ConsensusFactory(object):
+    """ConsensusFactory returns consensus modules by short name.
+    """
+
+    @staticmethod
+    def get_consensus_module(module_name):
+        """Returns a consensus module by name.
+
+        Args:
+            module_name (str): The name of the module to load.
+
+        Returns:
+            module: The consensus module.
+
+        Raises:
+            UnknownConsensusModuleError: Raised if the given module_name does
+                not correspond to a consensus implementation.
+        """
+        if module_name == 'devmode':
+            return importlib.import_module(
+                'sawtooth_validator.journal.consensus.dev_mode.'
+                'dev_mode_consensus')
+        elif module_name == 'poet1':
+            return importlib.import_module(
+                'sawtooth_validator.journal.consensus.poet1.poet_consensus')
+        else:
+            raise UnknownConsensusModuleError(
+                'Consensus module "{}" does not exist.'.format(module_name))
--- a/validator/sawtooth_validator/journal/consensus/consensus_factory.py
+++ b/validator/sawtooth_validator/journal/consensus/consensus_factory.py
@@ -15,6 +15,7 @@
 import importlib
 
 from sawtooth_validator.exceptions import UnknownConsensusModuleError
+from sawtooth_validator.state.config_view import ConfigView
 
 
 class ConsensusFactory(object):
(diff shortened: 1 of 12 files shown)


validator/sawtooth_validator/journal/consensus/consensus_factory.py
```diff
@@ -0,0 +1,47 @@
...
+
+        Raises:
+            UnknownConsensusModuleError: Raised if the given module_name does
+                not correspond to a consensus implementation.
+        """
+        if module_name == 'devmode':
+            return importlib.import_module(
+                'sawtooth_validator.journal.consensus.dev_mode.'
+                'dev_mode_consensus')
+        elif module_name == 'poet1':
```
jjason: Not necessarily directly a responsibility of this PR, but since we don't have two versions of PoET any more, do we want to ditch "poet1" and go back to "poet" for both the module name above as well as the module path in line 44 below (meaning also change the directory name in the repo)????

cintel: If you want to do the rename, now is the time. I don't have a strong feeling.

validator/sawtooth_validator/server/core.py
```diff
@@ -187,7 +188,7 @@ def __init__(self, network_endpoint, component_endpoint, peer_list,
         self._journal = Journal(
             consensus_module=dev_mode_consensus,
```
cintel: The journal and journal components have not been updated to use the on chain configuration.

validator/sawtooth_validator/journal/genesis.py
```diff
@@ -154,13 +162,20 @@ def start(self, on_done):
         LOGGER.debug('Produced state hash %s for genesis block.',
                      state_hash)
 
+        block_builder = GenesisController._generate_genesis_block()
         block_builder.add_batches(genesis_batches)
         block_builder.set_state_hash(state_hash)
 
+        block_publisher = self._get_block_publisher(state_hash)
+        block_publisher.initialize_block(block_builder)
```
cintel: This should probably be the raw block or a BlockWrapper instead of the BlockBuilder.

validator/sawtooth_validator/journal/consensus/consensus_factory.py
cintel: that was directed @jjason, since he has a Poet1 porting task in front of him.

validator/sawtooth_validator/journal/publisher.py
```diff
@@ -221,7 +222,7 @@ def on_check_publish_block(self, force=False):
                                                             _candidate_block):
```
cintel: This should be updated to pass a BlockWrapper or raw Block

validator/sawtooth_validator/journal/publisher.py
```diff
@@ -221,7 +222,7 @@ def on_check_publish_block(self, force=False):
...
                     candidate = self._candidate_block
                     self._candidate_block = None
-                    candidate = self._finalize_block(candidate)
+                    self._finalize_block(candidate)
```
cintel: This should be updated to pass a BlockWrapper or raw Block

validator/sawtooth_validator/server/core.py
peterschwarz: As noted in the PR description, it's part of a forthcoming PR

validator/sawtooth_validator/journal/publisher.py
peterschwarz: Is it find to call `builder.bulid_block()` at this point?  I changed this code because it wasn't following the doc from the consensus interfaces.

cintel: This call is fine ( I got it confused with the call above on line 199). This call sets the consensus data and generates the signature necessary before build block is called.

validator/sawtooth_validator/journal/consensus/consensus_factory.py
vaporos: I agree with @jjason, we should go back to 'poet' in all instances.  (Not necessarily in this PR though.)

validator/sawtooth_validator/journal/genesis.py
```diff
@@ -179,6 +194,34 @@ def start(self, on_done):
...
+            'sawtooth.consensus.algorithm', default_value='devmode')
+        try:
+            consensus_module = ConsensusFactory.get_consensus_module(
+                consensus_module_name)
+            return consensus_module.BlockPublisher(
+                BlockCache(self._block_store),
+                state_view=state_view)
+        except UnknownConsensusModuleError:
+            raise InvalidGenesisStateError(
+                'Unable to load consensus module "{}"'.format(
```
vaporos: Should this re-use the msg set on UnknownConsensusModuleError?  "Unable to load consensus module {}: {}".format(consensus_module_name, str(ucme)?

validator/sawtooth_validator/journal/consensus/consensus_factory.py
```diff
@@ -0,0 +1,47 @@
...
+            module_name (str): The name of the module to load.
+
+        Returns:
+            module: The consensus module.
+
+        Raises:
+            UnknownConsensusModuleError: Raised if the given module_name does
+                not correspond to a consensus implementation.
+        """
+        if module_name == 'devmode':
```
vaporos: Suggest separating the 'if' statement on determining the module to import from the actual import.  In the case that module_name isn't devmode/poet1, it should just import module_name assuming that it is the fully qualified python package.  This will facilitate working on consensus modules that are not part of core.

validator/sawtooth_validator/journal/consensus/consensus_factory.py
```diff
@@ -0,0 +1,47 @@
...
+            return importlib.import_module(
```
vaporos: importlib should be wrapped with try/except ImportError.

peterschwarz:
Updated with changes needed for Journal, as well as the minor changes to ConsensusFactory requested by @vaporos

PR #119, opened 2017-02-24
Update to work with PoET1

jjason:
Remove keyword arguments when calling into PoET enclave.
Renamed block_digest to block_hash everywhere to stay consistent with enclave.
Passing enclave wait timer object when creating wait certificate.
Fixed struct unpacking in PoET simulator that didn't work on Windows (but inexplicably worked on Linux).
Fix identifier property in wait certificate work across enclaves.
Force encoded PoET public key to string.
Fixed bug in wait timer duration creation.

### DIFF
--- a/validator/sawtooth_validator/consensus/poet1/poet_enclave_simulator/poet_enclave_simulator.py
+++ b/validator/sawtooth_validator/consensus/poet1/poet_enclave_simulator/poet_enclave_simulator.py
@@ -343,7 +343,7 @@ class _PoetEnclaveSimulator(object):
                         previous_certificate_id,
                         cls._seal_private_key))
 
-            tagd = float(struct.unpack('L', tag[-8:])[0]) / (2**64 - 1)
+            tagd = float(struct.unpack('Q', tag[-8:])[0]) / (2**64 - 1)
 
             # Now compute the duration
             duration = cls.__MINIMUM_DURATTION - local_mean * math.log(tagd)
--- a/validator/sawtooth_validator/consensus/poet1/poet_enclave_simulator/poet_enclave_simulator.py
+++ b/validator/sawtooth_validator/consensus/poet1/poet_enclave_simulator/poet_enclave_simulator.py
@@ -383,7 +383,7 @@ class _PoetEnclaveSimulator(object):
     @classmethod
     def create_wait_certificate(cls,
                                 wait_timer,
-                                block_digest):
+                                block_hash):
         with cls._lock:
             # If we don't have a PoET private key, then the enclave has not
             # been properly initialized (either by calling create_signup_info
@@ -453,7 +453,7 @@ class _PoetEnclaveSimulator(object):
                 EnclaveWaitCertificate.wait_certificate_with_wait_timer(
                     wait_timer=cls._active_wait_timer,
                     nonce=nonce,
-                    block_digest=block_digest)
+                    block_hash=block_hash)
             wait_certificate.signature = \
                 signing.sign(
                     wait_certificate.serialize(),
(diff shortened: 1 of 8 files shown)


cintel:
What is the motivation for moving to positional arguments?

jjason:
Swig seems to be puking on it.

jjason:
Want to get a fix in so @aludvik can proceed.  I personally like keyword arguments so will revisit to see if there is a better fix than simply going to positional arguments.

validator/sawtooth_validator/consensus/poet1/poet_enclave_simulator/poet_enclave_simulator.py
```diff
@@ -343,7 +343,7 @@ def create_wait_timer(cls,
                         previous_certificate_id,
                         cls._seal_private_key))
 
-            tagd = float(struct.unpack('L', tag[-8:])[0]) / (2**64 - 1)
+            tagd = float(struct.unpack('Q', tag[-8:])[0]) / (2**64 - 1)
```
peterschwarz: Should be part of a separate commit, which explains the change from 'L' to 'Q' (this has nothing to do with keyword args).

validator/sawtooth_validator/consensus/poet1/wait_certificate.py
```diff
@@ -120,10 +125,7 @@ def wait_certificate_from_serialized(cls,
 
     @property
     def enclave_wait_certificate(self):
-        return \
-            self.poet_enclave.deserialize_wait_certificate(
-                self._serialized_certificate,
-                self.signature)
+        return self._enclave_wait_certificate
```
peterschwarz: Like wise, this change seems to be unrelated to the keyword args change.  Not clear on why the change is needed.

validator/sawtooth_validator/consensus/poet1/wait_timer.py
```diff
@@ -150,6 +150,14 @@ def compute_local_mean(cls, certificates):
 
         return local_mean
 
+    @property
+    def enclave_wait_timer(self):
```
peterschwarz: Ditto

validator/tests/unit/poet1/test_wait_certificate.py
```diff
@@ -78,7 +78,7 @@ def test_create_wait_certificate_before_wait_timer_expires(self):
                 certificates=[])
         wc = \
             WaitCertificate.create_wait_certificate(
-                wait_timer=None,
+                wait_timer=wt,
```
peterschwarz: Ditto

jjason: Will break out into separate commits.

jjason:
@aludvik Was having problem with pickle and enclave wait certificate object.  Backed out a change that affected that.

If this works, will migrate any necessary changes to master.

validator/sawtooth_validator/consensus/poet1/wait_certificate.py
```diff
@@ -75,11 +75,16 @@ def create_wait_certificate(cls,
...
-            cls.poet_enclave.create_wait_certificate(
-                wait_timer=wait_timer,
-                block_digest=block_digest)
+        enclave_certificate = None
+        try:
+            enclave_certificate = \
+                cls.poet_enclave.create_wait_certificate(
+                    wait_timer.enclave_wait_timer,
+                    block_hash)
+        except Exception as ex:
```
aludvik: Should this swallow all exceptions here? Could cause a problem with debugging later. Is there a specific type of exception that you are expecting to see here for which you want enclave_certificate to be set to None?

jjason: Yeah....it somehow got past the lint filters in the 0-7 branch, but got caught in master.  I need to "backport" tha fix.

PR #120, opened 2017-02-24
Add block and batch request messages

agunde406:
The best way to test that this works (since we do not have network tests yet) is the following. You will see one validator create the genesis block. When you start loading transaction the non-genesis validator will not process any blocks and will send out a GossipBlockRequest message. Once it recives the genesis block, you will see it start processing batches. Currently, for this to work the validators need to be peered in both directions and they will continue to send batches back and forth forever. 

In terminal:
sawtooth admin genesis
validator -vv --peers tcp://0.0.0.0:8801

In new terminal:
validator --network-endpoint tcp://0.0.0.0:8801 --component-endpoint tcp://0.0.0.0:40001 --peers tcp://0.0.0.0:8800 -vv

In new terminal:
tp_intkey_python tcp://localhost:40000 -vv

In new terminal:
tp_intkey_python tcp://localhost:40001 -vv

Finally: 
intkey populate
intkey load

### DIFF
--- a/protos/network.proto
+++ b/protos/network.proto
@@ -1,4 +1,4 @@
-// Copyright 2016 Intel Corporation
+// Copyright 2016, 2017 Intel Corporation
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -45,4 +45,31 @@ message NetworkAcknowledgement {
     }
 
     Status status = 1;
+}
+
+message GossipBlockRequest {
+    // The id of the block that is being requested
+    string block_id = 1;
+
+    // The identity of the validator that is requesting the block
+    bytes node_id = 2;
+
+}
+
+message GossipBatchByBatchIdRequest {
+    // The id of the batch that is being requested
+    string id = 1;
+
+    // The identity of the validator that is requesting the batch
+    bytes node_id = 2;
+
+}
+
+message GossipBatchByTransactionIdRequest {
+    // The id's of the transaction that are in the batches requested
+    repeated string ids = 1;
+
+    // The identity of the validator that is requesting the batches
+    bytes node_id = 2;
+
 }
--- a/validator/sawtooth_validator/gossip/gossip.py
+++ b/validator/sawtooth_validator/gossip/gossip.py
@@ -13,6 +13,10 @@
 # limitations under the License.
 # ------------------------------------------------------------------------------
 from sawtooth_validator.protobuf.network_pb2 import GossipMessage
+from sawtooth_validator.protobuf.network_pb2 import GossipBatchByBatchIdRequest
+from sawtooth_validator.protobuf.network_pb2 import \
+    GossipBatchByTransactionIdRequest
+from sawtooth_validator.protobuf.network_pb2 import GossipBlockRequest
 from sawtooth_validator.protobuf import validator_pb2
 from sawtooth_validator.protobuf.network_pb2 import PeerRegisterRequest
 
@@ -26,17 +30,40 @@ class Gossip(object):
             content_type="BLOCK",
             content=block.SerializeToString())
 
-        self._broadcast(gossip_message)
+        self.broadcast(gossip_message, validator_pb2.Message.GOSSIP_MESSAGE)
+
+    def broadcast_block_request(self, block_id):
+        # Need to define node identity to be able to route directly back
+        block_request = GossipBlockRequest(block_id=block_id)
+        self.broadcast(block_request,
+                       validator_pb2.Message.GOSSIP_BLOCK_REQUEST)
 
     def broadcast_batch(self, batch):
         gossip_message = GossipMessage(
             content_type="BATCH",
             content=batch.SerializeToString())
 
-        self._broadcast(gossip_message)
+        self.broadcast(gossip_message, validator_pb2.Message.GOSSIP_MESSAGE)
 
-    def _broadcast(self, gossip_message):
-        message_type = validator_pb2.Message.GOSSIP_MESSAGE
+    def broadcast_batch_by_transaction_id_request(self, batch_id):
+        # Need to define node identity to be able to route directly back
+        batch_request = GossipBatchByTransactionIdRequest(
+            id=batch_id
+        )
+        self.broadcast(
+            batch_request,
+            validator_pb2.Message.GOSSIP_BATCH_BY_TRANSACTION_ID_REQUEST)
+
+    def broadcast_batch_by_batch_id_request(self, batch_id):
+        # Need to define node identity to be able to route directly back
+        batch_request = GossipBatchByBatchIdRequest(
+            id=batch_id
+        )
+        self.broadcast(
+            batch_request,
+            validator_pb2.Message.GOSSIP_BATCH_BY_BATCH_ID_REQUEST)
+
+    def broadcast(self, gossip_message, message_type):
         for connection in self._network.connections:
             connection.send(message_type, gossip_message.SerializeToString())
 
--- a/validator/sawtooth_validator/journal/chain.py
+++ b/validator/sawtooth_validator/journal/chain.py
@@ -167,7 +167,7 @@ class BlockValidator(object):
                 return False
         return True
 
-    def _validate_block(self, blkw, committed_txn):
+    def validate_block(self, blkw, committed_txn):
         try:
             if blkw.status == BlockStatus.Valid:
                 return True
@@ -335,7 +335,7 @@ class BlockValidator(object):
             valid = True
             for block in reversed(new_chain):
                 if valid:
-                    if not self._validate_block(block, committed_txn):
+                    if not self.validate_block(block, committed_txn):
                         LOGGER.info("Block validation failed: %s", block)
                         valid = False
                 else:
@@ -384,7 +384,8 @@ class ChainController(object):
                  executor,
                  transaction_executor,
                  on_chain_updated,
-                 squash_handler):
+                 squash_handler,
+                 chain_id_manager):
         """Initialize the ChainController
         Args:
             algorithm to use.
@@ -418,6 +419,7 @@ class ChainController(object):
         self._blocks_pending = {}  # set of blocks that the previous block
         # is being processed. Once that completes this block will be
         # scheduled for validation.
+        self._chain_id_manager = chain_id_manager
 
         try:
             self._chain_head = self._block_store.chain_head
@@ -508,6 +510,10 @@ class ChainController(object):
                     # do we already have this block
                     return
 
+                if self.chain_head is None:
+                    self._set_genesis(block)
+                    return
+
                 self._block_cache[block.identifier] = block
                 self._blocks_pending[block.identifier] = []
                 LOGGER.debug("Block received: %s", block)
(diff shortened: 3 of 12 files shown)


validator/sawtooth_validator/journal/completer.py
```diff
@@ -27,38 +27,51 @@
...
         self.batch_cache = TimedCache()
         self.block_cache = BlockCache(block_store)
+        # avoid throwing away the genesis block
+        self.block_cache[NULL_BLOCK_IDENTIFIER] = None
         self._on_block_received = None
         self._on_batch_received = None
 
     def _check_block(self, block):
         # currently only accepting finalized blocks
         # in the future if the blocks will be built
```
jsmitchell: I know this isn't part of your change, but this comment is a sentence fragment. Recommend changing this to docstring which describes what _check_block does.

validator/sawtooth_validator/journal/completer.py
```diff
@@ -27,38 +27,51 @@
...
+    def __init__(self, block_store, gossip):
+        self.gossip = gossip
         self.batch_cache = TimedCache()
         self.block_cache = BlockCache(block_store)
+        # avoid throwing away the genesis block
+        self.block_cache[NULL_BLOCK_IDENTIFIER] = None
         self._on_block_received = None
         self._on_batch_received = None
 
     def _check_block(self, block):
```
jsmitchell: Since this method now has actions it takes, _check_block() probably isn't a great name for it.

validator/sawtooth_validator/journal/completer.py
```diff
@@ -27,38 +27,51 @@
...
-                return False
+                LOGGER.debug("Request missing batch: %s",
+                             block.header.batch_ids[i])
+                self.gossip.broadcast_batch_request(block.header.batch_ids[i],
+                                                    "batch")
+                valid = False
 
-        self.block_cache[block.header_signature] = block
-        return True
+        if valid:
```
jsmitchell: Even though _check_block() now does things rather than just evaluating conditions and returning true/false, I'd still move this down to the caller, rather than having it as a side effect of _check_block()

validator/sawtooth_validator/journal/completer.py
```diff
@@ -75,6 +88,22 @@ def add_batch(self, batch):
         self.batch_cache[batch.header_signature] = batch
         self._on_batch_received(batch)
 
+    def get_block(self, block_id):
+        if block_id in self.block_cache:
+            return self.block_cache[block_id]
+        return None
+
+    def get_batch(self, batch_id=None, transaction_id=None):
```
jsmitchell: Consider splitting these into two methods get_batch and get_batch_by_transaction

validator/sawtooth_validator/journal/journal.py
```diff
@@ -103,7 +103,8 @@ def __init__(self,
                  block_sender,
                  transaction_executor,
                  squash_handler,
-                 block_cache=None  # not require, allows tests to inject a
+                 data_dir,
+                 block_cache=None,  # not require, allows tests to inject a
```
jsmitchell: s/require/required

validator/sawtooth_validator/journal/responder.py
```diff
@@ -0,0 +1,101 @@
...
+class Responder(object):
+    def __init__(self, completer):
+        self.completer = completer
+
+    def check_for_block(self, block_id):
+        # Ask Completer
+        block = self.completer.get_block(block_id)
+        return block
+
+    def check_for_batch(self, batch_id=None, transaction_id=None):
```
jsmitchell: same here, consider check_for_batch() and check_for_batch_by_transaction()

validator/sawtooth_validator/gossip/gossip.py
```diff
@@ -26,20 +28,37 @@ def broadcast_block(self, block):
...
+                        validator_pb2.Message.GOSSIP_BATCH_REQUEST)
 
-    def _broadcast(self, gossip_message):
-        message_type = validator_pb2.Message.GOSSIP_MESSAGE
+    def _broadcast(self, gossip_message, message_type):
         for connection in self._network.connections:
             connection.send(message_type, gossip_message.SerializeToString())
 
+    def broadcst(self, gossip_message, message_type):
+        self._broadcast(gossip_message, message_type)
```
cintel: Why do we need a public method that just calls the private method? Can we just have a single public method?

validator/sawtooth_validator/journal/chain.py
```diff
@@ -473,3 +480,27 @@ def on_block_received(self, block):
...
+        if block.previous_block_id == NULL_BLOCK_IDENTIFIER:
+            self._block_store.set_chain_head(block.identifier)
+
+            self._block_store[block.identifier] = block
+            self._chain_head = block
+            self._notify_on_chain_updated(self._chain_head)
+
+            self._save_block_chain_id(block.identifier)
+        else:
+            LOGGER.warning("Cannot set intial chain head, this is not a "
```
cintel: s/intial/initial/

validator/sawtooth_validator/journal/chain.py
```diff
@@ -473,3 +480,27 @@ def on_block_received(self, block):
...
+                           "genesis block. %s", block.identifier)
```
cintel: s/block.identifier/block/ let the block wrapper formatter take care of the string representation.

validator/sawtooth_validator/journal/chain.py
```diff
@@ -473,3 +480,27 @@ def on_block_received(self, block):
         # pylint: disable=broad-except
         except Exception as exc:
             LOGGER.exception(exc)
+
+    def _set_chain_head(self, block):
+        # This is used by journal when it has recieved the genesis block from
+        # the genesis validator
+        if block.previous_block_id == NULL_BLOCK_IDENTIFIER:
+            self._block_store.set_chain_head(block.identifier)
```
cintel: We should check against the block identity against the block-chain-id file if it exists.

validator/sawtooth_validator/journal/chain.py
```diff
@@ -473,3 +480,27 @@ def on_block_received(self, block):
...
+            self._block_store[block.identifier] = block
+            self._chain_head = block
+            self._notify_on_chain_updated(self._chain_head)
+
+            self._save_block_chain_id(block.identifier)
+        else:
+            LOGGER.warning("Cannot set intial chain head, this is not a "
+                           "genesis block. %s", block.identifier)
+
+    def _save_block_chain_id(self, block_chain_id):
```
cintel: I don't like that the Journal/ChainController has to know the data_dir and now we have two separate locations in the code maintaining the block_chain_id_file. Where should the management of these data dir files be?

validator/sawtooth_validator/journal/chain.py
cintel: Also we need to validate the block to build the required state? Not sure what happens if the block publisher tries to build a block on top of a block that the state_root_hash is not present in the merkle tree.

validator/sawtooth_validator/journal/completer.py
```diff
@@ -27,38 +27,51 @@
...
 from sawtooth_validator.networking.dispatch import HandlerResult
 from sawtooth_validator.networking.dispatch import HandlerStatus
+from sawtooth_validator.journal.block_wrapper import NULL_BLOCK_IDENTIFIER
 
 LOGGER = logging.getLogger(__name__)
 
 
 class Completer(object):
-    def __init__(self, block_store):
+    def __init__(self, block_store, gossip):
```
peterschwarz: Can you add some doc string here, since you changed the signature?

validator/sawtooth_validator/journal/chain.py
agunde406: I will separate out the handling of the block-chain-id file, check the block identity against the block-chain-id, and add validation of the genesis block.

validator/sawtooth_validator/journal/chain_id_manager.py
```diff
@@ -0,0 +1,50 @@
...
+    def __init__(self, data_dir):
+        self._data_dir = data_dir
+
+    def save_block_chain_id(self, block_chain_id):
+        LOGGER.debug('writing block chain id')
+        block_chain_id_file = os.path.join(self._data_dir, 'block-chain-id')
+        try:
+            with open(block_chain_id_file, 'w') as f:
+                f.write(block_chain_id)
+        except IOError:
```
jsmitchell: What should happen to the validator in this instance? Should this be a fatal error? If so, consider raising some more specific exception (like LocalConfigurationError) to main.

validator/sawtooth_validator/journal/chain_id_manager.py
```diff
@@ -0,0 +1,50 @@
...
+            LOGGER.error('The block-chain-id is unwriteable')
```
jsmitchell: Recommend ("Unable to write to %s", block_chain_id_file)

validator/sawtooth_validator/journal/chain_id_manager.py
agunde406: Updated to use LocalConfigurationError

protos/network.proto
```diff
@@ -46,3 +46,24 @@ message NetworkAcknowledgement {
...
+    string block_id = 1;
+
+    // The identity of the validator that is requesting the block
+    bytes node_id = 2;
+
+}
+
+message GossipBatchRequest {
+    // Need to clarify if the id belongs to a batch or to a transaction
+    string id_type = 1;
```
vaporos: Should this be an enum?

It seems like it could be two different requests - GossipBatchByBatchIdRequest, GossipBatchByTransactionIdRequest.

In the case where we are requesting based on transaction ids, it seems like sending one at a time will be very chatty, since 1000 txns could all be in the same batch.

protos/network.proto
```diff
@@ -46,3 +46,24 @@ message NetworkAcknowledgement {
...
+
+    // The id of the object the correlates to the batch that is being requested
```
vaporos: The comment needs rewording.

validator/sawtooth_validator/journal/chain.py
```diff
@@ -473,3 +479,38 @@ def on_block_received(self, block):
...
         except Exception as exc:
             LOGGER.exception(exc)
+
+    def _set_genesis(self, block):
+        # This is used by a non-genesis journal when it has recieved the
+        # genesis block from the genesis validator
+        if block.previous_block_id == NULL_BLOCK_IDENTIFIER:
+            chain_id = self._chain_id_manager.get_block_chain_id()
+            if chain_id != block.identifier:
+                LOGGER.warning("Block id does not match block chain id."
```
vaporos: These two sentences would be concatenated with no space between them.  The chain id / block id should be logged as part of the message.

validator/sawtooth_validator/journal/chain_id_manager.py
vaporos: This will only be caught by main() of it is thrown in the MainThread.

agunde406:
Updated to have two batch request messages, one for batch id and one for multiple txn id's. If using txn id the responder will check every txn id in the message if a batch is found it will be broadcast. If it was not found that txn will be added to a new request message.

validator/sawtooth_validator/journal/chain_id_manager.py
agunde406: added a logging statement

validator/sawtooth_validator/journal/chain_id_manager.py
```diff
@@ -0,0 +1,55 @@
...
+        if not Path(block_chain_id_file).is_file():
+            return None
+
+        try:
+            with open(block_chain_id_file, 'r') as f:
+                block_chain_id = f.read()
+                return block_chain_id if block_chain_id else None
+
+        except IOError:
+            LOGGER.error('The block-chain-id file exists, but is unreadable')
```
vaporos: You will get a better message here if you use: "Unable to read block-chain-id: {}".format(str(e))" where e is the exception.

Should (almost) never log and re-throw the exception, since the exception is likely to be caught and logged.

PR #125, opened 2017-02-24
Bringing in Fixes to PoET1 code before major re-implementation occurs

jjason:
Changes include:
* Rename wait certificate block_digest to block_hash to make everything consistent with PoET enclave
* Remove keyword arguments when calling into PoET enclave as apparently SWIGged C++ code does not like it
* Correctly provide enclave wait timer to PoET enclave when trying to create a wait certificate


jjason:
Pushed a couple of minor changes based upon testing with 0-7 branch

PR #131, opened 2017-02-28
Update JS-SDK to current protobuf messages

peterschwarz:
Update the JS SDK with the protobuf message changes that removed MessageList.

Additionally:

- Fix the tests
- Fix linting issues

### DIFF
--- a/sdk/javascript/client/state.js
+++ b/sdk/javascript/client/state.js
@@ -18,8 +18,6 @@
 'use strict'
 
 const {Entry, TpStateGetRequest, TpStateGetResponse, TpStateSetRequest, TpStateSetResponse, Message} = require('../protobuf')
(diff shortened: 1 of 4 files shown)


sdk/javascript/client/state.js
```diff
@@ -18,8 +18,6 @@
 'use strict'
 
 const {Entry, TpStateGetRequest, TpStateGetResponse, TpStateSetRequest, TpStateSetResponse, Message} = require('../protobuf')
```
delventhalz: This line is 126 characters. Probably not in this PR, but at some point we should refactor these files to enforce a character limit.

peterschwarz: We should revisit the linting once we get the JS SDK tests running in CI

PR #132, opened 2017-02-28
Add bidirectional message dispatch and respond on receiving connection

jsmitchell:
This changes the Dispatcher's single send_message to a dict indexed by connection identifier, and sets up new message dispatch on the receiving side (messages sent from ROUTER -> DEALER which are not responses).

### DIFF
--- a/validator/sawtooth_validator/networking/interconnect.py
+++ b/validator/sawtooth_validator/networking/interconnect.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 # ------------------------------------------------------------------------------
 
+import sys
 import asyncio
 import hashlib
 import logging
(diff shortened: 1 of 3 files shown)


validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -86,32 +91,43 @@ def _receive_message(self):
...
+                    self._dispatcher.dispatch(self._connection,
+                                              identity,
+                                              message)
                 else:
-                    LOGGER.info(
-                        "received a first message on the zmq dealer.")
+                    # Because this is a zmq.DEALER socket, there is no
+                    # outbound identity
+                    self._dispatcher.dispatch(self._connection,
+                                              None,
```
peterschwarz: If identity is something optional, should the dispatch method's arguments be arranged as such (with any defaults for identity as needed).  This would match our other uses of Identity in the system.

PR #133, opened 2017-02-28
Fix BlockStore: get_block_by_batch_id, __iter__, __len__

delventhalz:
Fix get_block_by_batch_id to properly return a block.

Previously, iterating over a BlockStore instance would throw an
error, and calling `len` on one would return the total number of
indexed keys (including `"chain_head_id"`), which is not especially
meaningful. Fix these to now more correctly throw `NotImplementedError`s.

### DIFF
--- a/validator/sawtooth_validator/journal/block_store.py
+++ b/validator/sawtooth_validator/journal/block_store.py
@@ -56,10 +56,12 @@ class BlockStore(MutableMapping):
         return x in self._block_store
 
     def __iter__(self):
-        return iter(self._block_store)
+        # Required by abstract base class, but implementing is non-trivial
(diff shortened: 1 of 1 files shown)


validator/sawtooth_validator/journal/block_store.py
```diff
@@ -56,10 +56,10 @@ def __contains__(self, x):
         return x in self._block_store
 
     def __iter__(self):
-        return iter(self._block_store)
+        raise NotImplementedError('BlockStore is not iterable')
```
peterschwarz: Perhaps a comment here on this being non-trivial, and only here as a requirement of the abstract base class.

PR #138, opened 2017-02-28
Journal fixes for BlockStore integration.

cintel:

### DIFF
--- a/validator/sawtooth_validator/journal/chain.py
+++ b/validator/sawtooth_validator/journal/chain.py
@@ -249,8 +249,7 @@ class BlockValidator(object):
         return (new_blkw, cur_blkw)
 
     def _find_common_ancestor(self, new_blkw, cur_blkw, new_chain, cur_chain):
-        """
-        Finds a common ancestor of the two chains.
+        """ Finds a common ancestor of the two chains.
         """
         while cur_blkw.identifier != \
                 new_blkw.identifier:
@@ -275,9 +274,8 @@ class BlockValidator(object):
             cur_blkw = \
                 self._block_cache[cur_blkw.previous_block_id]
 
-    def _compare_forks(self, new_chain, cur_chain):
-        """
-        Compare the two chains and determine which should be the head.
+    def _test_commit_new_chain(self):
+        """ Compare the two chains and determine which should be the head.
         """
         fork_resolver = self._consensus_module.\
             ForkResolver(block_cache=self._block_cache)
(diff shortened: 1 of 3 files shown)


validator/sawtooth_validator/journal/chain.py
```diff
@@ -348,7 +346,7 @@ def run(self):
                 return
 
             # 4) Evaluate the 2 chains to see which is the one true chain.
```
peterschwarz: This comment doesn't quite make sense anymore, as the "2 chains" are no longer local passed into this function.  In fact, it's no longer clear how two forks are getting into this function. Perhaps a better function name is needed, like `is_forked()` (but maybe better?).

validator/sawtooth_validator/journal/publisher.py
```diff
@@ -74,9 +74,10 @@ def __init__(self,
         self._block_sender = block_sender
         self._pending_batches = []  # batches we are waiting for validation,
         # arranged in the order of batches received.
-        self._committed_txn = None  # Look-up cache for transactions that are
-        # committed in the current chain. Cache is used here so that we can
-        # support opportunistically building on top of a block we published.
+        self._committed_txn = TransactionCache(self._block_cache.block_store)
```
peterschwarz: This should be named as `_commited_txn_cache` or `_commited_txns`.  This name implies a single value.

validator/sawtooth_validator/journal/publisher.py
```diff
@@ -125,10 +126,6 @@ def _build_block(self, chain_head):
         self._scheduler = self._transaction_executor.create_scheduler(
             self._squash_handler, chain_head.state_root_hash)
 
-        self._transaction_executor.execute(self._scheduler)
-        for batch in self._pending_batches:
-            self._validate_batch(batch)
-
         # build the TransactionCache
         self._committed_txn = TransactionCache(self._block_cache.block_store)
```
peterschwarz: Double setting the value of the committed txn cache

validator/sawtooth_validator/journal/publisher.py
```diff
@@ -357,23 +370,24 @@ def on_check_publish_block(self, force=False):
...
+                    if not self._finalize_block(candidate):
+                        return
+
                     # if no batches are in the block, do not send it out
                     if len(candidate.batches) == 0:
-                        LOGGER.info("No Valid batches added to block, " +
-                                    " dropping %s",
-                                    candidate.identifier[:8])
+                        LOGGER.info("No valid batches added to block, " +
+                                    " dropping %s", candidate)
```
agunde406: Will never reach this logging statement. If now blocks are added to a block (state_hash is None) you return False so this function is returned in line 385.

validator/sawtooth_validator/journal/publisher.py
cintel: I don't see the double set here, The TransactionCache is being rebuilt to be in the state of the new block and with the addition of transactinons of a block not yet in the chain.

peterschwarz: This was probably my mis-reading the diff, and not seeing the fact that it was in a separate function.

PR #140, opened 2017-03-01
test case to verify paramaters of sawtooth stats

nconde:


validator/tests/sawtooth_suites/ts_nightly_poet1.py
```diff
@@ -45,7 +47,7 @@ def test_suite(self):
             suite.addTest(TestConvergence('test_bootstrap', urls))
             suite.addTest(TestIntegration('test_intkey_load_ext', urls))
             suite.addTest(TestIntegration('test_missing_dependencies', urls))
-            suite.addTest(TestSawtoothStats('test_sawtooth_stats', urls))
+            suite.addTest(TestSawtoothStatsTwistedThread('test_sawtooth_stats_twisted_thread', urls))
```
agunde406: Is there any way to stop the print out from becoming distorted after running this test? It makes it very difficult to see the output for any following tests.

nconde: I believe that'd involve modifying the stats_print from stats_clients - Tom do you think that's changeable?

PR #142, opened 2017-03-01
Manage gossip identities and broadcast to ROUTER connected peers

jsmitchell:
This adds an identity registration mechanism to gossip and broadcasts out to connected identities on the router socket.

### DIFF
--- a/validator/sawtooth_validator/gossip/gossip.py
+++ b/validator/sawtooth_validator/gossip/gossip.py
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ------------------------------------------------------------------------------
+import logging
+from threading import Condition
+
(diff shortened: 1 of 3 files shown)


validator/sawtooth_validator/gossip/gossip.py
```diff
@@ -20,10 +22,40 @@
 from sawtooth_validator.protobuf import validator_pb2
 from sawtooth_validator.protobuf.network_pb2 import PeerRegisterRequest
 
+LOGGER = logging.getLogger(__name__)
+
 
 class Gossip(object):
     def __init__(self, network):
         self._network = network
+        self._identities = []
```
boydjohnson: My threading intuition would be that since _identities is being mutated in a Handler implementation/subclass it should be threadsafe.

jsmitchell:
@boydjohnson yeah, this is in advance of the peering work. I like your suggestion of a peer class which abstracts the two types.

PR #146, opened 2017-03-01
Add requests for batch status to REST and ZMQ APIs

delventhalz:
Makes it possible to query for the status of a batch by id from either API. Batches can have one of three statuses:
      • `COMMITTED` - batch was found in block chain
      • `PENDING` - batch found in batch cache, but not in chain
      • `REJECTED` - batch not found

Also includes unit tests for the new requests to both APIs

### DIFF
--- a/rest_api/sawtooth_rest_api/openapi.yaml
+++ b/rest_api/sawtooth_rest_api/openapi.yaml
@@ -113,13 +113,15 @@ paths:
 
   /batch_status:
     get:
-      summary: Fetches a list of the committed statuses for a set of batches
+      summary: Fetches the committed statuses for a set of batches
       description: >
-        Fetches a list of statuses for a batch or batches. The
-        batch(es) you want to check can be specified using the
-        `id` filter parameter. Note that this route does not
-        return a full resource, and the response body will
-        _not_ include the `head` or `paging` properties.
+        Fetches an object with a status for each batch requested. The keys of
+        the data object will be the ids of the batch(es), while the values will
+        be a string status with the values `'COMMITTED'`, `'INVALID'`,
+        `'PENDING'`, or `'UNKNOWN'`.
+        The batch(es) you want to check can be specified using the `id` filter
+        parameter. Note that as this route does not return a full resource, and
+        the response body will _not_ include the `head` or `paging` properties.
       parameters:
         - name: id
           in: query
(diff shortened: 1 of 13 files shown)


vaporos:
Seems like we should have four states: COMMITTED - it is committed in the current chain; PENDING - it is in the block publisher's pending queue; INVALID - dropped because it was determined to be invalid during execution (by the block publisher sending it to the scheduler, etc.); UNKNOWN - there is no record of this batch.

delventhalz:
@vaporos 

Re: Rejected vs Invalid
I wonder if batches can be rejected for reasons other than being invalid? Those would probably be exceptional cases, so I'll make the switch, and rejected status(es) can be added later if needed.

Re: Unknown status
Agreed. Unfortunately to distinguish between `UNKNOWN` and `INVALID`, some significant architecture needs to be added. But, I will refactor the current implementation to return `UNKNOWN` _instead_ of `INVALID`/`REJECTED`, as that more accurately reflects what is known.

rest_api/sawtooth_rest_api/openapi.yaml
```diff
@@ -113,13 +113,15 @@ paths:
...
+      summary: Fetches the committed statuses for a set of batches
       description: >
-        Fetches a list of statuses for a batch or batches. The
-        batch(es) you want to check can be specified using the
-        `id` filter parameter. Note that this route does not
-        return a full resource, and the response body will
-        _not_ include the `head` or `paging` properties.
+        Fetches an object with a status for each batch requested. The keys of
+        the data object will be the ids of the batch(es), while the values will
+        be a string status with the values `'COMMITTED'`, `'REJECTED'`, or
```
vaporos: This text should be updated to match the new values.

rest_api/sawtooth_rest_api/openapi.yaml
```diff
@@ -563,6 +563,16 @@ definitions:
         items:
           $ref: "#/definitions/Error"
 
+  BatchStatuses:
+    properties:
+      "{batch_id}":
+        type: string
+        example: PENDING
+        enum:
```
vaporos: This text should be updated to match the new statuses.

delventhalz: Updated

PR #149, opened 2017-03-01
Cluster stop

nick-drozd:
Aside from being slow, `sawtooth cluster stop` prints its `Stopping: validator-***` messages all at once, without regard for when the validators actually stop. `start` does the same thing. Those messages have been moved to track the actual starting and stopping of validators.


aludvik:
@vaporos I agree, but I think this should go through now while Nick works on a better solution. The processes are getting killed anyways and this will save time for anyone using cluster.

PR #150, opened 2017-03-01
Pass BatchSender to BlockPublisher consensus interface, enables Conse…

cintel:
…nsus algorithms to broadcast batches.

### DIFF
--- a/validator/sawtooth_validator/journal/consensus/dev_mode/dev_mode_consensus.py
+++ b/validator/sawtooth_validator/journal/consensus/dev_mode/dev_mode_consensus.py
@@ -25,11 +25,12 @@ from sawtooth_validator.journal.consensus.consensus\
 class BlockPublisher(BlockPublisherInterface):
     """ DevMode BlockPublisher implementation.
     """
-    def __init__(self, block_cache, state_view):
-        super().__init__(block_cache, state_view)
+    def __init__(self, block_cache, state_view, batch_publisher):
+        super().__init__(block_cache, state_view, batch_publisher)
 
         self._block_cache = block_cache
         self._state_view = state_view
+        self.batch_publisher = batch_publisher
         self._num_batches = 1
         self._count = 0
 
--- a/validator/sawtooth_validator/journal/genesis.py
+++ b/validator/sawtooth_validator/journal/genesis.py
@@ -48,7 +48,8 @@ class GenesisController(object):
                  state_view_factory,
                  identity_key,
                  data_dir,
-                 chain_id_manager):
+                 chain_id_manager,
+                 batch_sender):
         """Creates a GenesisController.
 
         Args:
@@ -62,6 +63,9 @@ class GenesisController(object):
                 factory for creating state views during processing.
             identity_key (str): A private key used for signing blocks, in hex.
             data_dir (str): The directory for data files.
+            chain_id_manager (ChainIdManager): utility class to manage the
+            chain id file.
+            batch_sender: interface to broadcast batches to the network.
         """
         self._context_manager = context_manager
         self._transaction_executor = transaction_executor
@@ -71,6 +75,7 @@ class GenesisController(object):
         self._identity_priv_key = identity_key
         self._data_dir = data_dir
         self._chain_id_manager = chain_id_manager
+        self._batch_sender = batch_sender
 
     def requires_genesis(self):
         """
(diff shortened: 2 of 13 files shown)


validator/sawtooth_validator/journal/consensus/dev_mode/dev_mode_consensus.py
```diff
@@ -32,11 +32,12 @@ class BlockPublisher(BlockPublisherInterface):
     other consensus algorithms
 
     """
-    def __init__(self, block_cache, state_view):
-        super().__init__(block_cache, state_view)
+    def __init__(self, block_cache, state_view, transaction_sender):
```
jsmitchell: What's an example of a scenario where consensus would send a transaction? Registry attestation for Poet?

jsmitchell: There are some other consensus impl's in journal/consensus that probably need this change given the change to the interface.

validator/sawtooth_validator/journal/publisher.py
```diff
@@ -35,6 +37,30 @@
 LOGGER = logging.getLogger(__name__)
 
 
+class TransactionSender(object):
```
jsmitchell: Is the publisher module the right place for TransactionSender?

jsmitchell: TransactionSender seems like a bit of a misnomer. Transactions don't get sent, they get packaged into batches which get sent via the BatchSender. Maybe BatchBuilder and then BatchSender would be more natural?

validator/sawtooth_validator/journal/consensus/dev_mode/dev_mode_consensus.py
cintel: A Poet node needs to broadcast SignupInfo to the network and wait for it to be added to a block before it can claim blocks.

validator/sawtooth_validator/journal/publisher.py
cintel: It is a local helper for the BlockPublisher.

jsmitchell: It's also used by journal/consensus though, right?

cintel: From the consensus point of view it is a TransactionSender, a consensus algorithm hands it a set of transactions and they get sent.

validator/sawtooth_validator/journal/consensus/dev_mode/dev_mode_consensus.py
cintel: dev_mode is the only working consensus in journal/consensus. The PoET consensus algorithm there and the implementation of this is still in progress, no implementation of the consensus interfaces are present yet.....

validator/sawtooth_validator/journal/publisher.py
cintel: The networking layer has moved in the direction of not providing central services or routing for components. I was following the the pattern of the BlockSender but saw no reason to encumber consensus algorithms with the details on how to build blocks.

cintel: It is the interface implementation for transaction sending the BlockPublisher passes to the Journal/Consensus BlockPublishers. The consensus algorithms are unaware of the implementation details or how to construct one.

validator/sawtooth_validator/journal/genesis.py
```diff
@@ -212,11 +217,17 @@ def _get_block_publisher(self, state_hash):
         """
         state_view = self._state_view_factory.create_view(state_hash)
         try:
+            class BatchPublisher(object):
+                def send(self, transactions):
+                    raise InvalidGenesisConsensusError(
```
cintel: The SignupInfo for the genesis controller will be included in the genesis node. It has to otherwise no  nodes can claim blocks. The c test (min age of sign up info) won't apply until there are at least C blocks are in the chain.

jjason: Agreed.  The system needs to be bootstrapped with the first validator.  And, as @cintel states, we cannot apply the C block policy until there are enough blocks in the blockchain.  It is going to have to be some factor of c and the number of validators as it also has to play nice with the K policy (i.e., a validator has to refresh signup information after claiing K blocks) and the z-Test, which prevents a validator from winning block elections too often.

peterschwarz: Is there a way we can document this?  It's not entirely clear from the source alone.

jjason: Not certain it should be documented here as it is very PoET specific.

cintel:
@peterschwarz genesis comment in place for your review

validator/sawtooth_validator/journal/genesis.py
```diff
@@ -212,11 +217,25 @@ def _get_block_publisher(self, state_hash):
         """
         state_view = self._state_view_factory.create_view(state_hash)
         try:
+            class BatchPublisher(object):
+                def send(self, transactions):
+                    # Consensus implementations are expected to have handling
+                    # inplace for genesis operation. This should includes
```
peterschwarz: inplace -> in place

validator/sawtooth_validator/journal/genesis.py
```diff
@@ -212,11 +217,25 @@ def _get_block_publisher(self, state_hash):
...
+                    # adding any authorization and registrations required
```
peterschwarz: authorization should probably be plural, as well

validator/sawtooth_validator/journal/genesis.py
```diff
@@ -212,11 +217,25 @@ def _get_block_publisher(self, state_hash):
...
+                    # for the genesis node to the Genesis Batch list and
+                    # detecting validation of the Genesis Block and handle it
```
peterschwarz: handle -> handling (match the same present participle form)

cintel:
@peterschwarz the update is in place!

PR #151, opened 2017-03-02
JS SDK: Handle disconnects and SIGTERM

peterschwarz:
The Javascript client can handle disconnects from the server, and reconnects, as well as handles SIGTERM by sending the Unregister event before shutting down.

Additionally, updates the Debian package for ZMQ from `libzmq-dev` to `libzmq3-dev`.  This is needed  as the former is version 2.x, which is missing the monitor feature, and the later is (currently) 4.x, which has it. The monitor feature is required for handling validator disconnects.

### DIFF
--- a/sdk/javascript/client/deferred.js
+++ b/sdk/javascript/client/deferred.js
@@ -1,5 +1,5 @@
 /**
- * Copyright 2016 Intel Corporation
+ * Copyright 2017 Intel Corporation
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,44 +17,21 @@
 
 'use strict'
 
-//
-// TODO: Timeouts
+function Deferred () {
+  if (Promise.defer) {
+    return Promise.defer()
+  } else {
+    this.resolve = null
+    this.reject = null
 
-const _handleResult = (self, f) => (v) => {
-  self._isDone = true
-  self._value = v
-  return f(v)
-}
-
-class Future {
-  constructor () {
-    let self = this
-    this._promise = new Promise((resolve, reject) => {
-      self._resolver = resolve
+    this.promise = new Promise((resolve, reject) => {
+      this.resolve = resolve
+      this.reject = reject
     })
-  }
 
-  get (f) {
-    if (this._isDone) {
-      return f(this._value)
-    } else {
-      return this._promise.then(_handleResult(this, f))
-    }
-  }
-
-  set (value) {
-    if (!this._isDone) {
-      this._resolver(value)
-    }
-  }
-
-  then (f) {
-    return this._promise.then(_handleResult(this, f))
-  }
-
-  catch (f) {
-    return this._promise.catch(f)
+    Object.freeze(this)
   }
 }
 
-module.exports = Future
+module.exports = Deferred
+
--- a/sdk/javascript/client/stream.js
+++ b/sdk/javascript/client/stream.js
@@ -25,7 +25,8 @@ const util = require('util')
 const assert = require('assert')
 
 const {Message} = require('../protobuf')
-const Future = require('./future')
+const Deferred = require('./deferred')
+const {ValidatorConnectionError} = require('../processor/exceptions')
 
 const _encodeMessage = (messageType, correlationId, content) => {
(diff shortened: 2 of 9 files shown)


sdk/javascript/client/deferred.js
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2016 Intel Corporation
+ * Copyright 2017 Intel Corporation
```
delventhalz: I believe the correct header for files made in 2016, but updated in 2017 is _"Copyright 2016, 2017"_

sdk/javascript/client/stream.js
```diff
@@ -47,42 +48,64 @@ const _generateId = () =>
...
+    })
+
+    this.connect(this._onConnectCb)
   }
 
   send (type, content) {
-    const correlationId = _generateId()
-    let future = new Future()
-    this._futures[correlationId] = future
+    if (this._socket) {
```
delventhalz: Does `send` just return `undefined` if there is no socket? Is that the desired behavior?

sdk/javascript/client/deferred.js
boydjohnson: My understanding is that if a file was originally made in 2016, but then updated in 2017 it would be "Copyright 2016, 2017 Intel Corporation"

sdk/javascript/client/stream.js
```diff
@@ -47,42 +48,64 @@ const _generateId = () =>
...
 
-    this._socket.send(_encodeMessage(type, correlationId, content))
+      this._socket.send(_encodeMessage(type, correlationId, content))
 
-    return future
+      return deferred.promise
+    }
   }
 
   sendBack (type, correlationId, content) {
```
delventhalz: I think I would prefer `send` be polymorphic and handle calls both with and without a `correlationId`.

sdk/javascript/client/deferred.js
peterschwarz: This looks like a git rename, but the only thing that is the same in that file is the copyright header.  Everything else is pretty much new.

boydjohnson: Ok, sounds good.

sdk/javascript/processor/index.js
```diff
@@ -43,71 +48,83 @@ class TransactionProcessor {
...
+        })
+        .catch(e => {
+          let {transactionFamilyName: familyName, version, encoding} = handler
+          console.log(`Registration of [${familyName} ${version} ${encoding}] Failed!`, e)
+        })
       })
-     */
+    })
+
+    process.on('SIGINT', () => {
```
boydjohnson: I see only SIGINT here, but SIGTERM was mentioned in the git commit message.

sdk/javascript/client/stream.js
boydjohnson: It currently follows the python and java sdks, but we could update any/all of them.

sdk/javascript/processor/index.js
```diff
@@ -43,72 +48,87 @@ class TransactionProcessor {
...
+            family: handler.transactionFamilyName,
+            version: handler.version,
+            encoding: handler.encoding,
+            namespaces: handler.namespaces
+          }).finish()
+        )
+        .then(content => TpRegisterResponse.decode(content))
+        .then(ack => {
+          let {transactionFamilyName: familyName, version, encoding} = handler
+          let status = ack.status === 0 ? 'succeded' : 'failed'
```
boydjohnson: 'succeded' should be 'succeeded'.

boydjohnson:
While running two js intkey processors under load, when I killed the validator the tps didn't always recover. When the last statement was "Received 5" so probably it had just received a TP_PROCESS_REQUEST, the js sdk didn't always recover. It depends how rock solid we want this to be. I'm unsure on how to improve on what I am seeing.

peterschwarz:
Part of that issue was due to recycling the socket id, which seemed to create odd behaviour (which I've changed).  The other part that will continue to be noticeable is due to the fact that it uses more of a polling model to monitor for disconnects - there isn't the same access to the FD, such that more low-level monitoring can be done.  So, it may take a second or two to register the disconnect (even if the validator has restarted), but when it does, it will reconnect immediately.

PR #153, opened 2017-03-02
Add "propose" sawtooth config command

peterschwarz:
Adds the subcommand `propose` under the `sawtooth config` command tree.
This command submits proposals to the validator, in order to change the
on-chain configuration settings of a running validator (and therefore
validator network).

### DIFF
--- a/cli/sawtooth_cli/config.py
+++ b/cli/sawtooth_cli/config.py
@@ -13,7 +13,9 @@
 # limitations under the License.
 # ------------------------------------------------------------------------------
 import datetime
+import getpass
 import hashlib
+import os
 
 from sawtooth_cli.exceptions import CliException
 from sawtooth_cli.rest_client import RestClient
@@ -139,7 +141,8 @@ def _read_signing_keys(key_filename):
     """Reads the given file as a WIF formatted key.
 
     Args:
-        key_filename: The filename where the key is stored.
+        key_filename: The filename where the key is stored. If None,
+            defaults to the default key for the current user.
 
     Returns:
         tuple (str, str): the public and private key pair
@@ -147,8 +150,15 @@ def _read_signing_keys(key_filename):
     Raises:
         CliException: If unable to read the file.
     """
+    filename = key_filename
+    if filename is None:
+        filename = os.path.join(os.path.expanduser('~'),
+                                '.sawtooth',
+                                'keys',
+                                getpass.getuser() + '.wif')
+
     try:
-        with open(key_filename, 'r') as key_file:
+        with open(filename, 'r') as key_file:
             wif_key = key_file.read().strip()
             signing_key = signing.encode_privkey(
                 signing.decode_privkey(wif_key, 'wif'), 'hex')
--- a/cli/sawtooth_cli/config.py
+++ b/cli/sawtooth_cli/config.py
@@ -16,6 +16,7 @@ import datetime
 import hashlib
 
 from sawtooth_cli.exceptions import CliException
+from sawtooth_cli.rest_client import RestClient
 
 from sawtooth_cli.protobuf.config_pb2 import ConfigPayload
 from sawtooth_cli.protobuf.config_pb2 import ConfigProposal
@@ -34,9 +35,16 @@ def add_config_parser(subparsers, parent_parser):
     """
     parser = subparsers.add_parser('config')
 
+    parser.add_argument(
+        '--url',
+        type=str,
+        help="the URL of the validator's REST API",
+        default='http://localhost:8080')
+
     config_parsers = parser.add_subparsers(title="subcommands",
                                            dest="subcommand")
     config_parsers.required = True
+
     set_parser = config_parsers.add_parser(
         'set',
         help='creates batches of sawtooth-config transactions')
@@ -56,12 +64,33 @@ def add_config_parser(subparsers, parent_parser):
         nargs='+',
         help='configuration setting, as a key/value pair: <key>=<value>')
 
+    propose_parser = config_parsers.add_parser(
+        'propose',
+        help='creates and submits a Propose configuration transaction')
+
+    propose_parser.add_argument(
+        '-k', '--key',
+        type=str,
+        help='the signing key for the resulting batch')
+
+    propose_parser.add_argument(
+        'setting',
+        type=str,
+        help='the configuration setting key')
+
+    propose_parser.add_argument(
+        'value',
+        type=str,
+        help='the proposed value of the setting key')
+
 
 def do_config(args):
     """Executes the config commands subcommands.
     """
     if args.subcommand == 'set':
         _do_config_set(args)
+    elif args.subcommand == 'propose':
+        _do_config_propose(args)
     else:
         raise AssertionError(
             '"{}" is not a valid subcommand of "config"'.format(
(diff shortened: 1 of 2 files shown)


cli/sawtooth_cli/config.py
```diff
@@ -75,41 +106,87 @@ def _do_config_set(args):
...
             batch_file.write(batch_list)
-    except:
-        raise CliException('Unable to write to {}'.format(args.output))
+    except IOError as ioe:
+        raise CliException(
+            'Unable to write to {}: {}'.format(args.output, ioe))
+
+
+def _do_config_propose(args):
+    """Executes the 'propose subcommand. Given a key file, a URL and a setting
```
jjason: Missing a closing quote after propose I presume...

cli/sawtooth_cli/config.py
```diff
@@ -75,41 +106,87 @@ def _do_config_set(args):
...
+    key and value, it generates a ConfigPayload.PROPOSE sawtooth_config
+    transaction and submits it via the validator REST api.
```
jjason: A nit, but since API is an acronym, should it be capitalized????

cli/sawtooth_cli/config.py
```diff
@@ -75,41 +106,87 @@ def _do_config_set(args):
...
+
+    Raises:
+        CliException: If unable to read the file.
+    """
+    filename = key_filename
+    if filename is None:
+        filename = os.path.join(os.path.expanduser('~'),
+                                '.sawtooth',
+                                'keys',
+                                getpass.getuser() + '.wif')
```
jjason: Does there need to be some centrally-located function that defaults the location of sawtooth configuration files so that every utility does not have to construct the path like this?

For example, when I fire up a validator, if not provided with a key, it assumes this is the default:

~/sawtooth/etc/keys/validator.wif

If I run 'sawtooth keygen' it defaults to the above.  Seems to me that there should be one place, and one place only, that defines the default key file name.

cli/sawtooth_cli/rest_client.py
```diff
@@ -40,19 +40,44 @@ def get_leaf(self, address, head=None):
         queries = RestClient._remove_nones(head=head)
         return self._get('/state/' + address, queries)
 
+    def send_batches(self, batch_list):
+        """Sends a list of batches to the validator.
+
+        Args:
+            batch_list (:obj:`BatchList`): the list of batches
```
jjason: Not meaning for you to fix anything, but I have seen several ways in which doc strings have been formatted.  Seems like we should settle on a "sawtooth standard"

cli/sawtooth_cli/config.py
peterschwarz: We can revisit this when the local config file stuff is in.  Particularly, it includes some path-related configuration.

cli/sawtooth_cli/rest_client.py
peterschwarz: I agree.  If we can, though, it should be something that can be linted.

cli/sawtooth_cli/config.py
vaporos: ~/sawtooth and ~/.sawtooth are different directories.  ~/sawtooth is the location we set SAWTOOTH_HOME, and is related to the validator.  ~/.sawtooth is for client CLI commands (it's the user's personal .sawtooth directory/keys).

cli/sawtooth_cli/config.py
```diff
@@ -75,41 +106,87 @@ def _do_config_set(args):
...
     batch_list = BatchList(batches=[batch]).SerializeToString()
 
     try:
         with open(args.output, 'wb') as batch_file:
             batch_file.write(batch_list)
-    except:
-        raise CliException('Unable to write to {}'.format(args.output))
+    except IOError as ioe:
+        raise CliException(
+            'Unable to write to {}: {}'.format(args.output, ioe))
```
vaporos: IOError will often provide the filename in it's message, so I'd expect the above to contain the name of the file twice.  This can be tested by setting the permissions on the directory such that you can't write to it.

cli/sawtooth_cli/rest_client.py
```diff
@@ -40,19 +40,44 @@ def get_leaf(self, address, head=None):
...
+            method='POST')
+
+        result = self._submit_request(batch_request)
+        return json.loads(result.read().decode())
+
     def _get(self, path, queries=None):
         query_string = '?' + urlencode(queries) if queries else ''
 
+        response = self._submit_request(self._base_url + path + query_string)
+        result = response.read().decode('utf-8')
```
vaporos: decode()'s default is utf-8, so it doesn't need to be explicitly set.

cli/sawtooth_cli/rest_client.py
jsmitchell: One of the reasons for choosing the format we did in the 0.7 code was that it was parseable by the sphinx plugin for doc generation.

cli/sawtooth_cli/config.py
```diff
@@ -75,41 +106,87 @@ def _do_config_set(args):
...
+                signing.decode_privkey(wif_key, 'wif'), 'hex')
+            pubkey = signing.encode_pubkey(
+                signing.generate_pubkey(signing_key), 'hex')
+
+            return pubkey, signing_key
+    except IOError:
+        raise CliException('Unable to read {}'.format(key_filename))
+
+
+def _create_batch(pubkey, signing_key, transactions):
```
jsmitchell: Add docstring

cli/sawtooth_cli/rest_client.py
```diff
@@ -40,19 +40,44 @@ def get_leaf(self, address, head=None):
...
+        return json.loads(result.read().decode())
+
     def _get(self, path, queries=None):
         query_string = '?' + urlencode(queries) if queries else ''
 
+        response = self._submit_request(self._base_url + path + query_string)
+        result = response.read().decode()
+        return json.loads(result)
+
+    def _submit_request(self, url_or_request):
```
jsmitchell: Add docstring

cli/sawtooth_cli/rest_client.py
```diff
@@ -40,19 +40,57 @@ def get_leaf(self, address, head=None):
...
+
+        Args:
+            url_or_request (str or `urlib.request.Request`): the request to
+                send.
+
+        Returns:
+            `http.client.HTTPResponse`: The response from the request.
+
+        Raises:
+            `CliException`: If any issues occurren when making the request or
```
jsmitchell: s/occurren/occur/

cli/sawtooth_cli/config.py
```diff
@@ -75,41 +106,99 @@ def _do_config_set(args):
...
+    batch = _create_batch(pubkey, signing_key, txns)
 
     batch_list = BatchList(batches=[batch]).SerializeToString()
 
     try:
         with open(args.output, 'wb') as batch_file:
             batch_file.write(batch_list)
-    except:
-        raise CliException('Unable to write to {}'.format(args.output))
+    except IOError:
```
vaporos: IOError usually has a nice user-friendly string providing the specific reason, including the filename, so a better string might be "Unable to write to batch file: {}".format(str(e))

PR #156, opened 2017-03-02
Port protogen to python

aludvik:

### DIFF
--- a/bin/protogen
+++ b/bin/protogen
@@ -1,6 +1,6 @@
-#!/bin/bash
+#!/usr/bin/env python3
 #
-# Copyright 2016, 2017 Intel Corporation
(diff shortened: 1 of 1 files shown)


aludvik:
Please also test this on a Windows machine if you have one. 

You will need to have grpcio-tools installed, which  you can get with `pip install grpcio-tools` if you have python3 installed on windows with pip on the path.

jjason:
By the way, protogen script seems to only work, at least on Windows, if executed from sawtooth-core directory (for example, `cd bin', 'python protogen` fails for me on Windows).

aludvik:
Pushed a fixup commit which fixes that issue @jjason. Will need to be squashed before merge.

PR #158, opened 2017-03-03
Add check for signed-off-by to Jenkinsfile

rbuysse:


Jenkinsfile
```diff
@@ -26,7 +26,30 @@ node ('master') {
 
         stage("Clone Repo") {
             checkout scm
+            // Check commits for signed-off-by
+            if (BRANCH_NAME == "master" || BRANCH_NAME == "0-7") {
+            } else {
```
aludvik: Why not `if (BRANCH_NAME != "master" && BRANCH_NAME != "0-7") {`?

Jenkinsfile
```diff
@@ -26,7 +26,30 @@ node ('master') {
 
         stage("Clone Repo") {
             checkout scm
+            // Check commits for signed-off-by
+            if (BRANCH_NAME == "master" || BRANCH_NAME == "0-7") {
```
vaporos: What does BRANCH_NAME here mean?  Of the destination or the source?  If the source, then wouldn't this break if the PR was created from user/sawtooth-core:master->hyperledger/sawtooth-core:master

PR #160, opened 2017-03-03
Cluster parallel

nick-drozd:


cli/sawtooth_cli/cluster.py
```diff
@@ -301,6 +300,10 @@ def do_cluster_start(args):
 
     save_state(state)
 
+    # Make sure the docker cluster network is up before starting nodes
+    if isinstance(node_controller, DockerNodeController):
```
vaporos: Use of isinstance() like this is an antipattern.  It may be better to always call the method on the controller but make the implementation on the non-docker ones "pass" (which could be done in the base class by default).

manage/sawtooth_manage/docker.py
```diff
@@ -154,20 +157,23 @@ def start(self, node_config):
         yaml.dump(compose_dict,
                   open(os.path.join(compose_dir, 'docker-compose.yaml'),
                        mode='w'))
+
+        print("Starting: {}".format(node_name))
```
vaporos: Printing within library code is usually not desirable, as you can then not use the library without it producing output.  A couple options: use logging instead, do it in a callback method.

manage/sawtooth_manage/vnm.py
```diff
@@ -23,8 +25,20 @@ def __init__(self,
...
         for command in commands:
-            command.execute(self._node_controller)
+            process = Process(
+                target=command.execute,
+                args=(self._node_controller,))
+            process.start()
+            processes.append(process)
+
+        # block until all processes have completed or timed out
+        wait_time = 30
```
vaporos: If you wanted to bound the time by a total of 30s, not 30s*len(processes), then you would want a different approach.  One similar problem is solved here:

https://github.com/vaporos/sawtooth-core/commit/ddae6c479b2cab210f3ae1f699358537608c785f#diff-935cde9b4842b2d10086c6304d39945bR315

That's not necessarily a perfect solution either, but you can see how it would be easy to bound the loop to a maximum number of iterations.

manage/sawtooth_manage/vnm.py
```diff
@@ -13,6 +13,8 @@
 # limitations under the License.
 # ------------------------------------------------------------------------------
 
+from multiprocessing import Process
```
vaporos: You could do this with threading, which would probably be desirable.  The primary use case for multiprocessing is offloading CPU.

manage/sawtooth_manage/docker.py
nick-drozd: Should we want to take it out completely? We could leave it in cluster, but then it wouldn't be accurate timewise (though that isn't as much of a problem if things are running in parallel).

cli/sawtooth_cli/cluster.py
nick-drozd: What about changing the test condition to something like `if manage_type == 'docker':`?

manage/sawtooth_manage/docker.py
nick-drozd: docker is already pretty talkative when it creates the containers.

cli/sawtooth_cli/cluster.py
```diff
@@ -302,6 +301,10 @@ def do_cluster_start(args):
 
     save_state(state)
 
+    # Make sure the docker cluster network is up before starting nodes
+    if manage_type == 'docker':
+        node_controller.start_cluster_network()
+
```
aludvik: As discussed yesterday: I think this should be moved into `manage/sawtooth_manage/docker.py` Since, as you mentioned, this would cause a race condition where all the threads try to create a docker network, you will need to use a condition variable or something similar: https://en.wikipedia.org/wiki/Monitor_(synchronization)#Condition_variables

manage/sawtooth_manage/docker.py
```diff
@@ -318,3 +304,40 @@ def is_running(self, node_name):
...
+
+    return invalid
+
+def _get_built_images():
+    docker_img_cmd = ['docker', 'images', '--format', '{{.Repository}}']
+    return subprocess.check_output(docker_img_cmd).decode().split('\n')
+
+def _built_in_processor_types():
+    image_data_dir = os.path.join(
+        os.path.dirname(__file__), '..', '..', 'docker')
```
vaporos: This is no longer a valid way to get the built-in types as these docker files will not be installed.

manage/sawtooth_manage/vnm.py
```diff
@@ -23,8 +26,21 @@ def __init__(self,
...
+        # execute commands in parallel
+        threads = []
         for command in commands:
-            command.execute(self._node_controller)
+            thread = Thread(
+                target=command.execute,
+                args=(self._node_controller,))
+            thread.start()
+            threads.append(thread)
+            time.sleep(1)
```
vaporos: The sleep here shouldn't be required.  Was this exposing a bug?

manage/sawtooth_manage/vnm.py
```diff
@@ -23,8 +26,21 @@ def __init__(self,
...
+
+        # block until all threads have completed or timed out
+        wait_time = 10
+        for thread in threads:
+            thread.join(wait_time)
```
vaporos: If this times out, are there any negative impacts (such as the command not stopping)?

manage/sawtooth_manage/vnm.py
```diff
@@ -23,8 +26,21 @@ def __init__(self,
 
     def update(self):
         commands = self._node_command_generator.get_commands()
+
+        # execute commands in parallel
+        threads = []
         for command in commands:
-            command.execute(self._node_controller)
+            thread = Thread(
```
vaporos: Wasn't the previous implementation using a threadpool?  That seemed like a good approach.

nick-drozd: Previously I was doing it with multiprocessing, but otherwise the same as it is now.

manage/sawtooth_manage/vnm.py
nick-drozd: Not that I've seen. The timeout seems to get reached only if there's an error, like there being no available address.

manage/sawtooth_manage/docker.py
nick-drozd: Should changing that be a separate commit?

manage/sawtooth_manage/vnm.py
vaporos: Sure, it could be.

manage/sawtooth_manage/docker.py
vaporos: Sure, it could be.

PR #162, opened 2017-03-03
Update POST /batches to match spec

delventhalz:
Update the `POST /batches` REST API endpoint to match the OpenAPI spec.

Includes:
- Add `link` to response envelope with appropriate `/batch_status` url
- Send back `400` when invalid batches are submitted
- Unit tests for `POST /batches`

### DIFF
--- a/protos/client.proto
+++ b/protos/client.proto
@@ -28,15 +28,21 @@ message Leaf {
     bytes data = 2;
 }
 
-// Corresponds to messagetype CLIENT_BATCH_SUBMIT_REQUEST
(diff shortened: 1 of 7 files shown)


validator/etc/txnvalidator.js
```diff
@@ -0,0 +1,70 @@
+{
```
peterschwarz: This file should not have been checked in - it's old and you should just delete your local copy.

delventhalz: Dammit.

PR #164, opened 2017-03-06
Fix PoET simulator based upon canonical key representation in signing module

jjason:
Changes include:
* Updated PoET simulator to no longer encode/decode private/public keys as the signing module uses string representing WIF/hex encodings for keys, respectively.
* Updated validator registry transaction processor accordingly to match PoET simulator.
* Re-enabled some unit tests for wait timer and certificate as they are not dependent upon final integration of new PoET consensus into the new architecture.

### DIFF
--- a/poet/sawtooth_poet/validator_registry/processor/handler.py
+++ b/poet/sawtooth_poet/validator_registry/processor/handler.py
@@ -121,11 +121,9 @@ class ValidatorRegistryTransactionHandler(object):
         '5Jz5Kaiy3kCiHE537uXcQnJuiNJshf2bZZn43CrALMGoCd3zRuo'
 
     def __init__(self):
-        self._report_private_key = \
-            signing.encode_privkey(
-                signing.decode_privkey(self.__REPORT_PRIVATE_KEY_WIF, 'wif'),
-                'hex')
-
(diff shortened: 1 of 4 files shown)


poet/sawtooth_poet/validator_registry/processor/handler.py
```diff
@@ -121,11 +121,9 @@ class ValidatorRegistryTransactionHandler(object):
         '5Jz5Kaiy3kCiHE537uXcQnJuiNJshf2bZZn43CrALMGoCd3zRuo'
 
     def __init__(self):
-        self._report_private_key = \
-            signing.encode_privkey(
-                signing.decode_privkey(self.__REPORT_PRIVATE_KEY_WIF, 'wif'),
-                'hex')
-
+        # Since signing works with WIF-encoded private keys, we don't have to
```
jsmitchell: It feels like most of these comments are explanations of the change (commit entry type stuff), rather than a description of what the code is doing. I'd recommend removing all the comments about not having to encode/decode keys.

jjason: Normally, I would agree, but after spending part of my Friday afternoon tracking down why the simulator was failing, and it was because of this canonical key representation change, I would prefer to leave the encode/decode comments in.

PR #166, opened 2017-03-06
On rebroadcast of a gossip message, do not resend to node who you received it from

agunde406:
This stops the state were validators will continuously send gossip messages back and forth forever.

### DIFF
--- a/validator/sawtooth_validator/gossip/gossip.py
+++ b/validator/sawtooth_validator/gossip/gossip.py
@@ -61,12 +61,13 @@ class Gossip(object):
                 LOGGER.debug("Attempt to unregister identity %s failed: "
                              "identity was not registered")
 
-    def broadcast_block(self, block):
+    def broadcast_block(self, block, exclude=None):
         gossip_message = GossipMessage(
             content_type="BLOCK",
             content=block.SerializeToString())
 
-        self.broadcast(gossip_message, validator_pb2.Message.GOSSIP_MESSAGE)
+        self.broadcast(
+            gossip_message, validator_pb2.Message.GOSSIP_MESSAGE, exclude)
 
     def broadcast_block_request(self, block_id):
         # Need to define node identity to be able to route directly back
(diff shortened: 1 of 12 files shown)


validator/sawtooth_validator/gossip/gossip.py
```diff
@@ -99,16 +102,35 @@ def broadcast_batch_by_batch_id_request(self, batch_id):
             batch_request,
             validator_pb2.Message.GOSSIP_BATCH_BY_BATCH_ID_REQUEST)
 
-    def broadcast(self, gossip_message, message_type):
+    def broadcast(self, gossip_message, message_type, exclude=None):
```
peterschwarz: This signature is starting to get complicated enough, that we should probably docstring it. Particularly, we should document this new parameter exclude, as it takes a bit of searching through the code to understand what it is (a list) and what is in it (pairs of a particular order).  Something about that starts to feel a little brittle...

validator/sawtooth_validator/gossip/gossip.py
```diff
@@ -99,16 +102,35 @@ def broadcast_batch_by_batch_id_request(self, batch_id):
...
         # Gossip broadcasts are sent out in two different ways.
         # 1) To connected identities on our server socket
         # 2) To outbound connections we have originated
+        identities = copy.copy(self._identities)
+        connections = copy.copy(self._network.connections)
+
+        # Remove any identities that are in the exclude list
         for identity in self._identities:
+            if exclude is not None:
```
peterschwarz: This could defaulted at the start of the function to an empty list., should make this block, and the connection version a bit clearer.

validator/sawtooth_validator/gossip/gossip.py
```diff
@@ -99,16 +102,45 @@ def broadcast_batch_by_batch_id_request(self, batch_id):
...
+                if peer[1] == identity:
+                    identities.remove(identity)
+
+        # Remove any connections in the exclude list
+        for connection in self._network.connections:
+            for peer in exclude:
+                if peer[0] == connection.local_id:
+                    connections.remove(connection)
+
+        for identity in identities:
```
jsmitchell: instead of making copies and modifying list objects, you could probably do set subtraction here in the loop or just put a conditional check in to determine if you send or not

jsmitchell:
```eg.

excluded_inbound_peers = [peer[1] for peer in exclude]
for identity in list(set(identities) - set(excluded_inbound_peers)):

or just:
for identity in identities:
    if identity not in excluded_inbound_peers:
        send()

PR #167, opened 2017-03-06
Update intkey smoke test

nick-drozd:
Signed-off-by: Nick Drozd <nicholasx.drozd@intel.com>

### DIFF
--- a/integration/sawtooth_integration/tests/test_intkey_smoke.py
+++ b/integration/sawtooth_integration/tests/test_intkey_smoke.py
@@ -24,10 +24,9 @@ import json
 from base64 import b64decode
 
 import cbor
-# ------------------------------
 
 from sawtooth_integration.intkey.intkey_message_factory \
-    import IntkeyMessageFactory # used by verifier
+    import IntkeyMessageFactory
 
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


integration/sawtooth_integration/tests/test_intkey_smoke.py
```diff
@@ -24,10 +24,10 @@
 from base64 import b64decode
 
 import cbor
-# ------------------------------
 
 from sawtooth_integration.intkey.intkey_message_factory \
-    import IntkeyMessageFactory # used by verifier
+    import IntkeyMessageFactory
+# ------------------------------
```
vaporos: This line could be removed.

integration/sawtooth_integration/tests/test_intkey_smoke.py
```diff
@@ -138,8 +140,8 @@ def _query_rest_api(suffix='', data=None, headers={}):
 
 class IntkeyTestVerifier:
     def __init__(self,
-                 valid=('cow', 'pig', 'sheep', 'goat', 'horse'),
-                 invalid=('lark', 'thrush', 'jay', 'wren', 'finch'),
+                 valid=('lark', 'thrush', 'jay', 'wren', 'finch'),
```
vaporos: Why are these switched?

nick-drozd: I decided I would rather have the passerine birds be the valid transactions. But I can change it back if you prefer the farm animals. It's _udderly_ inconsequential.

PR #168, opened 2017-03-06
Add warning for Edge/IE that those browsers are not supported

delventhalz:

### DIFF
--- a/extensions/bond/ui/client/src/main/bond/components/auth.cljs
+++ b/extensions/bond/ui/client/src/main/bond/components/auth.cljs
@@ -18,6 +18,7 @@
             [sablono.core :as html :refer-macros [html]]
             [sawtooth.state :refer [app-state]]
             [sawtooth.ledger.keys :refer [address get-key-pair]]
+            [sawtooth.utils :refer [browser]]
             [sawtooth.router :as router]
             [sawtooth.components.core
              :refer-macros [when-new-block handle-event]]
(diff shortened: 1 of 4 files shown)


extensions/bond/ui/client/src/main/bond/components/auth.cljs
```diff
@@ -43,7 +44,17 @@
           [:p "Sawtooth Bond is a proof-of-concept bond trading platform,
               built on the Sawtooth Lake distributed ledger."]
           [:p "To begin you will need to create an identity by generating a new
-              Wallet Import Format (WIF) key or importing an existing one."]]
+              Wallet Import Format (WIF) key or importing an existing one."]
+           (when (or (= :edge (browser)) (= :ie (browser)))
```
peterschwarz: Should we also show this warning if the browser is `nil` (unknown)?

extensions/bond/ui/client/src/main/sawtooth/utils.cljs
```diff
@@ -63,7 +63,9 @@
         agent-contains #(re-find (re-pattern %) user-agent)]
         (cond (agent-contains "Firefox") :firefox
               (agent-contains "OPR") :opera
+              (or (agent-contains "MSIE") (agent-contains "Trident")) :ie
+              ; Order important: Edge lists Chrome/Safari, Chrome lists Safari
+              (agent-contains "Edge") :edge
               (agent-contains "Chrome") :chrome
               (agent-contains "Safari") :safari
-              (or (agent-contains "MSIE") (.-documentMode js/document)) :ie
               :default nil)))))
```
peterschwarz: Perhaps we should return `:unknown`

delventhalz: Maybe. I like the semantics of `nil` though. `:unknown` feels somewhat arbitrary, while `nil`, as a falsey value, clearly communicates to me that the browser couldn't be found.

PR #169, opened 2017-03-07
Implement heartbeat ping on inbound connections

jsmitchell:
This moves GOSSIP_PING and GOSSIP_ACK to lower level NETWORK
message types, moves the PingHandler from the gossip package to
the networking package, and implements heartbeat pings over
connections on the ZMQ ROUTER socket.

### DIFF
--- a/validator/sawtooth_validator/networking/interconnect.py
+++ b/validator/sawtooth_validator/networking/interconnect.py
@@ -20,6 +20,7 @@ import logging
 from threading import Condition
 from threading import Thread
 import uuid
+import time
 
 import zmq
 import zmq.auth
@@ -29,6 +30,7 @@ import zmq.asyncio
 from sawtooth_validator.exceptions import LocalConfigurationError
 from sawtooth_validator.protobuf import validator_pb2
 from sawtooth_validator.networking import future
+from sawtooth_validator.protobuf.network_pb2 import PingRequest
 
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -69,12 +75,44 @@ def __init__(self, connection, address, futures, identity=None,
...
+        self._connected_identities = {}
+
+    @asyncio.coroutine
+    def _send_heartbeat(self):
+        with self._condition:
+            self._condition.wait_for(lambda: self._socket is not None)
+
+        ping = PingRequest()
+
+        while True:
```
peterschwarz: How does this shutdown?

jsmitchell: It's an asyncio coroutine. It yields until the event loop stops servicing it.

There is a broader question here about cleanup of connections in the dict once they are no longer live. This PR doesn't attempt to address that. Detecting stale connections and doing appropriate connection cleanup will occur in a future PR.

validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -246,7 +301,8 @@ def __init__(self,
                     identity=identity,
                     secured=secured,
                     server_public_key=server_public_key,
-                    server_private_key=server_private_key)
+                    server_private_key=server_private_key,
+                    heartbeat=heartbeat)
```
boydjohnson: It appears to me that the dealer socket can never send a PingRequest, but Connections are always Dealers so the _send_heartbeat is getting setup on the event loop but is a conditional that is always false and then an asyncio.sleep.

jsmitchell: Setting up the ensure_future is also conditional:
```
if self._heartbeat:
    asyncio.ensure_future(self._send_heartbeat(),
                                          loop=self._event_loop)

boydjohnson: But heartbeat is True for the network instance set in core.py and that True is passed in on every Connection. It should probably just not be passed in to Connection.

jsmitchell: Oh, right you are. Change incoming.

PR #173, opened 2017-03-07
Add wait-for-commit to batch status requests and batch submits

delventhalz:
Add the capability to specify that requests for batch status, or submissions of new batches, should wait to return until all the batches have been committed to the blockchain, or a timeout has expired.

This covers both the REST and ZMQ APIs, and adds unit tests for all new capabilities.


delventhalz:
PR has been updated to include wait-for-commit capabilities to new batch submissions.

PR #176, opened 2017-03-08
Graceful validator shutdown

boydjohnson:
After this PR a validator will be able to gracefully shutdown with one SIGINT or SIGTERM if given time, from a resting state. This doesn't handle shutdown under load, as that still isn't always graceful.

The run_docker_test script is modified for not stopping until the test container exits, so that other containers can exit which should be orchestrated by the test container.

### DIFF
--- a/bin/run_docker_test
+++ b/bin/run_docker_test
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from subprocess import run, PIPE, CalledProcessError
+from subprocess import run, PIPE, CalledProcessError, Popen
 import os
 import sys
 import argparse
@@ -46,12 +46,13 @@ def main():
     ]
 
     compose_up = compose + [
-        'up', '--abort-on-container-exit'
+        'up', '-d'
     ]
 
-    compose_down = compose + ['down', '--remove-orphans']
+    docker_wait = ['docker', 'wait',
+                    "{}_{}_1".format(isolation_id, args.test_service)]
 
-    compose_kill = compose + ['kill']
+    compose_down = compose + ['down', '--remove-orphans']
 
     inspect = [
         'docker', 'inspect',
@@ -65,6 +66,9 @@ def main():
             "Test service '{}' does not exist in compose file: '{}'".format(
                 args.test_service, compose_file))
 
+    compose_logs = compose + ['logs', '--follow'] + \
+                   [s for s in compose_dict['services']]
+
     start_time = time.time()
     timeout = args.timeout
 
@@ -76,6 +80,12 @@ def main():
         try:
             print("Running: {}".format(compose_up))
             run(compose_up, check=True, timeout=timeout)
+            handle = Popen(args=compose_logs, stdout=PIPE,
+                           stderr=PIPE,
+                           universal_newlines=True)
+            print("Running: {}".format(docker_wait))
+            run(docker_wait, check=True, timeout=timeout)
+
         except CalledProcessError:
             raise RunDockerTestError(
                 "Failed to start all docker containers. Do `docker ps -a` to "
@@ -113,6 +123,8 @@ def main():
                 "Failed to remove all docker containers. Do `docker ps -a` to "
                 "list residual containers.")
 
+        sys.stdout.write(handle.stdout.read())
+        sys.stdout.write(handle.stderr.read())
         exit(int(success))
 
     except KeyboardInterrupt:
--- a/bin/run_tests
+++ b/bin/run_tests
@@ -222,6 +222,8 @@ test_signing() {
 
 test_validator() {
     run_docker_test ./validator/tests/unit_validator.yaml -s validator
+    run_docker_test shutdown-smoke.yaml -s integration_test
 }
+
 
 main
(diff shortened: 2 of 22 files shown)


bin/run_docker_test
```diff
@@ -71,6 +75,11 @@ def main():
         try:
             print("Running: {}".format(compose_up))
             run(compose_up, check=True, timeout=timeout)
+            handle = Popen(args=compose_logs, stdout=PIPE,
```
aludvik: I think you may also want to add `, stderr=PIPE` and add `sys.stdout.write(handle.stderr.read())` below.

aludvik:
I have verified that if you run genesis, startup a validator, startup a transaction processor, shutdown the validator with Ctrl+C, clear out the state, re-run genesis, and restart the validator, everything works correctly and gracefully.

Things that do not work but might be outside the scope of the PR:
1. Running `intkey load` on the validator, waiting for the batch to finish, then shutting down the validator.
2. Not clearing out the state and re-running genesis after shutting down the validator and restarting it.

bin/run_tests
```diff
@@ -140,6 +140,7 @@ test_signing() {
 test_validator() {
     echo "[---Running validator tests---]"
     run_docker_test ./validator/tests/unit_validator.yaml -s validator -p $ISOLATION_ID
+    run_docker_test shutdown-smoke.yaml -s integration_test -p $ISOLATION_ID
 }
```
aludvik: You can omit the .yaml so it's just `shutdown-smoke`.

bin/run_docker_test
boydjohnson: Yeah, I think you are right.

bin/run_tests
boydjohnson: In that file there are a mix of .yaml and not. It seems since the "compose_file" arg is being passed directly into docker-compose, in run_docker_test, we would be relying on docker-compose to figure out that a particular file has .yaml at the end or .yml. It does do this, but I don't think we should rely upon it.

validator/sawtooth_validator/execution/context_manager.py
```diff
@@ -308,8 +308,8 @@ def _squash(state_root, context_ids):
         return _squash
 
     def stop(self):
-        self._context_writer.join(1)
-        self._context_reader.join(1)
+        self._address_queue.put_nowait(-1)
```
vaporos: Declaring a module-level variable for -1 and using it here and below would make it clearer.

validator/sawtooth_validator/execution/processor_iterator.py
```diff
@@ -32,6 +32,17 @@ class ProcessorIteratorCollection(object):
     (1 transaction processor can have multiple Handlers which will be stored
     as Processors)
 
+    Attributes:
```
vaporos: Internal attributes probably shouldn't be in the API doc string (as they are not part of the API), but rather a comment within __init__() itself as #-style comments above each attribute.

validator/sawtooth_validator/execution/processor_iterator.py
```diff
@@ -112,9 +123,30 @@ def remove(self, processor_identity):
     def __repr__(self):
         return ",".join([repr(k) for k in self._processors.keys()])
 
-    def wait_to_process(self, item):
+    def cancellable_wait(self, processor_type, cancelled_event):
+        """Waits for a particular processor type to register or until
```
vaporos: Text wrapping/indentation is odd with this comment, and there should be an empty line prior to Args:.

validator/sawtooth_validator/execution/executor.py
```diff
@@ -301,3 +341,21 @@ def __setitem__(self, key, value):
...
+    def __len__(self):
+        with self._lock:
+            return len(self._waiters)
+
+    def values(self):
+        """
+        Returns (list of _Waiter): All the _Waiter objects.
+        """
+        with self._lock:
+            return self._waiters.values()
```
vaporos: In python 3, this will return a dict_values object, so this code is probably not thread-safe currently.  You probably want: return list(self._waiters.values()).

validator/sawtooth_validator/execution/executor.py
```diff
@@ -301,3 +341,21 @@ def __setitem__(self, key, value):
...
+
+    def keys(self):
+        """
+        Returns (list of _ProcessorType): All the ProcessorTypes waited for
+        """
+        with self._lock:
+            return self._waiters.keys()
```
vaporos: This should probably be return list(self._waiters.keys()).

integration/sawtooth_integration/tests/test_shutdown_smoke.py
```diff
@@ -0,0 +1,111 @@
...
+
+import logging
+import subprocess
+import time
+
+LOGGER = logging.getLogger(__name__)
+LOGGER.setLevel(logging.DEBUG)
+
+
+class TestShutdownSmoke(unittest.TestCase):
```
vaporos: It might be worth (in a future PR) using the manage library to abstract the implementation specifics so the test itself isn't docker specific.  (This is not a request to update this PR.)

PR #177, opened 2017-03-08
Avoid committing empty state values

peterschwarz:
Avoid committing state values from the context that are None.  This
saves the context manager from writing all values from a transactions
output list, even if only some of those values have been written to by
the transaction processor.

### DIFF
--- a/validator/sawtooth_validator/execution/context_manager.py
+++ b/validator/sawtooth_validator/execution/context_manager.py
@@ -224,8 +224,13 @@ class ContextManager(object):
             merged_updates.update(context.get_writable_address_value_dict())
 
         new_root = merkle_root
-        add_value_dict = {address: value.result()
-                          for address, value in merged_updates.items()}
+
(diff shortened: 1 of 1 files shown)


validator/sawtooth_validator/execution/context_manager.py
```diff
@@ -224,8 +224,13 @@ def commit_context(self, context_id_list, virtual):
             merged_updates.update(context.get_writable_address_value_dict())
 
         new_root = merkle_root
-        add_value_dict = {address: value.result()
-                          for address, value in merged_updates.items()}
```
cintel: is this better remaining as a comprehension:
add_value_dict = {address: value.result()
                          for address, value in merged_updates.items() if value.result() is not None}

peterschwarz: value result has a `condition.wait(2)` in it, so it isn't a clean getter

PR #178, opened 2017-03-08
Add validator protobuf output directory for validator registry protobuf result

jjason:
In order to build validator registry transactions in the PoET consensus block publisher, it needs the resulting protobuf Python for the validator registry transaction to be in the same directory as all other protobuf resulting files used by the validator.

### DIFF
--- a/bin/protogen
+++ b/bin/protogen
@@ -50,6 +50,7 @@ def main():
     # 3. Generate PoET protos
     proto_dir = JOIN(TOP_DIR, "poet/sawtooth_poet/protos")
     protoc(proto_dir, "poet", "sawtooth_poet/protobuf")
+    protoc(proto_dir, "validator", "sawtooth_validator/protobuf")
 
 
 def protoc(src_dir, base_dir, pkg):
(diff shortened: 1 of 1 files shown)


bin/protogen
```diff
@@ -50,6 +50,7 @@ def main():
     # 3. Generate PoET protos
     proto_dir = JOIN(TOP_DIR, "poet/sawtooth_poet/protos")
     protoc(proto_dir, "poet", "sawtooth_poet/protobuf")
+    protoc(proto_dir, "validator", "sawtooth_validator/protobuf")
```
peterschwarz: You can put this into `sawtooth_poet/protobuf` as well, so we can avoid an possible `PYTHONPATH` conflicts

jjason:
Pushed up new change that fixes lint error

bin/protogen
peterschwarz: Misunderstood what was happening here.

jjason:
I'm trying to inflate my commit count ;)

peterschwarz:
He didn't introduce the lint error, so now the commits can be cherry picked, if need be.

PR #185, opened 2017-03-09
Move validator registry view

jjason:
Because this is purpose-written for PoET consensus, moving to live with it.

Signed-off-by: Jamie Jason <jamie.jason@intel.com


jjason:
Pushing again...for some reason, passed unit tests on local system, but was failing on build system.

PR #190, opened 2017-03-10
Update BlockVerifierInterface.verify_block signature and documentation to reflect actual usage

jjason:


peterschwarz:
Since we can't enforce this type change, did you verify that this works everywhere this function is used?

jjason:
@peterschwarz Verified that calling sights provide BlockWrapper and also verified with @cintel that this is the case.  Learned the hard way when implementing PoET consensus block verifier that it takes a BlockWrapper and not a Block.

PR #192, opened 2017-03-10
Add connection collection to Interconnect and stop Rebroadcasting to originally sender

agunde406:
This includes replacing keeping track of both outgoing and incoming connections separately and consolidates them into one dictionary owned by interconnect. This allows gossip handlers to know nothing about the type of connection that it is broadcasting to. 

Also fixes the problem with peered validators rebroadcasting the same messages over and over again.

### DIFF
--- a/validator/sawtooth_validator/networking/interconnect.py
+++ b/validator/sawtooth_validator/networking/interconnect.py
@@ -45,8 +45,8 @@ def get_enum_name(enum_value):
 
 
 class _SendReceive(object):
-    def __init__(self, connection, address, futures, identity=None,
-                 dispatcher=None, secured=False,
+    def __init__(self, connection, address, futures, connections,
+                 zmq_identity=None, dispatcher=None, secured=False,
                  server_public_key=None, server_private_key=None,
                  heartbeat=False, heartbeat_interval=10):
         """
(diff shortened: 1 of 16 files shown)


validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -253,7 +276,7 @@ class Interconnect(object):
     def __init__(self,
                  endpoint,
                  dispatcher,
-                 identity=None,
+                 zmq_identity=None,
```
peterschwarz: Is this different from the `connection_id` that we see elsewhere?  If not, we should use the same name.

agunde406: Yes, they are different. This is the zmq_identity which represents incoming connections. The connection_id is either a hash of this identity or a hash of the connection string for an OutboundConnection object.

validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -56,6 +56,14 @@ def __init__(self, connection, address, futures, identity=None,
             connection (str): A locally unique identifier for this
                 thread's connection. Used to identify the connection
                 in the dispatcher for transmitting responses.
+            futrues (future.FutureCollection): A Map of correlation ids to
```
vaporos: typo here - futrues

PR #196, opened 2017-03-11
Add capability for clients to request batches

delventhalz:
End-to-end solution for batch fetching. Includes:
  - `batch list` and `batch show` commands for the Sawtooth CLI
  - `GET /batches` and `GET /batches/{batch_id}` routes for the REST API
  - `ClientBatchList` and `ClientBatchGet` messages for the ZMQ API
  - Unit tests for REST and ZMQ APIs
  - Major refactor to `client_handlers.py` to be more straightforward, better documented
  - Various minor bug fixes and cleanup

### DIFF
--- a/cli/sawtooth_cli/batch.py
+++ b/cli/sawtooth_cli/batch.py
@@ -1,0 +1,176 @@
+# Copyright 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the 'License');
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+
+import sys
+import csv
+import json
+import yaml
+
+from sawtooth_cli import tty
+from sawtooth_cli.rest_client import RestClient
+from sawtooth_cli.exceptions import CliException
+
+
+def add_batch_parser(subparsers, parent_parser):
+    """Adds arguments parsers for the batch list and batch show commands
+
+        Args:
+            subparsers: Add parsers to this subparser object
+            parent_parser: The parent argparse.ArgumentParser object
+    """
+    parser = subparsers.add_parser('batch')
+
+    grand_parsers = parser.add_subparsers(title='grandchildcommands',
+                                          dest='subcommand')
+    grand_parsers.required = True
+    epilog = '''details:
+        Lists committed batches from newest to oldest, including their id (i.e.
+    header signature), transaction count, and their signer's public key.
+    '''
+
+    list_parser = grand_parsers.add_parser('list', epilog=epilog)
+    list_parser.add_argument(
+        '--url',
+        type=str,
+        help="the URL of the validator's REST API")
+    list_parser.add_argument(
+        '-F', '--format',
+        action='store',
+        default='default',
+        choices=['csv', 'json', 'yaml', 'default'],
+        help='the format of the output, options: csv, json or yaml')
+
+    epilog = '''details:
+        Shows the data for a single batch, or for a particular property within
+    that batch or its header. Displays data in YAML (default), or JSON formats.
+    '''
+    show_parser = grand_parsers.add_parser('show', epilog=epilog)
+    show_parser.add_argument(
+        'batch_id',
+        type=str,
+        help='the id (i.e. header_signature) of the batch')
+    show_parser.add_argument(
+        '--url',
+        type=str,
+        help="the URL of the validator's REST API")
+    show_parser.add_argument(
+        '-k', '--key',
+        type=str,
+        help='specify to show a single property from the batch or header')
+    show_parser.add_argument(
+        '-F', '--format',
+        action='store',
+        default='yaml',
+        choices=['yaml', 'json'],
+        help='the format of the output, options: yaml (default), or json')
+
+
+def do_batch(args):
+    """Runs the batch list or batch show command, printing output to the console
+
+        Args:
+            args: The parsed arguments sent to the command at runtime
+    """
+
+    rest_client = RestClient(args.url)
+
+    def print_json(data):
+        print(json.dumps(
+            data,
+            indent=2,
+            separators=(',', ': '),
+            sort_keys=True))
+
+    def print_yaml(data):
+        print(yaml.dump(data, default_flow_style=False)[0:-1])
+
+    if args.subcommand == 'list':
+        batches = rest_client.list_batches()
+        keys = ('batch_id', 'txns', 'signer')
+        headers = (k.upper() for k in keys)
+
+        def get_data(batch):
+            return (
+                batch['header_signature'],
+                len(batch.get('transactions', [])),
+                batch['header']['signer_pubkey'])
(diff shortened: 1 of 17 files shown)


cli/sawtooth_cli/block.py
```diff
@@ -94,50 +91,52 @@ def print_yaml(data):
...
 
-        def get_block_data(block):
+        def get_data(block):
             batches = block.get('batches', [])
-            txn_count = reduce(
-                lambda t, b: t + len(b.get('transactions', [])),
-                batches,
-                0)
-
+            txns = [t for b in batches for t in b.get('transactions', [])]
```
boydjohnson: This works, but it is more readable if it goes txns = [t for t in b.get('transactions', []) for b in batches].

delventhalz: I think I tried something like that, but it didn't work. Must have had it slightly off. Damn list comprehensions...

rest_api/sawtooth_rest_api/routes.py
```diff
@@ -196,6 +199,44 @@ def block_get(self, request):
...
+
+        batches = [RouteHandler._expand_batch(b) for b in response['batches']]
+        return RouteHandler._wrap_response(
+            data=batches,
+            metadata=RouteHandler._get_metadata(request, response))
+
+    @asyncio.coroutine
+    def batch_get(self, request):
+        """
+        Fetch a particular from the validator
```
agunde406: Missing word batch?

cli/sawtooth_cli/block.py
boydjohnson: I retract my comment. Apparently the more readable way doesn't work.

delventhalz: It should. But it doesn't.

cli/sawtooth_cli/batch.py
```diff
@@ -0,0 +1,151 @@
+# Copyright 2016 Intel Corporation
```
peterschwarz: 2017

cli/sawtooth_cli/batch.py
```diff
@@ -0,0 +1,151 @@
...
+    show_parser.add_argument(
+        '-f', '--format',
+        action='store',
+        default='yaml',
+        choices=['yaml', 'json'],
+        help='the format of the output, options: yaml (default), or json')
+
+
+def do_batch(args):
+
```
peterschwarz: docstring

cli/sawtooth_cli/block.py
```diff
@@ -94,50 +91,52 @@ def print_yaml(data):
...
-                block['header']['signer_pubkey']
-            )
+                len(txns),
+                block['header']['signer_pubkey'])
 
         if args.format == 'default':
-            # Fit within 150 chars, without truncating block id
-            print('{:<3}  {:128.128}  {:<4}  {:<4}  {:11.11}'.format(*headers))
+            # Set column widths based on window and data size
+            _, window_width = os.popen('stty size', 'r').read().split()
```
peterschwarz: Maybe not a worry for this PR, but this will prevent the CLI's from running on Windows.

cli/sawtooth_cli/state.py
```diff
@@ -96,32 +95,38 @@ def print_yaml(data):
...
                 leaf['address'],
-                len(decoded_data),
-                decoded_data if use_decoded else data)
+                len(decoded),
+                str(decoded) if decode else leaf['data'])
 
         if args.format == 'default':
-            print(format_string.format(*headers))
+            # Set column widths based on window and data size
+            _, window_width = os.popen('stty size', 'r').read().split()
```
peterschwarz: Same Win Concern

cli/sawtooth_cli/state.py
```diff
@@ -131,21 +136,18 @@ def get_leaf_data(leaf, use_decoded=True):
...
+                'data': [{k: d for k, d in
+                         zip(keys, get_leaf_data(l, False))}
+                         for l in leaves]}
 
             if args.format == 'json':
                 print_json(state_data)
             else:
                 print_yaml(state_data)
 
-        else:
```
peterschwarz: This is a protection against developer error, it should stay.

cli/sawtooth_cli/block.py
```diff
@@ -30,8 +29,7 @@ def add_block_parser(subparsers, parent_parser):
     grand_parsers = parser.add_subparsers(title='grandchildcommands',
                                           dest='subcommand')
     grand_parsers.required = True
-    epilog = '''
-    details:
+    epilog = '''details:
```
vaporos: The epilog output doesn't look right when output. I think we need to use another formatter (the raw one probably) so that epilog is output properly.  If we don't address it here, we should open up a JIRA task to go through the CLI and make it all consistent.

cli/sawtooth_cli/batch.py
```diff
@@ -0,0 +1,151 @@
...
+    header signature), transaction count, and their signer's public key.
+    '''
+
+    list_parser = grand_parsers.add_parser('list', epilog=epilog)
+    list_parser.add_argument(
+        '--url',
+        type=str,
+        help="the URL of the validator's REST API")
+    list_parser.add_argument(
+        '-f', '--format',
```
vaporos: '-f' is commonly used to mean input 'file' or 'force' on many Linux commands, so avoiding using it for other purposes is usually desirable.  Could use '-F', assuming we want '-F' to mean format across the CLI and not something else (we also have --filter in some places).

cli/sawtooth_cli/block.py
vaporos: We should retain Windows support for the CLI.

We should also avoid using the tty size when the output is being redirected to a file "sawtooth block list > foo", as well as running without a tty such as with ssh: "ssh localhost sawtooth block list", in which case stty gives different error output based on platform.  sys.stdout.isatty() might be useful for this purpose.

validator/sawtooth_validator/state/client_handlers.py
```diff
@@ -31,176 +31,316 @@
...
+    Args:
+        request_proto (class): Protobuf class of the request to be handled
+        response_proto (class): Protobuf class of the response to be sent
+        response_type (enum): Message status of the response
+        tree (MerkleDatabase, optional): State tree to be queried
+        block_store (BlockStoreAdapter, optional): Block chain to be queried
+        batch_cache (TimedCache, optional): A cache of Batches being processed
+
+    Attributes:
+        _status (class): Convenience ref to response_proto for accessing enums
```
delventhalz: I think I actually want to keep this. Our linter allows inherited classes to access `_` methods, but will complain if any outside code tries to. As `handle` is the only method I want used outside of this module, that is exactly the sort of behavior I'm looking for.

Since there doesn't seem a strong style preference in the repo to break that protection by writing inherited methods without an `_`, I think it should stay in this case.

cli/sawtooth_cli/block.py
delventhalz: I can write a fallback.

cli/sawtooth_cli/block.py
delventhalz: Will do

cli/sawtooth_cli/batch.py
```diff
@@ -0,0 +1,176 @@
...
+                print(batch[args.key])
+            elif args.key in batch['header']:
+                print(batch['header'][args.key])
+            else:
+                raise CliException(
+                    'key "{}" not found in batch or header'.format(args.key))
+        else:
+            if args.format == 'yaml':
+                print_yaml(batch)
+            if args.format == 'json':
```
vaporos: This could be elif, with an else clause as an assertion.  (Or the default in the else clause.)

PR #197, opened 2017-03-12
Fix POST /batches sometimes using wrong status

delventhalz:
With `including_default_value_fields` set to `True` in the `MessageToDict` function, `batch_statuses` will always come back with a value, even if not set.


peterschwarz:
It's possible that we can undo the use of `including_default_value_fields`, as the root cause was the ContextManager not throwing away unused state values (fixed in #177).

peterschwarz:
Although testing that, its still an issue for the cli tools.  This fix is probably correct, then.

PR #198, opened 2017-03-12
Add test for CLI keygen command

ojalatodd:
Test written for command 'sawtooth admin keygen'.
Tests for valid key as output of keygen, and
test of --force option working to overwrite
existing key files.

### DIFF
--- a/cli/tests/test_admin_keygen.py
+++ b/cli/tests/test_admin_keygen.py
@@ -1,0 +1,120 @@
+# Copyright 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+
+import argparse
+import os
+import unittest
+
+from sawtooth_signing import secp256k1_signer as signing
+from sawtooth_cli.admin_command import keygen
+from sawtooth_cli.admin_command.utils import ensure_directory
+from sawtooth_cli.exceptions import CliException
+
+
+class TestKeygen(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        cls._parser = None
+        cls._key_dir = ensure_directory('etc/keys', '/etc/sawtooth/keys')
+        cls._key_name = 'test_key'
+        cls._wif_filename = os.path.join(cls._key_dir,
+                                         cls._key_name + '.wif')
+        cls._addr_filename = os.path.join(cls._key_dir,
+                                          cls._key_name + '.addr')
+
+    def setUp(self):
+        self._parser = argparse.ArgumentParser(add_help=False)
+
+        parent_parser = argparse.ArgumentParser(prog='test_keygen',
+                                                add_help=False)
+
+        subparsers = self._parser.add_subparsers(title='subcommands',
+                                                 dest='command')
+
+        keygen.add_keygen_parser(subparsers, parent_parser)
+
+    def tearDown(self):
+        self.remove_key_files()
+
+    def _parse_keygen_command(self, *args):
+        cmd_args = ['keygen']
+        cmd_args += args
+        return self._parser.parse_args(cmd_args)
+
+    def test_writes_valid_key(self):
+        """Remove existing test files before creating new keys.
+        """
+        self.remove_key_files()
+
+        args = self._parse_keygen_command(self._key_name)
+        keygen.do_keygen(args)
+
+        pubkey, signing_key = _read_signing_keys(self._wif_filename)
+
+        self.assertIsNotNone(pubkey)
+        self.assertIsNotNone(signing_key)
+
+    def test_force_write(self):
+        """Write key files to be overwritten by test of --force option.
+        """
+        self.remove_key_files()
+        args = self._parse_keygen_command(self._key_name)
+        keygen.do_keygen(args)
+
+        args = self._parse_keygen_command('--force', self._key_name)
+        keygen.do_keygen(args)
+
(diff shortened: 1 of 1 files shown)


ojalatodd:
Tests currently failing on a different test...

aludvik:
Looks like you will need to rebase after 197 goes in.

ojalatodd:
Will rebase and retry. 
@aludvik But otherwise the test looks OK?

cli/tests/test_admin_keygen.py
```diff
@@ -0,0 +1,112 @@
...
+        parent_parser = argparse.ArgumentParser(prog='test_keygen',
+                                                add_help=False)
+
+        subparsers = self._parser.add_subparsers(title='subcommands',
+                                                 dest='command')
+
+        keygen.add_keygen_parser(subparsers, parent_parser)
+
+    def tearDown(self):
+        # shutil.rmtree(self._temp_dir)
```
aludvik: Remove commented code if it's not needed.

cli/tests/test_admin_keygen.py
```diff
@@ -0,0 +1,112 @@
...
+        self._parser = None
+        self._key_dir = ensure_directory('etc/keys', '/etc/sawtooth/keys')
+        self._key_name = 'test_key'
+        self._wif_filename = os.path.join(self._key_dir,
+                                          self._key_name + '.wif')
+        self._addr_filename = os.path.join(self._key_dir,
+                                           self._key_name + '.addr')
+
+    def setUp(self):
+        # self._temp_dir = tempfile.mkdtemp()
```
aludvik: Remove commented code if not needed.

cli/tests/test_admin_keygen.py
```diff
@@ -0,0 +1,112 @@
...
+        # shutil.rmtree(self._temp_dir)
+        pass
+
+    def _parse_keygen_command(self, *args):
+        cmd_args = ['keygen']
+        cmd_args += args
+        return self._parser.parse_args(cmd_args)
+
+    def test_writes_valid_key(self):
+        # Remove existing test files before creating new keys
```
aludvik: Can these comments be doc strings like here: https://github.com/hyperledger/sawtooth-core/blob/master/rest_api/tests/unit/test_rest_api.py#L83

cli/tests/test_admin_keygen.py
```diff
@@ -0,0 +1,112 @@
...
+
+        self._parser = argparse.ArgumentParser(add_help=False)
+
+        parent_parser = argparse.ArgumentParser(prog='test_keygen',
+                                                add_help=False)
+
+        subparsers = self._parser.add_subparsers(title='subcommands',
+                                                 dest='command')
+
+        keygen.add_keygen_parser(subparsers, parent_parser)
```
aludvik: Why does the test need an argument parser? I am probably just missing something.

cli/tests/test_admin_keygen.py
peterschwarz: Make these doc strings

cli/tests/test_admin_keygen.py
```diff
@@ -0,0 +1,112 @@
...
+from sawtooth_cli.admin_command.utils import ensure_directory
+from sawtooth_cli.exceptions import CliException
+
+
+class TestKeygen(unittest.TestCase):
+    def __init__(self, test_name):
+        super().__init__(test_name)
+        self._temp_dir = None
+        self._parser = None
+        self._key_dir = ensure_directory('etc/keys', '/etc/sawtooth/keys')
```
peterschwarz: Were you not able to use the tmp dir for this?

cli/tests/test_admin_keygen.py
peterschwarz: If you are not using the tmp dir, be sure to clean up any files created (putting them in the tmp dir made this easier)

cli/tests/test_admin_keygen.py
peterschwarz: This could be changed to run the `main` function of the cli, with the right parameters.

cli/tests/test_admin_keygen.py
ojalatodd: @peterschwarz The command does not use a temp directory. This is the admin subcommand. 
(Note: the non-admin keygen command should also be tested. I can do that and just add a point jira to do that this spring)

ojalatodd: Note: I will add code to "clean up" created files as suggested below.

cli/tests/test_admin_keygen.py
ojalatodd: Removed.

cli/tests/test_admin_keygen.py
ojalatodd: Will consider that in next iteration

ojalatodd:
Changes made, @aludvik and @peterschwarz 
Docstrings added, cleaned up written files during tear down, etc.
Note that this command can not use a temp directory.

cli/tests/test_admin_keygen.py
```diff
@@ -0,0 +1,120 @@
...
+from sawtooth_signing import secp256k1_signer as signing
+from sawtooth_cli.admin_command import keygen
+from sawtooth_cli.admin_command.utils import ensure_directory
+from sawtooth_cli.exceptions import CliException
+
+
+class TestKeygen(unittest.TestCase):
+    def __init__(self, test_name):
+        super().__init__(test_name)
+        self._temp_dir = None
```
boydjohnson: Looks like _temp_dir is not being used for anything. Also my understanding is that the constructor on a unittest.TestCase subclass should not be used. If there is shared state between all tests, then the setupClass method should be used.

cli/tests/test_admin_keygen.py
```diff
@@ -0,0 +1,120 @@
...
+        """Remove existing test files before creating new keys.
+        """
+        self.remove_key_files()
+
+        args = self._parse_keygen_command(self._key_name)
+        keygen.do_keygen(args)
+
+        pubkey, signing_key = _read_signing_keys(self._wif_filename)
+
+        self.assertIsNotNone(pubkey)
```
nick-drozd: Can anything more substantial be asserted about the keys? Just asserting that they exist seems like a low bar.

ojalatodd:
@boydjohnson Changes you suggested made.

PR #199, opened 2017-03-13
Add CLI tool to list Pending Proposals 

peterschwarz:
Add a `sawtooth config proposals` subcommand to list the currently pending proposals, when a validator is in "Ballot" mode for setting application.

Additionally:

* Allow for differentiation of return values by status.
* Only check approval_threshold > 1


cli/sawtooth_cli/config.py
```diff
@@ -164,6 +196,72 @@ def _do_config_create(args):
         raise AssertionError('No target for create set.')
 
 
+def _do_config_proposal_list(args):
+    """Excutes the 'proposal list' subcommand.
```
vaporos: Typo - Excutes

PR #209, opened 2017-03-13
Test validator with two TPs

nick-drozd:


integration/sawtooth_integration/docker/intkey-smoke.yaml
```diff
@@ -49,7 +49,7 @@ services:
     entrypoint: "bash -c \"\
         ./bin/sawtooth admin keygen && \
         ./bin/sawtooth admin genesis && \
-        ./bin/validator -vv\""
```
boydjohnson: So without the debug logging on it is hard to tell why it has failed. It could be because something in the genesis process isn't right and so the validator never starts building blocks, or it could be something else entirely. The transaction processor logging seems less important, but the validator logging in the test is essential.

integration/sawtooth_integration/docker/two_fams.yaml
```diff
@@ -0,0 +1,96 @@
...
+    stop_signal: SIGKILL
+
+  integration_test:
+    image: sawtooth-test:$ISOLATION_ID
+    volumes:
+      - ../../../:/project/sawtooth-core
+    expose:
+      - 8080
+    working_dir: /project/sawtooth-core/integration/sawtooth_integration/tests
+    entrypoint: python3 -m http.server
```
boydjohnson: This is an accident, that this initially got in to the intkey and xo smokes, so shouldn't be here.

integration/sawtooth_integration/docker/xo-smoke.yaml
```diff
@@ -48,7 +48,7 @@ services:
     entrypoint: "bash -c \"\
         ./bin/sawtooth admin keygen && \
         ./bin/sawtooth admin genesis && \
-        ./bin/validator -vv\""
```
boydjohnson: Seems like an accident that these are being changed, as well.

integration/sawtooth_integration/docker/intkey-smoke.yaml
nick-drozd: This test does everything that intkey smoke does and more, so unless something very strange happens, intkey smoke won't fail unless this one also does (and the same mostly goes for xo smoke too). Personally I find the test output unpleasantly cluttered as it is, so I would prefer not to duplicate it.

PR #213, opened 2017-03-14
Move handshake to network connect, chain callbacks to Gossip registration, etc

jsmitchell:
Also makes heartbeat timeouts relevant for both inbound and outbound connections, and makes a request for the head of the chain once peer registration succeeds.

### DIFF
--- a/validator/sawtooth_validator/gossip/gossip.py
+++ b/validator/sawtooth_validator/gossip/gossip.py
@@ -14,6 +14,7 @@
 # ------------------------------------------------------------------------------
 import logging
 from threading import Condition
+from functools import partial
 
 from sawtooth_validator.protobuf.network_pb2 import GossipMessage
 from sawtooth_validator.protobuf.network_pb2 import GossipBatchByBatchIdRequest
@@ -22,6 +23,7 @@ from sawtooth_validator.protobuf.network_pb2 import \
 from sawtooth_validator.protobuf.network_pb2 import GossipBlockRequest
 from sawtooth_validator.protobuf import validator_pb2
 from sawtooth_validator.protobuf.network_pb2 import PeerRegisterRequest
+from sawtooth_validator.protobuf.network_pb2 import NetworkAcknowledgement
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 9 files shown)


validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -424,8 +423,9 @@ def __init__(self,
         self._thread = None
 
     @property
-    def local_id(self):
-        return self._send_receive_thread.connection
+    def connection_id(self):
+        return hashlib.sha512(
```
peterschwarz: Is this frequently called?  Just thinking that a sha512 isn't cheap...

validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -463,21 +463,32 @@ def __init__(self,
...
+            future.Future
         """
         message = validator_pb2.Message(
             correlation_id=_generate_id(),
             content=data,
             message_type=message_type)
 
-        fut = future.Future(message.correlation_id, message.content)
+        fut = future.Future(message.correlation_id, message.content,
+                            has_callback=True if callback is not None
```
peterschwarz: Seems like we could just pass the callback here, and save a step, and some possible mismatch of `has_callback=True`, with no callback added.

validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -402,14 +402,19 @@ def send(self, message_type, data, connection_id, has_callback=False):
                 message_type=message_type)
 
             fut = future.Future(message.correlation_id, message.content,
-                                has_callback=has_callback)
+                                has_callback=True if callback is not None
```
peterschwarz: Again, there's this parallel logic issue.

validator/sawtooth_validator/networking/interconnect.py
jsmitchell: This is how the future implementation currently works. I agree just passing the callback would be easier, but that seems like a separate story.

validator/sawtooth_validator/networking/interconnect.py
jsmitchell: This is how the future implementation currently works. I agree just passing the callback would be easier, but that seems like a separate story.

validator/sawtooth_validator/gossip/gossip.py
```diff
@@ -122,13 +135,35 @@ def broadcast(self, gossip_message, message_type, exclude=None):
...
+            LOGGER.debug("Peering request to %s was NOT successful",
+                         connection_id)
+        elif ack.status == ack.OK:
+            LOGGER.debug("Peering request to %s was successful",
+                         connection_id)
             self._peers.append(connection_id)
+            self.broadcast_block_request("HEAD")
+
+    def _connect_success_callback(self, connection_id):
+        LOGGER.critical("Connection to %s succeeded", connection_id)
```
vaporos: Probably should be info instead of critical.

validator/sawtooth_validator/gossip/gossip.py
```diff
@@ -122,13 +135,35 @@ def broadcast(self, gossip_message, message_type, exclude=None):
...
+
         register_request = PeerRegisterRequest()
-        self.broadcast(
-            register_request,
-            validator_pb2.Message.GOSSIP_REGISTER)
+        self._network.send(validator_pb2.Message.GOSSIP_REGISTER,
+                           register_request.SerializeToString(),
+                           connection_id,
+                           callback=partial(self._peer_callback,
```
vaporos: Could just use lambda here probably.

validator/sawtooth_validator/journal/completer.py
```diff
@@ -281,6 +282,8 @@ def add_batch(self, batch):
 
     def get_block(self, block_id):
         with self.lock:
+            if block_id == "HEAD":
```
vaporos: Overloading block_id here could be avoided by adding an explicit get_chain_head() or similar.

validator/sawtooth_validator/networking/interconnect.py
jsmitchell: Cached in inbound squashed commit

validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -113,17 +114,42 @@ def _send_heartbeat(self):
...
-                        correlation_id=_generate_id(),
-                        content=ping.SerializeToString(),
-                        message_type=validator_pb2.Message.NETWORK_PING)
-                    fut = future.Future(message.correlation_id,
-                                        message.content,
-                                        has_callback=False)
-                    self._futures.put(fut)
-                    yield from self._send_message(zmq_identity, message)
+                    if time.time() - \
+                            self._connected_identities[zmq_identity] > \
```
jsmitchell: Fixed in inbound squashed commit

validator/sawtooth_validator/gossip/gossip.py
jsmitchell: Fixed in inbound squashed commit

validator/sawtooth_validator/journal/completer.py
jsmitchell: Fixed in inbound squashed commit

PR #215, opened 2017-03-14
Add CLI subcommand `config proposal vote`

peterschwarz:
Add the CLI subcommand to submit a vote for an outstanding proposal. The vote is guarded against double voting, and against missing proposals.  A vote is submitted to a validator, where it is tallied by the transaction family

Additionally:

- Reject setting proposal vote ties
- Correct typos
- code cleanup

### DIFF
--- a/cli/sawtooth_cli/config.py
+++ b/cli/sawtooth_cli/config.py
@@ -27,6 +27,7 @@ from sawtooth_cli.rest_client import RestClient
 
 from sawtooth_cli.protobuf.config_pb2 import ConfigPayload
 from sawtooth_cli.protobuf.config_pb2 import ConfigProposal
+from sawtooth_cli.protobuf.config_pb2 import ConfigVote
 from sawtooth_cli.protobuf.config_pb2 import ConfigCandidates
 from sawtooth_cli.protobuf.setting_pb2 import Setting
 from sawtooth_cli.protobuf.transaction_pb2 import TransactionHeader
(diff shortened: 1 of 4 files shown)


cli/sawtooth_cli/config.py
```diff
@@ -262,6 +273,43 @@ def _accept(candidate, public_key, prefix):
...
+        if candidate.proposal_id == args.proposal_id:
+            proposal = candidate
+
+    if proposal is None:
+        raise CliException('No proposal exists with the given id')
+
+    for vote_record in proposal.votes:
+        if vote_record.public_key == pubkey:
+            raise CliException(
+                'A vote has already been recored with this signing key')
```
ojalatodd: Typo in error message: change "recored" to "recorded"

core_transactions/config/sawtooth_config/processor/handler.py
```diff
@@ -204,6 +204,7 @@ def apply(self, transaction, state):
         if auth_type == 'Ballot':
```
aludvik: Typo in commit message. Two "of"s

cli/sawtooth_cli/config.py
```diff
@@ -262,6 +272,43 @@ def _accept(candidate, public_key, prefix):
         raise AssertionError('Unknown format {}'.format(args.format))
 
 
+def _do_config_proposal_vote(args):
+    """Executes the 'proposal vote' subcommand.  Given a key file, a proposal
+    id and a vote value,, it generates a batch of sawtooth_config transactions
+    in a BatchList instance.  The BatchList is file or submitted to a
+    validator.
```
aludvik: Typos: two commas and "The BatchList is (output to) file..."?

cli/sawtooth_cli/config.py
```diff
@@ -262,6 +272,43 @@ def _accept(candidate, public_key, prefix):
...
+    """
+    pubkey, signing_key = _read_signing_keys(args.key)
+    rest_client = RestClient(args.url)
+
+    proposals = _get_proposals(rest_client)
+
+    proposal = None
+    for candidate in proposals.candidates:
+        if candidate.proposal_id == args.proposal_id:
+            proposal = candidate
```
aludvik: Should this exit the loop after finding a match? Can there ever be more than one match?

cli/sawtooth_cli/config.py
```diff
@@ -317,6 +364,29 @@ def _do_config_list(args):
...
 
 
+def _get_proposals(rest_client):
+    state_leaf = rest_client.get_leaf(
+        _key_to_address('sawtooth.config.vote.proposals'))
+
+    config_candidates = ConfigCandidates()
+
+    if state_leaf is not None:
+        setting_bytes = b64decode(state_leaf['data'])
```
aludvik: How does this fail if the decode fails? Will the sawtooth command just blow up?

peterschwarz: If that fails, somethings wrong with the rest api.  We would have failed much earlier, because we wouldn't have parsed the JSON.  We could catch it, but I'd still see it as a developer error and probably throw an AssertionError

cli/sawtooth_cli/config.py
peterschwarz: Fixed

peterschwarz:
rebased

PR #216, opened 2017-03-14
sawtooth cluster start error message

ojalatodd:
Error message output by 'sawtooth cluster start" when
a processor is not specified has been made more
specific, explaining where to find the processors
that come with Sawtooth Lake, etc.

### DIFF
--- a/cli/sawtooth_cli/cluster.py
+++ b/cli/sawtooth_cli/cluster.py
@@ -15,6 +15,7 @@
 
 from __future__ import print_function
 
+import argparse
 from argparse import Namespace
 import logging
 import os
@@ -66,7 +67,17 @@ def add_cluster_status_parser(subparsers, parent_parser):
 
 
 def add_cluster_start_parser(subparsers, parent_parser):
-    parser = subparsers.add_parser('start', parents=[parent_parser])
+
+    epilog = '''details:
+    The available transaction processors specified by -P depend upon the
+    method (specified by -m). When using docker, the included processors
+    can be found using the command 'docker images'.
+    '''
+
+    parser = subparsers.add_parser(
+        'start', parents=[parent_parser],
+        epilog=epilog,
+        formatter_class=argparse.RawDescriptionHelpFormatter)
 
     parser.add_argument(
         '--count',
@@ -221,6 +232,7 @@ def do_cluster_start(args):
 
     if args.processors is None:
         raise CliException("Use -P to specify one or more processors")
+
     state['Processors'] = args.processors
 
     node_controller = get_node_controller(state, args)


cintel:
The original error message is insufficient. It provides no information on what should be specified after a -P argument or how to find a valid parameter. Even better if it provided a list of valid known options.

ojalatodd:
@cintel and @vaporos How about this compromise:
We can't provide a strict list of choices using the "choices" container of the argparse addarguments method. That would prevent the user from choosing their own home-grown transaction processors.

How about if the error message includes the current list of provided processors? We will need to try to keep this up to date, or add some kind of automation to create the list. 
I will start by just a list of processors that needs to be manually updated.

cintel:
Providing a list of the transaction processors provided would be good.

vaporos:
Since there is no natural list of TPs to display (what TPs are in core is completely irrelevant, as it may not match the installation), we should focus more on the text of the message being clear.  At minimum, "Use -P to specify one or more transaction processors".

We could also augment the usage statement for the command such that it describes this further in an epilog section.  Something like:

details:
  The available transaction processors specified by -P depends upon the method (specified by -m).
  When using docker, ... (and some hint about how to find them with docker ps).

cli/sawtooth_cli/cluster.py
```diff
@@ -252,7 +252,12 @@ def do_cluster_start(args):
     state["DesiredState"] = "Running"
 
     if args.processors is None:
-        raise CliException("Use -P to specify one or more processors")
+        error_string = "Use -P to specify one or more processors.\n"
+        error_string += "Included processors are found in sawtooth-core/bin " \
```
dcmiddle: Is that where they will be if the system is installed rather than run from a dev environment?

ojalatodd: Good point. Trying to address this PR today.

ojalatodd:
@cintel and @vaporos  I have attempted to make a good error message and a good epilog that is accurate. Please take a look!

ojalatodd:
Great. So we have an OK from @dcmiddle . @vaporos  or @cintel ?

ojalatodd:
@vaporos and @cintel I took into account Shawn's feedback from last week.
Please take a look at this compromise, maybe we can close or merge this PR today! ;)

cli/sawtooth_cli/cluster.py
```diff
@@ -66,7 +67,18 @@ def add_cluster_status_parser(subparsers, parent_parser):
 
 
 def add_cluster_start_parser(subparsers, parent_parser):
-    parser = subparsers.add_parser('start', parents=[parent_parser])
+
+    epilog = '''details:
+    The available transaction processors specified by -P depend upon the
+    method (specified by -m). When using docker, the included processors
+    can be found using the command 'docker images', and include the string
+    tp_ within the image name.
```
vaporos: The tp_ prefix is our convention but not a requirement.  Suggest we omit the content after the last comma.

cli/sawtooth_cli/cluster.py
```diff
@@ -220,7 +232,12 @@ def do_cluster_start(args):
     state["DesiredState"] = "Running"
 
     if args.processors is None:
-        raise CliException("Use -P to specify one or more processors")
+
+        error_string = '''Use -P to specify one or more processors.\n
+        Enter command "sawtooth cluster start -h" for more details."'''
```
vaporos: Remove the newline and 2nd line.  It is not common for Linux commands to point out -h or --help in error messages as they are very standard options.

cli/sawtooth_cli/cluster.py
ojalatodd: Changed.

cli/sawtooth_cli/cluster.py
ojalatodd: Changed.

PR #217, opened 2017-03-14
Support for parallel scheduler in the context manager

boydjohnson:


validator/sawtooth_validator/execution/context_manager.py
```diff
@@ -210,112 +272,120 @@ def commit_context(self, context_id_list, virtual):
...
-            value = val_fut.result()
-            if value is not None:
-                add_value_dict[k] = value
+            updates.update(effective_updates)
 
-        new_root = tree.update(set_items=add_value_dict, virtual=virtual)
+        state_hash = tree.update(updates, virtual=False)
+        # clean up all contexts that are involved in being squashed.
+        base_c_ids = []
+        for i in context_id_list:
```
peterschwarz: `i` -> `id` (`i` should be reserved for numerical indexes)

validator/sawtooth_validator/execution/context_manager.py
```diff
@@ -42,31 +43,33 @@ class SquashException(Exception):
 
 
 class StateContext(object):
+    """A datastructure holding address-_ContextFuture pairs and the addresses
```
peterschwarz: data structure

validator/sawtooth_validator/execution/context_manager.py
```diff
@@ -210,112 +272,120 @@ def commit_context(self, context_id_list, virtual):
...
-            context = self._contexts.get(context_id)
-            context.can_set(address_value_list)
-            add_value_dict = {}
-            for d in address_value_list:
-                for add, val in d.items():
-                    add_value_dict[add] = val
-            context.set_futures(add_value_dict)
+
+        if context_id not in self._contexts:
+            LOGGER.info("Context_id not in contexts, %s", context_id)
```
peterschwarz: Is info the right level here?

validator/sawtooth_validator/execution/context_manager.py
boydjohnson: Good call. I will change it to something short but more descriptive. id is a python builtin (https://docs.python.org/3/library/functions.html#id) and so variable names shouldn't shadow builtins.

validator/sawtooth_validator/execution/context_manager.py
```diff
@@ -210,112 +272,120 @@ def commit_context(self, context_id_list, virtual):
...
                     value = val_fut.result()
                     if value is not None:
                         effective_updates[k] = value
 
                 updates.update(effective_updates)
 
             state_hash = tree.update(updates, virtual=False)
+            # clean up all contexts that are involved in being squashed.
+            base_c_ids = []
+            for i in context_ids:
```
peterschwarz: This doesn't look changed: based on previous converstion, would suggest `i` -> `c_id`, which is consistent with what you're doing else where

boydjohnson: Changing L384 right now. I also found another place where I used an i for a variable name in a loop. There is no i in team, and there won't be i's in the diff either.

validator/sawtooth_validator/execution/context_manager.py
```diff
@@ -210,112 +272,120 @@ def commit_context(self, context_id_list, virtual):
...
-                for add, val in d.items():
-                    add_value_dict[add] = val
-            context.set_futures(add_value_dict)
+
+        if context_id not in self._contexts:
+            LOGGER.warning("Context_id not in contexts, %s", context_id)
+            return False
+
+        context = self._contexts.get(context_id)
+        context.can_set(address_value_list)
```
jsmitchell: A comment on this can_set is probably needed. It's unclear that this is intended to raise an exception.

PR #219, opened 2017-03-15
Improve build process UX and consistency

aludvik:


bin/build_all
```diff
@@ -15,37 +15,161 @@
...
+javascript
+python
+"
+fi
+
+
+set -e
+
+docker_build() {
+  image=$1
```
vaporos: We should keep an indent of 4 for bash scripts.

aludvik:
For developers who have access to the drive, the doc is here: https://docs.google.com/document/d/1hCS_dwxnJvTclUH7oq-v2TVJN_UJsBPV-H9xaFyt6K8

Also reproduced below:

# Docker Build Update

With the recent changes, it should now be possible to build sawtooth and run tests from an arbitrary environment that has docker installed and configured to run with a linux kernel. Some substantial changes have gone into the build process to accomplish this, and this document is a quick overview of the process for current developers.

## tl;dr
If you are just doing python development for sawtooth, you should now be able to build and run the tests from vagrant or Mac by doing the following:

```
$ ./bin/build_all -l python
$ ./bin/run_tests -s java_sdk -s javascript_sdk
```

## Building Sawtooth
The old build_all script has been moved to build_python and docker_build_all has been moved to build_all.

The “new” build_all script handles building all the docker images that are usually needed by a developer*. Some of these build images are also used to build the source code in sawtooth. This includes running protogen, maven, yarn, etc. To make development in one language easier, it is now possible to pass flags to the build_all script.

```
sawtooth-lake/sawtooth-core » ./bin/build_all -h
Usage:
  ./bin/build_all [-l Language] [-s Language]

Options:
  -h            print usage and exit
  -l Language   only build the specified language(s)
  -s Language   skip the specified language(s)
```

The -l flag can be used multiple times to hand pick the languages that should be built. The -s flag can be used to pick languages that should be skipped. For example, to only build the python components of sawtooth you would do:

`$ build_all -l python`

Or to skip building java but still build python and javascript:

`$ build_all -s java`

## Running Tests
The run_tests script has also been changed to accommodate finer grained control of which tests should run.

```
sawtooth-lake/sawtooth-core » ./bin/run_tests -h
Usage:
  ./bin/run_tests [-m Module] [-s Module]
  (-m and -s can be passed multiple times)

Options:
  -h          print usage and exit
  -m Module   only run tests of specified module
  -s Module   skip the tests of specified module

Current modules:
cli
java_sdk
javascript_sdk
poet
python_sdk
rest_api
signing
validator
```

The -m flag can now be passed multiple times to hand pick which modules should be tested. Alternatively, the -s flag can also be passed multiple times to skip certain modules. For example, to only run the python_sdk and poet modules, you would do:

`$ run_tests -m python_sd -m poet`

Or to skip the java and javascript sdk tests do:

`$ run_tests -s java_sdk -s javascript_sdk`

## Running Lint
The run_lint script has not changed, however it is now being run inside the sawtooth-build-python container on Jenkins. You can still run the script locally from outside of any container (in Vagrant) but to get the same results as you would on Jenkins, you can instead do (after build_all):

`$ docker run -v $(pwd):/project/sawtooth-core sawtooth-build-python ./bin/run_lint`

bin/build_all
aludvik: Sorry, not sure how I managed to do that. Force pushed a fix.

Jenkinsfile
```diff
@@ -58,74 +58,52 @@ node ('master') {
...
-                    sh './bin/run_lint'
-                }
-            }
+            sh 'docker run --rm -v $(pwd):/project/sawtooth-core sawtooth-build-python:$ISOLATION_ID ./bin/run_lint'
         }
 
         // Run the tests
         stage("Run Tests") {
-            sh 'ISOLATION_ID=$(printf $BUILD_TAG | sha256sum | cut -c1-64) ./bin/run_tests'
+            sh './bin/run_tests'
```
peterschwarz: Perhaps, until we get the java sdk properly working, we can have this skip the java sdk tests for now.

aludvik: Add -x flag for java sdk.

aludvik:
Remove VOLUME commands from dockerfiles and turned up the verbosity of the java build steps per @agunde406 request.

PR #221, opened 2017-03-15
Add a minimum wait time check for on_check_publish_block callbacks to…

cintel:
… the BlockPublisher

### DIFF
--- a/validator/sawtooth_validator/journal/journal.py
+++ b/validator/sawtooth_validator/journal/journal.py
@@ -36,28 +36,32 @@ class Journal(object):
     This object provides the threading and event queue for the processors.
     """
     class _ChainThread(Thread):
-        def __init__(self, chain_controller, block_queue, block_cache):
+        def __init__(self, chain_controller, block_queue, block_cache,
+                     block_cache_purge_frequency):
             Thread.__init__(self)
             self._chain_controller = chain_controller
             self._block_queue = block_queue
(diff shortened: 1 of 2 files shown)


validator/sawtooth_validator/journal/journal.py
```diff
@@ -70,22 +70,34 @@ def stop(self):
             self._exit = True
 
     class _PublisherThread(Thread):
+        CHECK_PUBLISH_BLOCK_FREQUENCY = 0.1
```
jsmitchell: recommend passing this in as a keyword argument to __init__ with a default. Pass it in from the top level validator class down to whatever instantiates _PublisherThread so that we can establish it via a config file in the future.

jsmitchell:
Can you summarize what problem this is addressing?

validator/sawtooth_validator/journal/journal.py
dcmiddle: I like configurable parameters but I dislike kwargs.

jsmitchell: To clarify, I'm recommending this `def __init__(self, block_publisher, batch_queue, check_publish_block_frequency=0.1)`

cintel:
This makes the BlockPublisher behavior on checking if a block should be published consistent with respect to incoming batch load from the network.

PR #222, opened 2017-03-15
Remove broken path config tests which set os.name

vaporos:
These tests do not work consistently as they require the ability to
update os.name, which does not always work (it is sensitive to test
runner).

In particular, these did not work when the following was run in the
vagrant environment:

   nose2-3 -s /project/sawtooth-core/validator/tests/unit3/test_config/
--log-capture -v


vaporos:
Couldn't think of a good way to get the same test without API modification, which would make the code more complex.

PR #223, opened 2017-03-16
remove TNG verbiage from readme

dcmiddle:


vaporos:
This file should be removed.  All docs should go in sphinx except the top-level README.md, etc.

PR #224, opened 2017-03-16
Rename simulator to workload and it to the python sdk. Also add intkey workload.

agunde406:
to test, have a validator running and run the following command:
intkey workload

if you want to run it on  multiple validators run:
intkey workload --urls "tcp://0.0.0.0:40000,tcp://0.0.0.0:40001,etc"

usage: intkey workload [-h] [-v] [--rate RATE] [-d DISPLAY_FREQUENCY]
                       [-u URLS]

optional arguments:
  -h, --help            show this help message and exit
  -v, --verbose         enable more verbose output
  --rate RATE           Batch rate in batches per minute. Should be greater
                        then 0.
  -d DISPLAY_FREQUENCY, --display-frequency DISPLAY_FREQUENCY
                        time in seconds between display of batches rate
                        updates.
  -u URLS, --urls URLS  comma separated urls of validators to connect to.


sdk/examples/intkey_python/sawtooth_intkey/client_cli/workload.py
```diff
@@ -0,0 +1,194 @@
...
+
+LOGGER = logging.getLogger(__name__)
+
+IntKeyState = namedtuple('IntKeyState', ['name', 'stream', 'value'])
+
+
+class IntegerKeyWorkload(SawtoothWorkload):
+    """
+    This workload is for the Sawtooth Integer Key transaction family.  In
+    order to guarantee that transactions/batches are submitted at a relatively-
```
peterschwarz: "batches of transactions" or "single-transaction batches" could be more accurate.

jsmitchell:
If you ctrl-c the validator while the workload is still running, you get ```[02:14:00 ERROR   base_events] Future exception was never retrieved
future: <Future finished exception=TypeError('on_batch_not_yet_committed() takes 1 positional argument but 2 were given',)>
  File "/usr/lib/python3.5/concurrent/futures/thread.py", line 55, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/project/sawtooth-core/sdk/python/sawtooth_sdk/workload/workload_generator.py", line 166, in _check_on_batch
    self._workload.on_batch_not_yet_committed(batch.id)
TypeError: on_batch_not_yet_committed() takes 1 positional argument but 2 were given```

sdk/python/sawtooth_sdk/workload/sawtooth_workload.py
```diff
@@ -0,0 +1,133 @@
...
+        final cleanup before the simulator tears down.
+
+        Returns:
+            Nothing
+        """
+        pass
+
+    def on_validator_discovered(self, url):
+        """
+        Called by the workload generatorto let the workload  know that it has
```
aludvik: typo "generator to"

sdk/python/sawtooth_sdk/workload/sawtooth_workload.py
```diff
@@ -0,0 +1,133 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+
+
+class SawtoothWorkload(object):
+    """
+       This is meant to be an abstract base class for all workloads.
```
aludvik: Is this the correct way to declare an abstract base class? I don't know the answer but the manage library does it differently using the abc module:
https://github.com/hyperledger/sawtooth-core/blob/master/manage/sawtooth_manage/node.py
https://docs.python.org/3.6/library/abc.html

sdk/python/sawtooth_sdk/workload/sawtooth_workload.py
```diff
@@ -0,0 +1,133 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+
+
+class SawtoothWorkload(object):
```
aludvik: Can this just be called Workload?

sdk/examples/intkey_python/sawtooth_intkey/client_cli/workload.py
```diff
@@ -0,0 +1,194 @@
...
+    constant rate, when the transaction callbacks occur the following
+    actions occur:
+
+    1.  If there are no pending batches (on_all_batches_committed),
+        a new key is created.
+    2.  If a batch is committed, the corresponding key is > 1000000
+        either a increment is made (if < 10000000) or a new
```
aludvik: typo "an increment"

sdk/examples/intkey_python/sawtooth_intkey/client_cli/workload.py
```diff
@@ -0,0 +1,194 @@
...
+from sawtooth_sdk.protobuf.validator_pb2 import Message
+from sawtooth_intkey.client_cli.create_batch import create_intkey_transaction
+from sawtooth_intkey.client_cli.create_batch import create_batch
+
+LOGGER = logging.getLogger(__name__)
+
+IntKeyState = namedtuple('IntKeyState', ['name', 'stream', 'value'])
+
+
+class IntegerKeyWorkload(SawtoothWorkload):
```
aludvik: Seems like the name for intkey should be standard. IntKey, IntegerKey, and intkey all appear in this file.

sdk/examples/intkey_python/sawtooth_intkey/client_cli/workload.py
```diff
@@ -0,0 +1,194 @@
...
+                        default=10)
+    parser.add_argument('-d', '--display-frequency',
+                        type=int,
+                        help='time in seconds between display of batches '
+                             'rate updates.',
+                        default=30)
+    parser.add_argument('-u', '--urls',
+                        help='comma separated urls of validators to connect '
+                        'to.',
+                        default="tcp://0.0.0.0:40000")
```
aludvik: I think the default should be "tcp://127.0.0.1:40000" or "tcp://127.0.0.1:40000". Typically 0.0.0.0 is used for listening on all addresses whereas the workload is sending packets at a specific address.
https://en.wikipedia.org/wiki/0.0.0.0
https://en.wikipedia.org/wiki/Localhost

aludvik:
Tested locally, and it appears to work. Requested some minor changes to the implementation. Is there a way to manually verify that transactions are being committed?

jsmitchell:
@aludvik beat me to the punch on some of my code comments. I was also able to test this locally last night. Very nice! I also have questions about the committed stats. I noticed that even if I killed the transaction processor, the workload generator kept reporting batches as committed. Maybe it was just catchup or due to the rolling average? Do we have a sense of how the batch status requests impact the performance of the system? Maybe it would be good to have a flag to turn those off and just fire load blindly.

agunde406:
Stats of committed batches is fixed, updated to use transactions per second instead of transactions per minute, and other requested changes have been addressed.

PR #230, opened 2017-03-17
Add id filter parameter to /batches and /blocks, update POST

delventhalz:
Add an id filter parameter to the `GET /batches` and `GET /blocks` endpoints in the REST API. Allows you to fetch multiple resource by specifying arbitrary ids.

Use this parameter to update the response envelope to a `POST /batches`, so that when all batches are committed, a `/batches` link is sent, rather than a `/batch_status` link.

### DIFF
--- a/validator/sawtooth_validator/state/client_handlers.py
+++ b/validator/sawtooth_validator/state/client_handlers.py
@@ -203,36 +203,67 @@ class _ClientRequestHandler(Handler, metaclass=abc.ABCMeta):
 
         return head_id
 
-    def _list_blocks(self, head_id, xform=lambda b: [b]):
-        """Traverses the blockchain, building a list of blocks or derived data.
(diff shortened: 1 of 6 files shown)


validator/sawtooth_validator/state/client_handlers.py
```diff
@@ -374,7 +405,18 @@ def __init__(self, block_store):
 
     def _respond(self, request):
         head_id = self._get_head_block(request).header_signature
-        blocks = self._list_blocks(head_id)
+        blocks = self._list_store_resources(
+            request,
+            head_id,
+            request.block_ids,
+            lambda filter_id: self._block_store[filter_id].block,
+            lambda block: [block])
```
peterschwarz: While this is very clear, it could just as easily be the function `list`.

PR #232, opened 2017-03-17
CLI submit command

ojalatodd:
CLI command 'sawtooth submit' command created.
Uses rest api.
Submits batches from file, such as intkey or
config batches.

### DIFF
--- a/cli/sawtooth_cli/submit.py
+++ b/cli/sawtooth_cli/submit.py
@@ -1,0 +1,77 @@
+# Copyright 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


cli/sawtooth_cli/submit.py
```diff
@@ -0,0 +1,70 @@
...
+import sawtooth_cli.protobuf.batch_pb2 as batch_pb2
+
+LOGGER = logging.getLogger(__file__)
+
+
+def _split_batch_list(batch_list):
+    new_list = []
+    for batch in batch_list.batches:
+        new_list.append(batch)
+        if len(new_list) == 100:
```
peterschwarz: This should either be a named constant, or add a CLI flag `--batch-size-limit` that defaults to 100.

ojalatodd: Done. I added a "batch size limit flag", since it seems of possible use to some people.

PR #234, opened 2017-03-20
Nml 2127 - Update PoET simulator and Validator Registry Transaction Processor to Support Enclave Quote Structure

jjason:
Updates the PoET simulator and validator registry transaction processor:

* Python wrapper classes created for supporting the enlcave quote and embedded structures found in IAS attestation verification report. Classes support serializing/deserializing between Python objects and binary layout of corresponding C/C++ structs.
* Simulator uses SGX structs wrapper to create enclave quote layout the same as used by SGX enclave implementation
* Simulator and validator registry transaction processor verify enclave quote in attestation verification report in the same was as SGX enclave implementation
* Validator registry transaction processor unit tests updated for new enclave quote layout

### DIFF
--- a/consensus/poet/common/sawtooth_poet_common/sgx_structs/_sgx_attributes.py
+++ b/consensus/poet/common/sawtooth_poet_common/sgx_structs/_sgx_attributes.py
@@ -1,0 +1,88 @@
+# Copyright 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+
+import struct
+
+from sawtooth_poet_common.sgx_structs._sgx_struct import SgxStruct
+
+
+class SgxAttributes(SgxStruct):
+    """
+    Provide a wrapper around sgx_attributes_t
+
+    typedef struct _attributes_t
+    {
+        uint64_t flags; /* 0 */
+        uint64_t xfrm;  /* 8 */
+    } sgx_attributes_t;
+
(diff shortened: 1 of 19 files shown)


consensus/poet/families/sawtooth_validator_registry/validator_registry/processor/handler.py
```diff
@@ -121,6 +122,17 @@ def _set_data(state, address, data):
 
 class ValidatorRegistryTransactionHandler(object):
 
+    # The basename and enclave measurement values we expect to find
+    # in the enclave quote in the attestation verification report.
+    __VALID_BASENAME__ = \
```
jsmitchell: Where do BASENAME and ENCLAVE_MEASUREMENT come from? Can they change? Can you give some commentary on hardcoding?

consensus/poet/common/sawtooth_poet_common/sgx_structs/_sgx_attributes.py
```diff
@@ -0,0 +1,93 @@
+#
+# INTEL CONFIDENTIAL
```
peterschwarz: This is not the same license we are using in the rest of the project - needs to be the Apache 2.0 license

consensus/poet/common/sawtooth_poet_common/sgx_structs/_sgx_basename.py
```diff
@@ -0,0 +1,90 @@
+#
```
peterschwarz: Fix license in all locations.

consensus/poet/common/sawtooth_poet_common/sgx_structs/_sgx_attributes.py
jjason: Crud.  Will do.  Stupid copy/paste error as I took original file from sawtooth-poet repo.

consensus/poet/families/sawtooth_validator_registry/validator_registry/processor/handler.py
jjason: These are actual values from an IAS verification report generated by SGX enclave.  They are currently placeholders (see comments in code when they are checked in simulator or validator registry transaction processor).  I think that the intent is going to be that these values have to somehow be configured for the validators - I don't now if the config transaction family is the place to put these values, but we are probably going to have some small set  of values that are valid as I believe that the enclave measurement will be different whenever we update the SGX PoET enclave.

jjason:
Updated copyrights and pushed new PR.

PR #238, opened 2017-03-21
Create/update 0.8 Packages

aludvik:
Create remaining 0.8 packages and update existing packages to be (mostly) accurate. 
To test locally, you can do:

```
# Build the docker image used to build the sawtooth debs, which includes building debs 
for all the external packages not available
$ docker build . -f ci/sawtooth-build-debs -t sawtooth-build-debs

# Build all the sawtooth debs
$ docker run -v $(pwd):/project/sawtooth-core sawtooth-build-debs

# Install the sawtooth debs using apt and let them pull in all the dependencies they need 
automatically
$ docker build --no-cache . -f ci/sawtooth-test-debs -t sawtooth-test-debs

# Start up a docker image with sawtooth debs installed
$ docker run -it sawtooth-test-debs
root@[container]# sawtooth admin keygen
root@[container]# sawtooth admin genesis
root@[container]# validator -vv &
root@[container]# tp_intkey_python -vv &
...
```

Component specific docker containers to follow soon...


cli/setup.py
```diff
@@ -73,12 +73,15 @@ def version(default):
       url='https://github.com/hyperledger/sawtooth-core',
       packages=find_packages(),
       install_requires=[
-          'sawtooth-signing',
           'colorlog',
-          'protobuf'
+          'grpcio-tools',
```
peterschwarz: I thought we were trying to not rely on grpc.

aludvik: `grpcio-tools` is the package that was being pip3 installed, so that is the package I built a deb for and installed.

I can test it locally to see if it still works with just protobuf...

core_transactions/config/setup.py
```diff
@@ -0,0 +1,80 @@
+# Copyright 2016 Intel Corporation
```
peterschwarz: 2017

sdk/examples/intkey_python/setup.py
```diff
@@ -0,0 +1,84 @@
+# Copyright 2016, 2017 Intel Corporation
```
peterschwarz: 2017

aludvik:
Swapped grpcio-tools for protobuf and my manual test still works. We still need grpcio-tools to build the protobuf files before building the debs, but we can fix that in another PR.

aludvik:
Updated all the Copyrights to 2017 in the setup.py files.

jjason:
Should there be a setup.py for consensus/poet/cli?

aludvik:
@jjason I followed the packaging structure you setup in [your poet restructuring PR](https://github.com/hyperledger/sawtooth-core/commit/82c3ea9755204f5a6807612c2889f0fb4ba81e5f)

Is the intent to have sawtooth-poet-cli define its own set of CLI tools, or to extend the sawtooth-cli packages "sawtooth" command?

aludvik:
Rebased on master and added "poet" entrypoint to sawtooth-poet-cli package.

jjason:
Is this supposed to work behind proxy?  If not, then that would explain why docker build failed on step 8 (apt-get) failed.  If so, then maybe I am just having problems with Intel proxy again.

aludvik:
This should work behind a proxy if you change the docker build steps to:
````
$ docker build . -f ci/sawtooth-build-debs -t sawtooth-build-debs --build-arg https_proxy=$https_proxy --build-arg http_proxy=$http_proxy --build-arg HTTPS_PROXY=$HTTPS_PROXY --build-arg HTTP_PROXY=$HTTP_PROXY
```

This is done in our other build scripts, eg. `build_all` but I didn't think to include it in the comment above.

aludvik:
Discovered two more issues while testing.

1. I think the sawtooth-config package should add tp_config to the PATH, which I forgot to do.
2. sawtooth-xo depends on colorlog, which I missed.

aludvik:
Fixed the two issues above and am working with @jjason on the proxy issue. Seems unrelated to this PR.

aludvik:
Added a commit that fixes the proxy issue. See commit message for details.

jjason:
Verified it works locally with proxy.  Great job Adam.

PR #239, opened 2017-03-22
Return Authorization Error status to Transaction Processor

agunde406:
When the context_manager returns an AuthorizationException, TpStateGetResponse and TpStateSetResponse will set have the status of AUTHORIZATION_ERROR. The sdk's version of state should then throw an InvalidTransactionError. 

Also add a missing logging statement in context manager and missing step when building javascript.

### DIFF
--- a/sdk/java/src/main/java/sawtooth/sdk/client/State.java
+++ b/sdk/java/src/main/java/sawtooth/sdk/client/State.java
@@ -18,6 +18,7 @@ import com.google.protobuf.ByteString;
 import com.google.protobuf.InvalidProtocolBufferException;
 
 import sawtooth.sdk.processor.exceptions.InternalError;
+import sawtooth.sdk.processor.exceptions.InvalidTransactionException;
 import sawtooth.sdk.protobuf.Entry;
 import sawtooth.sdk.protobuf.Message;
 import sawtooth.sdk.protobuf.TpStateGetRequest;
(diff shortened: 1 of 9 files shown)


validator/sawtooth_validator/execution/context_manager.py
```diff
@@ -159,6 +159,8 @@ def can_set(self, address_value_list):
         for add_value_dict in address_value_list:
             for address in add_value_dict.keys():
                 if address not in self._write_list:
+                    LOGGER.warning("Authorization exception, address: %s",
```
peterschwarz: Given that we properly handle this exception, we don't really need to log here. It could just create log noise, that no one can really deal with.

peterschwarz: We should create a story that actually sends back the error that occurred in the txn processor.

sdk/java/src/main/java/sawtooth/sdk/client/State.java
```diff
@@ -77,6 +79,10 @@ public State(Stream stream, String contextId) {
     }
     Map<String, ByteString> results = new HashMap<String, ByteString>();
     if (getResponse != null) {
+      if (getResponse.getStatus() == TpStateGetResponse.Status.AUTHORIZATION_ERROR) {
+        throw new InvalidTransactionException(
+          "Tried to get unauthorized address" + addresses.toString()) ;
```
jjason: A nit, but you will probably want a space after "...address" to make string readable.  Same for line 135 below.

PR #240, opened 2017-03-22
Poet smoke

nick-drozd:


consensus/poet/simulator/sawtooth_poet_simulator/poet_enclave_simulator/poet_enclave_simulator.py
```diff
@@ -46,7 +46,7 @@
 
 LOGGER = logging.getLogger(__name__)
 
-TIMER_TIMEOUT_PERIOD = 3.0
+TIMER_TIMEOUT_PERIOD = 30.0
```
jjason: You can get rid of this.  In a PR I merged, I have changed this to 30 already.

integration/sawtooth_integration/docker/poet-smoke.yaml
```diff
@@ -0,0 +1,216 @@
...
+    expose:
+      - 40000
+      - 8800
+    entrypoint: "bash -c \"\
+        ./bin/sawtooth admin keygen --force && \
+        ./bin/sawtooth config proposal create \
+          -k /etc/sawtooth/keys/validator.wif \
+          sawtooth.consensus.algorithm=poet \
+          sawtooth.poet.target_wait_time=5 \
+          -o config.batch && \
```
jjason: Not certain, but you may also want to add call to `./bin/poet genesis ...` to create key pair for genesis validator.  Then you will also want to include that batch file in the `./bin/sawtooth admin genesis` below.

integration/sawtooth_integration/tests/test_poet_smoke.py
```diff
@@ -0,0 +1,154 @@
...
+
+        3) Loop through the validators, sending each one $BATCH_COUNT
+           batches of intkey 'inc' transactions.
+
+        4) Assert that the validators are in consensus with each.
+        '''
+        populate, increment = _make_txns()
+
+        # wait for validators to register
+        time.sleep(INITIAL_WAIT_TIME)
```
jjason: Should this be a sleep or is there any way through REST API to find out if they have registered?  Or does this mean something other than registering with the validator registry?

nick-drozd: It needs to wait for all the validators to get the genesis block, since a validator can't be queried for its blocks if it doesn't have any. I don't know if there's anything better to do than to wait for them.

integration/sawtooth_integration/docker/poet-smoke.yaml
nick-drozd: Added

consensus/poet/simulator/sawtooth_poet_simulator/poet_enclave_simulator/poet_enclave_simulator.py
nick-drozd: Thanks to the magic of git, this took care of itself.

cli/sawtooth_cli/rest_client.py
```diff
@@ -56,7 +56,11 @@ def send_batches(self, batch_list):
         Returns:
             dict: the json result data, as a dict
         """
-        data_bytes = batch_list.SerializeToString()
+        try:
+            data_bytes = batch_list.SerializeToString()
+        except AttributeError:
+            data_bytes = batch_list
+
```
aludvik: Why is this needed?

nick-drozd: Some of the sawtooth commands send batch lists to the client, and the expect the client to convert them to strings. The intkey message factory converts batch lists to strings itself, so the client needs to be able to handle strings too.

This is a hack really; somebody needs to decide exactly what the interface should be.

aludvik: We should probably pick one interfaces then, rather than trying to implement both.

delventhalz: As the Rest Client is currently just an internal helper for CLI, tests and the like, I think a whatever-is-most-convenient implementation is fine. If we want to package the Rest Client for external usage, there is probably a fair amount of design and refactoring that should happen anyway.

jsmitchell: @delventhalz can you add a Jira story to simplify this down to a single argument type?

PR #242, opened 2017-03-22
Mark the schedule as complete on set response

peterschwarz:
The serial scheduler marks the schedule as complete when it recieves
responses, not when its queue is empty.  If the queue size is used to
determine comleteness, the schedule is marked as complete before a
transaction processor has replied with a result - this results in
genesis batch processing failing.

### DIFF
--- a/validator/tests/unit3/test_scheduler/tests.py
+++ b/validator/tests/unit3/test_scheduler/tests.py
@@ -180,6 +180,96 @@ class TestSerialScheduler(unittest.TestCase):
         with self.assertRaises(StopIteration):
             next(iterable)
 
+    def test_completion_on_finalize_only_when_done(self):
+        """Tests that iteration will stop when finalized is called on an
+        otherwise complete scheduler.
+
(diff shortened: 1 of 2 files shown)


validator/tests/unit3/test_scheduler/tests.py
```diff
@@ -181,6 +181,59 @@ def test_completion_on_finalize(self):
...
+        that they are returned in the appropriate order when using an iterator.
+        Test that the value of `complete` is false until the last value.
+
+        This test also finalizes the scheduler and verifies that StopIteration
+        is thrown by the iterator, and the complete is true in the at the end.
+        """
+        private_key = signing.generate_privkey()
+        public_key = signing.generate_pubkey(private_key)
+
+        context_manager = ContextManager(dict_database.DictDatabase())
```
boydjohnson: The context_manager now needs to be explicitly shutdown. In other scheduler tests there are setUp and tearDown methods doing that.

peterschwarz:
Added another test that will prevent a regression

PR #244, opened 2017-03-22
Add graceful error handling when finding handlers in TP

agunde406:
Logs if the wrong type of message is received by the transaction processor or if we do not have a handler to deal with the header family, encoding, and version. 

### DIFF
--- a/sdk/python/sawtooth_sdk/processor/core.py
+++ b/sdk/python/sawtooth_sdk/processor/core.py
@@ -53,16 +53,23 @@ class TransactionProcessor(object):
         """
         self._handlers.append(handler)
 
+    def _matches(self, handler, header):
+        return header.family_name == handler.family_name \
+            and header.family_version in handler.family_versions \
+            and header.payload_encoding in handler.encodings
+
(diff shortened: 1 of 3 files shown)


sdk/python/sawtooth_sdk/processor/core.py
```diff
@@ -59,10 +59,14 @@ def _find_handler(self, header):
         :param header transaction_pb2.TransactionHeader:
         :return: handler
         """
-        return list(filter(lambda h: header.family_name == h.family_name and
-                           header.family_version in h.family_versions and
-                           header.payload_encoding in h.encodings,
-                           self._handlers))[0]
+        try:
+            return list(filter(lambda h: header.family_name == h.family_name
```
peterschwarz: this lambda is pretty complicated - it could be pulled out into a function for readability's sake. This could also be done via:
```
next(handler for handler in self._handlers if _matches(handler, header))
```
Which is much faster than the filter-list combo you have there. You still would catch, but in this case, a `StopIteration` for the case with no match.

PR #246, opened 2017-03-23
Cxx sdk

cintel:
This is the initial draft of the C++ sdk for early review, I expect to update the following things before this is ready for merging:

1) improve error handling and error types in the TransactionProcessor
a) address the missing error handling in get state and look-ups of handlers
b) review all the functions for proper pointer management. 
2) migrate cout to logging framework (log4cpp?) 
3) Factor TransactionHandler to return an Applicator object for dispatch to thread pool. 
4) Complete IntKey implementation (needs cbor dependency) 
5) Improve connect and disscontect from validator handling

### DIFF
--- a/sdk/cxx/include/sawtooth/address_mapper.h
+++ b/sdk/cxx/include/sawtooth/address_mapper.h
@@ -1,0 +1,95 @@
+/* Copyright 2017 Intel Corporation
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+------------------------------------------------------------------------------*/
+#pragma once
+
+#include <memory>
+#include <string>
+
+#include "sawtooth/message_stream.h"
+
+namespace sawtooth {
+
+// Raised if any of the addresses generated violate
+// the merkel trie address constraints.
+class AddressFormatError: public std::runtime_error {
+ public:
+    explicit AddressFormatError(std::string const& error)
+        : std::runtime_error(error)
+    {}
+};
+
+
+// Helper class to provide mappings from domain identifiers
+// to merkle trie addresses. This class is designed to be overloaded, such
+// that implementers of TransactionProcessors can define their own mappings
+// from their native object ids to merkle trie addresses.
+//
+// The default implementation uses a SHA512 hash to map both the key and the
+// namespace to the mekle trie address. Collisions in the key mappings should
+// be uncommon but still possible, data storage behind the keys.
+//
+// Each instance of the mapper holds a namespace that is mapped to a
+// namespace prefix when it is constructed. If your transaction processor
+// operates on multiple namespaces, it is recommended that you use an instance
+// for each namespace.
+class AddressMapper {
+ public:
+    // Constructor for an address mapping object, Takes the
+    // unencoded namespace name, it will be mapping objects, to as a parameter.
+    // The namespace argument passed will be mapped to a namespace_prefix
+    explicit AddressMapper(const std::string& namespace_);
+    virtual ~AddressMapper() {}
+
+    AddressMapper (const AddressMapper&) = delete;
+    AddressMapper (const AddressMapper&&) = delete;
+    AddressMapper& operator= (const AddressMapper&) = delete;
+
+    // Maps an namespace string to an namespace address prefix.
+    // this provides a default implementation using SHA512.
+    // it is intended to be overriden by subclass implementations
+    // that can provide their own mapping of namspace strings to
+    // prefixes. The requirements are that the mapping needs to
+    // return a prefix that is 6 characters in length and consists
+    // of only lower case hexadecimal characters(0123456789abcdef).
+    virtual std::string MapNamespace(const std::string& key) const; 
+
+    // Maps an key string to a address, this provides a default
+    // implementation using SHA512.
+    // It is intended to be overriden by subclass implementations
+    // that can provide their own mappings. The requirements are that
+    // the mapping needs to return a string that is 128 characters
+    // in length and consists of only lower case hexadecimal characters.
+    virtual std::string MapKey(const std::string& key) const;
+
+
+    // returns the namespace prefix generated when the class was constructed.
+    std::string GetNamespacePrefix();
+
+    // Maps the key passed to the function and concatenates it with the
+    // namespace prefix this object was initialized with and returns
+    // a fully qualified Merkle Trie address. These addresses are always
+    // 134 characters in length and consist entirely of lower case hexidecimal
+    // characters(0123456789abcdef).
+    std::string MakeAddress(const std::string& key);
+ private:
+    bool namespace_initialized;
+    std::string namespace_;
+    std::string namespace_prefix;
+};
+
+typedef std::shared_ptr<AddressMapper> AddressMapperPtr;
+typedef std::unique_ptr<AddressMapper> AddressMapperUPtr;
+
+}  // namespace sawtooth
--- a/sdk/cxx/include/sawtooth/future_message.h
+++ b/sdk/cxx/include/sawtooth/future_message.h
@@ -1,0 +1,97 @@
+
+/* Copyright 2017 Intel Corporation
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+------------------------------------------------------------------------------*/
+#pragma once
+
+#include <condition_variable>
+#include <cstdint>
+#include <memory>
+#include <mutex>
+#include <string>
+#include <vector>
+
+namespace sawtooth {
+
+// Utility types to provide memory management of the Protobuffer defined
+// Message class.
+typedef std::shared_ptr<Message> MessagePtr;
+typedef std::unique_ptr<Message> MessageUPtr;
+
+// FutureMessage is a promise of completion for a message to be delivered.
+// This is used internally in the sdk to match responses from the validator
+// to the sender.
+class FutureMessage final {
+ public:
+    explicit FutureMessage(const std::string in_correlation_id):
+        correlation_id_(in_correlation_id), msg(nullptr) {}
+    virtual ~FutureMessage() {}
+
+    FutureMessage (const FutureMessage&) = delete;
+    FutureMessage (const FutureMessage&&) = delete;
+    FutureMessage& operator= (const FutureMessage&) = delete;
+
+    // Retreive the message correlation Id.
+    std::string correlation_id() const {
+        return this->correlation_id_;
+    }
+
+    // Check if the reponse message has been delivered. This can used to
+    // check if GetMessage will block or not.
+    bool HasResponse() const {
+        return this->msg != nullptr;
+    }
+
+    // Get the message response and decode it in to the corresponding
+    // proto buffer type. This function will block until the message is
+    // availiable. Use HasResponse to test if the message has been delivered.
+    template<typename T>
+    void GetMessage(Message::MessageType msg_type, T* proto) {
+        std::unique_lock<std::mutex> lock(this->mutex);
+        if (!this->msg) {
+            while (!this->msg) {
+                this->condition.wait(lock);
+            }
+        }
+        if (this->msg->message_type() != msg_type) {
+            std::stringstream error;
+            error << "Error unexpected message reponse type Expected:"
+                << msg_type << " got: " << this->msg->message_type();
+            throw std::runtime_error(error.str());
+        }
+
+        const std::string& msg_data = this->msg->content();
+        proto->ParseFromArray(msg_data.c_str(), msg_data.length());
+    }
+
+    // Set the reponse message and signal any potential waiters that it has
+    // been received. The futureMessage takes ownership of the messages.
+    void SetMessage(MessageUPtr msg) {
+        std::unique_lock<std::mutex> lock(this->mutex);
+        this->msg = std::move(msg);
+        this->condition.notify_all();
+    }
+
+ private:
+    std::string correlation_id_;  // The correlation id of the message
+    std::condition_variable condition;  // the condition that is signaled when
+    // the message arrives.
+    std::mutex mutex;  // the mutex to use with the condition variable.
+    MessageUPtr msg;  // The message data, set once it has arrives. Owned by
+    // this object.
+};
+
+typedef std::shared_ptr<FutureMessage> FutureMessagePtr;
+
+}  // namespace sawtooth
--- a/sdk/cxx/include/sawtooth/global_state.h
+++ b/sdk/cxx/include/sawtooth/global_state.h
@@ -1,0 +1,76 @@
+/* Copyright 2017 Intel Corporation
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+------------------------------------------------------------------------------*/
+#pragma once
+
+#include <unordered_map>
+#include <memory>
+#include <string>
+
+#include "sawtooth/message_stream.h"
+
+namespace sawtooth {
+
+// Provides access to the global state
+class GlobalState {
+ public:
+    typedef std::pair<std::string, std::string> KeyValue;
+ 
+    // Takes the messages stream that it will use to communicate with the
+    // Validator to retrieve the current state. It is not expected that this
+    // will be called directly from a Transaction Processor. An instance will
+    // be provided to transaction handlers when a transaction is being
+    // processed.
+    explicit GlobalState(const MessageStreamPtr& message_stream,
+        const std::string& context_id);
+    virtual ~GlobalState() {}
+
+    // Retrieve a single value from global state. If you are retrieving
+    // multiple values, it is encouraged to use the batch get overload defined
+    // below. The function returns false if the value is not retreived, true if
+    // it is found in the GlobalStore, and throws an exception if an error occurs
+    bool Get(std::string* out_value, const std::string& address) const;
+
+    // Retreive multiple values from the global state. Values are returned in
+    // the out_values map. If a value is not present then there will be no
+    // entry in the returned map. So you must check for the precense of a
+    // returned value prior to accessing it.
+    void Get(std::unordered_map<std::string, std::string>* out_values,
+        const std::vector<std::string>& addresses) const;
+
+    // Set a single GlobalState value. Use the batch Set function defined
+    // below if you are setting multiple items.
+    void Set(const std::string& address, const std::string& value) const;
+    // Set multiple values in the global state. Each entry in the addresses
+    // param is a std::pair, with the first value representing the address and
+    // the second the value.
+    //      std::vector<GlobalState::KeyValue> addresses;
+    //      addresses.push(std::make_pair(address, value));
+    void Set(const std::vector<KeyValue>& addresses) const;
+
+    // Delete a single GlobalState value. Used the batch Delete function
+    // defined below if you are deleting multiple items.
+    void Delete(const std::string& address) const;
+
+    // Delete multiple entries from global state.
+    void Delete(const std::vector<std::string>& address) const;
+
+ private:
+    std::string context_id;
+    MessageStreamPtr message_stream;
+};
+typedef std::shared_ptr<GlobalState> GlobalStatePtr;
+typedef std::unique_ptr<GlobalState> GlobalStateUPtr;
+
+}  // namespace sawtooth
--- a/sdk/cxx/include/sawtooth/message_dispatcher.h
+++ b/sdk/cxx/include/sawtooth/message_dispatcher.h
@@ -1,0 +1,79 @@
+
+/* Copyright 2017 Intel Corporation
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+------------------------------------------------------------------------------*/
+#pragma once
+
+#include <atomic>
+#include <condition_variable>
+#include <cstdint>
+#include <unordered_map>
+#include <memory>
+#include <thread>
+#include <string>
+#include <vector>
+
+#include <zmqpp/context.hpp>
+#include <zmqpp/socket.hpp>
+
+#include "sawtooth/message_stream.h"
+
+namespace sawtooth {
+
+// Main message processing handler. This object owns the ZMQ socket connection
+// to the validator as well as the internal routing sockets to route messages
+// to the main thread for processing and listens for messages to route back to 
+// the validator.
+class MessageDispatcher {
+ public:
+    // Connect to the validator, create and bind to the internal routing sockets, and route the messages.
+    void Connect(const std::string& connection_string);
+    // Close all the open sockets and shutdown the message routing thread.
+    void Close();
+
+    // Creates a MessageStream object to allow processors(via GlobalState) to
+    // submit messages to be routed back to the validator.
+    MessageStreamPtr CreateStream();
+    zmqpp::context& context() { return this->context_; }
+ protected:
+    friend class TransactionProcessor;  
+    MessageDispatcher();
+
+ private:
+    void ReceiveMessage();
+    void SendMessage();
+
+    static void DispatchThread(MessageDispatcher* dispatcher);
+
+    zmqpp::context context_;
+    zmqpp::socket server_socket;
+    zmqpp::socket message_socket;  // inproc socket to recv messages to
+    // route to the validator
+    zmqpp::socket processessing_request_socket;  // inproc socket to send
+    // processing requests to the Transaction Processing
+    // thread(the main thread)
+
+    // Synchronization variables to coordinate with MessageStreams and the
+    // dispatch thread.
+    std::condition_variable condition;
+    std::mutex mutex;
+    std::atomic<bool> run;
+    bool thread_ready;
+
+    std::unordered_map<std::string, std::shared_ptr<FutureMessage>> message_futures;
+
+    std::thread dispatch_thread;
+};
+
+}  // namespace sawtooth
--- a/sdk/cxx/include/sawtooth/message_stream.h
+++ b/sdk/cxx/include/sawtooth/message_stream.h
@@ -1,0 +1,81 @@
+
+/* Copyright 2017 Intel Corporation
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+------------------------------------------------------------------------------*/
+#pragma once
+
+#include <atomic>
+#include <cstdint>
+#include <unordered_map>
+#include <memory>
+#include <queue>
+#include <sstream>
+#include <string>
+#include <thread>
+#include <vector>
+
+#include <zmqpp/context.hpp>
+#include <zmqpp/message.hpp>
+#include <zmqpp/socket.hpp>
+
+#include "proto/validator.pb.h"
+
+#include "sawtooth/future_message.h"
+
+
+namespace sawtooth {
+
+// The MessageStream provides a conduit for messages between the Transaction
+// processing instance and the validator. MessageStreams are closely tied to
+// the MessageDispatcher, the two objects coordinate the mapping of message
+// responses thru a number of shared state objects. As such MessageStreams
+// should only be created by the MessageDispatcher via the CreateStream method.
+class MessageStream {
+ public:
+    virtual ~MessageStream() {}
+
+    // Send a message to the validator and get back the FutureMessage object
+    // that will receive the response.
+    FutureMessagePtr Send(Message::MessageType type,
+        const std::string& data);
+
+    // This is a helper function that takes care of serializing a protobuffer
+    // defined class and then sending the serialized results to the validator.
+    template <typename T>
+    FutureMessagePtr SendMessage(Message::MessageType type, const T& proto) {
+        std::stringstream proto_stream;
+        proto.SerializeToOstream(&proto_stream);
+        return this->Send(type, proto_stream.str());
+    }
+
+ private:
+    friend class MessageDispatcher;
+    explicit MessageStream(zmqpp::context* context,
+        std::unordered_map<std::string, std::shared_ptr<FutureMessage>>&,
+        std::mutex& future_message_mutex);
+
+    std::string GenerateCorrelationId() const;
+
+    zmqpp::context* context_;
+    std::unique_ptr<zmqpp::socket> socket;
+
+    std::unordered_map<std::string, std::shared_ptr<FutureMessage>>& future_message_map;
+    std::mutex& future_message_mutex;
+
+    static std::atomic<int> correlation_counter;
+};
+
+typedef std::shared_ptr<MessageStream> MessageStreamPtr;
+
+}  // namespace sawtooth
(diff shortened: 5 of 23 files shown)


jsmitchell:
Can you add some instructions for building and testing this?

sdk/cxx/include/sawtooth/future_message.h
```diff
@@ -0,0 +1,72 @@
...
+
+typedef std::vector<uint8_t> DataBuffer;
+typedef std::shared_ptr<DataBuffer> DataBufferPtr;
+
+typedef std::shared_ptr<Message> MessagePtr;
+typedef std::unique_ptr<Message> MessageUPtr;
+
+class FutureMessage {
+ public:
+    explicit FutureMessage(const std::string in_correlation_id):
```
jjason: Should this be `const std::string&`

sdk/cxx/include/sawtooth/future_message.h
```diff
@@ -0,0 +1,72 @@
...
+        correlation_id_(in_correlation_id), msg(nullptr) {}
+    virtual ~FutureMessage() {}
+
+    std::string correlation_id() {
```
jjason: Mark method as `const`

Also, could return be `const std::string&` ?

sdk/cxx/include/sawtooth/future_message.h
```diff
@@ -0,0 +1,72 @@
...
+        return this->correlation_id_;
+    }
+
+    template<typename T>
+    void GetMessage(Message::MessageType msg_type, T* proto) {
+        if (!this->msg) {
+            std::unique_lock<std::mutex> lock(this->mutex);
+            this->condition.wait(lock);
+        }
```
jjason: I thought that the normal pattern for condition variables was something like:

```
std::unique_lock<std::mutex> lock(this->mutex);
while (!this->msg) {
  this->condition.wait(lock);
}
```

sdk/cxx/include/sawtooth/future_message.h
```diff
@@ -0,0 +1,72 @@
...
+        if (this->msg->message_type() != msg_type) {
+            std::cout << "Expected:" << msg_type << " got: "
+                << this->msg->message_type() << std::endl;
+            throw std::runtime_error("Unexpected Message Type");
+        }
+
+        const std::string& msg_data = this->msg->content();
+        proto->ParseFromArray(msg_data.c_str(), msg_data.length());
+    }
+    void SetMessage(MessageUPtr msg) {
```
jjason: Does this need:

`std::unique_lock<std:mutex> lock(this->mutex);`

??

sdk/cxx/include/sawtooth/global_state.h
```diff
@@ -0,0 +1,41 @@
...
+
+namespace sawtooth {
+
+class GlobalState {
+ public:
+    explicit GlobalState(const MessageStreamPtr& message_stream,
+        const std::string& context_id);
+    virtual ~GlobalState();
+
+    std::string Get(const std::string& address);
```
jjason: Should this method be marked `const`?

sdk/cxx/include/sawtooth/message_stream.h
```diff
@@ -0,0 +1,73 @@
...
+        // serialize  to buffer
+        std::stringstream proto_stream;
+        DataBuffer buffer(proto.ByteSizeLong());
+        proto.SerializeToOstream(&proto_stream);
+
+        return this->Send(type, proto_stream.str());
+    }
+
+ private:
+    std::string GenerateCorrelationId();
```
jjason: Mark method as `const` ?

sdk/cxx/include/sawtooth/transaction_handler.h
```diff
@@ -0,0 +1,41 @@
...
+#include <vector>
+
+#include "proto/transaction.pb.h"
+#include "sawtooth/global_state.h"
+
+namespace sawtooth {
+
+class TransactionHandler {
+ public:
+    virtual std::string transaction_family_name() = 0;
```
jjason: Should these return `const &` instead of copies???

Should they be marked as `const` methods?

sdk/cxx/src/message_dispatcher.cpp
```diff
@@ -0,0 +1,131 @@
...
+    this->server_socket.connect(connection_string.c_str());
+    this->message_socket.bind("inproc://send_queue");
+    this->processessing_request_socket.bind("inproc://request_queue");
+
+    // start thread
+    std::unique_lock<std::mutex> lock(this->mutex);
+    this->dispatch_thread = std::thread(MessageDispatcher::SendReceiveThread,
+        this);
+
+    this->condition.wait(lock);
```
jjason: Again, I think that it is recommended that you do:

```
while (!this->thread_is_ready) {
  this->condition.wait(lock);
}

```
Because waits can end for a reason other than the notiify()/notify_all() method being invoked on it.

sdk/cxx/src/message_dispatcher.cpp
```diff
@@ -0,0 +1,131 @@
...
+}
+
+
+MessageStreamPtr MessageDispatcher::CreateStream() {
+    return std::shared_ptr<MessageStream>(new MessageStream(&this->context_));
+}
+
+void MessageDispatcher::SendReceiveThread(MessageDispatcher* dispatcher) {
+    std::cout << "MessageDispatcher::SendReceiveThread" << std:: endl;
+    dispatcher->condition.notify_all();
```
jjason: Here is where you would set condition.

```
{
  std::unique_lock<std::mutex> lock(dispatcher->mutex);
  dispatcher->thread_is_ready = true;
  dispatcher->condition.notify_all();
}
```

sdk/cxx/src/message_stream.cpp
```diff
@@ -0,0 +1,77 @@
...
+    if (this->socket.send(zmsg, 0)) {
+        send_req.release();
+    }
+
+    return future;
+}
+
+std::string MessageStream::GenerateCorrelationId() {
+    std::stringstream stream;
+    stream << MessageStream::correlation_counter++;
```
jjason: Does this counter need to be protected with a mutex?

sdk/cxx/include/sawtooth/future_message.h
```diff
@@ -0,0 +1,72 @@
...
+#include <vector>
+
+namespace sawtooth {
+
+typedef std::vector<uint8_t> DataBuffer;
+typedef std::shared_ptr<DataBuffer> DataBufferPtr;
+
+typedef std::shared_ptr<Message> MessagePtr;
+typedef std::unique_ptr<Message> MessageUPtr;
+
```
jjason: General comment - I assume that there will be class- and method-level comments.

cintel:
@jsmitchell My normal pattern is to add a readme.md to the folder with details on the package. None of the other SDKs have docs on how to build and tests, so there is no pre-existing pattern to follow. Where should these instructions live?

jsmitchell:
@cintel I meant like here in a quick comment on the PR so we know how to build/test.

sdk/cxx/src/message_stream.cpp
cintel: correlation_counter is an atomic.

sdk/cxx/include/sawtooth/transaction_handler.h
cintel: no, I want the implementer to be able to return the values on the stack.

cintel:
Current build and run instructions:
from your vagrant box, /project/sawtooth-core directory
1) install the c++ dependencies
sudo tools/plugins/install_cxx_tools.sh

2) build the sdk 
cd sdk/cxx
mkdir build
cd build
cmake ..
make

These following steps will change once I get the intekey sample fully implmented. 
3) run the sdk
from sdk/cxx directory:
export PYTHONPATH=$PYTHONPATH:/project/sawtooth-core/integration
python3 tp_intkey_tester.py

4) Run the test transaction processor
from sdk/cxx directory:
./build/bin/tp
Execution progress will stop at the first Process_Request, tp does not implement IntKey apply. My next update to the PR will have the intkey sample completed.

sdk/cxx/include/sawtooth/address_mapper.h
```diff
@@ -0,0 +1,88 @@
...
+// namespace to the mekle trie address. Collisions in the key mappings should
+// be uncommon but still possible, data storage behind the keys.
+//
+// Each instance of the mapper holds a namespace that has is mapped to a
+// namespace prefix when it is constructed. If your transaction processor
+// operates on multiple namespaces, it is recommened that you use an instance
+// for each namespace.
+class AddressMapper {
+ public:
+    // Constructor for an aaddress mapping object, Takes the
```
aludvik: "aadress" -> "address"

sdk/cxx/include/sawtooth/address_mapper.h
```diff
@@ -0,0 +1,88 @@
...
+    // namesspace it will be mapping objects to as a parameter.
```
aludvik: "namesspace" -> "namespace"

sdk/cxx/include/sawtooth/address_mapper.h
```diff
@@ -0,0 +1,88 @@
...
+    explicit AddressMapper(const std::string& namespace_);
+    virtual ~AddressMapper() {};
+
+    // Maps an namespace string to a namespace address prefix.
```
aludvik: "an" -> "a"

sdk/cxx/include/sawtooth/address_mapper.h
```diff
@@ -0,0 +1,88 @@
...
+    // this provides a default implementation using SHA515.
```
aludvik: "SHA515" -> "SHA512"?

sdk/cxx/include/sawtooth/address_mapper.h
```diff
@@ -0,0 +1,88 @@
...
+    // it is intended to be overriden by subclass implementations
+    // that can provide their own mapping of namspace strings to
+    // prefixes. The requirements is that the mapping needs to
```
aludvik: "is" -> "are"

sdk/cxx/include/sawtooth/address_mapper.h
```diff
@@ -0,0 +1,88 @@
...
+    // return a prefix that is 6 characters in length and consists
+    // of only lower case hexadecimal characters(0123456789abcdef).
+    virtual std::string MapNamespace(const std::string& key);
+
+    // Maps an key string to an address, this provides a default
```
aludvik: "an" -> "a"

sdk/cxx/include/sawtooth/address_mapper.h
```diff
@@ -0,0 +1,88 @@
...
+    // implementation using SHA515.
```
aludvik: "SHA512"

sdk/cxx/include/sawtooth/address_mapper.h
```diff
@@ -0,0 +1,88 @@
...
+    // It is intended to be overriden by subclass implementations
+    // that can provide their own mappings. The requirements is that
```
aludvik: "are"

sdk/cxx/include/sawtooth/address_mapper.h
```diff
@@ -0,0 +1,88 @@
...
+//
+// Each instance of the mapper holds a namespace that has is mapped to a
+// namespace prefix when it is constructed. If your transaction processor
+// operates on multiple namespaces, it is recommened that you use an instance
+// for each namespace.
+class AddressMapper {
+ public:
+    // Constructor for an aaddress mapping object, Takes the
+    // namesspace it will be mapping objects to as a parameter.
+    explicit AddressMapper(const std::string& namespace_);
```
aludvik: Reading the comment I wasn't sure if this took "intkey" or "1cf126". From the source it looks like "intkey".

sdk/cxx/include/sawtooth/address_mapper.h
```diff
@@ -0,0 +1,88 @@
...
+    // in length and consists of only lower case hexadecimal characters.
+    virtual std::string MapKey(const std::string& key);
+
+
+    // returns the namespace prefix generated when the class was constructed.
+    std::string GetNamespacePrefix() {
+        return namespace_prefix;
+    }
+
+    // Map the key passed to the function and concatinates it with the
```
aludvik: "Maps", "concatenates"

sdk/cxx/include/sawtooth/global_state.h
```diff
@@ -0,0 +1,77 @@
...
+    //      std::vector<GlobalState::KeyValue> addresses;
+    //      addresses.push(std::make_pair(address, value));
+    void Set(const std::vector<KeyValue>& addresses);
+
+    // Delete a single GlobalState value. Used the batch Delete function
+    // defined below if you are deleting multiple items.
+    void Delete(const std::string& address) {
+        std::vector<std::string> addrs = { address };
+        this->Delete(addrs);
+    }
```
aludvik: Should this implementation be in global_state.c?

sdk/cxx/include/sawtooth/global_state.h
```diff
@@ -0,0 +1,77 @@
...
+
+#include "sawtooth/message_stream.h"
+
+namespace sawtooth {
+
+// Provides access to the global state
+class GlobalState {
+ public:
+    typedef std::pair<std::string, std::string> KeyValue;
+ public:
```
aludvik: Does this need two "pubic:" lines?

sdk/cxx/include/sawtooth/message_dispatcher.h
```diff
@@ -0,0 +1,79 @@
...
+#include <zmqpp/context.hpp>
+#include <zmqpp/socket.hpp>
+
+#include "sawtooth/message_stream.h"
+
+namespace sawtooth {
+
+// Main message processing handler. This object owns the ZMQ socket connection
+// to the validator as well as the internal routing sockets to route messages
+// processing request to the main thread for processing and listens for
```
aludvik: ~processing request~

sdk/cxx/include/sawtooth/message_dispatcher.h
```diff
@@ -0,0 +1,79 @@
...
+    MessageStreamPtr CreateStream();
+    zmqpp::context& context() { return this->context_; }
+
+ private:
+    void ReceiveMessage();
+    void SendMessage();
+
+    static void DispatchThread(MessageDispatcher* dispatcher);
+
+ private:
```
aludvik: Two "private:"?

sdk/cxx/include/sawtooth/message_stream.h
```diff
@@ -0,0 +1,82 @@
...
+#include <zmqpp/socket.hpp>
+
+#include "proto/validator.pb.h"
+
+#include "sawtooth/future_message.h"
+
+
+namespace sawtooth {
+
+// The MessageStream provides a conduit for message between the Transaction
```
aludvik: "messages"

sdk/cxx/include/sawtooth/message_stream.h
```diff
@@ -0,0 +1,82 @@
...
+class MessageStream {
+ public:
+    virtual ~MessageStream() {};
+
+    // Send a message to the validator and get back the FutureMessage object
+    // that will receive the response.
+    FutureMessagePtr Send(Message::MessageType type,
+        const std::string& data);
+
+    // This is a helper function that takes care of serializeing a protobuffer
```
aludvik: "serializing"

sdk/cxx/include/sawtooth/message_stream.h
```diff
@@ -0,0 +1,82 @@
...
+    // defined class and then sending the serialized results to the validator.
+    template <typename T>
+    FutureMessagePtr SendMessage(Message::MessageType type, const T& proto) {
+        std::stringstream proto_stream;
+        proto.SerializeToOstream(&proto_stream);
+        return this->Send(type, proto_stream.str());
+    }
```
aludvik: Should this be in "message_stream.cpp"?

sdk/cxx/include/sawtooth/transaction_handler.h
```diff
@@ -0,0 +1,111 @@
...
+    sawtooth::GlobalStateUPtr state;
+};
+typedef std::unique_ptr<TransactionApplicator> TransactionApplicatorUPtr;
+
+
+// Definition of a TransactionHandler to be registered with the validator.
+// Every Transaction Processor should register at least one of these. This
+// definition allows for the handler to declare multiple version and encodings
+// that it supports.
+class TransactionHandler {
```
aludvik: Why is TransactionApplicator a separate class from TransactionHandler?

sdk/cxx/src/address_mapper.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include "cryptopp/filters.h"
+#include "cryptopp/hex.h"
+#include "log4cxx/logger.h"
+
+#include "proto/state_context.pb.h"
+
+namespace sawtooth {
+
+static log4cxx::LoggerPtr  logger(log4cxx::Logger::getLogger
+    ("sawtooth.address_mapper"));
```
aludvik: Is there a way to generate this string instead of hard-coding it so it doesn't need to be updated when the filename changes?

sdk/cxx/src/address_mapper.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+// Helper function to generate an SHA 512 hash and return it as a hex
```
aludvik: "a SHA512"

sdk/cxx/src/address_mapper.cpp
```diff
@@ -0,0 +1,105 @@
...
+// encoded string.
+std::string SHA512(std::string message){
+    std::string digest;
+    CryptoPP::SHA512 hash;
+
+    CryptoPP::StringSource hasher(message, true,
+    new CryptoPP::HashFilter(hash,
+      new CryptoPP::HexEncoder (
+         new CryptoPP::StringSink(digest), false)));
```
aludvik: Seems expensive to create all these objects for each call. Could they be made persistent? Or are they cheap to create?

sdk/cxx/src/address_mapper.cpp
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// Check that a string only contains valid lower case hex characters.
+bool IsHex(const std::string& str) {
+    return str.find_first_not_of("0123456789abcdef", 2) ==
+        std::string::npos;
+}
+
+// Checks if an address string meets the constraints of an address.
+// These are that is exactly 134 characters long and contains only
```
aludvik: "it is"

sdk/cxx/src/address_mapper.cpp
```diff
@@ -0,0 +1,105 @@
...
+            << "characters: " << addr.length() << " != 134";
+        throw AddressFormatError(out.str());
+    } else if (!IsHex(addr)) {
+        throw AddressFormatError("Address must contain only " \
+            "lowercase hexadecimal characters");
+    }
+}
+
+// Checks if an address string meets the constraints of an address.
+// These are that is exactly 6 characters long and contains only
```
aludvik: "it is"

sdk/cxx/src/message_dispatcher.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+
+namespace sawtooth {
+
+static log4cxx::LoggerPtr logger(log4cxx::Logger::getLogger
+    ("sawtooth.MessageDispatcher"));
+
+MessageDispatcher::MessageDispatcher():
+        context_(),
+        server_socket(this->context_, zmqpp::socket_type::dealer),
```
aludvik: If you drink the ZMQ kool-aid, socket creation should be done in the thread where the socket is used, so in `MesageDispatcher::DispatchThread()`.

sdk/cxx/tp_intkey_tester.py
```diff
@@ -0,0 +1,109 @@
+# Copyright 2016 Intel Corporation
```
aludvik: This file shouldn't be needed, since it already exists at `sdk/examples/intkey_python/tests/test_tp_intkey.py`

sdk/examples/intkey_cxx/int_key_transaction_processor.cpp
```diff
@@ -0,0 +1,292 @@
...
+using namespace nlohmann;
+
+static log4cxx::LoggerPtr logger(log4cxx::Logger::getLogger
+    ("sawtooth.IntKeyCXX"));
+
+static const std::string INTKEY_NAMESPACE = "intkey";
+
+// Custom Exception type defined to indicate any
+// failure in process a transaction.
+class InvalidTransaction: public std::runtime_error {
```
aludvik: Can this error be part of the SDK instead of intkey since it will be used by other processors?

sdk/cxx/include/sawtooth/global_state.h
dcmiddle: I don't think it should have any pubic lines.

peterschwarz:
Is there no other way to include `sdk/examples/intkey_cxx/json.hpp` without committing it to the repo?

sdk/cxx/src/transaction_processor.cpp
```diff
@@ -0,0 +1,171 @@
...
+    try {
+        this->message_disptcher.Connect(this->connection_string);
+        this->response_stream = this->message_disptcher.CreateStream();
+
+        this->Register();
+
+        zmqpp::socket socket(this->message_disptcher.context(), zmqpp::socket_type::dealer);
+        socket.connect("inproc://request_queue");
+
+        while (this->run) {
```
jjason: Is `this->run` something that should be protected by mutex/condition variable?

sdk/examples/intkey_cxx/int_key_transaction_processor.cpp
```diff
@@ -0,0 +1,292 @@
...
+ public:
+    explicit InvalidTransaction(std::string const& error)
+        : std::runtime_error(error)
+    {}
+};
+
+
+// utility function to provide copy conversion from vector of bytes
+// to a stl string container.
+void ToString(std::string* out, const std::vector<std::uint8_t>& in) {
```
jjason: I would tend to prefer making this function return a `std::string` and let the C++ compiler's return value optimization kick in.

sdk/examples/intkey_cxx/int_key_transaction_processor.cpp
```diff
@@ -0,0 +1,292 @@
...
+    out->resize(in.size());
+    const char* data = reinterpret_cast<const char*>(&in[0]);
+    out->assign(data, data+in.size());
+}
+
+// utility function to provide copy conversion from stl string container
+// to a vector of bytes.
+void ToVector(std::vector<std::uint8_t>* out, const std::string& in) {
```
jjason: Same comment about returning `std::vector`

sdk/examples/intkey_cxx/int_key_transaction_processor.cpp
```diff
@@ -0,0 +1,292 @@
...
+    IntKeyApplicator(sawtooth::TransactionUPtr txn,
+        sawtooth::GlobalStateUPtr state) :
+        TransactionApplicator(std::move(txn), std::move(state)),
+            address_mapper(INTKEY_NAMESPACE) { }
+
+    void Apply() {
+        LOG4CXX_DEBUG(logger, "IntKeyApplicator::Apply");
+
+        const std::string& raw_data = this->txn->payload();
+        std::vector<uint8_t> data_vector(raw_data.begin(), raw_data.end());
```
jjason: You created a `ToVector` function above to do this???

sdk/examples/intkey_cxx/int_key_transaction_processor.cpp
```diff
@@ -0,0 +1,292 @@
...
+        json intkey_cmd = json::from_cbor(data_vector);
+
+        if (!intkey_cmd.is_object()) {
+            throw InvalidTransaction("CBOR Object as the encoded command");
+        }
+        auto verb_it = intkey_cmd.find("Verb");
+        if (verb_it == intkey_cmd.end()) {
```
jjason: A nit, but I would tend to write:

`if (intkey_cmd.end() == verb_it)` to prevent accidental assignment, although g++ is probably much better at flagging that warning that MSFT compilers of the past.

jjason: Same comment applies to all `==` checks with an lvalue on the left side.

sdk/examples/intkey_cxx/int_key_transaction_processor.cpp
```diff
@@ -0,0 +1,292 @@
...
+        if (name.length() == 0 || name.length() >= 20) {
+            throw InvalidTransaction("Name is invalid, name muste be between" \
+                " 1 and 20 characters in length");
+        }
+
+        auto value_it = intkey_cmd.find("Value");
+        if (value_it == intkey_cmd.end()) {
+            throw InvalidTransaction("Value is required");
+        }
+        int value = (*value_it).get<uint32_t>();
```
jjason: Is there value in using specific-sized integers - i.e., `int32_t`?  If for some reason the `int` is really 64-bit, then there is a potential glitch if you expect a negative number to be a "big" `uint32_t` and converting to `int` you expect it to become negative number.

jjason: I guess I am trying to convince myself that conversions from negative to `uint32_t` to negative work in all cases and `int` is ambiguous.

sdk/examples/intkey_cxx/int_key_transaction_processor.cpp
```diff
@@ -0,0 +1,292 @@
...
+    sawtooth::AddressMapper address_mapper;
+
+ private:
+    std::string MakeAddress(const std::string& name) {
+        return this->address_mapper.MakeAddress(name);
+    }
+
+    // Handle an IntKey 'set' verb action. This sets a IntKey value to
+    // the given value and
+    void DoSet(const std::string& name, int value) {
```
jjason: I missed the discussion about whether or not IntKey allowed a value to be negative at the end of the operation (i.e., 'set', 'inc', 'dec').  Is it okay for a value to be negative?

jjason:
Would have preferred more file/class-level comments.

sdk/cxx/include/sawtooth/global_state.h
```diff
@@ -0,0 +1,77 @@
...
+    // be provided to transaction handlers when a transaction is being
+    // processed.
+    explicit GlobalState(const MessageStreamPtr& message_stream,
+        const std::string& context_id);
+    virtual ~GlobalState() {};
+
+    // Retrieve a single value from global state. If you are retrieving
+    // multiple values, it is encouraged to use the batch get overload defined
+    // below.
+    bool Get(std::string* out_value, const std::string& address);
```
cintel: per the google c++ style guide: https://google.github.io/styleguide/cppguide.html#Reference_Arguments

cintel: I will update the function to return the value directly, I have not fully incorporated the move semantic yet.

sdk/cxx/include/sawtooth/message_stream.h
```diff
@@ -0,0 +1,82 @@
...
+
+ private:
+    friend class MessageDispatcher;
```
cintel: It is to limit who can construct this class. Open to a another pattern to do this if you can suggest one?

sdk/cxx/include/sawtooth/transaction_handler.h
cintel: To enable parallel processing of Transactions ( not enabled yet in the c++ sdk, just a thread-pool away). Without the restriction that the Apply method being re-entrant.

sdk/cxx/src/address_mapper.cpp
cintel: Can be set to __FILE__ if we want the logs to contain source file names. I don't find that the most friendly log content.

sdk/cxx/include/sawtooth/global_state.h
jjason: Gotcha - it seems that their standard argues for taking pointer instead of reference, and I can buy that.  I would still vote for the third then - return the std::string instead - since they don't seem to address it (although didn't read the entire spec and just searched for terms I thought would match the situation of deciding between taking a pointer for output versus returning a single value).

sdk/examples/intkey_cxx/int_key_transaction_processor.cpp
cintel: Short answer is they are not but there will be a follow on task for the c++ sdk along with the other sdks to bring them inline with the intkey spec.

sdk/cxx/include/sawtooth/address_mapper.h
cintel: The intention is that the base implementation is instantiate and usable but may be derived from for specialized mapping functionality.

sdk/cxx/include/sawtooth/global_state.h
```diff
@@ -0,0 +1,77 @@
...
+class GlobalState {
+ public:
+    typedef std::pair<std::string, std::string> KeyValue;
+ public:
+    // Takes the messages stream that it will use to communicate with the
+    // Validator to retrieve the current state. It is not expected that this
+    // will be called directly from a Transaction Processor. An instance will
+    // be provided to transaction handlers when a transaction is being
+    // processed.
+    explicit GlobalState(const MessageStreamPtr& message_stream,
```
cintel: I've always used this pattern of sharing the reference from the calling function. The called function can create a reference to it if so desire but is not required to.

sdk/cxx/src/address_mapper.cpp
cintel: This is the recommended pattern from the CryptoPP docs.  When we get to profiling the SDK we can address this pattern if it becomes a hotspot.

sdk/cxx/src/transaction_processor.cpp
cintel: Converted it to a atomic_bool.

sdk/cxx/src/message_dispatcher.cpp
```diff
@@ -0,0 +1,146 @@
...
+    this->server_socket.close();
+    this->message_socket.close();
+    this->processessing_request_socket.close();
+}
+
+void MessageDispatcher::ReceiveMessage() {
+    zmqpp::message zmsg;
+    this->server_socket.receive(zmsg);
+
+    MessageUPtr msg_proto(new Message());
```
cintel: in 5 years when it is mature I will move to c++14

sdk/cxx/src/message_dispatcher.cpp
cintel: Fix, socket is constructed on first use.

cintel:
@aludvik @jjason I have updated the instructions above to run the intkey unit tests. Let me know if  you hit any rough patches.

tools/plugins/install_cxx_tools.sh
```diff
@@ -0,0 +1,33 @@
...
+. /vagrant/conf.sh
+
+set -e
+
+if [ ! -f /etc/debian_version ]; then
+    echo "Skipping $0"
+    exit 0
+fi
+
+apt -y install cmake g++ libzmq-dev libzmqpp-dev libcrypto++-dev liblog4cxx-dev
```
aludvik: I didn't notice this before, but the other SDKs are using libzmq3-dev. Is there a reason for using libzmq-dev? apt is telling me they conflict.

aludvik: Remove "libzmq-dev"

sdk/cxx/include/sawtooth/address_mapper.h
```diff
@@ -0,0 +1,95 @@
...
+    // implementation using SHA512.
+    // It is intended to be overriden by subclass implementations
+    // that can provide their own mappings. The requirements are that
+    // the mapping needs to return a string that is 128 characters
+    // in length and consists of only lower case hexadecimal characters.
+    virtual std::string MapKey(const std::string& key) const;
+
+
+    // returns the namespace prefix generated when the class was constructed.
+    std::string GetNamespacePrefix();
```
jjason: Mark as `const`

sdk/cxx/include/sawtooth/address_mapper.h
```diff
@@ -0,0 +1,95 @@
...
+
+    // Maps the key passed to the function and concatenates it with the
+    // namespace prefix this object was initialized with and returns
+    // a fully qualified Merkle Trie address. These addresses are always
+    // 134 characters in length and consist entirely of lower case hexidecimal
+    // characters(0123456789abcdef).
+    std::string MakeAddress(const std::string& key);
```
jjason: Mark as `const`

sdk/cxx/include/sawtooth/future_message.h
```diff
@@ -0,0 +1,93 @@
...
+            throw std::runtime_error(error.str());
+        }
+
+        const std::string& msg_data = this->msg->content();
+        proto->ParseFromArray(msg_data.c_str(), msg_data.length());
+    }
+
+    // Set the reponse message and signal any potential waiters that it has
+    // been received. The futureMessage takes ownership of the messages.
+    void SetMessage(MessageUPtr msg) {
```
jjason: Accepted

sdk/cxx/include/sawtooth/global_state.h
jjason: Accepted

sdk/cxx/include/sawtooth/message_stream.h
jjason: Accepted

sdk/cxx/include/sawtooth/transaction_handler.h
```diff
@@ -0,0 +1,118 @@
...
+// processing of a single transaction. This object should make no assumptions
+// about the thread and timing of when it will execute. The
+// TransactionApplicator owns the reference to the transaction and the global
+// state it is passed.
+class TransactionApplicator {
+ public:
+    TransactionApplicator(sawtooth::TransactionUPtr txn,
+        sawtooth::GlobalStateUPtr state) : txn(std::move(txn)),
+        state(std::move(state)) {}
+    ~TransactionApplicator() {}
```
jjason: This needs to be marked `virtual` since you have a pure virtual function and the intent is that the class should be derived from.  Deleting from the base class pointer will lead to shearing.

sdk/cxx/src/message_dispatcher.cpp
jjason: Accepted =)

sdk/cxx/src/transaction_processor.cpp
jjason: Accepted

jjason:
Tests passed locally.  Three small changes requested above.

sdk/cxx/include/sawtooth/address_mapper.h
cintel: Cannot be, since the virtual function overrides cannot be called in the constructor. GetNamespacePrefix generates and caches the namespace_prefix on first call.

sdk/cxx/include/sawtooth/address_mapper.h
cintel: Cannot be. Calls GetNamespacePrefix.

sdk/cxx/include/sawtooth/address_mapper.h
jjason: I would argue that because the cache is not affecting the externally-visible state, that you mark the method as `const` and then mark the variable holding the "cached" value as `mutable`.  However, I leave the ultimate decision to you.

PR #251, opened 2017-03-23
Updated consensus module interfaces

jjason:
This update changes the consensus module interface for BlockPublisher, BlockVerifier, and ForkResolver to take a state view factory object on creation instead of a state view object.  It allows them to be able to create an arbitrary state view based upon that state root hash of any block in the store.

### DIFF
--- a/validator/sawtooth_validator/journal/consensus/dev_mode/dev_mode_consensus.py
+++ b/validator/sawtooth_validator/journal/consensus/dev_mode/dev_mode_consensus.py
@@ -35,22 +35,28 @@ class BlockPublisher(BlockPublisherInterface):
      DevMode Consensus (BlockPublisher) will read these settings
      from the StateView when Constructed.
     """
-
-    def __init__(self, block_cache, state_view, batch_publisher, data_dir):
-        super().__init__(block_cache, state_view, batch_publisher, data_dir)
+    def __init__(self,
+                 block_cache,
+                 state_view_factory,
+                 batch_publisher,
+                 data_dir):
+        super().__init__(
+            block_cache,
+            state_view_factory,
+            batch_publisher,
+            data_dir)
 
         self._block_cache = block_cache
-        self._state_view = state_view
+        self._state_view_factory = state_view_factory
 
         self._start_time = 0
         self._wait_time = 0
-        config_view = ConfigView(self._state_view)
-        self._min_wait_time = config_view.get_setting(
-            "sawtooth.consensus.min_wait_time", 0, int)
-        self._max_wait_time = config_view.get_setting(
-            "sawtooth.consensus.max_wait_time", 0, int)
-        self._valid_block_publishers = config_view.get_setting(
-            "sawtooth.consensus.valid_block_publishers", None, list)
+
+        # Set these to default values right now, when we asked to initialize
+        # a block, we will go ahead and check real configuration
+        self._min_wait_time = 0
+        self._max_wait_time = 0
+        self._valid_block_publishers = None
 
     def initialize_block(self, block_header):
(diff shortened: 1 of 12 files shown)


jjason:
Initial submission only to see if hopelessly broke unit tests....passing locally, but I no longer have any confidence in my local system.

jjason:
All test passing.  Note - that this has the side effect that currently any settings for the PoET consensus module are not applied as they are not in the state view at the point where the genesis block is given to the PoET block publisher.

This will be fixed in the next sprint by @peterschwarz story on Genesis Consensus.  It is a larger discussion about whether or not the PoET consensus periodically checks config settings to see if things like target wait time should be updated, although this would have a cascading effect upon things like local mean computation.

jjason:
Updated to work around genesis block issue until Genesis Consensus is available.

consensus/poet/core/sawtooth_poet/poet_consensus/utils.py
```diff
@@ -103,3 +105,19 @@ def build_certificate_list(block_header,
...
+    Args:
+        block (BlockWrapper): The block for which to return the state view.
+        state_view_factory (StateViewFactor): The state view factory
+            to be used to create the state view object
+
+    Returns:
+        StateView: The read-only state view for the block
+    """
+    state_root_hash = \
+        block.state_root_hash if block is not None else INIT_ROOT_KEY
```
peterschwarz: This `INIT_ROOT_KEY` stuff should maybe be pushed into the `create_view` function.  It feels like `INIT_ROOT_KEY` is leaking out here.  Pass in `None` as a root hash to `create_view`, get the view of `INIT_ROOT_KEY`.  In other words, users of the `StateViewFactory` shouldn't have to know about this detail.

validator/sawtooth_validator/journal/consensus/dev_mode/dev_mode_consensus.py
```diff
@@ -35,16 +37,33 @@ class BlockPublisher(BlockPublisherInterface):
...
         self._wait_time = 0
-        config_view = ConfigView(self._state_view)
+
+        # Using the current chain head, we need to create a state view and
+        # then from that a config view so we can get consensus configuration
+        # values
+        state_root_hash = \
+            self._block_cache.block_store.chain_head.state_root_hash if \
+            self._block_cache.block_store.chain_head is not None else \
+            INIT_ROOT_KEY
```
peterschwarz: Likewise - here it is again.

consensus/poet/core/sawtooth_poet/poet_consensus/utils.py
jjason: That function should go away.  I actually had to use a different hack ;)

validator/sawtooth_validator/journal/consensus/dev_mode/dev_mode_consensus.py
jjason: Same goes here.  I need to go back and fix this to account for genesis block hack until we have genesis consensus.

jjason:
PR updated.

consensus/poet/core/sawtooth_poet/poet_consensus/utils.py
peterschwarz: I don't see any changes to `StateViewFactory`.  Do you not need to worry about this anymore?

jjason:
@peterschwarz Removed dependency on INIT_ROOT_KEY.  I think that what you are requesting is a good idea, but a bigger chunk than I want to tackle in this PR.  I can take a note to look at updating this in the future.

PR #252, opened 2017-03-23
Add --wait flag to sawtooth cli submit command

delventhalz:
If `--wait` is set on `sawtooth submit`, the CLI will waits for commit before printing the timestamp. If the Api times out before the batches commit, it will print a table with the batch ids and their statuses.

### DIFF
--- a/cli/sawtooth_cli/rest_client.py
+++ b/cli/sawtooth_cli/rest_client.py
@@ -40,12 +40,24 @@ class RestClient(object):
         return self._get('/batches/' + safe_id)['data']
 
     def list_state(self, subtree=None, head=None):
-        queries = RestClient._remove_nones(address=subtree, head=head)
-        return self._get('/state', queries)
+        return self._get('/state', address=subtree, head=head)
 
     def get_leaf(self, address, head=None):
-        queries = RestClient._remove_nones(head=head)
-        return self._get('/state/' + address, queries)
+        return self._get('/state/' + address, head=head)
+
+    def get_statuses(self, batch_ids, wait=None):
+        """Fetches the committed status for a list of batch ids.
+
+        Args:
+            batch_ids (list of str): The ids to get the status of.
+            wait (optional, int): Indicates that the api should wait to
+                respond until the batches are committed or the specified
+                time in seconds has elapsed.
+
+        Returns:
+            dict: A statuses map, with ids as keys, and statuses values
+        """
+        return self._post('/batch_status', batch_ids, wait=wait)['data']
 
     def send_batches(self, batch_list):
         """Sends a list of batches to the validator.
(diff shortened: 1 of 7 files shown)


cli/sawtooth_cli/submit.py
```diff
@@ -43,7 +43,14 @@ def do_submit(args):
     start = time.time()
 
     for batch_list in _split_batch_list(args, batches):
-        rest_client.send_batches(batch_list)
+        response = rest_client.send_batches(batch_list, args.wait)
+
+        # If told to wait, but timed out, print out statuses of batches
```
jsmitchell: How do we know if we received a full response or timed out here? Should we print some kind of notification if we timed out?

delventhalz: If timed out the `POST /batch` route sends back `data` with batch statuses, but otherwise does not. So, as written it prints out a list of batches, along with their commit status (like committed or pending), in the case of a timeout. The comment could probably be more clear about how this works.

That said, Shawn just requested some fairly significant functionality changes, so all of this is going to get refactored pretty dramatically anyway.

delventhalz:
PR has been significantly updated to meet Shawn's requirements that `sawtooth submit --wait` be able to stay open indefinitely. Anyone who reviewed it should rereview it.

rest_api/sawtooth_rest_api/openapi.yaml
```diff
@@ -177,6 +177,43 @@ paths:
         503:
           $ref: "#/responses/503ServiceUnavailable"
 
+    post:
+      summary: Fetches the committed statuses for a set of batches
+      description: |
+        Identical to `GET /batch_status`, but takes ids of batches as a JSON
+        formatted POST body rather than a query parameter. This allows for many
+        more batches to be checked and should be used when querying more than
+        ~15 ids.
```
peterschwarz: Lose the `~` . Let's just make a recommendation here.

cli/sawtooth_cli/rest_client.py
```diff
@@ -56,31 +68,35 @@ def send_batches(self, batch_list):
...
+            headers = {'Content-Type': 'application/octet-stream'}
+        else:
+            data = json.dumps(data).encode()
+            headers = {'Content-Type': 'application/json'}
+        headers['Content-Length'] = '%d' % len(data)
 
-        code, json_result = self._submit_request(
-            self._base_url + path + query_string)
-        if code == 200:
+        batch_request = urllib.Request(
```
peterschwarz: This is a generic function, so probably should rename to just `request`

cli/sawtooth_cli/submit.py
```diff
@@ -47,17 +48,45 @@ def do_submit(args):
...
+        wait_time = 0
+        start_time = time.time()
+
+        while wait_time < args.wait:
+            statuses = rest_client.get_statuses(
+                batch_ids,
+                args.wait - int(wait_time))
+            wait_time = time.time() - start_time
+
+            if all(s == 'COMMITTED' for s in statuses.values()):
```
peterschwarz: Should we sleep a bit in the else case, so as not to hammer the rest_api?

delventhalz: Added.

PR #255, opened 2017-03-24
Fix Devmode ForkResolver issues

agunde406:
BlockWrapper.block_header is actually BlockWrapper._block_header. Should use BlockWrapper.header instead. Also need to encode signer_pubkey and header_signature before hashing. (Issues found while running a two node network with devmode)


jjason:
My only comment would be to have more descriptive commit messages....otherwise, looks good.

agunde406:
Commit message updated to have more information

jjason:
Thanks @agunde406

PR #256, opened 2017-03-24
Vagrant install from sawtooth

aludvik:
Replaces (most) of the pip installs done in vagrant with apt-get installs from the new sawtooth repo.


dcmiddle:
vpn/fw was blocking access to the ubuntu key server. the update to the PR forcing that request over port 80 seems to have cleared it up for me.

PR #257, opened 2017-03-24
Fix shutdown smoke

boydjohnson:
This reverts the changes to run_docker_test that were made in a prior pr. Also rewrites shutdown-smoke to not require containers to be named.


boydjohnson:
I'm going to fix this issue once there are docker images with installed packages, so that the test container doesn't need to find where the host put sawtooth-core so that it can mount that directory in a sibling container. Closing for now.

PR #254, opened 2017-03-24
Update sawtooth-signing imports to use generic signing module

dcmiddle:
Signed-off-by: dcmiddle <dan.middleton@intel.com>

PR #261, opened 2017-03-25
Simplify Signing API

dcmiddle:
Simplify API
Updated all callers.

API implicitly encodes and decodes parameters.
Format parameters removed from API.

Basic usage:
    import sawtooth_signing as signing

    msg = 'this is a message'
    priv = signing.generate_privkey()
    pub = signing.generate_pubkey(priv)
    sig = signing.sign(msg, priv)
    ver = signing.verify(msg, sig, pub)

Storing / Retrieving:
    keyfile = open(filename, 'w')
    keyfile.write(priv)
    keyfile.close()

    keyfile = open(filename, 'r')
    priv = keyfile.readline()
    pub = signing.generate_pubkey(priv)
    sig = signing.sign(msg, priv)
    send_foo(msg, sig, pub)

    def receive_foo(msg, sig, pub):
            if signing.verify(msg, sig, pub):
                    # rely on message authenticity
            else:
                    # reject

### DIFF
--- a/core/sawtooth/client.py
+++ b/core/sawtooth/client.py
@@ -54,8 +54,8 @@ def _sign_message_with_transaction(transaction, message_type, key):
     of a sha256 hexdigest.
     """
     transaction['Nonce'] = time.time()
-    pub = signing.encode_pubkey(signing.generate_pubkey(key), "hex")
-    transaction["PublicKey"] = pub
+    pub = signing.generate_pubkey(key)
+    transaction['PublicKey'] = pub
     sig = signing.sign(_dict2cbor(transaction), key)
     transaction['Signature'] = sig
 
@@ -389,14 +389,11 @@ class SawtoothClient(object):
 
         if keystring:
             LOGGER.debug("set signing key from string\n%s", keystring)
-            self._signing_key = signing.encode_privkey(
-                signing.decode_privkey(keystring, 'wif'), 'hex')
+            self._signing_key = keystring
         elif keyfile:
             LOGGER.debug("set signing key from file %s", keyfile)
             try:
-                self._signing_key = signing.encode_privkey(
-                    signing.decode_privkey(
-                        open(keyfile, "r").read().strip(), 'wif'), 'hex')
+                self._signing_key = open(keyfile, "r").read().strip()
             except IOError as ex:
                 raise ClientException(
                     "Failed to load key file: {}".format(str(ex)))
--- a/core/sawtooth/client.py
+++ b/core/sawtooth/client.py
@@ -30,7 +30,7 @@ import cbor
 from sawtooth.exceptions import ClientException
 from sawtooth.exceptions import InvalidTransactionError
 from sawtooth.exceptions import MessageException
-from sawtooth_signing import secp256k1_signer as signing
+import sawtooth_signing as signing
 
 
 LOGGER = logging.getLogger(__name__)
--- a/signing/sawtooth_signing/__init__.py
+++ b/signing/sawtooth_signing/__init__.py
@@ -17,6 +17,26 @@
 
     This module provides an interface to signing operations that insulates
     the caller from decisions on the underlying crypto system.
+    All keys are returned as serialized strings.
+    All methods expect that serialized format.
+
+    Basic Usage:
+        import sawtooth_signing as signing
+
+        msg = 'this is a message'
+        priv = signing.generate_privkey()
+        pub = signing.generate_pubkey(priv)
+        sig = signing.sign(msg, priv)
+        ver = signing.verify(msg, sig, pub)
+
+        # Store Private Key
+        keyfile = open(filename, 'w')
+        keyfile.write(priv)
+        keyfile.close()
+
+        # Retrieve Private Key
+        keyfile = open(filename, 'r')
+        priv = keyfile.readline()
 
     As new crypto packages are implemented this package serves as a build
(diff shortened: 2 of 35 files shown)


peterschwarz:
Is there an order to this and #254?

core/sawtooth/client.py
```diff
@@ -54,8 +54,8 @@ def _sign_message_with_transaction(transaction, message_type, key):
     of a sha256 hexdigest.
     """
     transaction['Nonce'] = time.time()
```
peterschwarz: Are these 0.7-style transactions?  Is this code still necessary?  Maybe not an item for this PR, but something we should be aware of.

signing/sawtooth_signing/secp256k1_signer.py
```diff
@@ -88,38 +76,11 @@ def _decode_privkey_to_bytes(encoded_privkey, encoding_format='wif'):
...
-        encoding_format: string indicating format such as 'wif'
-
-    Returns:
-        private key object useable with this module
-    """
-    priv = _decode_privkey_to_bytes(encoded_privkey, encoding_format)
     return secp256k1.PrivateKey(priv, ctx=__CTX__)
 
 
 def generate_pubkey(privkey):
```
peterschwarz: Given that the format of the private key is defaulted, and we're expecting an encoded private key (based on the function body), we should have a keyword arg here for `format='wif'`

core/sawtooth/client.py
dcmiddle: I think most of this directory was removed in the great purge of 69cbb. By the comment, remainder was left for future work to port to 0.8. https://github.com/hyperledger/sawtooth-core/commit/69cbb589febd5ab451cd9b8dd54dabe728f84ad4

signing/sawtooth_signing/secp256k1_signer.py
dcmiddle: One of the intents of this PR is to save the developer from figuring out formats. Consequently the public methods do not expose any format options. Adding a format override to this method would breakdown that pattern. Incidentally, in the near future we may want to drop wif, which would, in turn eliminate the dependency on pybitcointools. This API insulates the callers from that kind of change.

dcmiddle: Regarding #254, I could take that one down. This PR builds on it. I left 254 up because it was pretty small and I thought it might go through before this one.

jsmitchell:
@jjason didn't you have a bunch of comments in the poet code regarding the signing stuff that you kept in as reminders about the process? I don't see them removed here, nor did I see them in a cursory review of this branch's poet dirs. What happened to them?

jjason:
@jsmitchell They are still in the PoET simulator code.

Here: https://github.com/hyperledger/sawtooth-core/blob/master/consensus/poet/simulator/sawtooth_poet_simulator/poet_enclave_simulator/poet_enclave_simulator.py#L158

and here: https://github.com/hyperledger/sawtooth-core/blob/master/consensus/poet/simulator/sawtooth_poet_simulator/poet_enclave_simulator/poet_enclave_simulator.py#L269

signing/sawtooth_signing/__init__.py
```diff
@@ -12,5 +12,17 @@
...
+    the caller from decisions on the underlying crypto system.
+
+    As new crypto packages are implemented this package serves as a build
+    time switch to select the crypto package for that sawtooth network.
+
+    For example 'from ed25519_signer import *' would change sawtooth's
+    crypto selection without requiring changes to the API consumers.
+"""
+# pylint: disable=wildcard-import
+from sawtooth_signing.secp256k1_signer import *
```
jjason: @peterschwarz I am not a Python-ista, is it more Pythonic to have:

`from sawtooth_signing import secp256k1_signer as signing`

(then you don't need the `# pytlint: disable=wildcard-import`)

and then in the other files:

`from sawtooth_signing import signing`

???

core/sawtooth/client.py
jjason: I think I would just leave this as-is.  It is probably broken in so many other ways :)

signing/sawtooth_signing/secp256k1_signer.py
```diff
@@ -184,7 +111,7 @@ def _decode_pubkey(serialized_pubkey, encoding_format='hex'):
 
 
 def generate_identifier(pubkey):
```
jjason: Is this method used any more?  I thought this was a vestige of the bitcoin tools.

dcmiddle: There's a call in core/sawtooth/client.py
https://github.com/hyperledger/sawtooth-core/blob/master/core/sawtooth/client.py#L405

I can't see where the result is ever used though.  Whether or not that client code should be pulled I think this PR has to not break that client.

jjason: Okay.  Since that whole dir is going away....don't know if we need that any more.  I thought that we were using sign public signer key as the "identifier" now....

dcmiddle:
Good memory @jsmitchell. I added a commit to cleanup/remove those comments.

jjason:
@dcmiddle Those were my "I spent half my Friday afternoon tracking this down and frustrated about it" comments =)

dcmiddle:
@jjason ... "were"  :)

PR #262, opened 2017-03-26
Add Genesis Consensus Module

peterschwarz:
### 	Add genesis consensus module

Add a consensus module for handling consensus of genesis blocks only. These blocks have no state associated with them, and therefore have no special information needed during block publishing or validation.

### Add block id to consensus selection

Add block id to the consensus selection to facilitate the defaults at the boundaries of the genesis block and genesis+1 blocks. In the case of the `NULL_BLOCK_IDENTIFIER`, the genesis consensus is the default. After that, `devmode` is the default.

### DIFF
--- a/validator/sawtooth_validator/journal/consensus/genesis/genesis_consensus.py
+++ b/validator/sawtooth_validator/journal/consensus/genesis/genesis_consensus.py
@@ -1,0 +1,86 @@
+# Copyright 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


validator/sawtooth_validator/journal/consensus/genesis/genesis_consensus.py
```diff
@@ -0,0 +1,86 @@
...
+                 block_cache,
+                 state_view_factory,
+                 batch_publisher,
+                 data_dir):
+        super().__init__(block_cache,
+                         state_view_factory,
+                         batch_publisher,
+                         data_dir)
+
+    def initialize_block(self, block_header):
```
jjason: I realize that the consensus factory is supposed to only return genesis as the consensus if it is the first block but as a safety precaution, maybe have `initialize_block`, `check_publish_block`, and `finalize_block` only return true if the block ID is the `NULL_BLOCK_IDENTIFIER`....or put asserts here to catch programmer errors in case of changes in the future.

validator/sawtooth_validator/journal/consensus/genesis/genesis_consensus.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def finalize_block(self, block_header):
+        """Returns `True`, as the genesis block is always considered good by the
+        genesis node.
+        """
+        return True
+
+
+class BlockVerifier(BlockVerifierInterface):
+    """The Genesis BlockVerifier validate that this consensus is only used in
```
jjason: Spelling nit.  ....validates....

validator/sawtooth_validator/journal/consensus/genesis/genesis_consensus.py
peterschwarz: I've added these changes and tests to correspond with the changes.

PR #265, opened 2017-03-27
Update the PoET consensus module so that it stops on non-PoET block

jjason:
This updates the PoET consensus module so that if it encounters a block not created by a PoET consensus module, it will stop walking back the block chain.

### DIFF
--- a/consensus/poet/core/sawtooth_poet/poet_consensus/poet_block_verifier.py
+++ b/consensus/poet/core/sawtooth_poet/poet_consensus/poet_block_verifier.py
@@ -123,18 +123,27 @@ class PoetBlockVerifier(BlockVerifierInterface):
                     utils.deserialize_wait_certificate(
                         block=block_wrapper,
                         poet_enclave_module=poet_enclave_module)
-                wait_certificate.check_valid(
-                    poet_enclave_module=poet_enclave_module,
-                    certificates=certificates,
-                    poet_public_key=validator_info.signup_info.poet_public_key)
+                if wait_certificate is None:
+                    raise \
+                        ValueError(
+                            'Being asked to verify a block that was not '
+                            'created by PoET consensus module')
+
+                try:
+                    poet_public_key = \
+                        validator_info.signup_info.poet_public_key
+                    wait_certificate.check_valid(
+                        poet_enclave_module=poet_enclave_module,
+                        certificates=certificates,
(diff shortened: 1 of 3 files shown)


jsmitchell:
Tested locally with two validators configured with poet consensus set via the genesis batch.

consensus/poet/core/sawtooth_poet/poet_consensus/poet_block_verifier.py
```diff
@@ -123,6 +123,12 @@ def verify_block(self, block_wrapper):
                     utils.deserialize_wait_certificate(
                         block=block_wrapper,
                         poet_enclave_module=poet_enclave_module)
+                if wait_certificate is None:
+                    LOGGER.error(
```
jsmitchell: What is the consequence of this to the validator? Is this a fatal error?

jjason: Good question.  I don't know if it serious enough to "take down" the validator (i.e., raise an exception) or if it is sufficient to return False to mean that this block should not be added to the chain.

jjason:
Updated PR to address @jsmitchell comment.

consensus/poet/core/sawtooth_poet/poet_consensus/poet_block_verifier.py
```diff
@@ -123,18 +123,27 @@ def verify_block(self, block_wrapper):
...
+
+                try:
+                    poet_public_key = \
+                        validator_info.signup_info.poet_public_key
+                    wait_certificate.check_valid(
+                        poet_enclave_module=poet_enclave_module,
+                        certificates=certificates,
+                        poet_public_key=poet_public_key)
+                except ValueError as ve:
+                    LOGGER.error('Wait certificate is not valid: %s', str(ve))
```
peterschwarz: Is this an error or a warning?  Does it require an action of the user (the admin, in this case)?

jjason: Technically, it is just a placeholder until the PoET consensus module places nice with the `poet genesis` comment (i.e., loading the sealed signup data that was generated so that it can seed the enclave with the keys).  That is in the PoET consensus state stuff I am working on now.

PR #268, opened 2017-03-28
Make it a little easier to get state view for corresponding block

jjason:
* Added member and static methods to the `BlockWrapper` class to make it a little easier for code to get the state view associated with a particular block.
* Updated `StateViewFactory.create_view` to default the state hash to `None` if not provided and also to return the default Merkle database is the state root hash is `None`.  This hopefully will alleviate some of the code from having to know about/use `INIT_ROOT_KEY`

### DIFF
--- a/validator/sawtooth_validator/state/state_view.py
+++ b/validator/sawtooth_validator/state/state_view.py
@@ -32,13 +32,22 @@ class StateViewFactory(object):
         """
         self._database = database
 
-    def create_view(self, state_root_hash):
+    def create_view(self, state_root_hash=None):
         """Creates a StateView for the given state root hash.
 
+        Args:
+            state_root_hash (str): The state root hash of the state view
+                to return.  If None, returns the state view for the
         Returns:
             StateView: state view locked to the given root hash.
         """
-        return StateView(MerkleDatabase(self._database, state_root_hash))
+        # Create a default Merkle database and if we have a state root hash,
+        # update the Merkle database's root to that
+        merkle_db = MerkleDatabase(self._database)
+        if state_root_hash is not None:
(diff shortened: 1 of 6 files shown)


jjason:
Thanks @boydjohnson .  There are a couple of places (`chain.py` and `publisher.py`) that still have reference, but I have been discussing with @peterschwarz what we can do to mitigate that as well and keep `INIT_ROOT_KEY` to be within `merkle.py` completely.

validator/sawtooth_validator/state/state_view.py
```diff
@@ -32,12 +32,20 @@ def __init__(self, database):
...
 
+        Args:
+            state_root_hash (str): The state root hash of the state view
+                to return.  If None, returns the state view for the
         Returns:
             StateView: state view locked to the given root hash.
         """
+        # If we are not given a state root hash, then return the default Merkle
+        # database
+        if state_root_hash is None:
```
peterschwarz: This could be simplifed to:
```
db = MerkleDatabase(self._database)
if state_root_hash is not None:
    db.set_root_hash(state_root_hash)

return Stateview(db)
```
I might have the wrong function name above, but I think it should work.

jjason: I'll change as I'll assume that there isn't a big performance hit to setting the Merkle root twice every time a state view is created with a non-None `state_root_hash`.

jjason:
Updated PR with @peterschwarz comments.

jjason:
Argh!  Failing PoET smoke test.....

jjason:
That's what I get for a being responsible and rebasing....

jjason:
Third time was a charm.

PR #270, opened 2017-03-28
Add rest_api startup to cluster command

ojalatodd:
The cluster start command (sawtooth cluster start) now starts one instance of the 
rest_api. 
Added to both docker and subprocess.

### DIFF
--- a/manage/sawtooth_manage/docker.py
+++ b/manage/sawtooth_manage/docker.py
@@ -171,18 +171,18 @@ class DockerNodeController(NodeController):
                 'entrypoint': 'bin/{} tcp://{}:40000'.format(proc, node_name)
             }
 
-        # start the rest_api for the genesis node only
-        if node_config.genesis:
+        # start the rest_api for the first node only
+        if node_num == '000':
             compose_dict['services']['rest_api'] = {
                 'image': 'sawtooth-dev-rest_api',
-                    'expose': ['40000', '8080'],
-                    'links': ['validator'],
-                    'volumes': ['/project:/project'],
-                    'container_name': '-'.join([self._prefix, 'rest_api', node_num]),
-                    # 'entrypoint': './bin/rest_api --stream-url tcp://sawtooth-cluster-0-validator-000:40000',
-                    # Confirm that node_name works for the host name of validator in docker network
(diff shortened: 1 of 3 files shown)


ojalatodd:
Tests may not pass due to some issues having to do with docker-compose or docker, recent changes by @aludvik . We will collaborate.

ojalatodd:
Holding on assigning reviewers until I can test properly with fixed docker images.

ojalatodd:
The changes to the code now work in my local testing.
@aludvik The container name changes seem to have done the trick.

manage/sawtooth_manage/docker.py
```diff
@@ -163,14 +163,28 @@ def start(self, node_config):
...
-                'image': 'sawtooth-{}'.format(proc),
+                'image': 'sawtooth-dev-{}'.format(proc),
                 'expose': ['40000'],
                 'links': ['validator'],
                 'volumes': ['/project:/project'],
                 'container_name': '-'.join([self._prefix, proc, node_num]),
                 'entrypoint': 'bin/{} tcp://{}:40000'.format(proc, node_name)
             }
 
+        # start the rest_api for the first node only
```
nick-drozd: Why not start the rest api for all the nodes? If I want to query different nodes, I'm going to have to start it anyway.

ojalatodd: @nick-drozd The use case didn't call for that. The idea was to be able to query blocks, state etc. to one of the validators in the network. 

@delventhalz Could you or @peterschwarz interject?

ojalatodd: Main point: the use case was to have one rest_api active, in addition to wanting to avoid unnecessary complexity. That's my take. ;)

boydjohnson:
I would think we would want to have a rest api per validator, to see if there is consensus.

ojalatodd:
@nick-drozd  and I discussed the issue regarding multiple rest_api instances, one for each validator.
Also discussed with @peterschwarz . 
We agreed that we could go with this PR, and iterate in the next sprint with a version that runs one rest_api for each validator.

aludvik:
Tested locally, can submit xo transactions against a cluster.

PR #271, opened 2017-03-28
Stop publisher from calling check_claim_block() if initialize_block() returns False

jjason:
The publisher should not call `check_claim_block()` if the consensus module returns False from `initialize_block()`.  `Publisher._build_block()` is now returning `None` in this case so that it doesn't think that it has a block it should try to claim.


jjason:
Argh!   Thwarted by PoET smoke test again....

jjason:
Rebuild passed.

PR #274, opened 2017-03-29
Fix the PoET fork resolver to be deterministic

jjason:
For time being PoET will use the following in determine fork reoslution:

If new fork longer (i.e., block number greater), choose new fork.
Else, if current fork longer, choose current fork.
Else, if new fork header signature greater then current, choose new fork.
Else, choose current fork.


jjason:
This may be tied together with other work in NML-2130

PR #277, opened 2017-03-29
Change serial scheduler to limit writes to database

boydjohnson:


boydjohnson:
Closing as right now this PR writes on every batch, when the desired behavior is to write on every block end, in the block validator, while in the block publisher it is to get the resulting state hash for all valid batches but not write it to state.

PR #287, opened 2017-03-30
Fix CLI epilogs to diplay properly

ojalatodd:
Used argparse formatter_class.RawDescriptionHelpFormatter
to cause epilogs for cli commands with epilogs to display
properly. The epilogs are formatted correctly and should not
wrap.

### DIFF
--- a/cli/sawtooth_cli/batch.py
+++ b/cli/sawtooth_cli/batch.py
@@ -16,6 +16,7 @@
 import sys
 import csv
 import json
+import argparse
 import yaml
 
 from sawtooth_cli import tty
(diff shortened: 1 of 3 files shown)


cli/sawtooth_cli/batch.py
```diff
@@ -56,7 +59,9 @@ def add_batch_parser(subparsers, parent_parser):
         Shows the data for a single batch, or for a particular property within
```
delventhalz: I know these aren't yours, but could we remove the extra indent from the first line of these epilogs? It looks weird to me now that it is formatted properly.

PR #288, opened 2017-03-31
Add validator ID to consensus module interfaces

jjason:


jjason:
This is a really easy one.

PR #291, opened 2017-03-31
Create PoET consensus state store

jjason:
This PR:

* Creates PoET consensus state class to manage per-block consensus state
* Creates PoET consensus state store for managing consensus stat objects, mapping a block ID to the consensus state at that point in time
* Updates the PoET block verifier to create consensus state for a block if it successfully verifies it
* Updates the PoET genesis tool to persist the sealed signup data in the consensus store at a well-known location
* Updates the PoET block publisher to look in the consensus store for sealed signup data so that it can give it to the PoET enclave

### DIFF
--- a/consensus/poet/core/sawtooth_poet/poet_consensus/consensus_state.py
+++ b/consensus/poet/core/sawtooth_poet/poet_consensus/consensus_state.py
@@ -1,0 +1,228 @@
+# Copyright 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+
+from collections import namedtuple
+
+import cbor
+
+ValidatorState = \
+    namedtuple(
+        'ValidatorState',
+        ['key_block_claim_count',
+         'poet_public_key',
+         'total_block_claim_count'
+         ])
+""" Instead of creating a full-fledged class, let's use a named tuple for
+the validator state.  The validator state represents the state for a single
+validator at a point in time.  A validator state object contains:
+
+key_block_claim_count (int): The number of blocks that the validator has
+claimed using the current PoET public key
+poet_public_key (str): The current PoET public key for the validator
+total_block_claim_count (int): The total number of the blocks that the
+    validator has claimed
+"""
+
+
+class ConsensusState(object):
+    """Represents the consensus state at a particular point in time (i.e.,
+    when the block that this consensus state corresponds to was committed to
+    the block chain).
+
+    Attributes:
+        sealed_signup_data (str): The encoded sealed signup data that was
+            received from the most recent creation of signup information
+            for the validator
+        expected_block_claim_count (float): The number of blocks that a
+            validator, based upon the population estimate, would be expected
+            to have claimed
+    """
+    def __init__(self):
+        """Initialize a ConsensusState object
+
+        Returns:
+            None
+        """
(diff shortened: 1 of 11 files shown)


jjason:
Updated based upon @peterschwarz comments.

consensus/poet/core/sawtooth_poet/poet_consensus/poet_fork_resolver.py
```diff
@@ -107,4 +121,75 @@ def compare_forks(self, cur_fork_head, new_fork_head):
...
+            except KeyError:
+                pass
+
+            state_view = \
+                BlockWrapper.state_view_for_block(
+                    block_wrapper=previous_block,
+                    state_view_factory=self._state_view_factory)
+
+            poet_enclave_module = \
+                factory.PoetEnclaveFactory.get_poet_enclave_module(state_view)
```
dcmiddle: Is there an issue if the previous block isn't a PoET block?
I think later on it may raise an attribute error when we attempt to deserialize the wait certificate if this module is None.

consensus/poet/core/sawtooth_poet/poet_consensus/consensus_state.py
```diff
@@ -0,0 +1,228 @@
...
+from collections import namedtuple
+
+import cbor
+
+ValidatorState = \
+    namedtuple(
+        'ValidatorState',
+        ['key_block_claim_count',
+         'poet_public_key',
+         'total_block_claim_count'
```
dcmiddle: Do you want to store a block depth since registration for C and K tests?
I believe key_block_claim_count is for Z test. 
I'm no sure what total_block_claim_count is used for?

consensus/poet/core/sawtooth_poet/poet_consensus/utils.py
```diff
@@ -113,3 +118,155 @@ def build_certificate_list(block_header,
...
+    total_block_claim_count = 1
+
+    if current_validator_state is not None:
+        total_block_claim_count = \
+            current_validator_state.total_block_claim_count + 1
+
+        if validator_info.signup_info.poet_public_key == \
+                current_validator_state.poet_public_key:
+            key_block_claim_count = \
+                current_validator_state.key_block_claim_count + 1
```
dcmiddle: Does this need to be computed over a window for z-test? i.e. of the last 10 blocks this poet public key claimed 5 blocks. (but over the lifetime of that key it claimed 500 blocks)

consensus/poet/core/sawtooth_poet/poet_consensus/poet_fork_resolver.py
jjason: The factory (`importlib.import_module()`) raises an exception if it cannot load.  I don't believe that `poet_enclave_module` can ever be `None`.

consensus/poet/core/sawtooth_poet/poet_consensus/utils.py
jjason: This will be taken care of when those stories are done.

consensus/poet/core/sawtooth_poet/poet_consensus/consensus_state.py
dcmiddle: Sure, and what about total_block_claim_count?

jjason: The total block count _probably_ will be used with doing zTest.  Right now, I am just keeping counts.  It was more useful for testing so I could for example run a validator and issues some intkey transactions and then dump the consensus state store to see if things were working properly.

consensus/poet/core/sawtooth_poet/poet_consensus/poet_fork_resolver.py
dcmiddle: You are correct. I missed that.

consensus/poet/core/sawtooth_poet/poet_consensus/consensus_state.py
dcmiddle: ok

consensus/poet/core/sawtooth_poet/poet_consensus/utils.py
dcmiddle: ok

jjason:
Rebasing before merge.

PR #295, opened 2017-04-03
Only write to state on genesis and block validation

boydjohnson:
The constructor to the SerialScheduler is changed in this PR, to accept the optional keyword argument always_persist, which can be used in genesis block construction to write the state change to the underlying database no matter what the resulting merkle hash is. When always_persist is False, the only time the state change is written to the underlying database is when the merkle root provided on the block boundary in add_batch equals the computed merkle root from the state changes.

### DIFF
--- a/validator/sawtooth_validator/execution/context_manager.py
+++ b/validator/sawtooth_validator/execution/context_manager.py
@@ -15,7 +15,7 @@
 
 import hashlib
 import logging
-import time
+import uuid
 
 from threading import Condition
 from threading import Lock
@@ -69,10 +69,7 @@ class StateContext(object):
         self._state = {}
         self.base_context_ids = base_context_ids
 
-        self._id = hashlib.sha256((str(state_hash) + ":" +
-                                  str(read_list + write_list) + ":" +
-                                  time.time().hex()).encode()
-                                  ).hexdigest()
+        self._id = hashlib.sha256(uuid.uuid4().hex.encode()).hexdigest()
 
     @property
     def session_id(self):
@@ -251,61 +248,6 @@ class ContextManager(object):
             self._address_queue.put_nowait(
                 (context.session_id, state_hash, list(reads)))
         return context.session_id
-
-    def commit_context(self, context_id_list, virtual):
-        """ Only used in a test ---
-        Commits the state from the contexts referred to in context_id_list
-        to the merkle tree.
-
-        Args:
-            context_id_list (list of str): The context ids with state to
-                commit to the merkle tree.
-            virtual (bool): True if the data in contexts shouldn't be
-                written to the merkle tree, but just return a merkle root.
-
-        Returns:
-            state_hash (str): the new state hash after the context_id_list
-                              has been committed
-
-        """
-        if any([c_id not in self._contexts for c_id in context_id_list]):
-            raise CommitException("Context Id not in contexts")
-        first_id = context_id_list[0]
-
-        if not all([self._contexts[first_id].merkle_root ==
-                    self._contexts[c_id].merkle_root
-                    for c_id in context_id_list]):
-            raise CommitException(
-                "MerkleRoots not all equal, yet asking to merge")
-
-        merkle_root = self._contexts[first_id].merkle_root
-        tree = MerkleDatabase(self._database, merkle_root=merkle_root)
-        updates = dict()
-        for c_id in context_id_list:
-            context = self._contexts[c_id]
-            for add in context.get_state().keys():
-                if add in updates:
-                    raise CommitException(
-                        "Duplicate address {} in context {}".format(
-                            add, c_id))
-
-            effective_updates = {}
-            for k, val_fut in context.get_state().items():
-                value = val_fut.result()
-                if value is not None:
-                    effective_updates[k] = value
-
-            updates.update(effective_updates)
-
-        state_hash = tree.update(updates, virtual=False)
-        # clean up all contexts that are involved in being squashed.
-        base_c_ids = []
-        for c_id in context_id_list:
-            base_c_ids += self._contexts[c_id].base_context_ids
-        all_context_ids = base_c_ids + context_id_list
-        self.delete_context(all_context_ids)
-
-        return state_hash
 
     def delete_context(self, context_id_list):
         """Delete contexts from the ContextManager.
--- a/validator/sawtooth_validator/execution/context_manager.py
+++ b/validator/sawtooth_validator/execution/context_manager.py
@@ -313,7 +313,7 @@ class ContextManager(object):
         return True
 
     def get_squash_handler(self):
-        def _squash(state_root, context_ids):
+        def _squash(state_root, context_ids, persist):
             tree = MerkleDatabase(self._database, state_root)
             updates = dict()
             for c_id in context_ids:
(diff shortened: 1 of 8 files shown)


agunde406:
I tested by running the poet-smoke test locally and saw the following error. It seems like a timing issues and does not happen every time the test is run (2 out of 4 tests):

 | [20:49:40.121 CRITICAL publisher] on_chain_updated exception.
validator-0_1              | [20:49:40.122 ERROR    publisher] 'hash ceeb977a59bce59cc69c992c8f501687dbeb77d20fa8b2194017a26de874786e not found in database'
validator-0_1              | Traceback (most recent call last):
validator-0_1              |   File "/project/sawtooth-core/validator/sawtooth_validator/journal/publisher.py", line 322, in on_chain_updated
validator-0_1              |     self._candidate_block = self._build_block(chain_head)
validator-0_1              |   File "/project/sawtooth-core/validator/sawtooth_validator/journal/publisher.py", line 107, in _build_block
validator-0_1              |     self._state_view_factory)
validator-0_1              |   File "/project/sawtooth-core/validator/sawtooth_validator/journal/block_wrapper.py", line 131, in state_view_for_block
validator-0_1              |     return state_view_factory.create_view(state_root_hash)
validator-0_1              |   File "/project/sawtooth-core/validator/sawtooth_validator/state/state_view.py", line 48, in create_view
validator-0_1              |     merkle_db.set_merkle_root(state_root_hash)
validator-0_1              |   File "/project/sawtooth-core/validator/sawtooth_validator/state/merkle.py", line 66, in set_merkle_root
validator-0_1              |     self._root_node = self._get_by_hash(merkle_root)
validator-0_1              |   File "/project/sawtooth-core/validator/sawtooth_validator/state/merkle.py", line 77, in _get_by_hash
validator-0_1              |     raise KeyError("hash {} not found in database".format(key_hash))
validator-0_1              | KeyError: 'hash ceeb977a59bce59cc69c992c8f501687dbeb77d20fa8b2194017a26de874786e not dator-0_1              | KeyError: 'hash ceeb977a59bce59cc69c992c8f501687dbeb77d20fa8b2194017a26de874786e not found in database'

jsmitchell:
ditto @agunde406

validator/sawtooth_validator/execution/scheduler.py
```diff
@@ -101,9 +101,18 @@ def finalize(self):
         raise NotImplementedError()
 
     @abstractmethod
-    def complete(self, block):
+    def complete(self, block, force=False):
```
peterschwarz: Can we name this better?  `force_write_state` perhaps?  Likewise in the other signatures.  `force` by itself doesn't communicate much in the places where it's used.

jsmitchell: how about `persist`

peterschwarz: `persist` works for me

boydjohnson: Yeah, seems like a good change. I'll update to force_write_state.

validator/sawtooth_validator/journal/publisher.py
```diff
@@ -370,7 +370,8 @@ def _finalize_block(self, block):
                 else:
                     block.add_batch(batch)
                     self._committed_txn_cache.add_batch(batch)
-                state_hash = result.state_hash
+                if result.state_hash is not None:
+                    state_hash = result.state_hash
```
cintel: Is this check necessary? Does this handle the case where a transaction does not update the Merkle Trie?

boydjohnson: It is more about how data is being passed back to users of the scheduler. No single txn will get a new state hash, and no single batch will get a new state hash, but the result of all of the valid batches will get a state hash (unless it is happening on block validation). So, At most one of the result.state_hash will be not None, and only that one should be recorded in the variable state_hash.

cintel: This feels like a very awkward interface, that we have to fish through the batch results and look for a valid state hash.  It feels that there should be a property set on the scheduler once  self._scheduler.complete(block=True) returns that is the root_state_hash that was generated for all the batches validated.

validator/sawtooth_validator/execution/scheduler.py
boydjohnson: How about force_persist? Because there would be other times that the tree would be persisted (Like in block validation where a resulting state root is given and it matches what transaction execution finds to be the state root -- it will persist that).

validator/sawtooth_validator/journal/publisher.py
boydjohnson: I tend to agree. From talking with @vaporos and @jsmitchell it sounds like there might be some use in the future for a multi-block scheduler, and so it wouldn't be just one state hash returned. Since the SerialScheduler is more of a first pass until we get the parallel scheduler implemented (and they should be entirely interchangable), I don't think we should try to change the interface for the scheduler.

validator/sawtooth_validator/execution/scheduler_serial.py
```diff
@@ -42,54 +42,64 @@ def __init__(self, squash_handler, first_state_hash):
...
             if txn_signature not in self._txn_to_batch:
                 raise ValueError("transaction not in any batches: {}".format(
                     txn_signature))
+
+            batch_signature = self._txn_to_batch[txn_signature]
             if is_valid:
-                # txn is valid, get a new state hash
-                state_hash = self._squash(self._last_state_hash, [context_id])
-                self._last_state_hash = state_hash
+                self._last_context_id = context_id
```
peterschwarz: Is this the previous context id?  If so, we should use `_previous_context_id`, to differentiate from `_last_in_batch`: the two together make it seem like the final one in a list.

validator/sawtooth_validator/execution/scheduler_serial.py
```diff
@@ -42,54 +42,64 @@ def __init__(self, squash_handler, first_state_hash):
...
             else:
                 # txn is invalid, preemptively fail the batch
-                batch_signature = self._txn_to_batch[txn_signature]
                 self._batch_statuses[batch_signature] = \
                     BatchExecutionResult(is_valid=is_valid, state_hash=None)
             if txn_signature in self._last_in_batch:
-                batch_signature = self._txn_to_batch[txn_signature]
                 if batch_signature not in self._batch_statuses:
                     # because of the else clause above, txn is valid here
+                    self._last_valid_batch_c_id = self._last_context_id
```
peterschwarz: Likewise, with `_last...` to `_previous...`, if that's its meaning

boydjohnson: Yes, that is the meaning. I will change.

validator/sawtooth_validator/execution/context_manager.py
```diff
@@ -69,10 +69,7 @@ def __init__(self, state_hash, read_list, write_list, base_context_ids):
         self._state = {}
         self.base_context_ids = base_context_ids
 
-        self._id = hashlib.sha256((str(state_hash) + ":" +
-                                  str(read_list + write_list) + ":" +
-                                  time.time().hex()).encode()
-                                  ).hexdigest()
+        self._id = hashlib.sha256(uuid.uuid4().hex.encode()).hexdigest()
```
jsmitchell: sha256 is expensive. Can we get by without hashing uuid.uuid4()?

boydjohnson: I will look into it. Since we have a context id for each txn, and will hang onto them for an entire block, there could be potentially many at a time.

jsmitchell: Well, surely if you get a collision on uuid.uuid4(), hashing the result is not going to help :)

boydjohnson: Ah, definitely true. Thanks!

PR #297, opened 2017-04-04
Fix chain controller block processing issues

jjason:
Updated `ChainController.on_block_validated()` to properly manage the pending blocks queue based upon the result of the block validation.

To test this fix:

1.  In `validator/sawtooth_validator/execution/executor.py` line 230, replace:

`self._executing_threadpool = ThreadPoolExecutor(max_workers=5)`

with

`self._executing_threadpool = ThreadPoolExecutor(max_workers=100)`

2.  In `integration/sawtooth_integration/docker/poet-smoke.yaml`, after the line:

 add the following lines:

`sawtooth.poet.target_wait_time=5 \`
`sawtooth.poet.initial_wait_time=0 \`

3.  Run the PoET smoke test:

`cd integration/sawtooth_integration/docker`

Hopefully test will pass.

### DIFF
--- a/validator/sawtooth_validator/journal/chain.py
+++ b/validator/sawtooth_validator/journal/chain.py
@@ -627,6 +627,11 @@ class ChainController(object):
                     self._set_genesis(block)
                     return
 
+                # If we are already currently processing this block, then
+                # don't bother trying to schedule it again.
+                if block.identifier in self._blocks_processing:
+                    return
+
                 self._block_cache[block.identifier] = block
                 self._blocks_pending[block.identifier] = []
                 LOGGER.debug("Block received: %s", block)
--- a/validator/sawtooth_validator/journal/chain.py
+++ b/validator/sawtooth_validator/journal/chain.py
@@ -406,8 +406,8 @@ class BlockValidator(object):
             LOGGER.error("Block validation failed with unexpected error: %s",
                          self._new_block)
             LOGGER.exception(exc)
-            self._done_cb(False)  # callback to clean up the block out of the
-            # processing list.
+            # callback to clean up the block out of the processing list.
+            self._done_cb(False, self._result)
 
 
 class ChainController(object):
--- a/validator/sawtooth_validator/journal/chain.py
+++ b/validator/sawtooth_validator/journal/chain.py
@@ -483,27 +483,29 @@ class ChainController(object):
     def chain_head(self):
         return self._chain_head
 
-    def _verify_block(self, blkw):
-        state_view = BlockWrapper.state_view_for_block(
-            self.chain_head,
-            self._state_view_factory)
-        consensus_module = ConsensusFactory.get_configured_consensus_module(
-            self.chain_head.header_signature,
-            state_view)
+    def _submit_blocks_for_verification(self, blocks):
+        for blkw in blocks:
+            state_view = BlockWrapper.state_view_for_block(
+                self.chain_head,
+                self._state_view_factory)
+            consensus_module = \
+                ConsensusFactory.get_configured_consensus_module(
+                    self.chain_head.header_signature,
+                    state_view)
 
-        validator = BlockValidator(
-            consensus_module=consensus_module,
-            new_block=blkw,
-            chain_head=self._chain_head,
-            block_cache=self._block_cache,
-            state_view_factory=self._state_view_factory,
-            done_cb=self.on_block_validated,
-            executor=self._transaction_executor,
-            squash_handler=self._squash_handler,
-            identity_signing_key=self._identity_signing_key,
-            data_dir=self._data_dir)
-        self._blocks_processing[blkw.block.header_signature] = validator
-        self._executor.submit(validator.run)
+            validator = BlockValidator(
+                consensus_module=consensus_module,
+                new_block=blkw,
+                chain_head=self._chain_head,
+                block_cache=self._block_cache,
+                state_view_factory=self._state_view_factory,
+                done_cb=self.on_block_validated,
+                executor=self._transaction_executor,
+                squash_handler=self._squash_handler,
+                identity_signing_key=self._identity_signing_key,
+                data_dir=self._data_dir)
+            self._blocks_processing[blkw.block.header_signature] = validator
+            self._executor.submit(validator.run)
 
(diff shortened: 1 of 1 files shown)


jjason:
To test this fix:

1.  In `validator/sawtooth_validator/execution/executor.py` line 230, replace:

`self._executing_threadpool = ThreadPoolExecutor(max_workers=5)`

with

`self._executing_threadpool = ThreadPoolExecutor(max_workers=100)`

2.  In `integration/sawtooth_integration/docker/poet-smoke.yaml`, after the line:

 add the following lines:

`sawtooth.poet.target_wait_time=5 \`
`sawtooth.poet.initial_wait_time=0 \`

3.  Run the PoET smoke test:

`cd integration/sawtooth_integration/docker`

Hopefully test will pass.

jjason:
Anyone know why this didn't kick off a build?

validator/sawtooth_validator/journal/chain.py
```diff
@@ -505,6 +505,37 @@ def _verify_block(self, blkw):
         self._blocks_processing[blkw.block.header_signature] = validator
         self._executor.submit(validator.run)
 
+    def _remove_block_and_verify_descendants(self, parent_block):
```
cintel: _remove_block_and_verify_descendants is confusing as a name. '_verify_descendants' is clearer.

validator/sawtooth_validator/journal/chain.py
```diff
@@ -505,6 +505,37 @@ def _verify_block(self, blkw):
...
+            # pending block queue, we schedule them for verification.
+            LOGGER.debug(
+                'Verify descendant blocks: %s (%s)',
+                parent_block,
+                [block.identifier[:8] for block in pending_blocks])
+
+            for pending_block in pending_blocks:
+                self._verify_block(pending_block)
+
+        return pending_blocks
```
cintel: just return pending_blocks is not None

validator/sawtooth_validator/journal/chain.py
```diff
@@ -525,9 +556,27 @@ def on_block_validated(self, commit_new_block, result):
...
+
+                    # The chain head has advanced since work started.  Remove
+                    # this block from the pending blocks queue and verify any
+                    # pending blocks.  If there were none, then we are going
+                    # to add it back to the pending blocks queue (just to be
+                    # safe) and verify it again in order to give it another
+                    # chance to become the chain head.
+                    pending_blocks = \
+                        self._remove_block_and_verify_descendants(new_block)
+                    if not pending_blocks:
```
cintel: if not self._verify_decendants(new_block):

validator/sawtooth_validator/journal/chain.py
```diff
@@ -525,9 +556,27 @@ def on_block_validated(self, commit_new_block, result):
...
                 # if the head has changed, since we started the work.
                 if result["chain_head"] != self._chain_head:
-                    # chain has advanced since work started.
-                    # the block validation work we have done is saved.
-                    self._verify_block(new_block)
+                    LOGGER.info(
+                        'Chain head updated from %s to %s while processing '
+                        'block: %s',
+                        result["chain_head"].identifier[:8],
+                        self._chain_head.identifier[:8],
```
cintel: These should both be BlockWrappers, don't format the identifier let the str operator format the object
result["chain_head"],
self._chain_head

validator/sawtooth_validator/journal/chain.py
jjason: It was initially named `_verify_descendants()` and I went back and forth between names, and settled on `_remove_...()` because it was explicit about the removing part.  I can change back.

validator/sawtooth_validator/journal/chain.py
jjason: Can do, although I did this because all I really cared about in the message was the identifier part.

validator/sawtooth_validator/journal/chain.py
cintel: The 'remove_block is confusing as to where the block is being removed from.

jjason: If it wasn't for link, I could name it `_remove_block_from_pending_queue_and_process_descendants()` ;)

validator/sawtooth_validator/journal/chain.py
jjason: Can do, although I had it return the list of pending blocks in case the caller cared about them.

jjason:
PR pushed addressing @cintel comments.

validator/sawtooth_validator/journal/chain.py
```diff
@@ -543,11 +587,42 @@ def on_block_validated(self, commit_new_block, result):
...
+
+                # The block is otherwise valid, but we have determined we
+                # don't want it as the chain head.
+                else:
+                    LOGGER.info('Rejected new chain head: %s', new_block)
+
+                    # Remove the block from the pending blocks and verify
+                    # any immediate descendants that arrived while we were
+                    # processing it.
+                    self._verify_descendants(new_block)
```
peterschwarz: Calling this `_verify_descentants` really seems to hide what the code is doing.  The fact that there is a comment above the call (here and in the other places it's called), describing what the function is doing, tells us that this is a badly named function.  

Since it is called by itself here at this line, the expectation is that if it failed verification (the implied idea by using "verify" - there are situations where this could fail), it should thrown an exception.  Looking at the implementation, it returns false, but it has the side-effect of removing a block from the pending queue.  We're still not really looking at the result of return value in this instance.

The old name didn't hide this, but perhaps the reason it seems wrong is it is doing to many things.  This could be divided into two operations:

```
remaining_blocks = \
            self._blocks_pending.pop(new_block.identifier, [])
self._verify_all(remaining_blocks)
```
where verify_all does the "verify" portion of the operation.   It takes the descendants out of the equation, as well, and opens the function up to broader usage.  

Part of the confusion about this function name, for me anyway, may also be related to the overloading of the word "verify" that some members of the team think about in it's domain-specific way, and I'm still thinking about the use of "verify" in terms of an general operation .

jjason:
Updated PR with code for another bug @nick-drozd found and addressing @peterschwarz comments.

PR #300, opened 2017-04-04
Add pagination to the Rest Api

delventhalz:
Adds paging controls and responses to the Rest Api including:
- Tweaks to how the validator handles paging
- Major refactor of Route Handlers to update syntax, styling, and docstrings
- Sending `PagingControls` based on `min`, `max`, and `count` queries
- Building a `paging` response envelope, including `next` and `previous` urls
- Unit testing many possible paging controls
- Updating existing unit tests to expect paging responses where applicable
- Updating Api spec as needed
- Minor typo corrections

This is a massive PR. I highly recommend viewing it on a commit-by-commit basis!


rest_api/sawtooth_rest_api/openapi.yaml
```diff
@@ -600,21 +598,21 @@ definitions:
...
+      start_index:
+        type: integer
+        example: 1000
       total_count:
         type: integer
         example: 54321
       previous:
         type: string
-        example: https://api.sawtooth.com/state?max_position=1001
+        example: https://api.sawtooth.com/state?min=0&count=1000
```
peterschwarz: Shouldn't this include the head parameter?

rest_api/sawtooth_rest_api/route_handlers.py
```diff
@@ -245,19 +251,24 @@ def __init__(self, stream_url, timeout=DEFAULT_TIMEOUT):
...
             link: The link to this exact query, including head block
+            paging: Paging info and nav, like total resources and a next link
         """
-        head = request.url.query.get('head', None)
-        ids = self._get_filter_ids(request)
+        paging_controls = self._get_paging_controls(request)
+        validator_query = client_pb2.ClientBlockListRequest(
+            head_id=request.url.query.get('head', None),
+            block_ids=self._get_filter_ids(request),
+            paging=self._make_proto_paging(paging_controls))
```
peterschwarz: suggest renaming this to `_make_paging_message`; less tied to protobuf as a name.

rest_api/sawtooth_rest_api/route_handlers.py
```diff
@@ -467,6 +561,64 @@ def _parse_header(cls, header_proto, obj):
...
+        min_pos = request.url.query.get('min', None)
+        max_pos = request.url.query.get('max', None)
+        count = request.url.query.get('count', None)
+        controls = {}
+
+        if count == '0':
+            raise errors.BadCount()
+
+        try:
+            controls['count'] = int(count)
```
peterschwarz: Since we have a BadCount error, throw BadCount if not None

rest_api/sawtooth_rest_api/route_handlers.py
```diff
@@ -88,13 +90,12 @@ def batches_post(self, request):
...
 
-        return RouteHandler._wrap_response(
+        return self._wrap_response(
             data=data,
             metadata={'link': link},
             status=status)
 
-    @asyncio.coroutine
-    def status_list(self, request):
+    async def list_statuses(self, request):
```
peterschwarz: Does this change require a certain level of Python 3? (3.5, I think?). If we're fine with that constraint, we will need to be sure that we're documenting the minimum python requirements. if not, we can't make this change.

cli/sawtooth_cli/rest_client.py
```diff
@@ -78,14 +78,21 @@ def send_batches(self, batch_list):
         return self._post('/batches', batch_list)
 
     def _get(self, path, **queries):
-        code, json_result = self._submit_request(
+        code, js_result = self._submit_request(
```
peterschwarz: `json_result` was more correct.  `js_result` implies JavaScript

validator/sawtooth_validator/state/client_handlers.py
```diff
@@ -32,8 +31,7 @@
 
 LOGGER = logging.getLogger(__name__)
 DEFAULT_TIMEOUT = 300
-# Change to something reasonable once all clients implement paging controls
-MAX_PAGE_SIZE = maxsize
+MAX_PAGE_SIZE = 1000
```
peterschwarz: This change would make sense as a separate commit

delventhalz: Moved

peterschwarz:
Does the `@asyncio.coroutine` -> `async` change require a certain level of Python 3? (3.5, I think?). If we're fine with that constraint, we will need to be sure that we're documenting the minimum python requirements. if not, we can't make this change.

delventhalz:
`async` requires Python 3.5. It is already being used in all of the unit tests written for the Rest Api, and in `rest_api.py`. I'm not sure what version of Python we are targeting or have documented as a requirement, but if it is older than 3.5, we'll need to add a task to swap out the other `async` uses.

PR #301, opened 2017-04-05
Move tests

nick-drozd:
Move config and validator registry TP tests to integration so they can be with their other TP test brethren. Also move the various message factories to the same directory.


bin/run_tests
```diff
@@ -187,18 +183,14 @@ test_poet_core(){
...
-}
-
 test_poet_simulator(){
     run_docker_test ./consensus/poet/simulator/tests/test_simulator/unit_simulator.yaml -s poet
 }
 
 test_python_sdk() {
     # transaction processors
     run_docker_test tp-config -s validator
+    run_docker_test tp-validator-registry -s validator
```
peterschwarz: Technically, this is actually part of a separate module (poet)

nick-drozd: There are a variety of criteria by which the tests can be grouped. My thought was that all the TP tests should be grouped together because they use the same test apparatus. It doesn't matter much, and we can scrap this PR if we decide that's not how it should be.

aludvik: The purpose of being able to pick modules is so you can save yourself time locally by only testing things that are affected by what you're working on. So the questions are "If I'm working on the validator registry, do I need to be testing the rest of the SDK?" and "If I'm working on the SDK, do I need to be testing the rest all of the transaction processors?" In any case, all the tests will be run on Jenkins, so it's really just a time saver.

aludvik: The purpose of being able to pick modules is so you can save yourself time locally by only testing things that are affected by what you're working on. So the questions are "If I'm working on the validator registry, do I need to be testing the rest of the SDK?" and "If I'm working on the SDK, do I need to be testing the rest all of the transaction processors?" In any case, all the tests will be run on Jenkins, so it's really just a time saver.

PR #304, opened 2017-04-06
Implement configurable logging

agunde406:
This Pr allows you to customize your logging experience. 

If there is no logging config file in the config directory, it defaults to two files. There will be a validator-{pubkey}-debug.log and validator-{pubkey}-error.log when running the validator and {tp_name}_{zmq_id}-debug.log and {tp_name}_{zmq_id}-error.log when running a transaction processor. Theses files will be in the log directory defined by the path_config (most often the defaults for os or what is set in SAWTOOTH_HOME, look at validator/etc/path.toml.example for clarifaction).

To configure the logging for the validator, put a log config file in the defined config directory and name the file log_config.toml. This file should be a toml file that is simulator to the example located in validator/etc/log_config.toml.example.

To configure the logging for a intkey transaction processor, put a log config file in the defined config directory and name the file intkey_log_config.toml. This file should be a toml file that is simulator to the example located in validator/etc/log_config.toml.example.

Xo transaction processor is similar but the config file should be xo_log_config.toml

I will write up better docs about how to use this next.


validator/etc/log_config.toml.example
```diff
@@ -0,0 +1,29 @@
+version = 1
+disable_existing_loggers = false
+[formatters.simple]
```
jsmitchell: can toml tolerate linefeeds before the section headings (in brackets)?  This is a bit hard to read all crammed together.

agunde406: I will check.

jsmitchell:
Tested locally.

validator/etc/log_config.toml.example
agunde406: Updated to separate out the sections

PR #305, opened 2017-04-06
Initial dynamic topology buildout

jsmitchell:
This change introduces a number of new validator command line arguments, including a required argument `--public-uri` to specify the reachable endpoint that the validator should advertise to other nodes upon connection. This differs from the existing `--network-endpoint`, as `--network-endpoint` describes the interfaces and port binding configuration on the local machine.

This also introduces the optional `--peering` argument, which takes `static` and `dynamic` as arguments. `static` is the default behavior if `--peering` is not supplied. In `static` peering mode, the validator will attempt to peer with endpoints provided in the `--peers` list. Notably, the first validator to start will not attempt to peer with any other nodes in this mode, and will accept peering requests from others. In `dynamic` mode, the validator will honor peering attempts with the `--peers` list and it will connect with listed endpoints provided to `--join`. Upon successful connection, it will request a connected peers list and then attempt to build up to a minimum connectivity threshold via successive peering attempts against random members of the collected list.

For great victory:

`./bin/validator --public-uri tcp://localhost:8800 --peering dynamic -vv`
`./bin/validator --network-endpoint tcp://0.0.0.0:8801 --component-endpoint tcp://0.0.0.0:40001 --peering dynamic --public-uri tcp://localhost:8801 --join tcp://localhost:8800 -vv`
`./bin/validator --network-endpoint tcp://0.0.0.0:8802 --component-endpoint tcp://0.0.0.0:40002 --public-uri tcp://localhost:8802 --join tcp://localhost:8800 --peering dynamic -vv`
`./bin/validator --network-endpoint tcp://0.0.0.0:8803 --component-endpoint tcp://0.0.0.0:40003 --public-uri tcp://localhost:8803 --join tcp://localhost:8800 --peering dynamic -vv`


validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -184,7 +184,8 @@ def _received_from_identity(self, zmq_identity):
         self._last_message_times[zmq_identity] = time.time()
         connection_id = self._identity_to_connection_id(zmq_identity)
         if connection_id not in self._connections:
-            self._connections[connection_id] = ("ZMQ_Identity", zmq_identity)
+            self._connections[connection_id] = \
+                ("ZMQ_Identity", zmq_identity, None)
```
peterschwarz: Any way we could make this a named tuple?  The purpose of the values (especially with that last `None`) as the tuple grows.

validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -533,7 +535,7 @@ def send(self, message_type, data, connection_id, callback=None):
         if connection_id not in self._connections:
             raise ValueError("Unknown connection id: %s",
                              connection_id)
-        connection_type, connection = self._connections.get(connection_id)
+        connection_type, connection, _ = self._connections.get(connection_id)
```
peterschwarz: Again, haven't yet figured out what that third value is.

validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -576,11 +578,47 @@ def stop(self):
...
+        """Returns the connection id associated with a publically
+        reachable endpoint or raises KeyError if the endpoint is not
+        found.
+
+        Args:
+            endpoint (str): A zmq-style uri which identifies a publically
+                reachable endpoint.
+        """
+        for connection_id in self._connections:
+            (_, _, uri) = self._connections[connection_id]
```
peterschwarz: Ah, now I see.

validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -576,11 +578,47 @@ def stop(self):
...
+                         "exist.",
+                         endpoint,
+                         connection_id)
+
+    def _add_connection(self, connection, uri=None):
         connection_id = connection.connection_id
         if connection_id not in self._connections:
             self._connections[connection_id] = \
-                ("OutboundConnection", connection)
+                ("OutboundConnection", connection, uri)
```
peterschwarz: Is this `"OutboundConnection"` effectively an enum?

validator/sawtooth_validator/networking/interconnect.py
jsmitchell: It's a secret.

integration/sawtooth_integration/tests/test_shutdown_smoke.py
```diff
@@ -214,14 +214,19 @@ def _startup(self):
         try:
             genesis = self._run_genesis()
             self._wait_for_containers_exit_status([genesis])
-            validator_genesis = self._start_validator(genesis, [], [])
+            validator_genesis = self._start_validator(
+                genesis,
+                ['--name', 'validator-0'],
```
boydjohnson: Hard coded container names will fail on jenkins during parallel builds. It is really annoying since the pass locally always, and then will fail intermittenly on jenkins. My suggestion, until we have more of a dynamic integration testing framework, is that 'validator-0' be a random string.

integration/sawtooth_integration/docker/poet-smoke.yaml
```diff
@@ -32,7 +32,7 @@ services:
           -o config.batch && \
         ./bin/poet genesis -o poet.batch && \
         ./bin/sawtooth admin genesis config.batch poet.batch && \
-        ./bin/validator -v\""
+        ./bin/validator --public-uri tcp//validator-0:8800 -vv\""
```
boydjohnson: tcp//validator-0:8800 missing a colon.

validator/sawtooth_validator/gossip/gossip_handlers.py
```diff
@@ -23,13 +23,60 @@
...
+    def __init__(self, gossip):
+        self._gossip = gossip
+
+    def handle(self, connection_id, message_content):
+        request = GetPeersRequest()
+        request.ParseFromString(message_content)
+        LOGGER.debug("got peers request message "
+                     "from %s. sending ack", connection_id)
+
+        self._gossip.send_peers(connection_id)
```
boydjohnson: I don't understand this Handler. Within gossip.send_peers there is a network.send( call that is a first message with a response. If this handler returned a GOSSIP_GET_PEERS_RESPONSE, it would be sent directly back to connection_id by the dispatcher. I will leave a comment in gossip.py where a callback can be added to the network.send function to do make the GetPeersResponseHandler not necessary.

validator/sawtooth_validator/gossip/gossip.py
```diff
@@ -137,49 +240,278 @@ def broadcast(self, gossip_message, message_type, exclude=None):
...
+                self._gossip.unregister_peer(conn_id)
+
+    def _get_peers_of_peers(self, peers):
+        get_peers_request = GetPeersRequest()
+
+        for conn_id in peers:
+            self._network.send(
+                validator_pb2.Message.GOSSIP_GET_PEERS_REQUEST,
+                get_peers_request.SerializeToString(),
+                conn_id)
```
boydjohnson: If a callback was added to network.send using (with a slight modification) gossip.add_candidate_peer_endpoints( the same functionality would happen.

validator/sawtooth_validator/gossip/gossip_handlers.py
jsmitchell: I think we are sending a NETWORK_ACK as a response in every other case (on the network). I'd like to maintain that pattern for now.

validator/sawtooth_validator/gossip/gossip.py
jsmitchell: Per above, I'd like to maintain the pattern of msg->ack(ok | error), with separate dispatch action on the response types for now. I think this may support other uses that we haven't thought of yet.

PR #308, opened 2017-04-07
Enforce and abide by PoET key block claim limit

jjason:
Introduces the following changes:

* PoET consensus state statistics are now reset whenever a non-PoET block is encountered in the chain and the sealed signup data is removed from the consensus state as that needs to be maintained on a per-PoET public key basis.
* A utility class to manage and persist locally on a validator key state for its PoET key.  This includes the sealed signup data associated with it as well as whether or not it has been refreshed because the block claim limit has been reached with that key.
* The PoET genesis CLI and block publisher have been changed to use the PoET key state store for sealed signup data.
* A utility class to query PoET configuration settings.
* The PoET block verifier now enforces the key block claim limit when determining the validity of a block.
* The PoET block publisher abides by the key block claim limit when being asked to initialize a block.  If the limit has been reached with its current key, it will fail to initialize the block and if appropriate will create and register new signup information.

To test locally:

`-k /etc/sawtooth/keys/validator.wif \`

Add:

```
sawtooth.consensus.algorithm=poet \
sawtooth.poet.key_block_claim_limit=2 \
```

This should result in a bunch of blocks and validators having to refresh after successfully claiming 2 blocks.

Execute:

### DIFF
--- a/consensus/poet/cli/sawtooth_poet_cli/genesis.py
+++ b/consensus/poet/cli/sawtooth_poet_cli/genesis.py
@@ -20,10 +20,11 @@ import time
 
 from sawtooth_poet_cli import config
 from sawtooth_poet_cli.exceptions import CliException
-from sawtooth_poet.poet_consensus.consensus_state import ConsensusState
-from sawtooth_poet.poet_consensus.consensus_state_store \
-    import ConsensusStateStore
 from sawtooth_poet.poet_consensus.signup_info import SignupInfo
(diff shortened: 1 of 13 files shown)


jjason:
Ooops.  Inadvertently screwed something up when I rebased before pushing PR.  Let's try that again.

jsmitchell:
Tested locally.

jjason:
Rebasing before merge.

PR #310, opened 2017-04-07
Add transaction fetching to Rest Api, ZMQ Api, and CLI

delventhalz:
- Add `get_transaction` method to block store and refactors some error handling
- Add transaction client handlers to validator with unit tests
- Add `/transactions` route to Rest Api, with unit tests
- Refactor similar CLI commands to use new `format_utils` module with common formatting logic
- Add `transaction list` and `transaction show` commands to CLI
- Correct some typos

### DIFF
--- a/validator/sawtooth_validator/state/client_handlers.py
+++ b/validator/sawtooth_validator/state/client_handlers.py
@@ -601,3 +601,54 @@ class BatchGetRequest(_ClientRequestHandler):
             LOGGER.debug(e)
             return self._status.NO_RESOURCE
         return self._wrap_response(batch=batch)
+
+
+class TransactionListRequest(_ClientRequestHandler):
+    def __init__(self, block_store):
+        super().__init__(
+            client_pb2.ClientTransactionListRequest,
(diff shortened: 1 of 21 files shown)


validator/sawtooth_validator/state/client_handlers.py
```diff
@@ -601,3 +601,54 @@ def _respond(self, request):
...
+            request.transaction_ids,
+            self._block_store.get_transaction,
+            lambda block: [t for a in block.batches for t in a.transactions])
+
+        transactions, paging = _Pager.paginate_resources(
+            request,
+            transactions,
+            self._status.INVALID_PAGING)
+
+        if not transactions:
```
peterschwarz: Is this treating Empty as a no resource, as opposed to data of length 0?

delventhalz: Because of the way Protobufs work there is no way to tell the difference between an empty list of data and data that wasn't set at all. For that reason, when there is no data the List handlers explicitly return a NO_RESOURCE status.

PR #318, opened 2017-04-10
Fix CLI failing unexpectedly when getting a string from Rest Api

delventhalz:


cli/sawtooth_cli/rest_client.py
```diff
@@ -82,11 +82,12 @@ def _get(self, path, **queries):
             self._base_url + path + self._format_queries(queries))
 
         # concat any additional pages of data
-        while 'next' in json_result.get('paging', {}):
-            previous_data = json_result.get('data', [])
-            code, json_result = self._submit_request(
-                json_result['paging']['next'])
-            json_result['data'] = previous_data + json_result.get('data', [])
+        if isinstance(json_result, dict):
```
jsmitchell: isinstance is an anti-pattern. What are the situations where a non expected type is being returned?

delventhalz: Errors from from the Rest Api are just simple strings at the moment.

What is the desired pattern when checking types? Or do you just want a try/except?

aludvik: It looks like `isinstance()` is being used a lot in the REST API right now. If it is something that shouldn't be used, then maybe removing it and finding a better pattern should be a different story?

peterschwarz: Not part of this PR, but it makes me wonder if it makes sense that the REST api should _always_ return JSON objects.  Then, when parsed, the consumer simple checks for the field "error(s)".

jsmitchell: Ok, I agree with both @aludvik and @peterschwarz. @delventhalz can you open a story to address?

delventhalz: Yes. It is my error I'm working on this sprint.

delventhalz: I'll do a status check instead of `isinstance` in this case. Probably makes more sense anyway.

PR #320, opened 2017-04-10
Move/Rename consensus/poet unit test YAML files to make consistent.

jjason:
This update makes the consensus/poet directory consistent with regard to unit test YAML files.  Specifically:
* Moves `unit_simulator.yaml` to its parent directory
* Moves `unit_consensus.yaml` to its parent directory and renames it to `unit_core.yaml`
* Removes `unit_sgx_structs.yaml` and `unit_validator_registry_view.yaml` and replaces with `unit_common.yaml` in its parent directory
* Fixes YAML files so that volume mount point for `/project/sawtooth-core` is correct
* Fixes `bin/run_test` to account for new location of YAML files


aludvik:
Please see PR 312, it makes some changes to these files which should happen before this PR.

jjason:
Rebased after PR 312 merged.

PR #322, opened 2017-04-11
Enforce and abide by PoET block claim delay

jjason:
Per the PoET specification, whenever a validator adds or updates signup information to the validator registry, it is not allowed to claim a block until some number of blocks after the block in which its validator registry information was committed.  This PR:

* Updates the PoET configuration view to contain the configurable block claim delay value.
* Updates the consensus state to keep track of the total number of blocks claimed.  Note that it is not sufficient to simply rely upon a block's number as PoET consensus state statistics, etc., are reset whenever a non-PoET block is encountered.
* Updates the PoET utils to accurately keep track of the total block claim count.
* Updates the per-validator state to keep track of the block number in which a validator's current PoET public key committed to the blockchain.
* Refactors the PoET utils to make validator state and retrieval useful for enforcing both this policy as well as the other PoET policies.
* Updates the PoET block verifier to apply the block claim delay to fail verification of blocks which are attempted to be claimed before the block claim delay has been satisfied.
* Updates the PoET block publisher to apply the block claim delay to prevent initializing a block if it would be rejected by other validator's PoET block verifiers because the block was claimed before the block claim delay has been satisfied..

To test locally:

`-k /etc/sawtooth/keys/validator.wif \`

Add:

```
sawtooth.consensus.algorithm=poet \
sawtooth.poet.key_block_claim_limit=2 \
```

This should result in a bunch of blocks and validators having to refresh after successfully claiming 2 blocks.

Execute:

### DIFF
--- a/consensus/poet/core/sawtooth_poet/poet_consensus/poet_block_verifier.py
+++ b/consensus/poet/core/sawtooth_poet/poet_consensus/poet_block_verifier.py
@@ -184,6 +184,65 @@ class PoetBlockVerifier(BlockVerifierInterface):
                             validator_state.key_block_claim_count,
                             poet_config_view.key_block_claim_limit))
 
+            # While having a block claim delay is nice, it turns out that in
+            # practice the claim delay should not be more than one less than
+            # the number of validators.  It helps to imagine the scenario
+            # where each validator hits their block claim limit in sequential
+            # blocks and their new validator registry information is updated
+            # in the following block by another validator, assuming that there
+            # were no forks.  If there are N validators, once all N validators
+            # have updated their validator registry information, there will
+            # have been N-1 block commits and the Nth validator will only be
+            # able to get its updated validator registry information updated
+            # if the first validator that kicked this off is now able to claim
+            # a block.  If the block claim delay was greater than or equal to
+            # the number of validators, at this point no validators would be
+            # able to claim a block.
+            number_of_validators = \
+                len(validator_registry_view.get_validators())
+            block_claim_delay = \
+                min(
+                    poet_config_view.block_claim_delay,
+                    number_of_validators - 1)
+
+            # While a validator network is starting up, we need to be careful
+            # about applying the block claim delay because if we are too
+            # aggressive we will get ourselves into a situation where the
+            # block claim delay will prevent any validators from claiming
+            # blocks.  So, until we get at least block_claim_delay blocks
+            # we are going to choose not to enforce the delay.
+            if consensus_state.total_block_claim_count <= block_claim_delay:
+                LOGGER.debug(
+                    'Skipping block claim delay check.  Only %d block(s) in '
+                    'the chain.  Claim delay is %d block(s). %d validator(s) '
+                    'registered.',
+                    consensus_state.total_block_claim_count,
+                    block_claim_delay,
+                    number_of_validators)
+                return True
+
+            blocks_since_registration = \
+                block_wrapper.block_num - \
+                validator_state.commit_block_number - 1
+
+            if block_claim_delay > blocks_since_registration:
+                raise \
+                    ValueError(
+                        'Validator {} claiming too early. Block: {}, '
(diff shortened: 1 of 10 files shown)


consensus/poet/core/sawtooth_poet/poet_consensus/poet_block_verifier.py
```diff
@@ -183,6 +184,65 @@ def verify_block(self, block_wrapper):
                             validator_state.key_block_claim_count,
                             poet_config_view.key_block_claim_limit))
 
+            # While having a block claim delay is nice, it turns out that in
```
jsmitchell: Maybe include this comment in poet_block_publisher too.

jjason: Can do.  My plan was that because these three PRs are intertwined, that in the third PR I was going to refactor the PoET policy tests to be in one place and the PoetBlockVerifier and PoetBlockPublisher would call them.

consensus/poet/core/sawtooth_poet/poet_consensus/utils.py
```diff
@@ -142,54 +142,95 @@ def build_certificate_list(block_header,
...
+                                consensus_state,
+                                block_cache):
+    """Fetches the current validator state for the validator, creating it if
+    necessary.
+
+    Args:
+        validator_info (ValidatorInfo): The validator information
+            corresponding to the consensus state
+        consensus_state (ConsensusState): The consensus state from which
+            validator state should be fecthed/derived.
```
peterschwarz: fecthed-> fetched

consensus/poet/core/sawtooth_poet/poet_consensus/utils.py
```diff
@@ -142,54 +142,95 @@ def build_certificate_list(block_header,
...
+            ValidatorState(
+                commit_block_number=enclosing_block.block_num,
+                key_block_claim_count=0,
+                poet_public_key=validator_info.signup_info.poet_public_key,
+                total_block_claim_count=0)
+
+    return validator_state
+
+
+def create_updated_validator_state(validator_info,
```
peterschwarz: simplify this to `update_validator_state` or `next_validator_state`: it looks like `current_validator_state` can no longer be `None` based on the doc string, so this name is more straightforward.  It's a just a non-destructive update.

consensus/poet/core/sawtooth_poet/poet_consensus/poet_block_verifier.py
```diff
@@ -184,6 +184,65 @@ def verify_block(self, block_wrapper):
...
+                    block_claim_delay,
+                    number_of_validators)
+                return True
+
+            blocks_since_registration = \
+                block_wrapper.block_num - \
+                validator_state.commit_block_number - 1
+
+            if block_claim_delay > blocks_since_registration:
+                raise \
```
peterschwarz: Any reason not to log this and return `False`?

consensus/poet/core/sawtooth_poet/poet_consensus/utils.py
jjason: I am okay with `next_validator_state` but reject `update_validator_state` as the latter implies updating in place. The reason the function name was prefixed with `create_` was to be explicit that it was non-destructive to existing/current validator state.  Would `create_next_validator_state` be acceptable?

peterschwarz: `create_next_...` is better

consensus/poet/core/sawtooth_poet/poet_consensus/poet_block_verifier.py
jjason: To be consistent with the previous PoET policy test that also raises ValueError on rejecting block.

jjason:
Updated based upon comments from @peterschwarz and @jsmitchell

PR #326, opened 2017-04-12
Remove 0.0.0.0

aludvik:


rest_api/sawtooth_rest_api/rest_api.py
```diff
@@ -29,10 +29,10 @@ def parse_args(args):
                         default=8080)
     parser.add_argument('--host',
                         help='The host for the api to run on',
-                        default="0.0.0.0")
+                        default="127.0.0.1")
     parser.add_argument('--stream-url',
                         help='The url to connect to a running Validator',
-                        default='tcp://localhost:40000')
+                        default='tcp://127.0.0.1:40000')
```
jsmitchell: This change is not required. While the bind() needs an interface definition, the connect() can take a dns name. I think the dns name is clearer.

aludvik: Hmm, I was trying to standardize on 127.0.0.1, rather than use localhost in some places and 127.0.0.1 in others. It isn't clear to me why localhost is clearer.

jsmitchell: It's clearer (to me) because in --stream-url, you are supplying something the process is connecting _to_, whereas in the rest of the arguments, you are supplying something the process is binding to/listening on.

aludvik:
Reverted `localhost` => `127.0.0.1` changes

aludvik:
Swapped the order of the commits.

PR #327, opened 2017-04-12
Change --peers and --join from argparse nargs='+' to append style arguments

jsmitchell:
This change allows multiple --peers and --join arguments and allows comma-delimited (but not space delimited) params provided to a single instance of the argument.


validator/sawtooth_validator/server/cli.py
```diff
@@ -118,10 +128,29 @@ def check_directory(path, human_readable_name):
     return errors
 
 
+def _split_comma_append_args(arg_list):
+    new_arg_list = []
+    for arg in arg_list:
+        if ',' in arg:
```
peterschwarz: split on a string with no commas will just return an array of the input string: you don't need to check for this.

PR #329, opened 2017-04-12
Cluster command fixed for public-uri parameter

ojalatodd:
Added --public-uri argument to cluster command.
Fixed for both subprocess and docker.

### DIFF
--- a/manage/sawtooth_manage/docker.py
+++ b/manage/sawtooth_manage/docker.py
@@ -135,14 +135,14 @@ class DockerNodeController(NodeController):
         if node_config.genesis:
             command = 'bash -c "sawtooth admin keygen && \
             sawtooth admin genesis && \
-            validator {} -v"'
+            validator {} -v --public-uri tcp://{}:8800"'
         else:
             command = 'bash -c "sawtooth admin keygen && \
-            validator {} -v"'
+            validator {} -v --public-uri tcp://{}:8800"'
         if len(peers) > 0:
-            command = command.format('--peers ' + ",".join(peers))
+            command = command.format('--peers ' + ",".join(peers), node_name)
         else:
-            command = command.format('')
+            command = command.format('', node_name)
 
         compose_dict = {
             'version': '2',
--- a/manage/sawtooth_manage/subproc.py
+++ b/manage/sawtooth_manage/subproc.py
@@ -103,14 +103,17 @@ class SubprocessNodeController(NodeController):
             if cmd == 'validator':
                 component = '--component-endpoint', url
                 network = '--network-endpoint', gossip_port
-                peer_list = ['tcp://0.0.0.0:' + str(base_gossip_port + i)
+                public_uri = '--public-uri', 'tcp://localhost:{}'.\
+                    format(gossip_port_num)
+                peer_list = ['tcp://localhost:' + str(base_gossip_port + i)
                              for i in range(node_num)]
                 peers = ['--peers']
-                peers.extend(peer_list)
+                peer_list_comma_sep = [",".join(peer_list)]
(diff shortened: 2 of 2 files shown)


manage/sawtooth_manage/docker.py
```diff
@@ -135,10 +135,10 @@ def start(self, node_config):
         if node_config.genesis:
             command = 'bash -c "sawtooth admin keygen && \
             sawtooth admin genesis && \
-            validator {} -v"'
+            validator {} -v --public-uri tcp://0.0.0.0:8800"'
```
jsmitchell: 0.0.0.0 is not valid for --public-uri. The intent is this is a reachable endpoint - it should be a hostname (preferably) or an ip address for how this docker container is reachable by other peers on the cluster.

ojalatodd: @jsmitchell Fixed, I believe.

ojalatodd: In subprocess, all the validators are running locally on the same machine, so subprocess should be OK, but take a look.

manage/sawtooth_manage/subproc.py
```diff
@@ -103,13 +103,14 @@ def start(self, node_args):
             if cmd == 'validator':
                 component = '--component-endpoint', url
                 network = '--network-endpoint', gossip_port
+                public_uri = '--public-uri', gossip_port
                 peer_list = ['tcp://0.0.0.0:' + str(base_gossip_port + i)
```
jsmitchell: These are not valid peer addresses.

manage/sawtooth_manage/subproc.py
```diff
@@ -103,13 +103,14 @@ def start(self, node_args):
             if cmd == 'validator':
                 component = '--component-endpoint', url
                 network = '--network-endpoint', gossip_port
+                public_uri = '--public-uri', gossip_port
```
jsmitchell: gossip_port, above is not a valid public-uri reachable endpoint

jsmitchell: If the purpose of this is a number of processes running on the same machine, then --public-uri should probably be something like "tcp://localhost:{}".format(gossip_port_num)

ojalatodd: Changed using your idea.

manage/sawtooth_manage/subproc.py
ojalatodd: Noted. Subprocess was originally coded with 0.0.0.0 as the network address, but I have changed it to "localhost," after consultation with Boyd and others.

ojalatodd:
Fixed merge conflict.

ojalatodd:
@jsmitchell Conflict resolved, changes made.

manage/sawtooth_manage/docker.py
```diff
@@ -135,14 +135,14 @@ def start(self, node_config):
...
             sawtooth admin genesis && \
-            validator {} -v"'
+            validator {} -v --public-uri tcp://{}:8800"'
         else:
             command = 'bash -c "sawtooth admin keygen && \
-            validator {} -v"'
+            validator {} -v --public-uri tcp://{}:8800"'
         if len(peers) > 0:
-            command = command.format('--peers ' + ",".join(peers))
+            command = command.format('--peers ' + " ".join(peers), node_name)
```
jsmitchell: This needs to be a ",".join based on the new peers behavior

ojalatodd: Fixed this for both docker and subprocess.

ojalatodd:
Note: tests pass locally, but not on the build server.

PR #331, opened 2017-04-13
Drop genesis block if not block-chain-id

peterschwarz:
Enforce the value of the block chain id (found in the data dir) by
dropping a received genesis block if it does not match.


boydjohnson:
I don't get the three cases in _set_genesis in this PR. Case 1, if the validator doesn't have the block-chain-id in it's data directory it creates it. Case 2, it does have it but it isn't the same as the genesis block it received. Case 3, It does have it and it is equal to the genesis block it received, so it validates the block. For the Poet Smoke test and any situation where a validator is joining a network where it doesn't already know about the genesis block it would have to receive the genesis block twice for that block to get into the block store.

peterschwarz:
Updated to address this comment: 

1. drop block if there is a block id mismatch
2. verify block:
    a. If valid, store block id
    b. if not, drop block

PR #332, opened 2017-04-13
Replace the use of Stream with the REST API in the intkey client_cli 

agunde406:
Will now need to start up a rest_api to use intkey load and intkey workload

### DIFF
--- a/rest_api/sawtooth_rest_api/exceptions.py
+++ b/rest_api/sawtooth_rest_api/exceptions.py
@@ -56,3 +56,9 @@ class ValidatorUnavailable(web.HTTPServiceUnavailable):
     def __init__(self):
         message = 'Could not reach validator, validator timed out'
         super().__init__(reason=message)
+
+
+class ValidatorDisconnect(web.HTTPServiceUnavailable):
+    def __init__(self):
+        message = 'The connection to the validator was lost'
+        super().__init__(reason=message)
--- a/sdk/examples/intkey_jvm_sc/sawtooth_intkey/cli/load.py
+++ b/sdk/examples/intkey_jvm_sc/sawtooth_intkey/cli/load.py
@@ -20,12 +20,41 @@ import time
 import argparse
 import cbor
 import sawtooth_sdk.protobuf.batch_pb2 as batch_pb2
-from sawtooth_sdk.protobuf.validator_pb2 import Message
-from sawtooth_sdk.messaging.stream import Stream
+import json
+import concurrent.futures
+from concurrent.futures import wait
+import urllib.request as urllib
+from urllib.error import URLError, HTTPError
+from http.client import RemoteDisconnected
 
 sys.path.insert(0, os.path.join(
     os.path.dirname(os.path.dirname(os.path.dirname(
         os.path.realpath(__file__)))), 'python'))
+
+
+def post_batches(url, batches):
+    data = batches.SerializeToString()
+    headers = {'Content-Type': 'application/octet-stream'}
+    headers['Content-Length'] = str(len(data))
+
+    request = urllib.Request(
+        url + "/batches",
+        data,
+        headers=headers,
+        method='POST')
+    try:
+        result = urllib.urlopen(request)
+        code, json_result = (result.status, json.loads(result.read().decode()))
+        if not (code == 200 or code == 201 or code == 202):
+            LOGGER.warning("(%s): %s", code, json_result)
+        return(code, json_result)
+    except HTTPError as e:
+        LOGGER.warning("(%s): %s", e.code, e.msg)
+        return(e.code, e.msg)
+    except RemoteDisconnected as e:
+        LOGGER.warning(e)
+    except URLError as e:
+        LOGGER.warning(e)
 
 
 def _split_batch_list(batch_list):
@@ -41,30 +70,25 @@ def _split_batch_list(batch_list):
 
 def do_load(args):
     print("Do load")
-    with open(args.filename, 'rb') as fd:
+    with open(args.filename, mode='rb') as fd:
         batches = batch_pb2.BatchList()
         batches.ParseFromString(fd.read())
 
-    stream = Stream(args.url)
-    futures = []
     start = time.time()
-
+    futures = []
+    executor = concurrent.futures.ThreadPoolExecutor(max_workers=5)
     for batch_list in _split_batch_list(batches):
-        future = stream.send(
-            message_type=Message.CLIENT_BATCH_SUBMIT_REQUEST,
-            content=batch_list.SerializeToString())
-        futures.append(future)
+        fut = executor.submit(post_batches, args.url, batch_list)
+        futures.append(fut)
 
-    for future in futures:
-        result = future.result()
-        assert(result.message_type == Message.CLIENT_BATCH_SUBMIT_RESPONSE)
+    # Wait until all futures are complete
+    wait(futures)
 
     stop = time.time()
+
     print("batches: {} batch/sec: {}".format(
         str(len(batches.batches)),
         len(batches.batches) / (stop - start)))
-
-    stream.close()
 
 
 def add_load_parser(subparsers, parent_parser):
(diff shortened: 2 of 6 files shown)


sdk/python/sawtooth_sdk/client/sdk_client.py
```diff
@@ -0,0 +1,68 @@
...
+        return (result.status, json.loads(result.read().decode()))
+    except HTTPError as e:
+        return (e.code, e.msg)
+    except RemoteDisconnected as e:
+        raise CliException(e)
+    except URLError as e:
+        raise CliException(e)
+
+
+def post_batches(url, batches):
```
vaporos: This throws away the error code (from a programatic perspective), so any code using post_batches is limited to success/failure.  The error message may or may not be formatted in a way that is good for users of the SDK.  These seems like a specific convenience wrapper that should live in app code (intkey, whatever) and not in the SDK.

The SDK is not CLI-specific, so CliException isn't appropriate.  This seems more like a BatchSubmissionException.

If we are going to abstract the use of the REST API at all, seems like we should do it in a way that provides the ability for a client to select between rest api and interconnect while keeping the same API.  That might mean rethinking the API of stream.

rest_api/sawtooth_rest_api/exceptions.py
```diff
@@ -56,3 +56,9 @@ class ValidatorUnavailable(web.HTTPServiceUnavailable):
     def __init__(self):
         message = 'Could not reach validator, validator timed out'
         super().__init__(reason=message)
+
+
+class ValidatorDisconnect(web.HTTPServiceUnavailable):
```
vaporos: I'm not seeing anything that catches this.  What happens when it is thrown?

sdk/examples/intkey_jvm_sc/sawtooth_intkey/cli/load.py
```diff
@@ -41,30 +42,20 @@ def _split_batch_list(batch_list):
...
     for batch_list in _split_batch_list(batches):
-        future = stream.send(
-            message_type=Message.CLIENT_BATCH_SUBMIT_REQUEST,
-            content=batch_list.SerializeToString())
-        futures.append(future)
-
-    for future in futures:
-        result = future.result()
-        assert(result.message_type == Message.CLIENT_BATCH_SUBMIT_RESPONSE)
+        post_batches(args.url, batch_list)
```
vaporos: Futures is a better API, because the SDK client might want to do other things while batches are being posted, like printing out status.

This also requires post_batches() to finish before going on to the next one, which worse than the code it replaces which allowed multiple batch lists to be sent before the previous had completely finished.

sdk/python/sawtooth_sdk/client/sdk_client.py
```diff
@@ -0,0 +1,68 @@
+# Copyright 2017 Intel Corporation
```
vaporos: The name of the module 'sdk_client' duplicates the namespace.  If we left it as-is, something like 'rest_api_wrapper_utils' would be an appropriate name.

rest_api/sawtooth_rest_api/exceptions.py
delventhalz: Aiohttp catches it, and sends an error response back to the client.

sdk/examples/intkey_jvm_sc/sawtooth_intkey/cli/load.py
```diff
@@ -41,31 +68,26 @@ def _split_batch_list(batch_list):
...
+    executor = concurrent.futures.ThreadPoolExecutor(max_workers=5)
     for batch_list in _split_batch_list(batches):
-        future = stream.send(
-            message_type=Message.CLIENT_BATCH_SUBMIT_REQUEST,
-            content=batch_list.SerializeToString())
-        futures.append(future)
+        fut = executor.submit(post_batches, args.url, batch_list)
+        futures.append(fut)
 
     for future in futures:
```
boydjohnson: I wonder if we can do something better than the prior implementation by using concurrent.futures.as_completed. If it was ```for future in as_completed(futures):``` if there was wide variation in the number of txns per batch or the complexity of those txns, the total time it would take to load would be less.

agunde406: Updated to use concurrent.futures.wait instead. This will return when all futures are completed.

sdk/examples/intkey_jvm_sc/sawtooth_intkey/cli/load.py
```diff
@@ -20,14 +20,42 @@
...
 
 sys.path.insert(0, os.path.join(
     os.path.dirname(os.path.dirname(os.path.dirname(
         os.path.realpath(__file__)))), 'python'))
 
 
+def post_batches(url, batches):
+    data = batches.SerializeToString()
+    headers = {'Content-Type': 'application/octet-stream'}
+    headers['Content-Length'] = '%d' % len(data)
```
peterschwarz: This could just be `str(len(data))`

sdk/examples/intkey_python/sawtooth_intkey/client_cli/load.py
```diff
@@ -16,17 +16,42 @@
...
-from sawtooth_sdk.protobuf.validator_pb2 import Message
-
 
 LOGGER = logging.getLogger(__file__)
 
 
+def post_batches(url, batches):
+    data = batches.SerializeToString()
+    headers = {'Content-Type': 'application/octet-stream'}
+    headers['Content-Length'] = '%d' % len(data)
```
peterschwarz: This could be `str(len(data))`

sdk/examples/intkey_python/sawtooth_intkey/client_cli/workload.py
```diff
@@ -16,23 +16,48 @@
...
 LOGGER = logging.getLogger(__name__)
 
-IntKeyState = namedtuple('IntKeyState', ['name', 'stream', 'value'])
+IntKeyState = namedtuple('IntKeyState', ['name', 'url', 'value'])
+
+
+def post_batches(url, batches):
+    data = batches.SerializeToString()
+    headers = {'Content-Type': 'application/octet-stream'}
+    headers['Content-Length'] = '%d' % len(data)
```
peterschwarz: Same `str(len(data))`

sdk/python/sawtooth_sdk/workload/workload_generator.py
```diff
@@ -73,64 +73,69 @@ def run(self):
 
     @asyncio.coroutine
     def _simulator_loop(self):
+        while True:
+            yield from asyncio.sleep(self._rate)
+            with self._lock:
```
peterschwarz: Looks like we're mixing async io event loops and threading locks - this will block the whole event loop, as we saw in the rest api case (#303).  This should probably be an `asyncio.Lock`

agunde406: The workload generator uses 1 coroutine and threads. As such asyncio.Lock should not be used in threads.

agunde406:
Rebased on master after SDK reorganization

sdk/python/sawtooth_sdk/workload/workload_generator.py
```diff
@@ -193,30 +194,43 @@ def on_new_batch(self, batch_id, stream):
...
 
-        except ValidatorConnectionError:
+            if code == 200 or code == 201 or code == 202:
+                status = list(json_result["data"].values())[0]
+                return status
+            else:
+                LOGGER.debug("(%s): %s", code, json_result)
+                return "UNKNOWN"
+        except HTTPError as e:
+            if "The connection to the validator was lost" in str(e.msg):
```
delventhalz: My error PR will simplify the error checking you did here (and break it). Since your PR is basically ready to merge, let's make sure it gets in first, and then I'll add the necessary refactor to mine.

PR #333, opened 2017-04-13
Enforce and abide by PoET block claim frequency (zTest) restrictions

jjason:
Per the PoET specification, validators are limited in how frequently they may claim blocks - i.e., the zTest.  If it is determined that allowing a validator to win an election (i.e., claim a block and have it added to the blockchain) would cause it to claim blocks more frequently than a configurable statistical rate, the block is rejected.

This PR updates:
* adds zTest block counts to both the consensus and validator state objects.  The zTest block count is the number of blocks claimed by all validators and a specific validator, respectively, that are covered by the zTest.  Additionally, code is added to properly maintain these values.
* updates the PoET block verifier to ensure that when verifying blocks it checks to make sure a validator is throttled by the block claim frequency allowed 
* updates the PoET block publisher to refuse to initialize a block if the block frequency check fails, indicating that other validators would reject the block
* refactors all PoET policy tests (key block claim limit, block claim delay, and block claim frequency)
* adds a short-circuit for blocks upon which block publisher has already determined it would not be allowed to build on top of

To test locally:

Add:

```
sawtooth.poet.block_claim_delay=1 \
sawtooth.poet.fixed_duration_block_count=5 \
sawtooth.poet.ztest_maximum_win_deviation=1.645 \
sawtooth.poet.ztest_minimum_win_count=2 \
```

This should increase the chance of validators refusing the publish blocks because it would violate the block claim frequency test.  You can look at the logs for error log messages.

To reduce the likelihood of zTest failures, you can change `sawtooth.poet.ztest_maximum_win_deviation`.  The values are based upon the confidence interval (i.e., how confident we are that we have truly detected a validator winning at a frequency we consider too frequent):

```
3.075 => 99.9%
2.575 => 99.5%
2.321 => 99%
1.645 => 95%
```

Execute:

### DIFF
--- a/consensus/poet/core/tests/test_consensus/test_poet_config_view.py
+++ b/consensus/poet/core/tests/test_consensus/test_poet_config_view.py
@@ -36,7 +36,7 @@ class TestPoetConfigView(unittest.TestCase):
         # pylint: disable=protected-access
         self.assertEqual(
             poet_config_view.key_block_claim_limit,
-            PoetConfigView._DEFAULT_KEY_CLAIM_LIMIT_)
+            PoetConfigView._KEY_BLOCK_CLAIM_LIMIT_)
 
         _, kwargs = \
             mock_config_view.return_value.get_setting.call_args
@@ -45,7 +45,7 @@ class TestPoetConfigView(unittest.TestCase):
         # pylint: disable=protected-access
         self.assertEqual(
             kwargs['default_value'],
-            PoetConfigView._DEFAULT_KEY_CLAIM_LIMIT_)
+            PoetConfigView._KEY_BLOCK_CLAIM_LIMIT_)
         self.assertEqual(kwargs['value_type'], int)
 
         # Underlying config setting is not a valid value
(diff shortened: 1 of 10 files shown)


consensus/poet/core/tests/test_consensus/test_poet_config_view.py
```diff
@@ -36,7 +36,7 @@ def test_key_block_claim_limit(self, mock_config_view):
         # pylint: disable=protected-access
         self.assertEqual(
             poet_config_view.key_block_claim_limit,
-            PoetConfigView._DEFAULT_KEY_CLAIM_LIMIT_)
+            PoetConfigView._KEY_BLOCK_CLAIM_LIMIT_)
```
peterschwarz: Not sure that I like this private access of these values.  If someone changes these constants, nothing breaks, and there's a lower barrier to think about why the values are set the way they are.  

In terms of code coverage, since you're testing the defaulting of the values, you're really not providing much other than that you've linked the correct constant.

PR #334, opened 2017-04-13
Remove legacy core/cli and references

dcmiddle:


bin/clean_all
```diff
@@ -20,7 +20,7 @@ top_dir=$(cd $(dirname $(dirname $0)) && pwd)
 set -x
 set -e
 
-for pkg in signing cli sdk sdk/python validator docs core extensions/arcade core_transactions rest_api \
+for pkg in signing cli sdk sdk/python validator docs extensions/arcade core_transactions rest_api \
```
vaporos: It would be nice to have clean_all clean up 0.7 stuff as well when switching between 0.7 and 0.8.

core/sawtooth/cli/admin_sub/clean.py
```diff
@@ -1,157 +0,0 @@
-# Copyright 2016 Intel Corporation
```
vaporos: We should leave clean's implementation until it's moved/implemented for 0.8 CLI.

core/sawtooth/cli/monitor.py
```diff
@@ -1,261 +0,0 @@
-# Copyright 2016 Intel Corporation
```
vaporos: monitor and stats should be kept if we intend to implement a substantially similar CLI commands.

dcmiddle: What if we just look in 0.7 for this code? That's more preferable to me than having dead code in the repo.

vaporos: It breaks 'git blame' and all such history (to remove then copy back in and commit as a new author from scratch), which is why it didn't get purged previously.

I wonder how much effort it would be to simplify the stats code to what's currently supported (which doesn't include most of the metrics) only.  The fork detect code, for example, should still work with what we currently get out of the REST API since it's using block id lists.

dcmiddle:
Close per review comment by @vaporos about still making use of bringing forward legacy items with intact history.

PR #337, opened 2017-04-14
Update docker install to use new procedure.

aludvik:


docker/sawtooth-dev-test
```diff
@@ -47,6 +51,7 @@ RUN echo "deb http://repo.sawtooth.me/ xenial universe" >> /etc/apt/sources.list
     python3-cryptography=1.7.2-1 \
     python3-lmdb=0.92-1 \
     python3-multidict=2.1.4-1 \
+    python3-nose2 \
```
peterschwarz: This should probably be locked to a version

PR #338, opened 2017-04-14
Peg remaining installs in Dockerfiles

aludvik:


aludvik:
Rebased on docker install fixes so the PR will pass.

aludvik:
Pegged packages added in new docker install...

PR #341, opened 2017-04-14
Add client modules to the JS SDK

delventhalz:
- Rename all existing `client` SDK modules (in Python, Java, _and_ JavaScript), to `interconnect` to make room for future external client focused modules.
- Add `signer` module to JS SDK with generates private/public keys and signs/verifies data
- Add `TransactionEncoder` module which packages data into validator ready Protobufs
- Include JSDoc formatted docstrings for the new modules
- Add unit tests for the new modules
- Generated `yarn.lock` file updated (350 additions)


sdk/javascript/client/encoder.js
```diff
@@ -0,0 +1,185 @@
...
+}
+
+const _makeNonce = () => {
+  const dateString = Date.now().toString(36).slice(-5)
+  const randomString = Math.floor(Math.random() * 46655).toString(36)
+  return dateString + ('00' + randomString).slice(-3)
+}
+
+/** Defines encoders which create multiple Transactions with common settings */
+class TransactionEncoder {
```
peterschwarz: Given that it deals with Transactions, Batches and Batch Lists, this name is too specific, making the batch and list methods a bit odd.  Either separate them into their own class, or give this a more generic name.

delventhalz: I think I would push back on this actually. The focus of the module is on creating a Transaction, and then wrapping it in whatever packaging is necessary to get Sawtooth to accept it (or to satisfy similar use cases, like creating a Transaction to be batched elsewhere). The focus is not on creating Batches, or BatchLists, those are incidental, only necessary to package Transactions. Every method with the exception of `list` (which is my worst named, and if anything, that should probably change), is named in relation to Transactions: "create _(a transaction)_," "encode _(a transaction)_," "batch _(transactions)."

I don't know what use cases there would be for doing operations on Batches and BatchLists independent of Transactions, but I don't think this module would satisfy them. I also don't think this functionality should be broken up, as there is none of the other methods make sense without the core `create` (transaction) method.

delventhalz:
I'm happy to bend on the `/interconnect` move. My impression was though, that we considered Stream/Future to be implementation details that are primarily, if not entirely, internal to the validator. If we think there is a case for clients to be using them, then I can leave them where they are and put `signer.js` and `encoder.js` in there with them.

delventhalz:
PR has been updated to reflect our offline discussions:
- `/interconnect` is now called `/messaging`
- client focused modules have been left in `/client`
- `state` modules have been moved to `/processor`
- There are now two encoders, Transaction and Batch, with equivalent functionality.

PR #342, opened 2017-04-15
Enforce and abide by PoET block claim frequency policy (zTest)

jjason:
Per the PoET specification, validators are limited in how frequently they may claim blocks - i.e., the zTest.  If it is determined that allowing a validator to win an election (i.e., claim a block and have it added to the blockchain) would cause it to claim blocks more frequently than a configurable statistical rate, the block is rejected.

This PR updates:
* refactors existing PoET policy tests (key block claim limit, block claim delay) to common utils
* updates the PoET block verifier to ensure that when verifying blocks it checks to make sure a validator is throttled by the block claim frequency allowed 
* updates the PoET block publisher to refuse to initialize a block if the block frequency check fails, indicating that other validators would reject the block
* removes the expected block claim count from the per-block consensus state as it is not necessary
* adds a short-circuit for blocks upon which block publisher has already determined it would not be allowed to build on top of

To test locally:

Add:

```
sawtooth.poet.block_claim_delay=1 \
sawtooth.poet.fixed_duration_block_count=5 \
sawtooth.poet.ztest_maximum_win_deviation=2.321 \
sawtooth.poet.ztest_minimum_win_count=2 \
```

This should increase the chance of validators refusing the publish blocks because it would violate the block claim frequency test.  You can look at the logs for error log messages.

To reduce the likelihood of zTest failures, you can change `sawtooth.poet.ztest_maximum_win_deviation`.  The values are based upon the confidence interval (i.e., how confident we are that we have truly detected a validator winning at a frequency we consider too frequent):

```
3.075 => 99.9%
2.575 => 99.5%
2.321 => 99%
1.645 => 95%
```

Execute:

### DIFF
--- a/consensus/poet/core/sawtooth_poet/poet_consensus/utils.py
+++ b/consensus/poet/core/sawtooth_poet/poet_consensus/utils.py
@@ -22,6 +22,7 @@ from sawtooth_validator.journal.block_wrapper import NULL_BLOCK_IDENTIFIER
 from sawtooth_poet_common.validator_registry_view.validator_registry_view \
     import ValidatorRegistryView
 
+from sawtooth_poet.poet_consensus.poet_config_view import PoetConfigView
 from sawtooth_poet.poet_consensus.wait_certificate import WaitCertificate
 from sawtooth_poet.poet_consensus.consensus_state import ConsensusState
 from sawtooth_poet.poet_consensus.consensus_state import ValidatorState
@@ -160,19 +161,14 @@ def get_current_validator_state(validator_info,
             point in time of the consensus state
     """
     # Fetch the validator state.  If it doesn't exist, then create an initial
-    # validator state object, including figuring out the block number in which
-    # the validator was most-recently registered.
+    # validator state object
     validator_state = \
         consensus_state.get_validator_state(
             validator_id=validator_info.id)
 
     if validator_state is None:
-        enclosing_block = \
-            block_cache.block_store.get_block_by_transaction_id(
-                validator_info.transaction_id)
         validator_state = \
             ValidatorState(
-                commit_block_number=enclosing_block.block_num,
                 key_block_claim_count=0,
                 poet_public_key=validator_info.signup_info.poet_public_key,
                 total_block_claim_count=0)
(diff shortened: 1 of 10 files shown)


jjason:
Let's try this again.  Rewrote the PoET zTest.

consensus/poet/core/tests/test_consensus/test_poet_config_view.py
```diff
@@ -22,6 +22,10 @@
 @patch('sawtooth_poet.poet_consensus.poet_config_view.ConfigView')
 class TestPoetConfigView(unittest.TestCase):
 
+    _KEY_BLOCK_CLAIM_LIMIT_ = 25
```
peterschwarz: Add `EXPECTED_DEFAULT_` for clarity

consensus/poet/core/sawtooth_poet/poet_consensus/utils.py
```diff
@@ -370,15 +362,348 @@ def get_consensus_state_for_block_id(
...
+        # number of wins necessary to trigger the zTest, then we are going to
+        # figure out if the validator is winning too frequently.
+        if estimate_info.validator_id == validator_info.id:
+            observed_wins += 1
+            if observed_wins > minimum_win_count and \
+                    observed_wins > expected_wins:
+                probability = expected_wins / block_count
+                standard_deviation = \
+                    math.sqrt(block_count * probability *
+                              (1.0 - probability))
```
dcmiddle: If you prefer the expression as is for readability with your referenced z-test article that's fine.
Otherwise, consider block_count algebraically cancels so you have s=sqrt(E-EP)

jjason: Yeah....that was left in only to try to make it easier to correlate code to spec.

consensus/poet/core/sawtooth_poet/poet_consensus/utils.py
```diff
@@ -370,15 +362,348 @@ def get_consensus_state_for_block_id(
...
+    observed_wins = 0
+    expected_wins = 0
+    block_count = 0
+    minimum_win_count = poet_config_view.ztest_minimum_win_count
+    maximum_win_deviation = poet_config_view.ztest_maximum_win_deviation
+
+    # We are now going to compute a "1 sample Z test" for each
+    # progressive range of results in the history.  Test the
+    # hypothesis that validator won elections (i.e., was able to
+    # claim blocks) with higher mean that expected.
```
jjason: Will fix.

jjason:
Fixed spelling mistake and rebased on master.

jsmitchell:
Tested locally

PR #343, opened 2017-04-17
Make Interconnect and Dispatcher more threadsafe

boydjohnson:
This commit also removes some cruft

### DIFF
--- a/validator/sawtooth_validator/networking/dispatch.py
+++ b/validator/sawtooth_validator/networking/dispatch.py
@@ -17,10 +17,12 @@ import enum
 from functools import partial
 import logging
 from threading import Condition
+from threading import Lock
 from threading import Thread
 import queue
 import uuid
 
+from sawtooth_validator.networking.interconnect import ThreadsafeDict
 from sawtooth_validator.networking.interconnect import get_enum_name
 from sawtooth_validator.protobuf import validator_pb2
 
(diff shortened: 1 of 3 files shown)


validator/sawtooth_validator/networking/dispatch.py
```diff
@@ -104,53 +103,59 @@ def add_handler(self, message_type, handler, executor):
...
+                            get_enum_name(message.message_type), connection_id,
+                            connection)
             self._process(message_id)
 
         elif future.result().status == HandlerStatus.RETURN:
-            with self._condition:
-                connection, connection_id,  \
-                    original_message, _ = self._message_information[message_id]
+            connection, connection_id,  \
+                original_message, _ = self._message_information[message_id]
```
peterschwarz: This get, coupled with the `del` later are no longer "atomic" w.r.t. the condition - is it ok to relax this?

validator/sawtooth_validator/networking/dispatch.py
```diff
@@ -195,11 +202,13 @@ class _ManagerCollection(object):
     def __init__(self, handler_managers):
         self._chain = handler_managers
         self._index = 0
+        self._lock = Lock()
 
     def __next__(self):
-        result = self._chain[self._index]
-        self._index += 1
-        return result
+        with self._lock:
```
peterschwarz: If this is a shared structure across threads, there's possible a lot wrong with this iteration, that a lock is not going to help:  the index will get advanced by each thread, which may result in them getting different iteration results.  Also, what raises `StopIteration` here?  Looks like right now, it will throw a ValueError instead.

boydjohnson: It will throw KeyError, which I could catch and raise StopIteration, but that would only be needed in a for-loop. Its current use is having next() directly called on it, and catching the KeyError if it is thrown. I don't see the "the index will get advanced by each thread, which may result in them getting different iteration results" as a bad thing, as this data structure is used in the message_information dict to see which Handler has been executed on a message.

validator/sawtooth_validator/networking/dispatch.py
boydjohnson: I think the .get plus del in these lines don't need to be atomic, since it would be fine for several .gets to happen across threads and then the dels. For any given message (tied to a message_id), the _determine_next function will be sequential.

boydjohnson:
@peterschwarz Do you have some time today to talk some of the threading considerations through about the Dispatcher in particular?

PR #345, opened 2017-04-17
Setup vagrant for Go Development

aludvik:
Test that everything (go, protogen, protobuf, zmq) is working by provisioning vagrant and then:
1. `protogen go`
2. `cd sdk/go/src/sawtooth_sdk`
3. `go run fake_validator.go`
4. `go run fake_processor.go`

### DIFF
--- a/protos/batch.proto
+++ b/protos/batch.proto
@@ -17,6 +17,7 @@ syntax = "proto3";
 
 option java_multiple_files = true;
 option java_package = "sawtooth.sdk.protobuf";
+option go_package = "batch";
 
 import "transaction.proto";
 
--- a/tools/plugins/install_go_tools.sh
+++ b/tools/plugins/install_go_tools.sh
@@ -1,0 +1,19 @@
+#!/bin/bash -x
+
+. /vagrant/conf.sh
+
+set -e
+
+apt-get install -y -q \
+    golang
+
+
+export GOPATH=/home/$VAGRANT_USER/go
+
+go get -u \
+    github.com/golang/protobuf/proto \
+    github.com/golang/protobuf/protoc-gen-go \
+    github.com/pebbe/zmq4
+
+chown -R $VAGRANT_USER:$VAGRANT_USER $GOPATH
+chown -R $VAGRANT_USER:$VAGRANT_USER $GOPATH/bin
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,7 @@
 /sdk/examples/intkey_java/dependency-reduced-pom.xml
 /sdk/examples/intkey_java/target/
 /sdk/java/target/
+/sdk/go/src/sawtooth_sdk/protobuf/
 
 /signing/build/
 /signing/sawtooth_signing.egg-info/
--- a/bin/protogen
+++ b/bin/protogen
@@ -19,12 +19,13 @@ import os
 import tempfile
 from glob import glob
 import re
+import sys
 
 
 try:
     from grpc.tools.protoc import main as _protoc
 except ImportError:
-    print("Error: protoc not found.")
+    print("Error: grpc.tools.protoc not found")
     exit(1)
 
 
@@ -32,7 +33,7 @@ JOIN = os.path.join
 TOP_DIR = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
 
 
-def main():
+def main(args=sys.argv[1:]):
     # 1. Generate and distribute top-level protos
     proto_dir = JOIN(TOP_DIR, "protos")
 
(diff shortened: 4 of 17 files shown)


protos/batch.proto
```diff
@@ -17,6 +17,7 @@ syntax = "proto3";
 
 option java_multiple_files = true;
 option java_package = "sawtooth.sdk.protobuf";
+option go_package = "sawtooth_sdk/protobuf";
```
peterschwarz: This commit should probably be sorted before the protogen changes.

aludvik: Hmm, can you do `git log` locally? I am seeing a discrepancy between the ordering here and local.

aludvik:
Added install_go_tools to default plugin list.

protos/batch.proto
aludvik: Or you can see the ordering here: https://github.com/aludvik/sawtooth-core/commits/install_go

vaporos: The go_package should perhaps include the proto name; for example "sawtooth_sdk/protobuf/batch", since having these all in the same namespace is awkward.

tools/plugins/install_go_tools.sh
```diff
@@ -0,0 +1,23 @@
...
+set -e
+
+GOPATH=/home/$VAGRANT_USER/go
+GOBIN=$GOPATH/bin
+
+export GOPATH
+echo "export GOPATH=$GOPATH:/project/sawtooth-core/sdk/go" >> /home/$VAGRANT_USER/.bashrc
+echo "export PATH=$PATH:$GOBIN" >> /home/$VAGRANT_USER/.bashrc
+
+apt-get install -y -q \
```
vaporos: This should be installed in the package_groups/ directory instead of a manual apt-get.

tools/plugins/install_go_tools.sh
```diff
@@ -0,0 +1,23 @@
...
+. /vagrant/conf.sh
+
+set -e
+
+GOPATH=/home/$VAGRANT_USER/go
+GOBIN=$GOPATH/bin
+
+export GOPATH
+echo "export GOPATH=$GOPATH:/project/sawtooth-core/sdk/go" >> /home/$VAGRANT_USER/.bashrc
+echo "export PATH=$PATH:$GOBIN" >> /home/$VAGRANT_USER/.bashrc
```
vaporos: This clobbers PATH since $PATH is expanded here.  Suggest adding GOPATH, PATH statements to local-env.sh instead.

sdk/go/src/sawtooth_sdk/fake_validator.go
```diff
@@ -0,0 +1,45 @@
+package main
```
vaporos: No need to commit a fake validator.  The fake processor works fine against the real validator if the validator is patched and the request is enclosed in a Message.

sdk/go/src/sawtooth_sdk/fake_processor.go
```diff
@@ -0,0 +1,44 @@
...
+	socket, _ := context.NewSocket(zmq.REQ)
+	defer context.Term()
+	defer socket.Close()
+
+	endpoint := "tcp://localhost:40000"
+	fmt.Printf("Connecting to %v...", endpoint)
+	socket.Connect(endpoint)
+	fmt.Println("done")
+
+	// Send registration request
```
vaporos: If we keep this file in this PR, let's update it to send a message the validator can understand.  Something like:

```	// Send registration request
	request := &protobuf.TpRegisterRequest{
		Family:     "test_family",
		Version:    "1.0",
		Encoding:   "protobuf",
		Namespaces: []string{"test_namespace"},
	}
	requestData, err := proto.Marshal(request)
	if err != nil {
		fmt.Printf("Failed to marshal: %v\n", err)
	}

	correlationId := "123"
	msg := &protobuf.Message{
		MessageType: protobuf.Message_TP_REGISTER_REQUEST,
		CorrelationId: correlationId,
		Content: requestData,
	}

	msgData, err := proto.Marshal(msg)
	if err != nil {
		fmt.Printf("Failed to marshal: %v\n", err)
	}
```

tools/plugins/install_go_tools.sh
aludvik: Should all the other apt-get statements for the other plugins be moved there as well? Creating new plugins has been the pattern I've been following for all additions, but maybe that was incorrect.

protos/batch.proto
aludvik: I tried this first, but was getting a lot of complaints from Go's protoc plugin. Will try again

aludvik:
1. Fixed vagrant setup.
2. Created separate packages for each .proto file
3. Changed fake_processor.go -> noop_processor.go which tries to register and unregister with a validator, but it doesn't totally work yet.

PR #348, opened 2017-04-18
Nml 2165 - Add rate batch limit to block publisher. 

cintel:
This PR contains 3 changes:
1) refactor the block publisher to move all of the logic managing a candidate block out of the BlockPublisher into a helper class. This also changed the policy for checking transaction dependencies to be done only in the context of a Candidate block that is being built. 

2) Update the block publisher to have a batch limit per block, to allow for block sizes to be controlled. 

3) Checks for duplicate transactions in batches. Look in the current chain and CandidateBlock to make sure that any transaction in a batch added has not been previously committed to the chain.

### DIFF
--- a/validator/sawtooth_validator/execution/scheduler_serial.py
+++ b/validator/sawtooth_validator/execution/scheduler_serial.py
@@ -49,6 +49,9 @@ class SerialScheduler(Scheduler):
         self._last_in_batch = []
         self._last_state_hash = first_state_hash
 
+    def __del__(self):
+        self.cancel()
+
     def __iter__(self):
         return SchedulerIterator(self, self._condition)
 
--- a/validator/sawtooth_validator/journal/publisher.py
+++ b/validator/sawtooth_validator/journal/publisher.py
@@ -72,6 +72,17 @@ class _CandidateBlock(object):
     def previous_block_id(self):
         return self._block_builder.previous_block_id
 
+    @property
+    def last_batch(self):
+        if self._pending_batches:
+            return self._pending_batches[-1]
+        return None
+
+    @property
+    def can_add_batch(self):
+        return self._max_batches == 0 or\
+            len(self._pending_batches) < self._max_batches
+
     def _check_batch_dependencies(self, batch, committed_txn_cache):
         """Check the dependencies for all transactions in this are present.
         If all are present the committed_txn is updated with all txn in this
@@ -295,7 +306,7 @@ class BlockPublisher(object):
 
         config_view = ConfigView(state_view)
         max_batches = config_view.get_setting(
-            'sawtooth.publisher.max_block_size',
+            'sawtooth.publisher.max_batches_per_block',
             default_value=0, value_type=int)
 
         consensus = consensus_module.\
@@ -328,7 +339,10 @@ class BlockPublisher(object):
                                                 block_builder,
                                                 max_batches)
         for batch in self._pending_batches:
-            self._candidate_block.add_batch(batch)
+            if self._candidate_block.can_add_batch:
+                self._candidate_block.add_batch(batch)
+            else:
+                break
 
     def on_batch_received(self, batch):
         """
@@ -339,7 +353,7 @@ class BlockPublisher(object):
         self._pending_batches.append(batch)
         # if we are building a block then send schedule it for
         # execution.
-        if self._candidate_block:
+        if self._candidate_block and self._candidate_block.can_add_batch:
             self._candidate_block.add_batch(batch)
 
     def _rebuild_pending_batches(self, committed_batches, uncommitted_batches):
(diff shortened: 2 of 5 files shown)


validator/sawtooth_validator/journal/publisher.py
```diff
@@ -32,10 +32,184 @@
...
 
+from sawtooth_validator.state.config_view import ConfigView
 
 LOGGER = logging.getLogger(__name__)
 
 
+class CandidateBlock(object):
+    """This is a helper class for the BlockPublisher. The CandidateBlock
+    tracks all the state associated with the Block that is being built.
+    This allows the BlockPublisher to
```
agunde406: Unfinished sentence

validator/sawtooth_validator/execution/scheduler_serial.py
```diff
@@ -49,6 +49,9 @@ def __init__(self, squash_handler, first_state_hash):
         self._last_in_batch = []
         self._last_state_hash = first_state_hash
 
+    def __del__(self):
```
boydjohnson: I don't see the __del__ method being called on the SerialScheduler (In CandidateBlock scheduler.cancel() is called, though). In general the only methods that should be called by external components are the ones defined in the scheduler.Scheduler abstract base class.

cintel: All components that hold scarce resources should clean them up automatically when they go away. When in python do you worry about an a file handle not being closed correctly. See the reference to __del__ here: https://docs.python.org/3/reference/datamodel.html. The CandidateBlock calls either finalize if the block is build or cancel if the block is discarded via the CandidateBlock.__del__ method.

validator/sawtooth_validator/journal/publisher.py
```diff
@@ -32,10 +32,185 @@
 from sawtooth_validator.protobuf.block_pb2 import BlockHeader
 from sawtooth_validator.protobuf.transaction_pb2 import TransactionHeader
 
+from sawtooth_validator.state.config_view import ConfigView
 
 LOGGER = logging.getLogger(__name__)
 
 
+class CandidateBlock(object):
```
jjason: Small nit....and all my comments probably are going to be....if this is meant to only be used within `publisher.py`, do you want to preprend `_` to name to indicate "private"?

validator/sawtooth_validator/journal/publisher.py
```diff
@@ -32,10 +32,185 @@
...
+
+    def finalize_block(self, identity_signing_key, pending_batches):
+        self._scheduler.finalize()
+        self._scheduler.complete(block=True)
+
+        # this is a transaction cache to track the transactions committed
+        # up to this batch. Only valid transactions that were processed
+        # by the scheduler are added.
+        committed_txn_cache = TransactionCache(self._block_store)
+        self._committed_txn_cache = TransactionCache(self._block_store)
```
jjason: I am confused why you create two caches here.  Maybe a comment on why it appears you are basically clearing out the candidate block object's committed transaction cache.

validator/sawtooth_validator/journal/publisher.py
```diff
@@ -306,97 +374,20 @@ def on_chain_updated(self, chain_head,
...
-                    self._consensus = None
-
-                    for batch in self._pending_batches:
-                        self._committed_txn_cache.add_batch(batch)
                 else:
+                    self._candidate_block = None
                     self._rebuild_pending_batches(committed_batches,
                                                   uncommitted_batches)
-                    self._candidate_block = self._build_block(chain_head)
+                    self._build_candidate_block(chain_head)
```
jjason: Could this be rewritten as follows:
```
# We can only build blocks if we have a chain head
self._candidate_block = None
if chain_head is not None:
  self._rebuild_pending_batches(committed batches, uncommitted_batches)
  self._build_candidate_block(chain_head)
```

validator/sawtooth_validator/journal/publisher.py
```diff
@@ -32,10 +32,200 @@
...
+                if not self._check_batch_dependencies(batch,
+                                                      committed_txn_cache):
+                    LOGGER.debug("Batch %s invalid, due to missing txn "
+                                 "dependency.", batch.header_signature)
+                    LOGGER.debug("Abandoning block %s:" +
+                                 "root state hash has invalid txn applied",
+                                 builder)
+                    pending_batches.remove(batch)
+                    pending_batches = pending_batches
+                    return None
```
jjason: ??? What does this do???

validator/sawtooth_validator/journal/publisher.py
```diff
@@ -32,10 +32,200 @@
...
+                # still succeed validation. In which case we do not want
+                # to add it to the batch.
+                if not self._check_batch_dependencies(batch,
+                                                      committed_txn_cache):
+                    LOGGER.debug("Batch %s invalid, due to missing txn "
+                                 "dependency.", batch.header_signature)
+                    LOGGER.debug("Abandoning block %s:" +
+                                 "root state hash has invalid txn applied",
+                                 builder)
+                    pending_batches.remove(batch)
```
jjason: When running locally with `sawtooth.poet.target_wait_time=5` and `sawtooth.poet.initial_wait_time=1`, got the following:
```
validator-2_1              | [17:25:10.247 INFO     chain] Starting block validation of : b6114567(34, S:82d27cee, P:c1d71581)
validator-2_1              | [17:25:10.651 INFO     poet_fork_resolver] Choose new fork b6114567: New fork aggregate local mean (170.000000) greater than current fork aggregate local mean (165.000000)
validator-2_1              | [17:25:10.673 INFO     chain] on_block_validated: b6114567(34, S:82d27cee, P:c1d71581)
validator-2_1              | [17:25:10.676 INFO     chain] Chain head updated to: b6114567(34, S:82d27cee, P:c1d71581)
validator-2_1              | [17:25:10.736 CRITICAL publisher] on_check_publish_block exception.
validator-2_1              | [17:25:10.737 ERROR    publisher] list.remove(x): x not in list
validator-2_1              | Traceback (most recent call last):
validator-2_1              |   File "/project/sawtooth-core/validator/sawtooth_validator/journal/publisher.py", line 453, in on_check_publish_block
validator-2_1              |     pending_batches)
validator-2_1              |   File "/project/sawtooth-core/validator/sawtooth_validator/journal/publisher.py", line 227, in finalize_block
validator-2_1              |     pending_batches.remove(batch)
validator-2_1              | ValueError: list.remove(x): x not in list
```
I can provide full log (although it is not a debug log) if you want.

validator/sawtooth_validator/journal/publisher.py
jjason: Oops...that should have been after the line:
`pending_batches = pending_batches`

validator/sawtooth_validator/execution/scheduler_serial.py
boydjohnson: Ah, cool. I was getting __del__ confused with __delitem__. The scheduler contains no external resources so doesn't have to release anything upon garbage collection. In the current code, the scheduler won't be garbage collected until either finalize or cancel is called, so calling .cancel again on garbage collection doesn't make much sense.

cintel: The scheduler is holding the execution of a thread in the thread pool. Since the condition held by the SerialScheduler must be signaled for the thread to be reclaimed. The penalty for not doing this is the lose of a thread in the Execution pools, so it is better to error on the side of caution cleaning up these resources then let these resources leak.

validator/sawtooth_validator/journal/publisher.py
dcmiddle: maybe you want to remove from `self._pending_batches` rather than `pending_batches`.

dcmiddle: actually maybe no maintenance of `self._pending_batches` is useful anyway. Seems there's no further use of it or the rest of this `_CandidateBlock` if this batch dependency failure condition is reached.

PR #349, opened 2017-04-18
Implement PoET Fork Resolution and Refactoring Goodness

jjason:
This update implements fork resolution using the same method previously used by PoET, with a few additions:

1.  If neither block is a PoET block, bail out.
2.  Otherwise, if one block is a PoET block and the other isn't, then the PoET block is chosen.
3.  Otherwise (both blocks are PoET blocks), if the blocks share the same immediate predecessor block, the one with the smaller wait duration is chosen.
4.  Otherwise (they have different predecessor blocks), the one with the greater aggregated local mean is chosen.
5.  Otherwise,  the one with the lexicographically greater block identifier (header signature) is chosen.

Criteria 1, 2, and 5 were added because prior to 0-8, non-PoET blocks were not a possibility.

In addition, the code has been refactored to better encapsulate the consensus state statistics and information used to make decision.

This PR updates:
* refactors existing PoET policy tests (key block claim limit, block claim delay) to common utils
* adds aggregate local mean to the consensus state, properly updates it when blocks are committed, and updates fork resolution based upon steps outlined previously
* most of the code in the common utils has been refactored to live inside the consensus state class to better encapsulate it
* as a side effect of the refactoring, all consensus state properties are now read-only as the refactoring allowed for updating of properties to always take place inside of consensus state object methods

To test locally:

Add:

```
sawtooth.poet.key_block_claim_count=5 \
sawtooth.poet.fixed_duration_block_count=10 \
```

Execute:

After the test has completed, you should be able to locate in the log, messages from `poet_fork_resolver` indicating why a particular fork was chosen.


jsmitchell:
Tested locally

jjason:
Final rebase (or freebase) before merge.

PR #350, opened 2017-04-18
Create new systemd unit files

aludvik:


aludvik:
Forgot to sign off on a commit

aludvik:
Rebased and added some extra logging.

aludvik:
Fixed isolation id not being in the environment

rest_api/packaging/ubuntu/postinst
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -e
+
+user="sawtooth"
+group="sawtooth"
+
+if ! getent passwd sawtooth > /dev/null; then
```
rbuysse: Should the getent tests in all the postinst scripts use the vars set above instead of "sawtooth"?

aludvik:
Fixed postinst files per @rbuysse comment

PR #351, opened 2017-04-18
Fixed config smoke test

ojalatodd:
Fixed config smoke test to account for PR #326
which changes arguments to validator.

Also added --host argument to rest_api container.


ojalatodd:
Added @vaporos and @peterschwarz at the suggestion of @nick-drozd as reviewers, since we added a new test module in run_tests.

PR #355, opened 2017-04-19
Documentation reorg

vaporos:
This is a stepping stone toward better 0.8 documentation.  It creates top-level sections which separate guides based on audience (app, core, sysadmin) and other top-level cleanups such as ordering and consistency between contents and index page.


docs/source/core_developers_guide.rst
```diff
@@ -0,0 +1,12 @@
+
+**********************
+Core Developer's Guide
+**********************
+
+.. toctree::
+   :maxdepth: 2
+
+   core_developers_guide/sawtooth_vagrant
+   core_developers_guide/intkey_workload.rst
```
delventhalz: Is there a reason this reference uses the `.rst` extension but the other three don't?

ojalatodd: Looks like core_developers_guide/sawtooth_vagrant path should be core_developers_guide/sawtooth_vagrant.rst, since there is no "sawtooth_vagrant" directory in core_developers_guide.

delventhalz: A directory would be `core_developers_guide/sawtooth_vagrant/`. I am no expert on sphinx or RST, but I assume you can leave off the `.rst` and it will be added automatically.

vaporos: The .rst is optional apparently, and was from a previous commit not in this PR.  I've added a commit to remove it as stylistically we should be consistent.

PR #357, opened 2017-04-19
Add Routed Batch and Block Response Handling to the Responder

agunde406:
The Responder will keep track of received requests that it could not fulfill. When it receives a response to the broadcasted request message, it will send the message back to the  originally requester. This is done using a TimedCache to store the requests, and two new handlers.

For example:
A broadcasts a  request to B
B cannot fulfill the request so B broadcast the request to C
C can fulfill the request and send a Response to B
B receives the response and sends it to A 

This PR also adds Responder unit tests for the above expected behavior.

### DIFF
--- a/validator/sawtooth_validator/journal/responder.py
+++ b/validator/sawtooth_validator/journal/responder.py
@@ -14,19 +14,28 @@
 # ------------------------------------------------------------------------------
 
 import logging
+import time
+from threading import RLock
 
 from sawtooth_validator.networking.dispatch import Handler
 from sawtooth_validator.networking.dispatch import HandlerResult
 from sawtooth_validator.networking.dispatch import HandlerStatus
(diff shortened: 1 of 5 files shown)


validator/sawtooth_validator/journal/responder.py
```diff
@@ -44,6 +53,31 @@ def check_for_batch_by_transaction(self, transaction_id):
...
         return batch
 
+    def add_request(self, requested_id, connection_id):
+        with self._lock:
+            if (requested_id in self.pending_requests) and \
+                    (connection_id not in self.pending_requests[requested_id]):
+                self.pending_requests[requested_id] += [connection_id]
+
+            else:
+                self.pending_requests[requested_id] = [connection_id]
```
agunde406: Good catch, that if should be separated into two, where if the connection id is already in the list, nothin should happen. Will update.

PR #359, opened 2017-04-19
Compute Local Mean in Consensus State and More Refactoring Goodness

jjason:
This _should_ be the last big PoET-related PR (fingers crossed).  It completes the transition from the 0-7-based architecture/implementation to the new 0-8 architecture.  This PR:
* adds the remaining generic PoET configuration settings (target wait time, et al) to the PoET configuration view, updates the PoET configuration view to cache values so it doesn't have to re-fetch them (as the state view is read-only and they cannot change for the particular instantiation of the PoET configuration view), and adds unit tests for checking them
* removes the `sawtooth.poet.fixed_duration_count` setting as it is redundant because of `sawtooth.poet.population_estimate_sample_size` and also brings code closer to spec (spec does not define a separate fixed duration count and instead has defined the population estimate sample size as the fixed duration count)
* updates the PoET enclave factory to use the PoET configuration view instead of directlying using the more-generic configuration view
* pulls the computation of the local mean inside the consensus state object, which includes encapsulating the keeping and updating of population samples within the consensus object
* updates the wait timer (when creating) and wait certificate (when validating) to use the consensus state for computing the local mean as well as removing the prior implementation of this in the WaitTimer class
* because everything is now encapsulated inside the consensus state class, the previous PoET configuration class variables (target wait time, et al) have been removed from the WaitTimer class

To test locally:

Add:

```
sawtooth.poet.fixed_population_estimate_sample_size=10 \
```

Execute:

### DIFF
--- a/consensus/poet/core/sawtooth_poet/poet_consensus/poet_config_view.py
+++ b/consensus/poet/core/sawtooth_poet/poet_consensus/poet_config_view.py
@@ -31,6 +31,7 @@ class PoetConfigView(object):
     _FIXED_DURATION_BLOCK_COUNT_ = 50
     _INITIAL_WAIT_TIME_ = 3000.0
     _KEY_BLOCK_CLAIM_LIMIT_ = 25
+    _MINIMUM_WAIT_TIME_ = 1.0
     _TARGET_WAIT_TIME_ = 20.0
     _ZTEST_MAXIMUM_WIN_DEVIATION_ = 3.075
     _ZTEST_MINIMUM_WIN_COUNT_ = 3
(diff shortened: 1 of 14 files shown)


jsmitchell:
Tested locally - I did see an error on one run, but I don't think it's related to these changes. The error was a state where nodes were repeatedly requesting a specific missing block. It looked like the responses were being sent, but then the block would be requested again. Perhaps the requested block was invalid somehow and being tossed out prior to hitting the completer (bad signature?).

consensus/poet/core/sawtooth_poet/poet_consensus/poet_config_view.py
```diff
@@ -51,6 +51,16 @@ def __init__(self, state_view):
 
         self._config_view = ConfigView(state_view)
 
+        self._block_claim_delay = None
```
peterschwarz: As long as the lifetimes of instances of this class are short, this is probably fine.  If this is too-long lived, changes in these values that occur between block commits will get lost.

jjason:
Rebase before merge

PR #360, opened 2017-04-19
Update of Vagrant file and tutorial for proxy 

ojalatodd:
Update of Vagrant file and tutorial. 
Tutorial explains how to set env variables used by proxy.
Vagrant file does not notify about proxyconf unless http_proxy 
or https_proxy are set.

### DIFF
--- a/docs/source/app_developers_guide/getting_started.rst
+++ b/docs/source/app_developers_guide/getting_started.rst
@@ -157,17 +157,24 @@ If you are behind a network proxy, follow these steps before continuing:
 
 If you are using the Bash shell, run the following commands:
 
+.. warning::
+
+  The example URLs and port numbers used below are examples only.
+  Please substitute the actual URL, with actual port numbers, used
+  in your environment. Contact your network administrator for the
+  information if necessary.
+
 .. code-block:: console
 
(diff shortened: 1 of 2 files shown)


docs/source/tutorial.rst
```diff
@@ -144,6 +144,40 @@ The following tools are required:
...
+1. Set the following environment variables:
+
+  * http_proxy
+  * https_proxy
+
+If you are using the Bash shell, run the following commands:
+
+.. code-block:: console
+
+  % export http_proxy=http://example-proxy-server.com:911
```
rbuysse: The port numbers after the colons should be changed to something more standard like 3128, 8080, 8000, or even just <port>.

ojalatodd: OK, I changed to more standard ports, and also put a warning box in to tell user to make sure they are using the actual settings for their environment.

ojalatodd:
@dcmiddle OK. The doc is very clear on using the correct port numbers now.
Regarding moving the changes to the "right file," the right file does not exist yet. 
I suggest keeping the change where it is now, after the rebase I just did, which is in the "Getting Started" part of the Application Developer's Guide. The proxy section **will** be put in the right place in the next sprint, when it exists. 
Sound good?

PR #361, opened 2017-04-19
Remove non-ballot mode from sawtooth config

peterschwarz:
Remove the non-ballot or non-voting mode from the sawtooth config transaction family reference implementation.  This change requires that there be at least on public key set in the authorized keys list before any setting can be changed.

The end result is a more secure model for handling configuration changes on-chain.

With the chain in the config transaction family requiring votes by authorized key, this PR introduces a new subcommand for `sawtooth config`: `genesis`.  Genesis provides a simple way to add authorized keys and set the vote threshold for accepting or rejecting a proposal.

Additionally:
 - config smoke adds the genesis step
 - poet smoke adds the genesis step

### DIFF
--- a/cli/sawtooth_cli/config.py
+++ b/cli/sawtooth_cli/config.py
@@ -505,7 +505,6 @@ def _config_inputs(key):
     setting key.
     """
     return [
-        _key_to_address('sawtooth.config.authorization_type'),
         _key_to_address('sawtooth.config.vote.proposals'),
         _key_to_address('sawtooth.config.vote.authorized_keys'),
         _key_to_address('sawtooth.config.vote.approval_threshold'),
--- a/cli/sawtooth_cli/config.py
+++ b/cli/sawtooth_cli/config.py
@@ -53,6 +53,34 @@ def add_config_parser(subparsers, parent_parser):
                                            dest="subcommand")
     config_parsers.required = True
 
+    # The following parser is for the `genesis` subcommand.
+    # This command creates a batch that contains all of the initial
+    # transactions for on-chain settings
+    genesis_parser = config_parsers.add_parser('genesis')
+    genesis_parser.add_argument(
+        '-k', '--key',
+        type=str,
+        help='the signing key for the resulting batches '
+             'and initial authorized key')
+
+    genesis_parser.add_argument(
+        '-o', '--output',
+        type=str,
+        default='config-genesis.batch',
+        help='the name of the file to output the resulting batches')
+
+    genesis_parser.add_argument(
+        '-T', '--approval-threshold',
+        type=int,
+        help='the required number of votes to enable a setting change')
+
+    genesis_parser.add_argument(
+        '-A', '--authorized-key',
+        type=str,
+        action='append',
+        help='a public key for an user authorized to submit '
+             'config transactions')
+
     # The following parser is for the `proposal` subcommand group. These
     # commands allow the user to create proposals which may be applied
     # immediately or placed in ballot mode, depending on the current on-chain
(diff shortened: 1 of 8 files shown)


cli/sawtooth_cli/config.py
```diff
@@ -53,6 +53,34 @@ def add_config_parser(subparsers, parent_parser):
...
+             'and inital authorized key')
+
+    genesis_parser.add_argument(
+        '-o', '--output',
+        type=str,
+        default='config-genesis.batch',
+        help='the name of the file to ouput the resulting batches')
+
+    genesis_parser.add_argument(
+        '--approval_threshold',
```
vaporos: Use - instead of _ in the command line arguments.

This one is quite long.  If a shorter argument (--threshold?) won't work, maybe add a short option (-T?).

cli/sawtooth_cli/config.py
```diff
@@ -53,6 +53,34 @@ def add_config_parser(subparsers, parent_parser):
...
+        type=int,
+        help='the required number of votes to enable a setting change')
+
+    genesis_parser.add_argument(
+        '--authorized_key',
```
vaporos: Use - instead of _.

cli/sawtooth_cli/config.py
```diff
@@ -366,6 +396,46 @@ def _do_config_list(args):
...
+    txns = []
+
+    txns.append(_create_propose_txn(
+        pubkey, signing_key,
+        ('sawtooth.config.vote.authorized_keys',
+         ','.join(authorized_keys))))
+
+    if args.approval_threshold is not None:
+        if args.approval_threshold < 1:
+            raise CliException('approval_threshold must be greater than 0')
```
vaporos: I'd remove the _ here and make it english instead of part of the CLI argument.

cli/sawtooth_cli/config.py
peterschwarz: Will do the same for `--authorized-key': `-A`

cli/sawtooth_cli/config.py
```diff
@@ -53,6 +53,34 @@ def add_config_parser(subparsers, parent_parser):
...
+        '-k', '--key',
+        type=str,
+        help='the signing key for the resulting batches '
+             'and inital authorized key')
+
+    genesis_parser.add_argument(
+        '-o', '--output',
+        type=str,
+        default='config-genesis.batch',
+        help='the name of the file to ouput the resulting batches')
```
agunde406: ouput -> output

cli/sawtooth_cli/config.py
```diff
@@ -366,6 +396,46 @@ def _do_config_list(args):
...
+        ('sawtooth.config.vote.authorized_keys',
+         ','.join(authorized_keys))))
+
+    if args.approval_threshold is not None:
+        if args.approval_threshold < 1:
+            raise CliException('approval threshold must be greater than 0')
+
+        if args.approval_threshold > len(authorized_keys):
+            raise CliException(
+                'approval_threshold must not be greater than the number of '
```
agunde406: approval threshold or approval_threshold? It should be consistent between the two CliException above

cli/sawtooth_cli/config.py
```diff
@@ -53,6 +53,34 @@ def add_config_parser(subparsers, parent_parser):
...
 
+    # The following parser is for the `genesis` subcommand.
+    # This command creates a batch that contains all of the initial
+    # transactions for on-chain settings
+    genesis_parser = config_parsers.add_parser('genesis')
+    genesis_parser.add_argument(
+        '-k', '--key',
+        type=str,
+        help='the signing key for the resulting batches '
+             'and inital authorized key')
```
agunde406: inital -> initial

peterschwarz:
Rebased from master

PR #362, opened 2017-04-20
Remove yarn

delventhalz:
As yarn's lock file generation has caused some docker build issues,
and it is probably too early to peg specific version numbers, and
we are not actually utlizing yarn's extra functionality, remove it
and use npm for JavaScript installs.


delventhalz:
Removing yarn has turned into a huge rabbit hole, as it seems to have been papering over and/or solving a number of dependency issues. I am closing this PR and opening a JIRA task for removing yarn. I'll push up a separate PR with fixed lock files.

PR #365, opened 2017-04-21
Regen yarn lock files to fix build issue

delventhalz:
Builds for installed were failing because new uncommitted lock
files were getting generated during building. This fixes that
problem.


peterschwarz:
Approved - but with the requirement that we have a future story that removes Yarn.

PR #366, opened 2017-04-21
Implement Validator Signup Recent Block Test Plus Unit Test Goodness

jjason:
This update adds an additional test in the PoET block verifier and publisher to check that the signup information for the validator was committed to the blockchain within a small number of blocks of it being created.  Specifically, when a validator generates its signup information it includes in the information a nonce, which is the block ID of the current head chain.  When validating the signup information is "fresh" other validators will look at the nonce and compare it to some number of previous blocks.  The number is configurable, but defaults to a delay of zero, meaning that the validator's signup information must be committed in the block immediately after the block referenced in the nonce.  This test can be relaxed by increasing the allowed delay.

Updates in this PR include:
* Adds some unit tests for the different PoET policies
* Updates the signup information in the valiator registry for the nonce, removing commented-out tests and code that no longer is necessary in the PoET simulator or validator registry transaction processor
* Updates the PoET config view to have a configuration setting to configure the maximum amount of signup commit delay allowed
* Updates the PoET consensus state to make the check of the signup information
* Updates the PoET block verifier to ensure signup information for a validator claiming a block passes the new check
* Updates the PoET block publisher to make the signup info check before initializing a block.  If the check fails, the PoET block publisher will attempt to sign up again

To test locally:

Add:

```
```

Execute:

### DIFF
--- a/consensus/poet/cli/sawtooth_poet_cli/genesis.py
+++ b/consensus/poet/cli/sawtooth_poet_cli/genesis.py
@@ -123,8 +123,8 @@ def do_genesis(args):
             signup_info=vr_pb.SignUpInfo(
                 poet_public_key=signup_info.poet_public_key,
                 proof_data=signup_info.proof_data,
-                anti_sybil_id=signup_info.anti_sybil_id),
-        )
+                anti_sybil_id=signup_info.anti_sybil_id,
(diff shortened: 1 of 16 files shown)


agunde406:
Tested locally

jjason:
Rebasing before merge

PR #368, opened 2017-04-21
Update PoET fork resolver with respect to non-PoET blocks.

jjason:
PoET fork resolution is changed to:
* if the new fork head is a non-PoET block, raises an exception as this is a completely unexpected turn of events
* if the current head is non-PoET but it is the immediate predecessor to the new fork head, it is reported as the blockchain changing to PoET consensus
* if the current head is non-PoET and it is not the immediate predecessor to the new fork head, it raise an exception as this is a completely unexpected turn of events

Remaining fork resolution continues as before.
To test locally:

Add:

```
```

Execute:


agunde406:
Tested Locally

PR #369, opened 2017-04-21
Update keygen artifacts to produce .pub & .priv files

ojalatodd:
Commands 'sawtooth admin keygen' and 'sawtooth keygen'
updated to produce files with extensions .priv & .pub
instead of .wif & .addr

Files ending in .priv are equivalent to the files that
ended in .wif.

### DIFF
--- a/extensions/arcade/sawtooth_battleship/battleship_cli.py
+++ b/extensions/arcade/sawtooth_battleship/battleship_cli.py
@@ -228,8 +228,8 @@ def do_init(args, config):
     save_config(config)
 
     wif_filename = config.get('DEFAULT', 'key_file')
-    if wif_filename.endswith(".wif"):
-        addr_filename = wif_filename[0:-len(".wif")] + ".addr"
+    if wif_filename.endswith(".priv"):
+        addr_filename = wif_filename[0:-len(".priv")] + ".addr"
     else:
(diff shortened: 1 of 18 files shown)


ojalatodd:
Notes:
I had intended to make the documentation update a separate commit, but mistakenly squashed it together. 

Second note: battleship and XO still use and create their own .addr files. Not sure if Battleship is currently even functional, but didn't want to touch code unnecessarily.  The creation of .addr files from the public key by the games seemed like something that didn't need to be changed.

extensions/arcade/sawtooth_battleship/battleship_cli.py
```diff
@@ -228,8 +228,8 @@ def do_init(args, config):
     save_config(config)
 
     wif_filename = config.get('DEFAULT', 'key_file')
-    if wif_filename.endswith(".wif"):
-        addr_filename = wif_filename[0:-len(".wif")] + ".addr"
+    if wif_filename.endswith(".priv"):
+        addr_filename = wif_filename[0:-len(".priv")] + ".addr"
```
peterschwarz: Probably for another PR, but this address can be generated from the contents of the `.pub` file

ojalatodd: Noting this.

PR #370, opened 2017-04-21
Update DevMode consensus to handle non-DevMode blocks.

jjason:
DevMode consensus is updated to, before looking at block header data:
* if the new fork head is not a DevMode block, raises an exception
* if the current fork head is not a DevMode block and is the immediate predecessor to the new fork, it immediately accepts the new fork head as that indicates that consensus mode is changing
* if the current fork head is not a DevMode block and also is not the immediate predecessor to the new fork head, it raises an exception

Easiest way I found to test was manually (fingers crossed these instructions work) in multiple xterms:

Start up a validator:
```
sawtooth admin keygen --force
sawtooth config proposal create -k /home/ubuntu/sawtooth/keys/validator.wif sawtooth.consensus.algorithm=poet -o config.batch
poet genesis -k /home/ubuntu/sawtooth/keys/validator.wif -o poet.batch
sawtooth admin genesis config.batch poet.batch
validator -v --public-uri tcp://0.0.0.0:8800
```
Start up a validator registry TP:
```
tp_validator_registry -v tcp://localhost:40000
```
Start up a config TP:
```
tp_config -v tcp://localhost:40000
```
Start up REST API:
```
rest_api
````
Propose a config setting for DevMode consensus:
```
/sawtooth config proposal create -k /home/ubuntu/sawtooth/keys/validator.wif sawtooth.consensus.algorithm=devmode
```
Ensure that after you create the config proposal that in the validator log you see that the blockchain is udpated (you should see a message about the new fork head being chosen).  Depending upon the wait timer chosen by the PoET consensus module, this may take a minute or so (it is random after all).

Propose a config setting for PoET consensus:
```
/sawtooth config proposal create -k /home/ubuntu/sawtooth/keys/validator.wif sawtooth.consensus.algorithm=poet
```
Monitor the log for the validator - you should see a message when you propose the config setting `sawtooth.consensus.algorithm=poet` that DevMode has chosen the new fork head and it should also log a message that it detected a switch to DevMode consensus.

Repeat the propose Poet/propose DevMode several times.  You should see DevMode detecting the switch.

Then propose a config setting that does not switch consensus modes:
```
/sawtooth config proposal create -k /home/ubuntu/sawtooth/keys/validator.wif sawtooth.poet.target_wait_time=5
```
You should see DevMode fork resolver choose the new fork, but no message about detecting a change as it should be building upon a DevMode block.

### DIFF
--- a/validator/sawtooth_validator/journal/consensus/dev_mode/dev_mode_consensus.py
+++ b/validator/sawtooth_validator/journal/consensus/dev_mode/dev_mode_consensus.py
@@ -16,6 +16,7 @@
 import time
 import random
 import hashlib
+import logging
 
 from sawtooth_validator.journal.block_wrapper import BlockWrapper
 from sawtooth_validator.journal.consensus.consensus\
@@ -26,6 +27,8 @@ from sawtooth_validator.journal.consensus.consensus\
     import ForkResolverInterface
 
 from sawtooth_validator.state.config_view import ConfigView
+
+LOGGER = logging.getLogger(__name__)
 
 
 class BlockPublisher(BlockPublisherInterface):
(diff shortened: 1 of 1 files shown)


jjason:
Rebased on master

jsmitchell:
Tested locally

jjason:
Rebase before merge

PR #372, opened 2017-04-24
Add Darian Plumb to MAINTAINERS list

dplumb94:


MAINTAINERS
```diff
@@ -45,3 +45,6 @@
 
 - name: Peter Schwarz
   user: peterschwarz
+
+- name: Darian Plumb
+  user: dplumb
```
peterschwarz: Does this need to be your github user name? If so, it should be `dplumb94`

PR #375, opened 2017-04-24
Add global state section to architecture doc

jsmitchell:


docs/source/architecture/global_state.rst
```diff
@@ -0,0 +1,99 @@
...
+Serialization Concerns
+======================
+
+In addition to questions regarding the encoding of addresses,
+namespace designers also need to define the mechanism of serialization
+and the rules for serializing/deserializing the data stored at addresses.
+The domain-specific Transaction Processor will be making get(address) and
+set(address, data) calls against a version of state that the validator
+provides. get(address) will return the byte array found at that address
+and set(address, data) will set the byte array stored at that address.
```
aludvik: I think these sentences switch tense which threw me off while reading. Should it be "The domain-specific Transaction Processor makes get(address) and set(address, data) calls against a version of state that the validator provides. get(address) returns the byte array found at that address and set(address, data) sets the byte array stored at that address."?

PR #377, opened 2017-04-25
[STL-24] Fix Golang SDK "index out of bound" panic

hacera:
Fixes https://jira.hyperledger.org/browse/STL-24

Signed-off-by: Jonathan Levi (HACERA) <jonathan@hacera.com>


vaporos:
It would be preferable to remove the [STL-24] from the first line of the commit message.

PR #383, opened 2017-04-25
Update Dockerfiles and Vagrant with new public apt repos

rbuysse:


aludvik:
After this is merged we probably want to blow away the repo at https://repo.sawtooth.me/ if nothing else is using it. Probably wait till after the hackathon though, so people there don't have issues.

PR #384, opened 2017-04-26
Store State Deltas

peterschwarz:
Creates a StateDeltaStore, where state changes can be stored.  These deltas are stored when the context manager performs the squash operation _and_ the state updates will be persisted.  

The purpose of storing these deltas is to support State Delta Subscriptions by external processes.

### DIFF
--- a/validator/sawtooth_validator/state/state_delta_store.py
+++ b/validator/sawtooth_validator/state/state_delta_store.py
@@ -1,0 +1,64 @@
+# Copyright 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


validator/sawtooth_validator/state/state_delta_store.py
```diff
@@ -0,0 +1,64 @@
...
+from sawtooth_validator.protobuf.state_delta_pb2 import StateDeltaSet
+
+
+class StateDeltaStore(object):
+    """A StateDeltaStore persists StateChange records to a provided database
+    implementation.
+    """
+
+    def __init__(self, delta_db):
+        """ Constructs a StateDeltaStore, backed by a given database
```
jsmitchell: Spurious space

PR #386, opened 2017-04-26
Setting up the dev environment tutorial

ojalatodd:
New tutorial added for setting up the dev
environment. This includes instructions that
comes from the App Dev guide Getting Started
tutorial, and the Vagrant guide from the Core
Dev guide.


docs/source/core_developers_guide/environment_setup.rst
```diff
@@ -144,3 +174,182 @@ install_sphinx
...
+   Occasionally, the configuration files used to create the vagrant
+   environment change and this can result in vagrant asking for a password when
+   doing `vagrant up` or `vagrant ssh`. This is usually a sign that your
+   environment is out of date. If this happens, please follow the instructions
+   below to reset your environment. This is especially common when switching
+   between versions of Sawtooth Lake, eg. 0.7 -> 0.8.
+
+
+
+Resetting The Environment (Optional)
```
delventhalz: _"Step Three:"_ missing from this header. That said, these don't really seem like a strictly ordered set of steps. I think we should probably remove the "Step" verbage from all headings.

docs/source/core_developers_guide/environment_setup.rst
```diff
@@ -144,3 +174,182 @@ install_sphinx
...
+   /vagrant and /project are shared with the host and will be preserved.
+
+
+Step Four: Build sawtooth-core
+==============================
+
+Most of the components of Sawtooth Lake depend partially on code that must
+first be built. This includes C++/swig code and generating protobuf classes for
+each language. Sawtooth Lake also includes SDKs for other languages, including
+Java and JavaScript, which require building. To build the python portion of
```
delventhalz: Strictly speaking, JavaScript is not _built_, it just needs dependencies installed. Unlike Java or C++, (which must be parsed into executable byte code before running), JavaScript runs right off the source code.

docs/source/core_developers_guide/environment_setup.rst
```diff
@@ -144,3 +174,182 @@ install_sphinx
...
+Sawtooth Lake within vagrant, do:
+
+.. code-block:: console
+
+  $ /project/sawtooth-core/bin/build_python
```
delventhalz: In the previous paragraph, you mostly talk about the non-python elements that need to be built, and then you only mention the script for Python. I think we should start with `build_all` as the example, and then explain in all likelihood, you do not need to wait for this lengthy build process, and that they should just `build_python` instead.

docs/source/core_developers_guide/environment_setup.rst
```diff
@@ -144,3 +174,182 @@ install_sphinx
...
+
+This will allow you to run and test Sawtooth Lake components manually. The
+automated tests for python and all other languages rely on docker to ensure
+reproducibility. To run the automated tests for python, first run the
+following:
+
+.. code-block:: console
+
+  $ /project/sawtooth-core/bin/build_all -l python
```
delventhalz: The difference between `build_python` and `build_all -l python` is not clear to me. These two commands could probably be explained a little more in-depth.

docs/source/core_developers_guide/environment_setup.rst
```diff
@@ -144,3 +174,182 @@ install_sphinx
...
+
+This will create docker images for all the python components and run
+`build_python` inside a build container. You can then run the automated tests
+with:
+
+.. code-block:: console
+
+  $ /project/sawtooth-core/bin/run_tests -x java_sdk -x javascript_sdk
```
delventhalz: Once again why are we using `-x java_sdk -x javascript_sdk`? What does that do? There is no context here.

Also `run_tests` seems a little out of place in this section. There shoudl either be a separate _"Running Tests"_ section, or this section should be renamed something like _"Building Components and Running Tests"_.

docs/source/core_developers_guide/environment_setup.rst
```diff
@@ -144,3 +174,182 @@ install_sphinx
...
+  % vagrant destroy
+  % vagrant up
+
+.. warning::
+
+   vagrant destroy will delete all contents within the VM. However,
+   /vagrant and /project are shared with the host and will be preserved.
+
+
+Step Four: Build sawtooth-core
```
delventhalz: How about _"Build Components"_? We know we are in `sawtooth-core/`, but we don't know what we're building.

docs/source/core_developers_guide/environment_setup.rst
delventhalz: Also, this heading uses the gerund form of the verb, while the others all use the imperative. _"Resetting the Environment"_ versus _"Configure Proxy"_.

docs/source/core_developers_guide/environment_setup.rst
```diff
@@ -144,3 +174,182 @@ install_sphinx
...
+
+2. Install the vagrant-proxyconf plugin:
+
+.. code-block:: console
+
+  % cd sawtooth-core/tools
+  % vagrant plugin install vagrant-proxyconf
+
+
+Step Three: Start VM
```
delventhalz: I would rather write out _"Virtual Machine"_ here. The verb _"Start"_ also seems a little inaccurate. Maybe, _"Build and Run Virtual Machine"_.

docs/source/core_developers_guide/environment_setup.rst
```diff
@@ -1,9 +1,37 @@
 
-Vagrant
-=======
+***************
+Getting Started
```
ojalatodd: Changed to Environment Setup

docs/source/core_developers_guide/environment_setup.rst
```diff
@@ -18,8 +46,9 @@ which can run validators within a few minutes.
 A quick introduction to using dev tools is available as a
 quickstart in this Developer's Guide.
 
+
 Layout of the sawtooth-dev-tools git Repository
-------------------------------------------------
+===============================================
```
ojalatodd: Changed to Layout of Dev Tools

docs/source/core_developers_guide/environment_setup.rst
```diff
@@ -144,3 +174,182 @@ install_sphinx
...
+* `Git install instructions <https://git-scm.com/book/en/v2/Getting-Started-Installing-Git>`_
+
+.. note:: 
+
+  When checking out Sawtooth Lake on Windows for use with vagrant, you should
+  take steps to ensure that Windows-style CRLF line endings are not added to
+  the code. The bash scripts used by your vagrant VM will not run correctly 
+  with CRLF line endings. Git uses a configuration setting, *core.autocrlf*,
+  to control whether or not LF-style line endings are automatically converted
+  to CRLF-style line endings. `This setting should be set in such a way that
```
ojalatodd: @delventhalz Good points. This "note" is already in our documentation, and went through a lot of iterations already. However, I will try to make it less awkward. For some reason, I was told to not put the exact setting to use, but just reference it (in case it changes?)

docs/source/core_developers_guide/environment_setup.rst
ojalatodd: Changed to your suggestion.

docs/source/core_developers_guide/environment_setup.rst
ojalatodd: Removed "Step" verbiage. The order should make it fairly obvious.
changed "Resetting" to "Reset."

docs/source/core_developers_guide/environment_setup.rst
ojalatodd: Used suggestion.

docs/source/core_developers_guide/environment_setup.rst
ojalatodd: Rewrote.

docs/source/core_developers_guide/environment_setup.rst
ojalatodd: Rewrote.

docs/source/core_developers_guide/environment_setup.rst
ojalatodd: Created new section. It should be expanded in another iteration (lots more to testing)

docs/source/core_developers_guide/environment_setup.rst
```diff
@@ -0,0 +1,374 @@
...
+
+   vagrant destroy will delete all contents within the VM. However,
+   /vagrant and /project are shared with the host and will be preserved.
+
+
+Build Components
+================
+
+Some of the components of Sawtooth depend partially on code that must
+first be built. This includes C++/swig code and generating protobuf classes
```
dcmiddle: Are we swig free now? 
@cintel a quick search of your impending c++ sdk pr didn't show any swig. Will that have any swigging?

docs/source/core_developers_guide/environment_setup.rst
```diff
@@ -0,0 +1,374 @@
...
+    plugins/
+    scripts/
+    tests/
+    win-bin/
+    Vagrantfile
+
+
+Vagrantfile
+  Vagrantfile is the main configuration file for Vagrant.  It is ruby
+  code executed by Vagrant and is executed every ime a vagrant command is
```
rbuysse: typo: "every ime"

docs/source/core_developers_guide/environment_setup.rst
```diff
@@ -0,0 +1,374 @@
...
+on the host:
+
+.. code-block:: console
+
+  % vagrant destroy
+  % vagrant up
+
+.. warning::
+
+   vagrant destroy will delete all contents within the VM. However,
```
rbuysse: Some places Vagrant is capitalized, and some not. Should probably always be capitalized when not referring to commands.

docs/source/core_developers_guide/environment_setup.rst
```diff
@@ -0,0 +1,374 @@
...
+This will create docker images for all the python components and run
+`build_python` inside a build container. You can then run the automated tests
+for python components, while excluding java and javascript components,
+with:
+
+.. code-block:: console
+
+  $ /project/sawtooth-core/bin/run_tests -x java_sdk -x javascript_sdk
+
+  .. note::
```
rbuysse: Extra spaces here are causing it to render incorrectly.

docs/source/core_developers_guide/environment_setup.rst
cintel: Not sure what I would be swigging in the C++ sdk, I can interface directly with the generated c++ protobuf interfaces.

docs/source/core_developers_guide/environment_setup.rst
ojalatodd: Fixed

docs/source/core_developers_guide/environment_setup.rst
ojalatodd: Done

docs/source/core_developers_guide/environment_setup.rst
ojalatodd: Fixed

docs/source/core_developers_guide/environment_setup.rst
```diff
@@ -0,0 +1,374 @@
...
+A convenient feature of Vagrant is the easy sharing of directories between
+the host machine and the guest virtual machine. By default the Vagrant
+configuration directory is mounted under /vagrant in the virtual machine.
+
+In the dev tools configuration a /project mount point is also
+defined which provides access to the Sawtooth repositories.
+
+
+Configuration Options
+=====================
```
aludvik: I think that a lot of this is out of date. Probably need stories to update `conf-local.sh.example` For example, `build_ubuntu_deps` isn't a plugin anymore and I am not aware of the INSTALL_TYPE or START_TXNVALIDATOR config values.

docs/source/core_developers_guide/environment_setup.rst
aludvik: Probably for this tutorial, it is enough to just instruct the user to run `build_all` and `run_tests` and then let them know about the flags and non-docker build scripts. Following it as it is written, someone will likely do `build_all`, then `build_python`, then `build_all -l python`

docs/source/core_developers_guide/environment_setup.rst
aludvik: Actually, you may want to just omit the Configuration Options section

ojalatodd: Agreed. Let's add a story if we think we need this re-added with updated info. I'll make a to-do for this for early next week.

PR #387, opened 2017-04-26
Update REST API docs, adding error documentation

delventhalz:
- Update REST API spec to remove /stats and /status endpoints
- Move current REST API docs into REST API section
- Add new error doc
- Hack sphinx

### DIFF
--- a/docs/source/rest_api/endpoint_specs.rst
+++ b/docs/source/rest_api/endpoint_specs.rst
@@ -1,0 +1,17 @@
+***********************
+Endpoint Specifications
+***********************
+
+These specifications document the available Sawtooth REST API endpoints. They
+are generated from the official `OpenAPI <http://swagger.io/specification/>`_
+formatted YAML specification, which can be found on the
+`Sawtooth Lake Github <https://github.com/hyperledger/sawtooth-core/blob/master/rest_api/sawtooth_rest_api/openapi.yaml>`_.
+
+
+.. note::
+
+   The query parameters *sort*, *fields*, *omit*, and *callback* have yet to
+   be implemented, and will not operate as specified.
+
+
+.. openapi:: ../../../rest_api/sawtooth_rest_api/openapi.yaml
--- a/docs/source/rest_api/error_codes.rst
+++ b/docs/source/rest_api/error_codes.rst
@@ -1,0 +1,165 @@
+***************
+Error Responses
+***************
+
+When the REST API encounters a problem, or receives notification that the
+validator has encountered a problem, it will notify clients with both an
+appropriate HTTP status code, and a more detailed JSON response.
+
+
+HTTP Status Codes
+=================
+
+.. list-table::
+   :widths: 4, 16, 60
+   :header-rows: 1
+
+   * - Code
+     - Title
+     - Description
+   * - 400
+     - Bad Request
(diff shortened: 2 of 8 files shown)


rest_api/sawtooth_rest_api/exceptions.py
```diff
@@ -194,7 +194,7 @@ class StatusIdQueryInvalid(_ApiError):
     status_code = 400
     title = 'Id Query Invalid or Missing'
     message = ("Requests for batch statuses sent as a GET request must have "
-               "an 'id' query parameter with a comma seperated list of "
+               "an 'id' query parameter with a comma separated list of "
```
peterschwarz: comma-separated

docs/source/rest_api/endpoint_specs.rst
```diff
@@ -0,0 +1,10 @@
+***********************
+Endpoint Specifications
+***********************
+
+These specifications document the available Sawtooth REST API endpoints. They are generated from the official `OpenAPI <http://swagger.io/specification/>`_ formatted YAML specification, which can be found on the `Sawtooth Lake Github <https://github.com/hyperledger/sawtooth-core/blob/master/rest_api/sawtooth_rest_api/openapi.yaml>`_.
+
+.. note::
+   This specification has not been fully implemented. Some query parameters may not function or behave as specified.
```
peterschwarz: Do we have a way to mark which have not been implemented?  It seems like the complete spec is the design, but the doc should only be what's there (e.g. we might decide not to implement any of the unimplemented params until v2)

docs/source/rest_api/error_codes.rst
```diff
@@ -0,0 +1,165 @@
...
+       have been corrupted.
+   * - 27
+     - Unable to Fetch Statuses
+     - The validator should always return some status for every batch
+       requested. An unknown error caused statuses to be missing, and should be
+       reported.
+   * - 30
+     - Submitted Batches Invalid
+     - The submitted BatchList failed initial validation by the validator. It
+       may be poorly formed, or (most likely) has a bad signature.
```
peterschwarz: I would order this message such that the more likely item is first:

It may have a bad signature, or be poorly formed.

docs/source/rest_api/error_codes.rst
```diff
@@ -0,0 +1,165 @@
...
+       'min', 'max', or 'count' query parameters were invalid or out of range.
+   * - 62
+     - Invalid State Address
+     - The state address submitted was invalid. Returned when attempting to
+       fetch a particular "leaf" from the state tree. When fetching specific
+       state data, the full 70-character address must be used.
+   * - 66
+     - Id Query Invalid or Missing
+     - If using a GET request to fetch batch statuses, an 'id' query parameter
+       must be specified, with a comma separated list of at least one batch id.
```
peterschwarz: comma-separated

ojalatodd: ditto

docs/source/rest_api/endpoint_specs.rst
delventhalz: I can manually add a note. I haven't in the past because I didn't want to maintain it every time a change was made. At this point it's only a few query params though, so I suppose it isn't too onerous.

PR #390, opened 2017-04-27
PoET Block Verifier failure tests

nconde:


jjason:
Forgot to add - tested locally

PR #394, opened 2017-04-28
Add CLI reference for various commands

vaporos:
This is an initial pass at adding usage output to the CLI
reference for the various commands. It uses the command -h output
for accuracy.


vaporos:
https://build.sawtooth.me/job/Sawtooth-Hyperledger/job/sawtooth-core/job/PR-394/3/artifact/docs/build/html/cli.html

docs/source/cli/sawtooth.rst
```diff
@@ -0,0 +1,282 @@
...
+at ``/var/lib/sawtooth``.
+
+Usage
+-----
+
+.. literalinclude:: output/sawtooth_admin_genesis_usage.out
+   :language: console
+   :linenos:
+
+Arguments
```
aludvik: Should "Arguments" and "Output" be omitted since they are part of the automation?

vaporos: This was an attempt to bring forward the content that was previously there, so I was assuming another iteration which would figure out what content was actually appropriate for each section.

PR #400, opened 2017-05-01
Context manager unit tests

boydjohnson:
Implement several parallel-case context manager unit tests, but there will need to be further work in the context manager for these tests to pass.

### DIFF
--- a/validator/tests/unit3/test_context_manager/tests.py
+++ b/validator/tests/unit3/test_context_manager/tests.py
@@ -24,6 +24,93 @@ from sawtooth_validator.protobuf.state_delta_pb2 import StateChange
 
 class TestContextManager(unittest.TestCase):
 
+    def setUp(self):
+        self.database_of_record = dict_database.DictDatabase()
+        self.state_delta_store = StateDeltaStore(dict_database.DictDatabase())
+        self.context_manager = context_manager.ContextManager(
+            self.database_of_record, self.state_delta_store)
+        self.first_state_hash = self.context_manager.get_first_root()
+
+        # used for replicating state hash through direct merkle tree updates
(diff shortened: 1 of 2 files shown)


validator/tests/unit3/test_context_manager/tests.py
```diff
@@ -206,3 +290,531 @@ def test_squash_no_updates(self):
         self.assertEqual(
             [StateChange(address='aaa', value=b'xyz', type=StateChange.SET)],
             [c for c in changes])
+
+    @unittest.skip("Support for context manager parallelization")
```
peterschwarz: Will the production code for these skipped tests be in a future PR?

boydjohnson: Not in this sprint, I don't think, but I made a jira story for making them pass.

validator/tests/unit3/test_context_manager/tests.py
```diff
@@ -15,15 +15,163 @@
...
+        # used for replicating state hash through direct merkle tree updates
+        self.database_results = dict_database.DictDatabase()
+
+    def tearDown(self):
+        self.context_manager.stop()
+
+    def _setup_context(self):
+        # 1) Create transaction data
+        first_transaction = {'inputs': ['aaaa', 'bbbb', 'cccc'],
+                             'outputs': ['llaa', 'aall', 'nnnn']}
```
jsmitchell: any non hex char (0-9,a-f) addresses will be invalid someday

boydjohnson: It might be nice to clean up these addresses now, since you are working this sprint on enforcing address characteristics. I'll update the PR to only use valid addresses.

PR #402, opened 2017-05-01
Handle receiving requests before response

aludvik:
When registering, some extra logic is needed to push work requests onto
the work queue before registration is confirmed.


agunde406:
I am  seeing nondeterministic test failures on intkey smoke. Looks like it is about half the time.

aludvik:
Need to do some more work on this.

PR #404, opened 2017-05-02
 PoET Block Verifier Success Case

nconde:


nconde:
This is the same as the previous branch, i was having some issues rebasing with upstream/master after merging the first branch. So i created this new branch.

PR #405, opened 2017-05-02
Version pin pylint in Vagrant

rbuysse:

### DIFF
--- a/tools/bootstrap.d/30-python-lib-install.sh
+++ b/tools/bootstrap.d/30-python-lib-install.sh
@@ -2,4 +2,6 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


rbuysse:
@boydjohnson Correct. We've got a story to make run_lint compatible with the newer version of pylint so this is just a temporary workaround.

aludvik:
re: @boydjohnson's comment, run_lint can be run inside a docker container, so this isn't 100% necessary but if people are still running lint in vagrant, this is a quick fix.

rbuysse:
We still have instructions for using the Vagrant dev environment, and I know more than one member of the core-dev team is still using Vagrant.

PR #406, opened 2017-05-02
Publish State Change Deltas on Block Commit

peterschwarz:
On block commit, the `ChainController` calls out to a `StateDeltaProcessor` instance to publish state changes to its subscribers.

### DIFF
--- a/validator/sawtooth_validator/journal/chain.py
+++ b/validator/sawtooth_validator/journal/chain.py
@@ -425,6 +425,7 @@ class ChainController(object):
                  on_chain_updated,
                  squash_handler,
                  chain_id_manager,
+                 state_delta_processor,
                  identity_signing_key,
                  data_dir):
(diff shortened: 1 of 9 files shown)


validator/sawtooth_validator/journal/chain.py
```diff
@@ -567,6 +573,9 @@ def on_block_validated(self, commit_new_block, result):
 
                     LOGGER.info("Chain head updated to: %s", self._chain_head)
 
+                    # Publish the state deltas
+                    self._state_delta_processor.publish_deltas(new_block)
```
cianx: This looks like it could be time consuming, We should do it as the last step in this function. Getting descendant blocks processing and the BlockPublisher building a new block should be done first.

peterschwarz: Moved

PR #407, opened 2017-05-02
Improve Go SDK internal message routing

aludvik:
This improves the intermittent Go SDK failure so it only happens ~10% of the time (I tested locally and had 27/30 passes).

I believe the problem is that the internal message routing is sometimes still a little inefficient and there is occasionally a delay in processing the requests. Increasing the sleep from 1 to 3 seconds gave me 30/30 successes. I already created a story in my backlog to improve the routing logic so I am planning to revisit this issue when I get to that story.


peterschwarz:
It's likely that the validator should also be fixed to send the acknowledgement _before_ adding the TP to the collection of available processors.

PR #411, opened 2017-05-02
Fix python sdk stream class hiding errors

boydjohnson:
This PR causes the python sdk stream class to not hide errors, and exceptions on connecting to the validator will cause the class to raise an exception on instantiation.


boydjohnson:
@peterschwarz Can you say more about this? The transaction processor has a more complicated shutdown, that includes calling stream.close(), than the rest api.

boydjohnson:
@peterschwarz Not sure if this is what you meant, but I updated the xo and intkey python transaction processors to catch the previously hidden errors that are now not being hidden by stream.

PR #413, opened 2017-05-03
Build Go "Correctly" + Simplify Docker

aludvik:
Run TP Intkey Go from the binary instead of `go run`
Copy this binary into an "installed" docker image for testing.

In the course of doing the above, I decided to reduce the number of dev/build Dockerfiles in the repository to one for every language. I believe this simplifies maintenance and addition of components.


.gitignore
```diff
@@ -53,6 +53,7 @@
 /sdk/go/src/github.com/
 /sdk/go/bin/
 /sdk/go/pkg/
+/sdk/examples/intkey_go/src/sawtooth_intkey/sawtooth_intkey
```
peterschwarz: Is this a generated artifact?  Should it go into a `int_key_go/bin` directory?

aludvik:
Incorporated @peterschwarz's suggestion

agunde406:
Tested locally

PR #415, opened 2017-05-04
Introduce a start for Burrow EVM as Sawtooth Transaction Processor

benjaminbollen:
1. Introduces a copy of the "minimal" packages for the Ethereum VM to be connected as a TP:
   - `burrow/common/math/integral`
   - `burrow/common/sanity`
   - `burrow/evm/abi`
   - `burrow/evm/opcodes`
   - `burrow/evm/sha3`
   - `burrow/evm`
   - `burrow/permission/types`
   - `burrow/version`
   - `burrow/word256`
2. comments out the event system (and as a consequence the tests that rely here on the events)
3. all import paths are set to `"sawtooth_burrow_evm/..."`; but this likely does not work as the gopath is set to `/home/ubuntu/go:/project/sawtooth-core/sdk/go`
4. a second commit copies into `sawtooth_burrow_evm/vendor` three (optional) dependencies for reference:
   - `github.com/stretchr/testify` used in disabled test files
   - `golang.org/x/crypto/ripemd160` for native contract at `0x00..03 ripemd160Func`
   -  `gopkg.in/fatih/set.v0` for analysing jump destinations


benjaminbollen:
Open considertions on my part:
1. how to deal with GOPATH, and dependencies (aim for none / vendored-in / package-management ) ?
2. I put these packages under directory `burrow` with the intent to be able to unfork at later stage more easily;  I intend to make these packages top-level packages in burrow - where they are not already, ie `evm`
3. at it's most basic we now need to make an implementation for `AppSate` that translates the semantics of this interface to context operations in the handler;
see https://github.com/benjaminbollen/sawtooth-core/blob/burrow-evm/families/burrow_evm/src/sawtooth_burrow_evm/burrow/evm/types.go#L46-L58

jsmitchell:
Is there a good way to bring the vendor/ code in at build time instead of being in the repo?

benjaminbollen:
I removed the commit with the vendor directory

aludvik:
@jsmitchell We should consider using glide for go dependency management. Looks like @benjaminbollen's team is already using it and I think it will give us more control than the `go get` command we are currently using. I will research it a little more to verify we can make the swap.

families/burrow_evm/src/sawtooth_burrow_evm/burrow/common/sanity/sanity.go
```diff
@@ -0,0 +1,48 @@
...
+
+//--------------------------------------------------------------------------------------------------
+// panic wrappers
+// NOTE: [ben] Fail fast and fail hard, these are wrappers that point to code that needs
+// to be addressed, but simplify finding them in the code;
+
+// A panic resulting from a sanity check means there is a programmer error
+// and some gaurantee is not satisfied.
+func PanicSanity(v interface{}) {
+	panic(fmt.Sprintf("Paniced on a Sanity Check: %v", v))
```
boydjohnson: Tiny spelling error. Paniced -> Panicked. I don't think we need to change this before the PR goes in, but I was just noting it.

benjaminbollen:
updated import path to reflect:

`families/burrow_evm/src/burrow`
`families/burrow_evm/src/burrow/evm`
`families/burrow_evm/src/sawtooth_burrow_evm/handler`

kiko333:
I do not catch very well the way of the integration burrow-STL.
a) As other kind of family with own language of Smart Contract but using the STL-Validator, 
b) or only to interconection of Burrow to the POET -STL-Consensus

PR #416, opened 2017-05-04
Implement Connection Maintenance 

agunde406:
Connections that are used for getting candidate peers or an
unsuccessful peer request are removed. Also, send unregister
peer request to peers and disconnect messages to all connection
on validator shutdown.

Update gossip in core to have min and max peer connectivity set to 1:

self._gossip = Gossip(self._network,
                              public_uri=public_uri,
                              peering_mode=peering,
                              initial_join_endpoints=join_list,
                              initial_peer_endpoints=peer_list,
                              minimum_peer_connectivity=1,
                              maximum_peer_connectivity=1,
                              topology_check_frequency=1)

To test, start up the following validators one at a time (waiting for it to be completely started up) : 
validator -vvv --public-uri tcp://localhost:8800 --peering dynamic

validator --network-endpoint tcp://0.0.0.0:8801 --component-endpoint tcp://0.0.0.0:40001 --join tcp://localhost:8800 -vv --public-uri tcp://localhost:8801 --peering dynamic

validator --network-endpoint tcp://0.0.0.0:8802 --component-endpoint tcp://0.0.0.0:40002 --join tcp://localhost:8801 -vv --public-uri tcp://localhost:8802 --peering dynamic

The first two should be peered and the the third will be trying to peer unsuccessfully. You should see "Closing connection to .....". Control C one of the peered validators and you should see the other acknowledge an unregister peer request and a disconnect message.

### DIFF
--- a/validator/sawtooth_validator/gossip/gossip.py
+++ b/validator/sawtooth_validator/gossip/gossip.py
@@ -20,6 +20,7 @@ from threading import Thread
 from threading import Condition
 from functools import partial
 
+from sawtooth_validator.protobuf.network_pb2 import DisconnectMessage
 from sawtooth_validator.protobuf.network_pb2 import GossipMessage
 from sawtooth_validator.protobuf.network_pb2 import GossipBatchByBatchIdRequest
 from sawtooth_validator.protobuf.network_pb2 import \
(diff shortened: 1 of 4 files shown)


validator/sawtooth_validator/gossip/gossip.py
```diff
@@ -375,6 +387,22 @@ def run(self):
 
     def stop(self):
         self._stopped = True
+        for connection_id in self._connection_statuses:
+            try:
+                if self._connection_statuses[connection_id] == "closed":
+                    LOGGER.debug("Already sent disconnet message to %s",
```
jsmitchell: s/disconnet/disconnect

jsmitchell:
Tested locally

validator/sawtooth_validator/gossip/gossip.py
```diff
@@ -375,6 +387,22 @@ def run(self):
 
     def stop(self):
         self._stopped = True
+        for connection_id in self._connection_statuses:
+            try:
+                if self._connection_statuses[connection_id] == "closed":
+                    LOGGER.debug("Already sent disconnect message to %s",
```
peterschwarz: Since this is the only place that we can mark a connection as closed, this log message doesn't seem particularly useful.

agunde406:
rebasing before merge

PR #418, opened 2017-05-05
Graceful shutdown Go

aludvik:
To test locally, in vagrant do:
```
$ build_go
$ tp_intkey_go -v
```
Then start up a validator and run intkey workload against it.
Then you can press Ctrl+C in the tp_intkey_go window and restart tp_intkey_go to your hearts content.

### DIFF
--- a/sdk/examples/intkey_go/src/sawtooth_intkey/main.go
+++ b/sdk/examples/intkey_go/src/sawtooth_intkey/main.go
@@ -22,6 +22,7 @@ import (
 	intkey "sawtooth_intkey/handler"
 	"sawtooth_sdk/logging"
 	"sawtooth_sdk/processor"
+	"syscall"
 )
 
 func main() {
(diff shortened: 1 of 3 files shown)


sdk/examples/intkey_go/src/sawtooth_intkey/main.go
```diff
@@ -51,6 +52,7 @@ func main() {
 	handler := intkey.NewIntkeyHandler(prefix)
 	processor := processor.NewTransactionProcessor(endpoint)
 	processor.AddHandler(handler)
+	processor.ShutdownOnSignal(syscall.SIGINT, syscall.SIGTERM)
```
peterschwarz: Is there a reason that this can't be contained within the TransactionProcessor class?

aludvik: Technically there isn't. I wasn't sure whether it should or not and am open to hearing other viewpoints.

My thought process was: Since this is an SDK it should be general and limit the assumptions made about how it is used. That includes leaving it up to the main package/function to decided how and when to shutdown. By making the Shutdown() function part of the API and making it easy to setup signal handling with ShutdownOnSignal(), the API allows the following use case while still making it easy to handle SIGINT in the simple case:
1. Starting more than one transaction processor in a single process. This would require a custom signal handler which would call Shutdown() on all the processors running.
2. Not shutting down on Ctrl+C or perhaps shutting down on reading from a channel or some other event.
3. Shutting down on signals other than or in addition to SIGINT

PR #419, opened 2017-05-05
xo fixes

vaporos:
Two minor fixes to xo implementation.

### DIFF
--- a/sdk/examples/xo_python/sawtooth_xo/xo_client.py
+++ b/sdk/examples/xo_python/sawtooth_xo/xo_client.py
@@ -16,6 +16,7 @@
 import hashlib
 import base64
 import requests
+import time
 import yaml
 
 import sawtooth_signing.secp256k1_signer as signing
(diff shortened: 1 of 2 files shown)


sdk/examples/xo_python/sawtooth_xo/xo_client.py
```diff
@@ -126,7 +127,8 @@ def _send_xo_txn(self, name, action, space=""):
             dependencies=[],
             payload_encoding="csv-utf8",
             payload_sha512=_sha512(payload),
-            batcher_pubkey=self._public_key
+            batcher_pubkey=self._public_key,
+            nonce=time.time().hex().encode()
```
boydjohnson: .encode() on a hex str, will make it bytes, which isn't what the proto definition is. I would expect it to work in most cases, but probably should have it be the right type.

boydjohnson: time.time().hex().encode() is what we have for intkey, so I'm sure it works. I can't remember if this is a python2/python3 difference with protobuf, or the qualities of a nonce only matter that you make it somewhat random.

vaporos: Ok, if we want to change the approach (I did copy this from intkey), then let's create a JIRA for it.

PR #421, opened 2017-05-05
Add IntegerKey transaction family specification

dplumb94:

### DIFF
--- a/docs/source/transaction_family_specifications/integerkey_transaction_family.rst
+++ b/docs/source/transaction_family_specifications/integerkey_transaction_family.rst
@@ -1,0 +1,139 @@
+***************************************************
+IntegerKey Transaction Family Specification
+***************************************************
+
+Overview
+=========
+The IntegerKey transaction family allows users to set, increment, and decrement
+the value of entries stored in a state dictionary.
+
+An IntegerKey family transaction request is defined by the following values:
+
+* A verb which describes what action the transaction takes 
+* A name of the entry that is to be set or changed
+* A value by which the entry will be set or changed
+
+The 'set' verb is used to create new entries. The initial value of the entry will be 
+set to the value specified in the transaction request. The 'inc' and 'dec' verbs are 
+used to change the value of existing entries in the state dictionary.
+
+State
+=====
+This section describes in detail how IntegerKey transaction information is stored 
+and addressed.
+
+The address values of IntegerKey transaction family entries are stored in state as
+a CBOR encoded dictionary, with the key being the *Name* and the value being an integer
+*Value*.
+
+.. code-block:: python
+
+    cbor.dumps({
+        # The name of the entry : The current value of the entry
+        'Name':'Value'
+    })
+
+\ *Names* and *Values* must conform to the following rules:
+
+* Valid *Names* are utf-8 encoded strings with a maximum of 20 characters
+* Valid *Values* must be integers in the range of 0 through 2\ :sup:`32` - 1 (32-bit unsigned int)
+
+Addressing
+----------
+IntegerKey data is stored in the state dictionary using addresses which are generated from the IntegerKey namespace prefix and the unique name of the IntegerKey entry. Addresses will adhere to the following format:
+
+- Addresses must be a 70 character hexadecimal string
+- The first 6 characters of the address are the first 6 characters of a sha512 hash of the IntegerKey namespace prefix: "intkey"
+- The following 64 characters of the address are the last 64 characters of a sha512 hash of the entry *Name*
+
+For example, an IntegerKey address could be generated as follows:
+
+.. code-block:: pycon
+
+    >>> hashlib.sha512('intkey'.encode('utf-8')).hexdigest()[0:6] + hashlib.sha512('name'.encode('utf-8')).hexdigest()[-64:]
+    '1cf126cc488cca4cc3565a876f6040f8b73a7b92475be1d0b1bc453f6140fba7183b9a'
+
+.. note:: Due to the possibility of hash collisions, there may be multiple *Names* in the state dictionary with the same address.
+
+Transaction Payload
+===================
+
+IntegerKey transaction request payloads are defined by the following CBOR data format:
+
+.. code-block:: python
+
+    cbor.dumps({
+        # Describes the action the transaction takes, either 'set', 'inc', or 'dec'
+        'Verb': 'verb',
+
+        # The variable name of the entry which is to be modified
+        'Name': 'name',
+
+        # The amount to set, increment, or decrement
+        'Value': 1234,
+    })
+
+Transaction Header
+==================
+
+Inputs and Outputs
+------------------
+
+The inputs for IntegerKey family transactions must include:
+
+* Address of the *Name* being changed or added
+
+
+The outputs for IntegerKey family transactions must include:
+
+* Address of the *Name* being changed or added
+
+Dependencies
+------------
+* A list of transaction *header_signatures* which are required dependencies and must be processed prior to processing this transaction
+
+.. note:: For example, 'inc' and 'dec' transactions must list the initial 'set' transaction for the entry. If an 'inc' or 'dec' transaction is ordered before the corresponding 'set' transaction (without listing the 'set' transaction as a dependency), they will be considered invalid (because *Name* will not exist when they are processed).
+
+Family
+------
+- family_name: "intkey"
+- family_version: "1.0"
+
+Encoding
+--------
+- payload_encoding: "application/cbor"
+
+.. note:: The CBOR encoding map used by IntegerKey is a definite map. For example, a transaction payload is encoded as follows:
+
+    .. code-block:: pycon
+
+        >>> cbor.dumps({'Verb':'verb', 'Name':'name', 'Value':1234})
+        b'\xa3dVerbdverbdNamednameeValue\x19\x04\xd2'
+
+    CBOR Specification: `RFC 7049 - Concise Binary Object Representation (CBOR) <https://tools.ietf.org/html/rfc7049>`_
+
+Execution
+=========
+
+The IntegerKey transaction processor receives a transaction request and a state dictionary.
+
+If the payload of the transaction request is empty, the transaction is invalid.
+
+The address for the transaction is generated using the algorithm stated in the Addressing 
+section of this document. If an encoding error occurs, the transaction is invalid.
+
+The transaction request *Verb* \, *Name*\ , and *Value* are checked. If any of these values are 
+empty, the transaction is invalid. *Verb* must be either 'set', 'inc', or 'dec'. 
+*Name* must be a utf-8 encoded string with a maximum of 20 characters. *Value* must be 
+a 32-bit unsigned integer. If any of these checks fail, the transaction is invalid.
+
+If the *Verb* is 'set', the state dictionary is checked to determine if the *Name* associated with the 
+transaction request already exists. If it does already exist, the transaction is invalid. 
+Otherwise the *Name* and *Value* are stored as a new entry in the state dictionary.
+
+If the *Verb* is 'inc', the *Name* specified by the transaction request is checked determine 
+if the entry exists in the state dictionary. If the *Name* does not exist in the state dictionary, 
+it is an invalid transaction. Otherwise, we attempt to increment the *Value* in the state dictionary by the *Value* specified in the transaction request. If this incrementation would result in a value outside the range of 0 through 2\ :sup:`32` - 1 it is considered an invalid transaction. Otherwise, the *Value* in the state dictionary is incremented.
+
+If the *Verb* is 'dec', the *Name* specified by the transaction request is checked determine 
+if the entry exists in the state dictionary. If the *Name* does not exist in the state dictionary, it is an invalid transaction. Otherwise, we attempt to decrement the *Value* in the state dictionary by the *Value* specified in the transaction request. If this decrementation would result in a value outside the range of 0 through 2\ :sup:`32` - 1, it is considered an invalid transaction. Otherwise, the *Value* in the state dictionary is decremented.
--- a/docs/source/transaction_family_specifications.rst
+++ b/docs/source/transaction_family_specifications.rst
@@ -3,7 +3,8 @@ Transaction Family Specifications
 *********************************
 
 .. toctree::
-   :maxdepth: 2
+   :maxdepth: 3
 
    transaction_family_specifications/config_transaction_family.rst
+   transaction_family_specifications/integerkey_transaction_family.rst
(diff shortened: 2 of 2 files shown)


docs/source/transaction_family_specifications/integerkey_transaction_family.rst
```diff
@@ -0,0 +1,134 @@
...
+* Address of the *Name* being changed or added
+
+
+The outputs for IntegerKey family transactions must include:
+
+* Address of the *Name* being changed or added
+
+Dependencies
+------------
+* A list of transaction *header_signatures* which are required dependencies and must be processed prior to processing this transaction
```
jsmitchell: You might want to add "For example, 'inc' and 'dec' transactions might want to list the initial 'set' transaction for the *Name*. If 'inc' or 'dec' transactions are ordered before the corresponding 'set' transaction (and don't list the 'set' transaction as a dependency, they will be considered invalid (because *Name* will not exist when they are processed)."

docs/source/transaction_family_specifications/integerkey_transaction_family.rst
```diff
@@ -0,0 +1,134 @@
...
+
+Execution
+=========
+
+The IntegerKey transaction processor receives a transaction request and a state.
+If the payload of the transaction request is empty, the transaction is invalid.
+
+The address for the transaction is generated using the algorithm stated in the Addressing 
+section of this document. If an encoding error occurs, it is considered an internal 
+error.
```
jsmitchell: "it is considered an internal error, and the transaction is invalid."

docs/source/transaction_family_specifications/integerkey_transaction_family.rst
```diff
@@ -0,0 +1,134 @@
...
+    })
+
+\ *Names* and *Values* must conform to the following rules:
+
+* Valid *Names* are utf-8 encoded strings with a maximum of 20 characters
+* Valid *Values* must be integers in the range of 0 through 2\ :sup:`32` − 1 (32-bit unsigned int)
+
+Addressing
+----------
+When an IntegerKey family transaction entry is stored, it can be accessed by addressing using
```
jsmitchell: Maybe reword this to "IntegerKey data is stored in state using addresses which are generated from the intkey namespace prefix and the unique *Name* of the intkey variable. These addresses are generated using the following algorithm:"

docs/source/transaction_family_specifications/integerkey_transaction_family.rst
```diff
@@ -0,0 +1,134 @@
...
+When an IntegerKey family transaction entry is stored, it can be accessed by addressing using 
+the following algorithm:
+
+The address of an IntegerKey family transaction entry must be a 70 character hexadecimal string. 
+Addresses for the IntegerKey transaction family are generated by adding the last 64 characters 
+of the sha512 hash of the entry *Name* to the first 6 characters of a sha512 hash of the 
+IntegerKey namespace (the IntegerKey namespace is the string "intkey"). For example, an 
+address could be generated like this:
+
+.. code-block:: pycon
```
ojalatodd: typo: pycon ==> python

docs/source/transaction_family_specifications/integerkey_transaction_family.rst
ojalatodd: Agree with @jsmitchell  on  a little example here.

docs/source/transaction_family_specifications/integerkey_transaction_family.rst
boydjohnson: @jsmitchell, So with InternalErrors, they are currently being treated as if the txn is invalid, but according to @vaporos that is not the desired intent. After this coming sprint they won't be the same as an invalid transaction. They should be used in situations where the txn handler can't determine the validity of the txn. @agunde406  has a story in the upcoming sprint to change the behavior on the validator side to keep trying indefinitely (we may have to come up with a better solution than this) the transaction with a transaction processor if it comes back with InternalError. We should be pretty careful where we use InternalError.

jsmitchell: In that case, it sounds like we should strike any language in this doc referring to InternalError, and just say "the transaction is invalid".

docs/source/transaction_family_specifications/integerkey_transaction_family.rst
```diff
@@ -0,0 +1,134 @@
...
+
+    >>> hashlib.sha512('intkey'.encode('utf-8')).hexdigest()[0:6] + hashlib.sha512('eBuPof'.encode('utf-8')).hexdigest()[:64]
```
boydjohnson: Up above it says 'last 64 characters of the sha512 hash...', but this code shows the __first__ 64 characters of the (hex digest of the) sha512 hash being taken. '[:64]' is the first 64 characters, while '[-64:]' would be the last 64 characters.

docs/source/transaction_family_specifications/integerkey_transaction_family.rst
```diff
@@ -0,0 +1,134 @@
...
+The 'set' verb is used to create new entries. The initial value of the entry will be 
+set to the value specified in the transaction request. The 'inc' and 'dec' verbs are 
+used to change the value of existing entries in state.
+
+State
+=====
+This section describes in detail how IntegerKey transaction information is stored 
+and addressed.
+
+The address values of IntegerKey transaction family entries are stored in state as
```
boydjohnson: I would mention, maybe as a note or an aside, that there could be multiple Names in the state dictionary for a particular address. This allows for the possibility of hash collisions which may be *possible*. Or we could do away with this requirement that we allow for the possibility of hash collisions, and mention that in the spec. Either way it should be mentioned in the spec.

docs/source/transaction_family_specifications/integerkey_transaction_family.rst
```diff
@@ -0,0 +1,134 @@
...
+a cbor encoded dictionary, with the key being the *Name* and the value being an integer 
+*Value*.
+
+.. code-block:: python
+
+    cbor({
```
boydjohnson: So the cbor specification is making more sense to me. The first type of map is described this way: ```0xa0..0xb7      | map (0x00..0x17 pairs of data items follow) ```. So that is the first hex representation of the initial bytes for a map that has 23 or less key-value pairs in a map. So when the maps have 23 or less pairs, this would specify the format of the bytes stored in state. The other map specifications in cbor have to do with more pairs of items, and then also an indefinite amount.

boydjohnson: ^ wasn't a good comment, but was more information. I would reference the cbor specification at https://tools.ietf.org/html/rfc7049 and say that we are using a definite map, with that snippet of code up above, and maybe give examples of possible bytes stored in state. So using a python3 repl, b'\xa1cfoo\x13' = cbor.dumps({'foo': 13})

jsmitchell:
@dplumb94 note that you are getting a docs build error on this.

```
! Package inputenc Error: Unicode char − (U+2212)
(inputenc)                not set up for use with LaTeX.

See the inputenc package documentation for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.6364 ...e range of 0 through 2$^{\text{32}}$ −
                                                   1 (32-bit unsigned int)
? 
! Emergency stop.
 ...                                              
                                                  
l.6364 ...e range of 0 through 2$^{\text{32}}$ −
                                                   1 (32-bit unsigned int)
!  ==> Fatal error occurred, no output PDF file produced!```

You can test this locally in your vagrant vm by making sure that `sudo tools/plugins/install_latex.sh` has been run and then doing a `make latexpdf` in sawtooth-core/docs/

File: docs/source/transaction_family_specifications/integerkey_transaction_family.rst (hunk 3 above)
Comment (peterschwarz) [inline docs/source/transaction_family_specifications/integerkey_transaction_family.rst]: Looks like the build failure is due to "2\ :sup:`32`"

File: docs/source/transaction_family_specifications/integerkey_transaction_family.rst [hunk 8]
```diff
@@ -0,0 +1,134 @@
...
+If the *Verb* is 'inc', the *Name* specified by the transaction request is checked determine 
+if the entry exists in state. If the *Name* does not exist in state, it is an invalid transaction. 
+Otherwise, we attempt to increment the *Value* in state by the *Value* specified in the transaction 
+request. If this incrementation would result in a value outside the range of 0 through 2\ :sup:`32` − 1 
+it is considered an invalid transaction. Otherwise, the *Value* in state is incremented.
+
+If the *Verb* is 'dec', the *Name* specified by the transaction request is checked determine 
+if the entry exists in state. If the *Name* does not exist in state, it is an invalid transaction. 
+Otherwise, we attempt to decrement the *Value* in state by the *Value* specified in the transaction 
+request. If this decrementation would result in a value outside the range of 0 through 2\ :sup:`32` − 1,
```
peterschwarz: Another instance of "2\ :sup:`32`" that is could be causing an issue

docs/source/transaction_family_specifications/integerkey_transaction_family.rst
```diff
@@ -0,0 +1,134 @@
...
+a 32-bit unsigned integer. If any of these checks fail, the transaction is invalid.
+
+If the *Verb* is 'set', the state is checked to determine if the *Name* associated with the 
+transaction request already exists. If it does already exist, the transaction is invalid. 
+Otherwise the *Name* and *Value* are stored as a new entry in state.
+
+If the *Verb* is 'inc', the *Name* specified by the transaction request is checked determine 
+if the entry exists in state. If the *Name* does not exist in state, it is an invalid transaction. 
+Otherwise, we attempt to increment the *Value* in state by the *Value* specified in the transaction 
+request. If this incrementation would result in a value outside the range of 0 through 2\ :sup:`32` − 1
```
peterschwarz: Another instance of "2\ :sup:`32`" that is could be causing an issue

docs/source/transaction_family_specifications/integerkey_transaction_family.rst
```diff
@@ -0,0 +1,133 @@
...
+set to the value specified in the transaction request. The 'inc' and 'dec' verbs are 
+used to change the value of existing entries in the state dictionary.
+
+State
+=====
+This section describes in detail how IntegerKey transaction information is stored 
+and addressed.
+
+The address values of IntegerKey transaction family entries are stored in state as
+a cbor encoded dictionary, with the key being the *Name* and the value being an integer
```
peterschwarz: In text, outside of the example, cbor should be all caps: CBOR

docs/source/transaction_family_specifications/integerkey_transaction_family.rst
```diff
@@ -0,0 +1,133 @@
...
+* Valid *Names* are utf-8 encoded strings with a maximum of 20 characters
+* Valid *Values* must be integers in the range of 0 through 2\ :sup:`32` - 1 (32-bit unsigned int)
+
+Addressing
+----------
+IntegerKey data is stored in the state dictionary using addresses which are generated from the IntegerKey namespace prefix and the unique name of the IntegerKey entry. These addresses are generated using the following algorithm:
+
+.. code-block:: pycon
+
+    >>> hashlib.sha512('intkey'.encode('utf-8')).hexdigest()[0:6] + hashlib.sha512('name'.encode('utf-8')).hexdigest()[-64:]
```
nick-drozd: Should that `[-64:]` be `[64:]`? `hashlib.sha512('name'.encode('utf-8')).hexdigest()` is 128 characters, so it works out the same, but I'm not sure which is intended. It might be helpful to include an English description of the algorithm.

docs/source/transaction_family_specifications/integerkey_transaction_family.rst
```diff
@@ -0,0 +1,133 @@
...
+
+Transaction Payload
+===================
+
+IntegerKey transaction request payloads are defined by the following CBOR data format:
+
+.. code-block:: python
+
+    cbor.dumps({
+        // Describes the action the transaction takes, either 'set', 'inc', or 'dec'
```
nick-drozd: Python comments use `#`, not `//`.

docs/source/transaction_family_specifications/integerkey_transaction_family.rst
```diff
@@ -0,0 +1,133 @@
...
+
+    cbor.dumps({
+        // The name of the entry : The current value of the entry
+        'Name':'Value'
+    })
+
+\ *Names* and *Values* must conform to the following rules:
+
+* Valid *Names* are utf-8 encoded strings with a maximum of 20 characters
+* Valid *Values* must be integers in the range of 0 through 2\ :sup:`32` - 1 (32-bit unsigned int)
```
nick-drozd: Are there any rules on how an intkey TP should handle floats? More generally, should a TP reject a non-integer value outright, or can it convert it somehow?

docs/source/transaction_family_specifications/integerkey_transaction_family.rst
dplumb94: `[-64:]` is correct because we are taking the final 64 characters from the hash. I originally had an English description here but I took it out. I can add it back in if you think that would clarify matters.

nick-drozd: I can never remember anything but the most basic Python string slicing syntax, so I think the English description should be included. I was under the impression that we were taking the **first** 64 characters of the hash, and I therefore thought that `[-64:]` was either a typo or a fancy trick that yielded the first 64 characters (but maybe that's just me?).

docs/source/transaction_family_specifications/integerkey_transaction_family.rst
```diff
@@ -0,0 +1,133 @@
...
+.. code-block:: python
+
+    cbor.dumps({
+        // The name of the entry : The current value of the entry
+        'Name':'Value'
+    })
+
+\ *Names* and *Values* must conform to the following rules:
+
+* Valid *Names* are utf-8 encoded strings with a maximum of 20 characters
```
dplumb94: This description contains the only restrictions for names.

docs/source/transaction_family_specifications/integerkey_transaction_family.rst
dplumb94: In the execution section, we mention that the value must be a 32-bit unsigned int or the transaction is invalid.

docs/source/transaction_family_specifications/integerkey_transaction_family.rst
dplumb94: I'm guessing that many folks have a similar recollection of Python string slicing. I'll add a concise English description.

PR #426, opened 2017-05-08
Replace jackson-dataformat-cbor with co.nstant.in.cobr in Intkey Java

agunde406:
This change allows us to encode the intkey state HashMap in
the correct format to work with the python intkey and javascript
intkey examples. Jackson was formatting it to be an indefinite
map instead of a definite map. This caused the state root hash
to be different on validation when running a mixed network.

To test:
build_all
start up a validator and connect both a tp_intkey_python -vv and a tp_intkey_java
run intkey workload

You should see both transaction processors receiving transactions and blocks being build.

### DIFF
--- a/sdk/examples/intkey_java/IntegerKeyHandler.java
+++ b/sdk/examples/intkey_java/IntegerKeyHandler.java
@@ -16,8 +16,11 @@ package sawtooth.examples.intkey;
 
 import com.google.protobuf.ByteString;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.dataformat.cbor.CBORFactory;
+import co.nstant.in.cbor.CborBuilder;
+import co.nstant.in.cbor.CborDecoder;
+import co.nstant.in.cbor.CborEncoder;
+import co.nstant.in.cbor.CborException;
+import co.nstant.in.cbor.model.DataItem;
 
 import sawtooth.sdk.processor.State;
 import sawtooth.sdk.processor.TransactionHandler;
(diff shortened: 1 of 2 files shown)


sdk/examples/intkey_java/IntegerKeyHandler.java
```diff
@@ -80,6 +82,60 @@ public String getVersion() {
     return namespaces;
   }
 
+  /**
+   * Helper function to decode the Payload of a transactions.
```
jsmitchell: s/transactions/transaction/

sdk/examples/intkey_java/IntegerKeyHandler.java
```diff
@@ -80,6 +82,60 @@ public String getVersion() {
...
+          data.get(keys[i]).toString());
+    }
+    return result;
+  }
+
+  /**
+   * Helper function to decode State retrieved from the address of the name.
+   * Convert the co.nstant.in.cbor.model.Map to a HashMap.
+   */
+  public  Map decodeState(byte[] bytes) throws CborException {
```
jsmitchell: remove extra space between "public  Map"

jsmitchell: This function looks identical to decodePayload(). If this is desired, maybe we could collapse it down to decodeCbor() or something?

peterschwarz: Should return `Map<String, Integer>`

This would show that they are separate functions.

sdk/examples/intkey_java/IntegerKeyHandler.java
```diff
@@ -80,6 +82,60 @@ public String getVersion() {
     return namespaces;
   }
 
+  /**
+   * Helper function to decode the Payload of a transactions.
+   * Convert the co.nstant.in.cbor.model.Map to a HashMap.
+   */
+  public Map decodePayload(byte[] bytes) throws CborException {
```
peterschwarz: Should return `Map<String, String>`

PR #432, opened 2017-05-09
Address and Namespace Document

ojalatodd:
As part of the App Dev Guide, created an Addressing
and Namespacing doc.

Put it into RST format and placed in App Dev Guide.


docs/source/app_developers_guide/address_and_namespace.rst
```diff
@@ -0,0 +1,147 @@
...
+--------------
+
+This is a simple transaction family that stores key value pairs, e.g. “x=1”,
+or “y=2”, similar to the included intkey transaction family.
+
+The transaction family prefix is:
+
+.. code-block:: python
+
+	hashlib.sha256("intkey".encode('utf-8')).hexdigest()[:6]
```
boydjohnson: This says it is an example of a transaction family like intkey and uses a hash of 'intkey' to find the prefix, but intkey uses sha-512 hash instead of sha-256. This might confuse people. I suggest either the example doesn't reference intkey, and the prefix could be the hash of 'simple-example', or the hash algorithm for address construction that intkey uses, is used.

ojalatodd: Changed to 'simple-example' to avoid confusion with actual intkey.

PR #434, opened 2017-05-09
Fix demo docker compose file to work with 0.8.2

rbuysse:

### DIFF
--- a/docker/compose/sawtooth-demo.yaml
+++ b/docker/compose/sawtooth-demo.yaml
@@ -18,7 +18,7 @@ version: "2.1"
 services:
 
   tp_config:
-    image: hyperledger/sawtooth-tp_config:unstable
+    image: hyperledger/sawtooth-tp_config:latest
     expose:
       - 40000
     depends_on:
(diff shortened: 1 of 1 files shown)


boydjohnson:
Since you fixed another bug (the rest api was listening on the default localhost instead of the container's ip address), I think the client container should depend on the rest api instead of the validator. Our cli tools for submitting batches used to all send them directly to the validator, but now all use the rest api, so starting up as it is currently wouldn't always allow someone to do 'intkey load --url rest_api' in the client container.

rbuysse:
@boydjohnson I made that change in a fixup commit. Good catch!

docker/compose/sawtooth-demo.yaml
```diff
@@ -18,52 +18,58 @@ version: "2.1"
...
       - 40000
+    ports:
+      - "40000:40000"
     # start the validator with an empty genesis batch
     entrypoint: "bash -c \"\
         sawtooth admin keygen && \
         sawtooth admin genesis && \
-        validator -vv\""
+        validator -vv \
+          --public-uri tcp://localhost:8800 \
```
boydjohnson: The --public-uri is where other validators can connect to this validator. Since this is a single validator, localhost will technically work, but it would probably be a good idea to use tcp://validator:8800 so that people don't get confused.

PR #435, opened 2017-05-09
Add proper logger to REST API

delventhalz:
Adds a logger to the REST API, and replaces the existing access logger with an expanded access logger that no longer uses `print` statements.

This PR does not close STL-69 (https://jira.hyperledger.org/browse/STL-69), but establishes the groundwork for a more robust logger which will.


agunde406:
Test locally

peterschwarz:
We should consider having the rest_api container instances in the various docker-compose files match the logging level of the other components.

delventhalz:
Added commit to up the logging level of all rest_api docker components

PR #437, opened 2017-05-10
Send "catch-up" State Delta Events on Subscribe

peterschwarz:
When a state delta export subscriber registers with a known block id that is older than the current chain head, send the subscriber all the events that have occurred in the interim.   This allows the subscriber to behave as if it received all the correct events.

Additionally:

* Add a Predecessor Iterator on block store

### DIFF
--- a/validator/sawtooth_validator/exceptions.py
+++ b/validator/sawtooth_validator/exceptions.py
@@ -65,3 +65,10 @@ class PeeringException(Exception):
     Indicates that a request to peer with this validator should not be allowed.
     """
     pass
+
+
+class PossibleForkDetectedError(Exception):
+    """Exception thrown when a possible fork has occurred while iterating
+    through the block store.
+    """
+    pass
--- a/validator/sawtooth_validator/journal/block_store.py
+++ b/validator/sawtooth_validator/journal/block_store.py
@@ -17,6 +17,9 @@ from time import time
 from threading import Condition
 # pylint: disable=no-name-in-module
 from collections.abc import MutableMapping
+
+from sawtooth_validator.exceptions import PossibleForkDetectedError
+from sawtooth_validator.journal.block_wrapper import NULL_BLOCK_IDENTIFIER
 from sawtooth_validator.journal.block_wrapper import BlockStatus
 from sawtooth_validator.journal.block_wrapper import BlockWrapper
 from sawtooth_validator.protobuf.batch_pb2 import BatchHeader
@@ -61,8 +64,7 @@ class BlockStore(MutableMapping):
         return x in self._block_store
 
     def __iter__(self):
-        # Required by abstract base class, but implementing is non-trivial
-        raise NotImplementedError('BlockStore is not iterable')
+        return _BlockPredecessorIterator(self, start_block=self.chain_head)
 
     def __len__(self):
         # Required by abstract base class, but implementing is non-trivial
@@ -116,6 +118,22 @@ class BlockStore(MutableMapping):
         Access to the underlying store dict.
         """
         return self._block_store
+
+    def get_predecessor_iter(self, starting_block=None):
+        """Returns an iterator that traverses blocks via their
+        previous_block_ids.
+
+        Args:
+            starting_block (:obj:`BlockWrapper`): the block from which
+                traversal begins
+
+        Returns:
+            An iterator
+        """
+        if not starting_block:
+            return _BlockPredecessorIterator(self, start_block=self.chain_head)
+        else:
+            return _BlockPredecessorIterator(self, start_block=starting_block)
 
     def wait_for_batch_commits(self, batch_ids=None, timeout=None):
         """Waits for a set of batch ids to be committed to the block chain,
@@ -249,3 +267,43 @@ class BlockStore(MutableMapping):
                     return txn
 
         raise ValueError('Transaction "%s" not in BlockStore', transaction_id)
+
+
+class _BlockPredecessorIterator(object):
+    """An Iterator for traversing blocks via a block's previous_block_id
+    """
+
+    def __init__(self, block_store, start_block):
+        """Iterates from a starting block, through its predecessors.
+
+        Args:
+            block_store (:obj:`BlockStore`): the block store, from which
+                the predecessors are found
+            start_block (:obj:`BlockWrapper`): the starting block, from which
+                the predecessors will be iterated over.
+        """
+        self._block_store = block_store
+        if start_block:
+            self._current_block_id = start_block.identifier
+        else:
+            self._current_block_id = None
+
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        if not self._current_block_id:
+            raise StopIteration()
+
+        try:
+            block = self._block_store[self._current_block_id]
+        except KeyError:
+            raise PossibleForkDetectedError(
+                'Block {} is no longer in the block store'.format(
+                    self._current_block_id[:8]))
+
+        self._current_block_id = block.header.previous_block_id
+        if self._current_block_id == NULL_BLOCK_IDENTIFIER:
+            self._current_block_id = None
+
+        return block
--- a/validator/sawtooth_validator/state/state_delta_processor.py
+++ b/validator/sawtooth_validator/state/state_delta_processor.py
@@ -15,6 +15,7 @@
 import logging
 from threading import Condition
 
+from sawtooth_validator.exceptions import PossibleForkDetectedError
 from sawtooth_validator.networking.dispatch import Handler
 from sawtooth_validator.networking.dispatch import HandlerResult
 from sawtooth_validator.networking.dispatch import HandlerStatus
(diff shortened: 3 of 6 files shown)


validator/sawtooth_validator/journal/block_store.py
```diff
@@ -249,3 +265,39 @@ def get_transaction(self, transaction_id):
...
 
         raise ValueError('Transaction "%s" not in BlockStore', transaction_id)
+
+
+class _BlockPredecessorIterator(object):
+    """An Iterator for traversing blocks via a block's previous_block_id
+    """
+
+    def __init__(self, block_store, start_block):
+        """Iterates from a starting block, through it's anscestors.
```
cintel: anscestors -> ancestors

validator/sawtooth_validator/journal/block_store.py
```diff
@@ -249,3 +265,39 @@ def get_transaction(self, transaction_id):
...
+
+        Args:
+            block_store (:obj:`BlockStore`): the block store to retrieve the
+                anscestors from
```
cintel: anscestors -> ancestors

validator/sawtooth_validator/state/state_delta_processor.py
```diff
@@ -119,6 +119,25 @@ def add_subscriber(
...
                      connection_id, address_prefix_filters)
 
+        if known_block_id != self._block_store.chain_head.identifier:
+            LOGGER.debug('Catching up Subscriber %s from %s to %s',
+                         connection_id,
+                         known_block_id,
+                         self._block_store.chain_head.identifier[:8])
+            catch_up_blocks = []
+
+            for block in self._block_store.get_predecessor_iter():
```
cintel: It looks like the _BlockPredecessorIterator will throw a stop iteration if the block store has changed status, like having switched forks and the predecessor is no longer in the BlockStore. How is that detected and handled here?

validator/tests/unit3/test_block_store/tests.py
```diff
@@ -0,0 +1,93 @@
...
+from sawtooth_validator.database.dict_database import DictDatabase
+from sawtooth_validator.journal.block_store import BlockStore
+from sawtooth_validator.journal.block_wrapper import NULL_BLOCK_IDENTIFIER
+from sawtooth_validator.journal.block_wrapper import BlockWrapper
+
+from sawtooth_validator.protobuf.block_pb2 import Block
+from sawtooth_validator.protobuf.block_pb2 import BlockHeader
+
+
+class BlockStorePredecessorIteratorTest(unittest.TestCase):
```
cintel: Should add a case where the BlockStore is updated during iteration, specifically with a predecessor that has been removed.

peterschwarz: Added.

peterschwarz:
Rebased on master for the shutdown test ignores

validator/sawtooth_validator/state/state_delta_processor.py
```diff
@@ -119,6 +120,37 @@ def add_subscriber(
...
 
+        if known_block_id != self._block_store.chain_head.identifier:
+            LOGGER.debug('Catching up Subscriber %s from %s to %s',
+                         connection_id,
+                         known_block_id,
+                         self._block_store.chain_head.identifier[:8])
+            catch_up_blocks = []
+
+            try:
+                for block in self._block_store.get_predecessor_iter():
```
delventhalz: Is there a reason not to just `for block in self._block_store`? That would be a bit more readable.

peterschwarz:
Rebased before merging

peterschwarz:
Added a new change that may require more review.

validator/sawtooth_validator/exceptions.py
```diff
@@ -65,3 +65,10 @@ class PeeringException(Exception):
     Indicates that a request to peer with this validator should not be allowed.
     """
     pass
+
+
+class PossibleForkDetectedError(Exception):
+    """Exception thrown when an possible fork has occurred while iterating
```
boydjohnson: an -> a

validator/sawtooth_validator/state/state_delta_processor.py
```diff
@@ -188,19 +236,41 @@ def handle(self, connection_id, message_content):
...
+                request.last_known_block_ids):
+            ack.status = ack.OK
+            return HandlerResult(
+                HandlerStatus.RETURN_AND_PASS,
+                message_out=ack,
+                message_type=self._msg_type)
+        else:
+            ack.status = ack.UNKNOWN_BLOCK
+            return HandlerResult(
+                HandlerStatus.DROP,
```
boydjohnson: HandlerStatus.DROP will stop processing on the message, and not send a message back, but since an ack is being sent back a HandlerStatus.RETURN is needed here. RETURN stops with that handler for that message, but sends a message back.

validator/tests/unit3/test_state_delta_processor/tests.py
```diff
@@ -76,28 +79,76 @@ def test_register_with_uknown_block_ids(self):
...
+
+        handler = StateDeltaAddSubscriberHandler(delta_processor)
+
+        request = RegisterStateDeltaSubscriberRequest(
+            last_known_block_ids=['a'],
+            address_prefixes=['000000']).SerializeToString()
+
+        response = handler.handle('test_conn_id', request)
+
+        self.assertEqual(HandlerStatus.DROP, response.status)
```
boydjohnson: Similar to my above comment, DROP -> RETURN.

peterschwarz: In this later case, there is no response and we don't want to process the request any more.

boydjohnson: Ok, thanks!

validator/tests/unit3/test_state_delta_processor/tests.py
```diff
@@ -125,22 +126,51 @@ def test_add_subscriber(self):
...
+
         delta_processor = StateDeltaProcessor(
             service=mock_service,
-            state_delta_store=Mock(),
+            state_delta_store=delta_store,
             block_store=block_tree_manager.block_store)
 
+        delta_store.save_state_deltas(
+            block_tree_manager.chain_head.state_root_hash,
+            [StateChange(address='deadbeefgenesis',
```
jsmitchell: this is an invalid address and may break on future changes

validator/tests/unit3/test_state_delta_processor/tests.py
```diff
@@ -125,22 +126,51 @@ def test_add_subscriber(self):
...
+        # test that it catches up, and receives the events from the chain head.
+        # In this case, it should just be once, as the chain head is the
+        # genesis block
+        mock_service.send.assert_called_with(
+            validator_pb2.Message.STATE_DELTA_EVENT,
+            StateDeltaEvent(
+                block_id=block_tree_manager.chain_head.identifier,
+                block_num=block_tree_manager.chain_head.block_num,
+                state_root_hash=block_tree_manager.chain_head.state_root_hash,
+                state_changes=[StateChange(address='deadbeefgenesis',
```
jsmitchell: same

validator/tests/unit3/test_state_delta_processor/tests.py
```diff
@@ -99,6 +101,54 @@ def test_register_with_uknown_block_ids(self):
...
+        request = RegisterStateDeltaSubscriberRequest(
+            last_known_block_ids=[block_tree_manager.chain_head.identifier],
+            address_prefixes=['0123456']).SerializeToString()
+
+        response = handler.handle('test_conn_id', request)
+        self.assertEqual(HandlerStatus.PASS, response.status)
+
+        self.assertEqual(['test_conn_id'], delta_processor.subscriber_ids)
+
+    def test_register_with_uknown_block_ids(self):
```
jsmitchell: s/uknown/unknown

validator/sawtooth_validator/state/state_delta_processor.py
peterschwarz: The reason I didn't use that, is I wasn't sure that using the predecessor iterator as the default iterator was necessarily the right path.   I wanted this use to be explicit, in case that changes.

validator/tests/unit3/test_state_delta_processor/tests.py
```diff
@@ -125,22 +176,51 @@ def test_add_subscriber(self):
...
+                block_id=block_tree_manager.chain_head.identifier,
+                block_num=block_tree_manager.chain_head.block_num,
+                state_root_hash=block_tree_manager.chain_head.state_root_hash,
+                state_changes=[StateChange(address='deadbeef0000000',
+                               value='my_genesis_value'.encode(),
+                               type=StateChange.SET)]
+            ).SerializeToString(),
+            connection_id='test_conn_id')
+
     def test_add_subscrber_known_block_id(self):
```
cintel: subscrber -> subscriber

PR #438, opened 2017-05-10
Disable shutdown smoke tests

peterschwarz:
Disable shutdown smoke tests until STL-120 is complete.  These test cause periodic failures due to the shutdown activities not being properly implemented.  While the argument could be made that we should keep things painful, so as to force ourselves to fix it, it makes the PR process (especially with external contributors) particularly painful.

See: https://jira.hyperledger.org/browse/STL-120


peterschwarz:
I disagree with that, @boydjohnson. First, commenting out the test leaves no record of the test in the build output.  Using `unittest.skip` gives us a line item for every skipped test. This is both a reminder and a public record of the test.  Commented tests get forgotten.

Secondly, skipping the test will still give some test failure feedback if the code changes, though not all, since the test methods won't be run, but it imports change, etc. This will keep the tests from getting to out of date.

PR #446, opened 2017-05-11
Update PoET Enclave Simulator to not Use kwargs for Configuration

jjason:
The legacy way in which the PoET enclave simulator determined configuration was by having a dictionary (aka, kwargs) passed in to its initialize function.  This PR replaces the kwargs with the location of the validator's configuration directory.  The enclave should use a `toml` file for any configuration and the file should be found in the configuration directory.

Changes:
* The consensus module block publisher, block verifier, and fork resolver interfaces has been updated to take a configuration directory parameter
* All code that touches these interfaces have been updated
* The PoET enclave simulator has been updated to use the config directory to locate its `toml` file (if it exists) and use it for configuration


jjason:
Rebasing before merge

PR #447, opened 2017-05-11
Add synchronization lock between the BlockPublisher and ChainControlle…

cintel:
…r to prevent ChainController(and encapsulaed Consensus algos) accessesing an updated blockstore.

### DIFF
--- a/validator/sawtooth_validator/journal/chain.py
+++ b/validator/sawtooth_validator/journal/chain.py
@@ -422,6 +422,7 @@ class ChainController(object):
                  state_view_factory,
                  executor,
                  transaction_executor,
+                 chain_head_lock,
                  on_chain_updated,
                  squash_handler,
                  chain_id_manager,
@@ -438,6 +439,12 @@ class ChainController(object):
              executor: The thread pool to process block validations.
              transaction_executor: The TransactionExecutor used to produce
              schedulers for batch validation.
+             chain_head_lock: Lock to hold while the chain head is being
+             updated, this prevents other components that depend on the chain
+             head and the BlockStore from having the BlockStore change under
+             them.  This lock is only for core Journal components
+             (BlockPublisher and ChainController), other components should
+             handle block not found errors from the BlockStore explicitly.
              on_chain_updated: The callback to call to notify the rest of the
              system the head block in the chain has been changed.
              squash_handler: a parameter passed when creating transaction
@@ -452,6 +459,7 @@ class ChainController(object):
             None
         """
         self._lock = RLock()
+        self._chain_head_lock = chain_head_lock
         self._block_cache = block_cache
         self._block_store = block_cache.block_store
         self._state_view_factory = state_view_factory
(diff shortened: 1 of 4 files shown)


cintel:
@vaporos The issue is the contents of the BlockStore. Queue updates to chain head does not solve the issue that the data structures are changed under the BlockPublisher. The BlockPublisher and encapsulated algorithms(consensus) expect the BlockStore to match the chain_head it is operating on.  The update to the BlockStore and the notification of the BlockPublisher need to be atomic.

peterschwarz:
It sounds like there's an intertwining of those members of Journal where there's a single concern (updating the chain) that's getting spread across to different modules. Is there way that there is single entity handling this atomic operation.  We have a lot of parameters that also start to need to be passed to everything - that seems like a code smell that says we might be missing an object somewhere.

vaporos:
It is logical that BlockStore matches chain head for the purposes of storage, as if we restart the validator, there is no point in having retained anything else.  However, the block publisher (or even a chain controller implementation which processes multiple potential chains heads in parallel) could reference in-memory (or cached) blocks which are separate from the chain head.  Ideally, the in-memory (or cached) blocks could be copy-on-write to facilitate read concurrency except when blocks are being purged (maybe due to lack of references to them).  This would avoid the need for the global lock, as multiple chains could co-exist at the same time with one of them being pointed to by chain head and stored in BlockStore.

cixce:
@vaporos @peterschwarz 
Do note this PR is essentially moving a acquisition of a lock up 2 lines of code to prevent a race condition on a data structure.  It is not a new functionality or change in the behavior of the interaction of the journal components. 

 ChainController is responsible for determining the current chain, which is specified by the chain-head. As part of this responsibility the ChainController is responsible for maintaining and updating the BlockStore.

The BlockStore is an expensive disk backed data structure that maintains a number of indexes in addition to the blocks in the store. Updates to it are atomic but iterators and other cached states are not aware of this.

The BlockPublisher role is to build a single block extending the current chain head. It is important to note that block publishing is a stateful activity that happens in the context of a current chain head, you cannot build or publish block with out that context. The BlockPublisher uses the BlockStore to determine if Batches or Transactions are already committed to the chain or not.  Consensus algorithms have access to the BlockStore, for instance PoET uses this for determining the local mean as well as other tests. 

Any work the BlockPublisher does not extending the current chain head works against consensus in the system as it will create forks and in particular forks that this validator does not beleive in, since it has already committed to a new chain head. 

It is important that the BlockStore remain consistant for the BlockPublisher and that the switch be made from the building on old chain head to building on the new chain head as quickly as possible.

So this PR expresses that by essentially moving the claiming of the lock on the BlockPublisher actvities up 2 lines to prevent the possibility of the BlockPublisher accessing the BlockStore during the chain-head update.

Now I could accomplish the same goal with the following change, which uses exactly the same locks just less efficiently. If that is less offensive I can update the PR.

self._notify_on_chain_updated(None) 
self._chain_head = new_block
# update the the block store to have the new chain
self._block_store.update_chain(result["new_chain"],
	result["cur_chain"])
LOGGER.info("Chain head updated to: %s", self._chain_head)
# tell the BlockPublisher else the chain is updated
self._notify_on_chain_updated(
	self._chain_head,
	result["committed_batches"],
	result["uncommitted_batches"])

cintel:
@vaporos @peterschwarz
Do note this PR is essentially moving a acquisition of a lock up 2 lines of code to prevent a race condition on a data structure. It is not a new functionality or change in the behavior of the interaction of the journal components.

ChainController is responsible for determining the current chain, which is specified by the chain-head. As part of this responsibility the ChainController is responsible for maintaining and updating the BlockStore.

The BlockStore is an expensive disk backed data structure that maintains a number of indexes in addition to the blocks in the store. Updates to it are atomic but iterators and other cached states are not aware of this.

The BlockPublisher role is to build a single block extending the current chain head. It is important to note that block publishing is a stateful activity that happens in the context of a current chain head, you cannot build or publish block with out that context. The BlockPublisher uses the BlockStore to determine if Batches or Transactions are already committed to the chain or not. Consensus algorithms have access to the BlockStore, for instance PoET uses this for determining the local mean as well as other tests.

Any work the BlockPublisher does not extending the current chain head works against consensus in the system as it will create forks and in particular forks that this validator does not beleive in, since it has already committed to a new chain head.

It is important that the BlockStore remain consistant for the BlockPublisher and that the switch be made from the building on old chain head to building on the new chain head as quickly as possible.

So this PR expresses that by essentially moving the claiming of the lock on the BlockPublisher actvities up 2 lines to prevent the possibility of the BlockPublisher accessing the BlockStore during the chain-head update.

Now I could accomplish the same goal with the following change, which uses exactly the same locks just less efficiently. If that is less offensive I can update the PR.
```python
self._notify_on_chain_updated(None) 
self._chain_head = new_block
# update the the block store to have the new chain
self._block_store.update_chain(result["new_chain"],
	result["cur_chain"])
LOGGER.info("Chain head updated to: %s", self._chain_head)
# tell the BlockPublisher else the chain is updated
self._notify_on_chain_updated(
	self._chain_head,
	result["committed_batches"],
	result["uncommitted_batches"])

```

vaporos:
By 'alternate fix', I meant one avoiding a shared lock.

PR #451, opened 2017-05-12
Add current serial scheduler test to parallel scheduler test

agunde406:
This includes consolidating the use of create_transaction and
create_batch, adding setUp and tearDown to TestParallelScheduler,
adding test to TestParallelScheduler that should also work from
TestSerialScheduler, and adding the previous context_id to
create_context in test_transaction_order_with_dependencies.

Two tests are currently failing.

### DIFF
--- a/validator/tests/unit3/test_scheduler/tests.py
+++ b/validator/tests/unit3/test_scheduler/tests.py
@@ -35,6 +35,7 @@ from sawtooth_validator.state.merkle import MerkleDatabase
 
 LOGGER = logging.getLogger(__name__)
 
+
 def create_transaction(name, private_key, public_key, inputs=None,
                         outputs=None, dependencies = None):
     payload = name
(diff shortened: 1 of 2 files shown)


boydjohnson:
I suggest that the test file should have the parallel scheduler and serial scheduler test classes right after one another, instead of having the predecessor tree tests in between.

This commit, aa93605a77e9466274edc789566761b7bb5edb34, from my yaml-based scheduler PR, refactors the txn and batch creation utilities. Since this PR refactors those same utilities, we can save ourselves some work if we refactor them in one way to decrease merge conflicts in the future.

I would suggest again that we are using the same test code to test the parallel scheduler and serial scheduler where we have determined that each scheduler implementation needs to pass those tests. That could be done with a joint test class that all scheduler implementations must pass with non-test methods that are the actual test code, that take a scheduler instance, and then test methods that create the scheduler instance and the pass it in to the non-test method with the current code. That way, when looking at an error, it would be possible to tell just from the test framework failure message which scheduler implementation failed, whether the test was one that was specific to the implementation, or whether it was general to all scheduler implementations.

validator/tests/unit3/test_scheduler/tests.py
```diff
@@ -35,18 +35,29 @@
 
 LOGGER = logging.getLogger(__name__)
 
-
-def create_transaction(name, private_key, public_key):
+def create_transaction(name, private_key, public_key, inputs=None,
```
boydjohnson: It would nice to have dependencies in the function signature, so it isn't alway [].

agunde406:
Updated to incorporate @boydjohnson suggestions

PR #457, opened 2017-05-12
Initial yaml-based scheduler tests

boydjohnson:
This PR implements a SchedulerTester that operates on yaml files with a particular data structure. A yaml file and associated scheduler test are included. The included test is a single block-single scheduler instance test.

Comments or suggestions on future features welcome.

### DIFF
--- a/validator/tests/unit3/test_scheduler/data/simple_scheduler_test.yaml
+++ b/validator/tests/unit3/test_scheduler/data/simple_scheduler_test.yaml
@@ -1,0 +1,166 @@
+# Copyright 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+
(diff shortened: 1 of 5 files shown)


validator/tests/unit3/test_scheduler/data/simple_scheduler_test.yaml
```diff
@@ -0,0 +1,166 @@
...
+      - ddtt:sha :
+      - ffff:sha :
+  -
+    inputs:
+      - dddd:sha
+    outputs:
+      - ssss:sha
+    addresses_to_set:
+      - ssss:sha :
+-
```
peterschwarz: Continue the pattern of commenting the batch-starts

validator/tests/unit3/test_scheduler/tests.py
```diff
@@ -31,11 +32,119 @@
...
+                always_persist=False)
+            tester = SchedulerTester(file_name,
+                                     serial_scheduler,
+                                     self._context_manager)
+            defined_batch_results_dict = tester.batch_results
+            batch_results = tester.run_scheduler()
+            self.assert_batch_validity(
+                defined_batch_results_dict,
+                batch_results)
+            self.assert_one_state_hash(batch_results=batch_results)
```
peterschwarz: Is there any way to add the expected results to the yaml file?  This might make it easier to define tests in an entirely declarative fashion.  Something to think about for a future PR with more tests.

boydjohnson: With how the tester currently works, a state_hash key can be added to a batch and would be used similar to the scheduler in the block validator. A assert_state_hash key that could be optionally added to the yaml, where appropriate, might be helpful.

boydjohnson:
I am going to make some changes so that I don't use a utils.py file, but leave the batch/txn creation functions in the tests.py file and make a yaml_scheduler_tester.py file for the SchedulerTester.

PR #458, opened 2017-05-15
Remove active key state class variable in PoET block publisher

jjason:
The PoET block publisher keeps track of its active PoET public key by storing it in a class variable so that it is persisted across creation of PoET block publisher objects.  While this works, it leads to headaches when running tests.  This update removes the class variable and puts it in the PoET key state store.


jjason:
Rebase before merge

PR #468, opened 2017-05-16
add yum notice to install guide

n1zyz:
Signed-off-by: n1zyz <mwagner@redhat.com>


ojalatodd:
@rbuysse ?

ojalatodd:
@dcmiddle Approving - we should make sure this feature gets into planning!

PR #479, opened 2017-05-17
Go Client SDK

aludvik:
```
$ ./bin/build_all -l go
$ ./bin/run_docker_test sdk/go/tests/unit_go_client.yaml -s go_client
```


sdk/go/src/sawtooth_sdk/client/encoder.go
```diff
@@ -0,0 +1,140 @@
...
+
+	header := &transaction_pb2.TransactionHeader{
+		BatcherPubkey:   Encode(self.pubkey, "hex"),
+		FamilyName:      self.familyName,
+		FamilyVersion:   self.familyVersion,
+		Inputs:          inputs,
+		Outputs:         outputs,
+		PayloadEncoding: self.payloadEncoding,
+		PayloadSha512:   hash,
+		SignerPubkey:    Encode(self.pubkey, "hex"),
```
aludvik: Need to add Nonce and Dependencies

peterschwarz: Is this a future PR item, or this PR?

sdk/go/src/sawtooth_sdk/client/encoder.go
```diff
@@ -0,0 +1,156 @@
...
+type Batch struct{ pb *batch_pb2.Batch }
+
+type TransactionEncoder struct {
+	privkey         []byte
+	pubkey          []byte
+	familyName      string
+	familyVersion   string
+	payloadEncoding string
+	addNonce        bool
+}
```
delventhalz: In the JS Client SDK, no decisions are made about what defaults are stored in the TransactionEncoder. So for example, even though `inputs` and `outputs` are likely to be different in each transaction, you have the option to set defaults if you like. Some possible use cases would be if wild cards are used to designate a whole subtree, or a Transaction Processor stores all of one user's data at one address.

Which is a long way of saying, I think you should be able to set defaults for `inputs` and `outputs`. There is more of a case for omitting `dependencies`, but at that point it is the odd-one-out, and I think just for consistency of the interface, it should be possible to set that too.

sdk/go/src/sawtooth_sdk/client/encoder.go
```diff
@@ -0,0 +1,156 @@
...
+type Transaction struct{ pb *transaction_pb2.Transaction }
+type Batch struct{ pb *batch_pb2.Batch }
+
+type TransactionEncoder struct {
+	privkey         []byte
+	pubkey          []byte
+	familyName      string
+	familyVersion   string
+	payloadEncoding string
+	addNonce        bool
```
delventhalz: Making auto-generated nonce's optional is a good idea. Do you need a boolean flag for it though? Why not just make it so if a nonce is set during transaction creation it is used, otherwise one is auto-generated?

sdk/go/src/sawtooth_sdk/client/encoder.go
```diff
@@ -0,0 +1,156 @@
...
+		familyName:      familyName,
+		familyVersion:   familyVersion,
+		payloadEncoding: payloadEncoding,
+		addNonce:        addNonce,
+	}
+}
+
+// NewTransaction Creates a new transaction and handles the construction and
+// signing of the transaction header.
+func (self *TransactionEncoder) NewTransaction(payload []byte, inputs, outputs, dependencies []string) Transaction {
```
delventhalz: How does the nonce get set if `addNonce` is false?

sdk/go/src/sawtooth_sdk/client/encoder.go
delventhalz: Also missing is the `payloadEncoder` option. Which I thought was a nice convenience, though others may disagree and/or Go may not be functional enough to support it.

sdk/go/src/sawtooth_sdk/client/encoder.go
```diff
@@ -0,0 +1,156 @@
...
+
+// GetId Returns the Transaction ID which can be used to specify this
+// transaction as a dependency for other transactions.
+func (self Transaction) Id() string {
+	return self.pb.GetHeaderSignature()
+}
+
+// NewBatch creates a new batch from the given transactions created by
+// NewTransaction. It handles the construction and signing of the batch header.
+func (self *TransactionEncoder) NewBatch(transactions []Transaction) Batch {
```
delventhalz: There should be a function for dumping Transactions into a serialized TransactionList. This further supports the use case where the user who creates the Transactions is different than the user who creates the Batches.

In that vein, `NewBatch` should be able to accept either instances of Transactions, or a serialized TransactionList.

sdk/go/src/sawtooth_sdk/client/encoder.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+}
+
+// NewTransaction Creates a new transaction and handles the construction and
+// signing of the transaction header.
+func (self *TransactionEncoder) NewTransaction(payload []byte, inputs, outputs, dependencies []string) Transaction {
+	hash := Encode(SHA512(payload), "hex")
+
+	header := &transaction_pb2.TransactionHeader{
+		BatcherPubkey:   Encode(self.pubkey, "hex"),
```
delventhalz: This makes it impossible for Transaction creators and Batchers to be separate users (that's the whole reason we have a batcher pubkey). `BatcherPubkey` should be a settable default in the TransactionEncoder, which if not set (there or during transaction creation), should default to the signer's pubkey.

sdk/go/src/sawtooth_sdk/client/signer.go
```diff
@@ -0,0 +1,106 @@
...
+		panic("Failed to generate private key")
+	}
+	return priv.Serialize()
+}
+
+// GenPubKey generates a new public key from a given private key and returns it
+// using the 33 byte compressed format
+func GenPubKey(privkey []byte) []byte {
+	_, pub := ellcurv.PrivKeyFromBytes(ellcurv.S256(), privkey)
+	return pub.SerializeCompressed()
```
delventhalz: Is there some utility in giving users the raw binary instead of a hex string? I see you are encoding it for them during transaction creation, but is a binary going to be the easiest format to store or transmit? Will people be confused by having a public key that is a binary when the documentation shows hex strings?

sdk/go/src/sawtooth_sdk/client/signer.go
```diff
@@ -0,0 +1,106 @@
...
+// SHA256 calculates a sha256 has from the given input byte slice
+func SHA256(input []byte) []byte {
+	hash := sha256.New()
+	hash.Write(input)
+	return hash.Sum(nil)
+}
+
+// ---
+
+func serializeCompact(sig *ellcurv.Signature) []byte {
```
delventhalz: Different casing than other functions? Not sure if typo or I just don't know Go styling.

sdk/go/src/sawtooth_sdk/client/encoder.go
```diff
@@ -0,0 +1,156 @@
...
+	"encoding/hex"
+	"fmt"
+	"github.com/golang/protobuf/proto"
+	"sawtooth_sdk/protobuf/batch_pb2"
+	"sawtooth_sdk/protobuf/transaction_pb2"
+	"strings"
+	"time"
+)
+
+// See sdk/go/tests/client_test.go for example usage.
```
delventhalz: Technically this line should be part of the next commit. If we rolled back to`Create Go Client SDK`, the file you are referencing would not exist.

sdk/go/src/sawtooth_sdk/client/signer.go
aludvik: Oddly enough, capitalization is used to determine whether something is exported or not. Starting with uppercase means it's exported, starting with lowercase means it isn't.

sdk/go/src/sawtooth_sdk/client/signer.go
aludvik: My thought here was that it would be a little faster to pass around the byte array directly rather than the string encoded key (which would need to be encoded/decoded constantly). My understanding is there will be a separate guide for using the Go SDK which would give examples where you are encoding the key.

sdk/go/src/sawtooth_sdk/client/encoder.go
```diff
@@ -0,0 +1,156 @@
...
+	familyName      string
+	familyVersion   string
+	payloadEncoding string
+	addNonce        bool
+}
+
+// NewTransactionEncoder constructs a new encoder which can be used to generate
+// transactions and batches, and to serialize batches for submitting to the
+// REST API.
+func NewTransactionEncoder(privkey []byte, familyName, familyVersion, payloadEncoding string, addNonce bool) *TransactionEncoder {
```
aludvik: https://github.com/golang/go/wiki/CodeReviewComments#line-length

sdk/go/src/sawtooth_sdk/client/encoder.go
aludvik: Removed the comment.

sdk/go/src/sawtooth_sdk/client/encoder.go
vaporos: Agree with @delventhalz

sdk/go/src/sawtooth_sdk/client/encoder.go
aludvik: I think you are right about not making assumptions about the defaults, so I am making changes based on that.

I don't think it makes sense to have a payloadEncoder option, at least not in go. Since Go is typed, this would require type casting at runtime by both the payloadEncoder and the NewTransaction functions. It also wouldn't benefit a developer that much, since they would be defining the function in either case.

sdk/go/src/sawtooth_sdk/client/encoder.go
aludvik: Good point

sdk/go/src/sawtooth_sdk/client/encoder.go
aludvik: I agree on the first point.

I disagree on the second point. Rather than overloading the interface (which in Go means either type casting or defining a new method) I think it makes more sense to provide a method for parsing the serialized TransactionList, which can then be passed to the NewBatch() method.

aludvik:
Updated the interface based on suggestions from @delventhalz

sdk/go/src/sawtooth_sdk/client/signer.go
delventhalz: Well in the case of public keys, the hex string is the form in which it is used. You would probably speed things up by giving them a hex string so you only encoded it once, not every time a transaction was created.

I chose to also give private keys back in hex string for the JS SDK so it was more consistent, but I could see going either way on that one.

sdk/go/src/sawtooth_sdk/client/encoder.go
delventhalz: I love this.

sdk/go/src/sawtooth_sdk/client/encoder.go
```diff
@@ -0,0 +1,270 @@
...
+
+// NewTransaction Creates a new transaction and handles the construction and
+// signing of the transaction header.
+func (self *Encoder) NewTransaction(payload []byte, p TransactionParams) *Transaction {
+	h := &transaction_pb2.TransactionHeader{
+		// Load defaults
+		FamilyName:      self.defaults.FamilyName,
+		FamilyVersion:   self.defaults.FamilyVersion,
+		PayloadEncoding: self.defaults.PayloadEncoding,
+		Nonce:           self.defaults.Nonce,
```
delventhalz: I see how this streamlines the encoder interface . . . but it would be awfully weird to have a default nonce.

sdk/go/src/sawtooth_sdk/client/encoder.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+	if p.PayloadEncoding != "" {
+		h.PayloadEncoding = p.PayloadEncoding
+	}
+	if p.Nonce != "" {
+		h.Nonce = p.Nonce
+	}
+	if p.BatcherPubkey != "" {
+		h.BatcherPubkey = p.BatcherPubkey
+	}
```
delventhalz: I know nothing about Go, but is there not a way you can loop through the keys in the header or the new params,  so you don't have to check each individually?

aludvik: Since the field types are different, I had to use a struct instead of a map, so there wasn't a good way I know of. I think of Go as just a prettier version of C with a GC

PR #482, opened 2017-05-17
Add Get State Deltas Handler

peterschwarz:
Add Get State Delta handler to request state delta events by block ID.

This is useful for pre-catching up a client, in the case where they don't want to flood the network with many event messages.

Adds tests to verify that the subscriber sending a list of known block ids will return true if the request contains some known ids.

### DIFF
--- a/validator/sawtooth_validator/state/state_delta_processor.py
+++ b/validator/sawtooth_validator/state/state_delta_processor.py
@@ -173,7 +173,7 @@ class StateDeltaProcessor(object):
             block (:obj:`BlockWrapper`): The block whose state delta will
                 be published.
         """
-        LOGGER.debug('Publishing state delta fro %s', block)
+        LOGGER.debug('Publishing state delta from %s', block)
         state_root_hash = block.header.state_root_hash
 
         deltas = self._get_delta(state_root_hash)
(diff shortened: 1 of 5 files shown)


validator/sawtooth_validator/state/state_delta_processor.py
```diff
@@ -294,3 +298,54 @@ def handle(self, connection_id, message_content):
...
+
+    def __init__(self, block_store, state_delta_store):
+        self._block_store = block_store
+        self._state_delta_store = state_delta_store
+
+    def handle(self, connection_id, message_content):
+        request = GetStateDeltaEventsRequest()
+        request.ParseFromString(message_content)
+
+        # Create a temporary subscriber for this repsonse
```
agunde406: Typo: repsonse -> response

validator/tests/unit3/test_state_delta_processor/tests.py
```diff
@@ -408,3 +517,81 @@ def test_publish_deltas_no_state_changes(self):
...
+        delta_processor = StateDeltaProcessor(
+            service=mock_service,
+            state_delta_store=delta_store,
+            block_store=block_tree_manager.block_store)
+
+        self.assertTrue(delta_processor.is_valid_subscription([
+            block_tree_manager.chain_head.identifier]))
+
+    def test_is_valid_subscription_known_old_chain(self):
+        """test that a check for a valid subscription where the known block id
```
agunde406: test should be Uppercase

peterschwarz:
Rebased

PR #491, opened 2017-05-18
Remove stacktraces on 404 from sawtooth show command

ojalatodd:
Previously, the Sawtooth CLI mishandled 404 errors when fetching
resources, causing a stack trace to be be printed to the console
when running a 'sawtooth show' command with a bad id.
This replaces the stack trace with a more sensible error message.

### DIFF
--- a/cli/sawtooth_cli/rest_client.py
+++ b/cli/sawtooth_cli/rest_client.py
@@ -95,7 +95,8 @@ class RestClient(object):
         if code == 200:
             return json_result
         elif code == 404:
-            return None
+            raise CliException('There is no resource with the identifier "{}"'.
+                               format(path.split('/')[-1]))
         else:
             raise CliException("({}): {}".format(code, json_result))
 


cli/sawtooth_cli/rest_client.py
```diff
@@ -95,7 +95,7 @@ def _get(self, path, **queries):
         if code == 200:
             return json_result
         elif code == 404:
-            return None
+            raise CliException("404 not found: {} ".format(path))
```
delventhalz: I would prefer a little more informative error message here. Something like: `"There is no resource at the requested address: {}"`. The fact that the HTTP designation happens to be `404 Not Found` seems utterly unimportant to me.

ojalatodd: Thanks. Great suggestion.

ojalatodd:
@delventhalz I improved the commit message

PR #492, opened 2017-05-19
Update Readme with name and status changes

ineffectualproperty:
Signed-off-by: Kelly Olson <kelly.m.olson@intel.com>


ineffectualproperty:
third time is the charm @dcmiddle

PR #494, opened 2017-05-20
Make xo implementations and tests conform to spec

nick-drozd:
The actual xo specification hasn't been written yet, so this PR expands the TP test to be spec-like and make the Python implementation clearer. It also fixes the xo state format to account for hash collisions.

Though it isn't strictly related to changes in the specification, the xo smoke test has been updated to actually verify results (previous, it just ran the xo cli commands and then passed if nothing blew up).

### DIFF
--- a/integration/sawtooth_integration/docker/two-families.yaml
+++ b/integration/sawtooth_integration/docker/two-families.yaml
@@ -23,8 +23,6 @@ services:
       - $SAWTOOTH_CORE:/project/sawtooth-core
     expose:
       - 40000
-    depends_on:
-      - validator
     command: tp_config -vv tcp://validator:40000
     stop_signal: SIGKILL
 
@@ -34,9 +32,7 @@ services:
       - $SAWTOOTH_CORE:/project/sawtooth-core
     expose:
       - 40000
-    depends_on:
-      - validator
-    command: tp_intkey_python -vv tcp://validator:40000
+    command: tp_intkey_python -v tcp://validator:40000
     stop_signal: SIGKILL
 
   tp_xo_python:
@@ -45,9 +41,7 @@ services:
       - $SAWTOOTH_CORE:/project/sawtooth-core
     expose:
       - 40000
-    depends_on:
-      - validator
-    command: tp_xo_python -vv tcp://validator:40000
+    command: tp_xo_python -v tcp://validator:40000
     stop_signal: SIGKILL
 
   validator:
@@ -57,11 +51,10 @@ services:
     expose:
       - 40000
       - 8800
-    # start the validator with an empty genesis batch
     command: "bash -c \"\
         sawtooth admin keygen && \
         sawtooth admin genesis && \
-        validator --public-uri tcp://validator:8800 -vv \
+        validator --public-uri tcp://validator:8800 -v \
             --component-endpoint tcp://eth0:40000 \
             --network-endpoint tcp://eth0:8800 \
     \""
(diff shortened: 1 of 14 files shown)


sdk/examples/xo_python/sawtooth_xo/xo_message_factory.py
```diff
@@ -13,57 +13,65 @@
...
+
+def make_xo_address(name):
+    return XO_NAMESPACE + hashlib.sha512(name.encode('utf-8')).hexdigest()[-64:]
+
+# encodings
+def encode_txn_payload(action, name, space=''):
+    return ','.join([str(data) for data in (action, name, space)]).encode()
+
+
 class XoMessageFactory:
```
peterschwarz: Could you add a commit to move this to a test directory - it shouldn't live with the production code.

sdk/examples/xo_javascript/xo_handler.js
```diff
@@ -29,13 +29,21 @@ const _hash = (x) =>
 const XO_FAMILY = 'xo'
 const XO_NAMESPACE = _hash(XO_FAMILY).substring(0, 6)
 
+const VALID_ACTIONS = ['create', 'take']
+const VALID_STATES = ['P1-NEXT', 'P2-NEXT', 'P1-WIN', 'P2-WIN', 'TIE']
+
+const MARK_1 = 'X'
```
peterschwarz: Name these `MARK_X` and `MARK_O` - no need to obfuscate them.

sdk/examples/xo_python/sawtooth_xo/processor/handler.py
```diff
@@ -21,16 +21,47 @@
...
+    return data.decode().split(',')
+
+def encode_state_data(data):
+    return ','.join(data).encode()
 
+# board
+SIDE_LENGTH = 3
+SPACE_COUNT = SIDE_LENGTH ** 2
+VALID_SPACES = range(1, SPACE_COUNT + 1)
+VALID_MARKS = MARK_1, MARK_2, EMPTY_SPACE = 'X', 'O', '-'
```
peterschwarz: Name these `MARK_X` and `MARK_O`; no need for the obfuscation.

integration/sawtooth_integration/docker/two-families.yaml
```diff
@@ -23,8 +23,6 @@ services:
       - $SAWTOOTH_CORE:/project/sawtooth-core
     expose:
       - 40000
-    depends_on:
```
boydjohnson: These 'depends_on' deletions shouldn't be part of this PR. I mentioned in RocketChat that I had spread some bad information about needing these statements in the docker compose yaml files: https://chat.hyperledger.org/channel/sawtooth?msg=cM4rrDvj95WfaPrrD. We should verify that all of our integration tests can pass reliably without depends_on, and then make a PR that removes all of them.

sdk/examples/xo_python/sawtooth_xo/xo_message_factory.py
nick-drozd: I can do that in this PR, or open another PR to do it for all the TPs -- intkey and registry follow this same pattern, while config has a different but still odd arrangement.

sdk/examples/xo_python/sawtooth_xo/processor/handler.py
nick-drozd: Should they be renamed PLAYER_1_MARK, etc? They're the marks for players 1 and 2. They happen to be 'X' and 'O', but they could be different.

nick-drozd:
@aludvik My hope was to get all the formal requirements into the TP test and then convert that into a document. The basic guiding principle is "do whatever the existing Python implementation does".

nick-drozd:
Clearly there are some problems with the structure of this PR's commits, so I'm going to reorganize it.

PR #495, opened 2017-05-20
Split on-chain settings keys into parts

peterschwarz:
Setting keys are broken into four parts, based on the dots in the string. For example, the address for the key `a.b.c` is computed based on `a`, `b`, `c` and the empty string.  A longer key, for example
`a.b.c.d.e`, is still broken into four parts, but the remain pieces are in the last part: `a`, `b`, `c` and `d.e`.

Each of these pieces has a short hash computed (the first 16 characters of its SHA256 hash in hex) and is joined into a single address, with the config namespace (`000000`) added at the beginning.

This strategy is employed at all locations where config keys are converted into addresses.

This allows subsets of settings keys to be queried from state by address prefix.


peterschwarz:
Updated the transaction family spec with the changes.

PR #497, opened 2017-05-20
Add no-op transaction processor and workload client

jsmitchell:
These commits add a Go noop transaction processor which immediately returns when invoked without parsing payload or accessing state. Also included is a python client which includes a workload generator to submit noop transactions. The transaction payload consist of an 8 byte random nonce to give the transactions unique signatures.

Submission of solely noop transactions results in moving the block chain forward without changes to the state root hash.

This is intended for use in performance/integration testing.

### DIFF
--- a/sdk/examples/noop_go/src/sawtooth_noop/handler/handler.go
+++ b/sdk/examples/noop_go/src/sawtooth_noop/handler/handler.go
@@ -1,0 +1,65 @@
+/**
+ * Copyright 2017 Intel Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ------------------------------------------------------------------------------
+ */
+
+package handler
+
+import (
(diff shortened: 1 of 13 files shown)


sdk/examples/noop_go/src/sawtooth_noop/handler/handler.go
```diff
@@ -0,0 +1,65 @@
...
+	namespace string
+}
+
+func NewNoopHandler(namespace string) *NoopHandler {
+	return &NoopHandler{
+		namespace: namespace,
+	}
+}
+
+var FAMILY_NAME = "noop"
```
aludvik: `const FAMILY_NAME = "noop"`

bin/build_all
```diff
@@ -213,6 +213,7 @@ then
     build_go() {
         docker_build docker/sawtooth-dev-go docker/ sawtooth-dev-go
         docker_build docker/sawtooth-dev-go docker/ sawtooth-tp_intkey_go
+        docker_build docker/sawtooth-dev-go docker/ sawtooth-tp_noop_go
         docker_run sawtooth-dev-go
```
aludvik: You created a sawtooth-int-tp_noop_go Dockerfile but you never build it. Since you don't have any tests for this TP, there isn't a reason to build this here. If you do, then you need to add a line for building the noop_go in installed mode (which is what runs on Jenkins) See line 247 below.

sdk/examples/noop_go/src/sawtooth_noop/handler/handler.go
vaporos: @aludvik should this be 'familyName'?

aludvik: If you make it const, I don't see a reason to keep it private. Here is an example of using all caps for consts https://golang.org/doc/effective_go.html#constants

PR #498, opened 2017-05-20
Create compose files for application development

aludvik:


aludvik:
I decided this is a bad idea and we should instead get stable packages published to appropriate package management tools for each language.

PR #502, opened 2017-05-22
Add Xo Go

vaporos:


sdk/examples/xo_go/src/sawtooth_xo/handler/handler.go
```diff
@@ -0,0 +1,321 @@
...
+
+func (self *XoHandler) Apply(request *processor_pb2.TpProcessRequest, state *processor.State) error {
+	// The xo player is defined as the signer of the transaction, so we unpack
+	// the transaction header to obtain the signer's public key, which will be
+	// used as the player's identity.
+	header, err := unpackHeader(request.Header)
+	if err != nil {
+		return err
+	}
+	player := header.SignerPubkey
```
aludvik: My convention has been to use the helper methods here, so instead I would do `header.GetSignerPubkey()`.

"There are getters that return a field's value if set, and return the field's default value if unset. The getters work even if the receiver is a nil message." -https://github.com/golang/protobuf

sdk/examples/xo_go/src/sawtooth_xo/handler/handler.go
```diff
@@ -0,0 +1,321 @@
...
+func isWin(board string, letter byte) bool {
+	var wins [8][]int
+	wins[0] = []int{1, 2, 3}
+	wins[1] = []int{4, 5, 6}
+	wins[2] = []int{7, 8, 9}
+	wins[3] = []int{1, 4, 7}
+	wins[4] = []int{2, 5, 8}
+	wins[5] = []int{3, 6, 9}
+	wins[6] = []int{1, 5, 9}
+	wins[7] = []int{3, 5, 7}
```
aludvik: This feels cleaner to me:
```
wins := [8][3]int{
    {1, 2, 3},
    {4, 5, 6},
    {7, 8, 9},
    {1, 4, 7},
    {2, 5, 8},
    {3, 6, 9},
    {1, 5, 9},
    {3, 5, 7},
}
```

sdk/examples/xo_go/src/sawtooth_xo/handler/handler.go
```diff
@@ -0,0 +1,321 @@
...
+		return nil, &processor.InvalidTransactionError{"Name is required"}
+	}
+
+	if len(payload.Action) < 1 {
+		return nil, &processor.InvalidTransactionError{"Action is required"}
+	}
+
+	if payload.Action == "take" {
+		space, err := strconv.Atoi(parts[2])
+		payload.Space = space
```
aludvik: This should be after `if err != nil {...}`

sdk/examples/xo_go/src/sawtooth_xo/handler/handler.go
```diff
@@ -0,0 +1,321 @@
...
+			fmt.Sprint("Failed to unmarshal TransactionHeader: %v", err)}
+	}
+	return header, nil
+}
+
+func unpackGame(gameData []byte) (*Game, error) {
+	parts := strings.Split(string(gameData), ",")
+	if len(parts) != 5 {
+		return nil, &processor.InternalError{
+			fmt.Sprintf("Failed to deserialize game data: '%v'", string(gameData))}
```
aludvik: Change error message to "Malformed game data: '%v'" to match language above?

docker/compose/sawtooth-minimal.yaml
```diff
@@ -0,0 +1,62 @@
+# Copyright 2017 Intel Corporation
```
aludvik: Should this file be here?

docker/sawtooth-int-tp_xo_go
```diff
@@ -0,0 +1,39 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+
+# Description:
+#   Builds an image with Sawtooth TP Intkey Go installed.
```
aludvik: Change Intkey to Xo

docker/sawtooth-int-tp_xo_go
```diff
@@ -0,0 +1,39 @@
...
+#
+# Build:
+#   This image should be built using `build_all installed`.
+#
+# Run:
+#   $ docker run sawtooth-tp_intkey_go
```
aludvik: Change intkey to xo

sdk/examples/xo_go/src/sawtooth_xo/handler/handler.go
```diff
@@ -0,0 +1,321 @@
...
+
+	if len(string(results[address])) > 0 {
+		game, err := unpackGame(results[address])
+		if err != nil {
+			return nil, err
+		}
+
+		// NOTE: Since the game data is stored in a Merkle tree, there is a
+		// small chance of collision. A more correct usage would be to store
+		// a dictionary of games so that multiple games could be store at
```
ojalatodd: (Grammar/typo): 
Change
 _a dictionary of games so that multiple games could be store at_
to 
 a dictionary of games so that multiple games could be **_stored_** at

sdk/examples/xo_go/src/sawtooth_xo/handler/handler.go
```diff
@@ -0,0 +1,321 @@
...
+	data := packGame(game)
+
+	addresses, err := state.Set(map[string][]byte{
+		address: data,
+	})
+	if err != nil {
+		return &processor.InternalError{fmt.Sprint("Failed to set new Value:", err)}
+	}
+	if len(addresses) == 0 {
+		return &processor.InternalError{"No addresses in set response"}
```
ojalatodd: You have more error handling here than the other implementations. Is that something we want to encourage or document in terms of writing a txn processor? Just a general comment.

sdk/examples/xo_go/src/sawtooth_xo/handler/handler.go
```diff
@@ -0,0 +1,321 @@
...
+		game, err := unpackGame(results[address])
+		if err != nil {
+			return nil, err
+		}
+
+		// NOTE: Since the game data is stored in a Merkle tree, there is a
+		// small chance of collision. A more correct usage would be to store
+		// a dictionary of games so that multiple games could be store at
+		// the same location. See the python intkey handler for an example
+		// of this.
```
dplumb94: In order to handle this scenario, we will be storing data at each address as a JSON string in the format: `{"game_name":"board, state, player1, player2", "another_game_name":"board,state,player1,player2"}`

We might want to incorporate this into the Go implementation.

sdk/examples/xo_go/src/sawtooth_xo/handler/handler.go
```diff
@@ -0,0 +1,321 @@
...
+		if game.Name != name {
+			return nil, &processor.InternalError{"Hash collision"}
+		}
+		return game, nil
+	}
+	return nil, nil
+}
+
+func saveGame(game *Game, state *processor.State) error {
+	address := namespace + hexdigest(game.Name)
```
dplumb94: Do we want to limit addresses to 70 characters (as with Intkey)?

docker/sawtooth-int-tp_xo_go
vaporos: I think the 'int' in the filename is short for integration test.

sdk/examples/xo_go/src/sawtooth_xo/handler/handler.go
vaporos: I just followed intkey go in this respect.  @aludvik might know why this specific additional check after err is necessary.  In general, the documentation needs to make it clear that there are two errors than can be returned, InvalidTransactionError and InternalError; they cause different behavior in the validator.

sdk/examples/xo_go/src/sawtooth_xo/handler/handler.go
vaporos: Agreed.

sdk/examples/xo_go/src/sawtooth_xo/handler/handler.go
vaporos: Yes, probably, but this PR needs to be compatible with the xo in master so when it merges they work together.  If that is part of @nick-drozd 's spec work and it gets merged first, then I'll update this PR.

integration/sawtooth_integration/docker/xo-go-smoke.yaml
```diff
@@ -0,0 +1,92 @@
...
+    image: sawtooth-validator:$ISOLATION_ID
+    volumes:
+      - $SAWTOOTH_CORE:/project/sawtooth-core
+    expose:
+      - 40000
+    # start the validator with an empty genesis batch
+    command: "bash -c \"\
+        sawtooth admin keygen && \
+        sawtooth admin genesis && \
+        validator -v \
```
agunde406: This needs to be updated to match the new validator command line options.

integration/sawtooth_integration/docker/xo-go-smoke.yaml
```diff
@@ -0,0 +1,92 @@
...
+  rest_api:
+    image: sawtooth-rest_api:$ISOLATION_ID
+    volumes:
+      - $SAWTOOTH_CORE:/project/sawtooth-core
+    expose:
+      - 40000
+      - 8080
+    depends_on:
+      - validator
+    command: rest_api -v --stream-url tcp://validator:40000 --host rest_api
```
agunde406: This needs to be updated to match the new rest_api command line options.

sdk/examples/xo_go/src/sawtooth_xo/handler/handler.go
```diff
@@ -0,0 +1,322 @@
...
+	if payload.Action == "take" {
+		space, err := strconv.Atoi(parts[2])
+		if err != nil {
+			return nil, &processor.InvalidTransactionError{
+				fmt.Sprintf("Invalid Space: '%v'", parts[2])}
+		}
+		payload.Space = space
+	}
+
+	if strings.Index(payload.Name, "|") != -1 {
```
aludvik: You used strings.Contains above but not here. Not sure why.
`strings.Contains(payload.Name, "|")`

sdk/examples/xo_go/src/sawtooth_xo/handler/handler.go
```diff
@@ -0,0 +1,322 @@
...
+func loadGame(name string, state *processor.State) (*Game, error) {
+	// Use the namespace prefix + the hash of the game name to create the
+	// storage address
+	address := namespace + hexdigest(name)[:64]
+
+	results, err := state.Get([]string{address})
+	if err != nil {
+		return nil, &processor.InternalError{fmt.Sprint("Error getting state:", err)}
+	}
+
```
aludvik: Missing pipe handling: https://github.com/hyperledger/sawtooth-core/blob/master/sdk/examples/xo_python/sawtooth_xo/processor/handler.py#L105

Also, the state.Get() omits addresses that have data of length 0: https://github.com/hyperledger/sawtooth-core/blob/master/sdk/go/src/sawtooth_sdk/processor/state.go#L113

Should look something like:
```
gameData, exists := results[address]
if !exists {
    return nil, nil
}

games := strings.Split(gameData, "|")
for g := range games {
    game, err := unpackGame(g)
    if err != nil {
        return nil, err
    }
    if game.Name == name {
        return game, nil
    }
}
return nil, nil
```

sdk/examples/xo_go/src/sawtooth_xo/handler/handler.go
aludvik: Hmm, maybe I got the level of abstraction wrong with the state.Set method. Would it make more sense if the State class returned an error when the SetResponse does not include one of the addresses that it tried to set, even if the return status is okay?

docker/sawtooth-int-tp_xo_go
```diff
@@ -0,0 +1,39 @@
...
+
+RUN apt-get update \
+ && apt-get install -y -q \
+    libzmq3-dev \
+ && apt-get clean \
+ && rm -rf /var/lib/apt/lists/*
+
+COPY tp_xo_go /usr/bin/tp_xo_go
+
+EXPOSE 40000/tcp
```
aludvik: 4004

sdk/examples/xo_go/src/sawtooth_xo/handler/handler.go
vaporos: Support for handling hash collision is a deeper change, so I decided not to extend the intent of this PR to include it.  Instead, I'll submit another PR that makes that modification.

sdk/examples/xo_go/src/sawtooth_xo/handler/handler.go
vaporos: No particular reason, I'll change it.

PR #504, opened 2017-05-22
Add sort query to REST API

delventhalz:
Results to endpoints that list resources can now be sorted validator-side (pre-pagination) using the `sort` query parameter. It accepts a comma-separated list of keys which can be modified with some useful syntax:

- dot-notation indicates nested keys
- the header can be explicitly or implicity referenced
- starting with a minus sign indicates descending order
- ending with ".length" indicates the lengths should be compared
- additional key strings are tie-breakers for earlier ones
- example: `sort=-header.transaction_ids.length,signer_pubkey`

Includes a full set of unit tests.

### DIFF
--- a/rest_api/sawtooth_rest_api/exceptions.py
+++ b/rest_api/sawtooth_rest_api/exceptions.py
@@ -189,6 +189,14 @@ class PagingInvalid(_ApiError):
                "out of range.")
 
 
+class SortInvalid(_ApiError):
+    api_code = 57
+    status_code = 400
+    title = 'Invalid Sort Query'
+    message = ("The sort request failed as written. Some of the keys "
+               "specified were not valid.")
+
+
 class InvalidStateAddress(_ApiError):
     api_code = 62
     status_code = 400
--- a/rest_api/sawtooth_rest_api/route_handlers.py
+++ b/rest_api/sawtooth_rest_api/route_handlers.py
@@ -206,6 +206,7 @@ class RouteHandler(object):
         validator_query = client_pb2.ClientStateListRequest(
             head_id=request.url.query.get('head', None),
             address=request.url.query.get('address', None),
+            sorting=self._get_sorting_message(request),
             paging=self._make_paging_message(paging_controls))
 
         response = await self._query_validator(
@@ -267,6 +268,7 @@ class RouteHandler(object):
         validator_query = client_pb2.ClientBlockListRequest(
             head_id=request.url.query.get('head', None),
             block_ids=self._get_filter_ids(request),
+            sorting=self._get_sorting_message(request),
             paging=self._make_paging_message(paging_controls))
 
         response = await self._query_validator(
@@ -322,6 +324,7 @@ class RouteHandler(object):
         validator_query = client_pb2.ClientBatchListRequest(
             head_id=request.url.query.get('head', None),
             batch_ids=self._get_filter_ids(request),
+            sorting=self._get_sorting_message(request),
             paging=self._make_paging_message(paging_controls))
 
         response = await self._query_validator(
@@ -378,6 +381,7 @@ class RouteHandler(object):
         validator_query = client_pb2.ClientTransactionListRequest(
             head_id=request.url.query.get('head', None),
             transaction_ids=self._get_filter_ids(request),
+            sorting=self._get_sorting_message(request),
             paging=self._make_paging_message(paging_controls))
 
         response = await self._query_validator(
@@ -496,6 +500,12 @@ class RouteHandler(object):
         try:
             if content.status == proto.INVALID_PAGING:
                 raise errors.PagingInvalid()
+        except AttributeError:
+            pass
+
+        try:
+            if content.status == proto.INVALID_SORT:
+                raise errors.SortInvalid()
         except AttributeError:
             pass
 
@@ -718,6 +728,37 @@ class RouteHandler(object):
             end_id=controls.get('end_id', None),
             start_index=start_index,
             count=count)
+
+    @staticmethod
+    def _get_sorting_message(request):
+        """Parses the sort query into a list of SortControls protobuf messages.
+        """
+        control_list = []
+        sort_query = request.url.query.get('sort', None)
+        if sort_query is None:
+            return control_list
+
+        for key_string in sort_query.split(','):
+            if key_string[0] == '-':
+                reverse = True
+                key_string = key_string[1:]
+            else:
+                reverse = False
+
+            keys = key_string.split('.')
+
+            if keys[-1] == 'length':
+                compare_length = True
+                keys.pop()
+            else:
+                compare_length = False
+
+            control_list.append(client_pb2.SortControls(
+                keys=keys,
+                reverse=reverse,
+                compare_length=compare_length))
+
+        return control_list
 
     def _set_wait(self, request, validator_query):
         """Parses the `wait` query parameter, and sets the corresponding
--- a/validator/sawtooth_validator/state/client_handlers.py
+++ b/validator/sawtooth_validator/state/client_handlers.py
@@ -15,6 +15,7 @@
 
 import abc
 import logging
+from functools import cmp_to_key
 # pylint: disable=import-error,no-name-in-module
 # needed for google.protobuf import
 from google.protobuf.message import DecodeError
@@ -26,6 +27,8 @@ from sawtooth_validator.networking.dispatch import HandlerStatus
 
 from sawtooth_validator.protobuf import client_pb2
 from sawtooth_validator.protobuf.block_pb2 import BlockHeader
+from sawtooth_validator.protobuf.batch_pb2 import BatchHeader
+from sawtooth_validator.protobuf.transaction_pb2 import TransactionHeader
 from sawtooth_validator.protobuf import validator_pb2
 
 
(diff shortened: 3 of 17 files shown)


validator/tests/unit3/test_client_request_handlers/test_block_handlers.py
```diff
@@ -389,6 +389,199 @@ def test_block_list_paginated_with_head (self):
...
+            {header: {block_num: 1 ...}, header_signature: 'B-1' ...},
+            {header: {block_num: 0 ...}, header_signature: 'B-0' ...},
+
+        Expects to find:
+            - a status of OK
+            - a head_id of 'B-2', the latest
+            - a paging response showing all 3 resources returned
+            - a list of blocks with 3 items
+            - the items are instances of Block
+            - the first item has a header_signature of 'B-0'
```
agunde406: Again first item is B-2 in this example and the last is B-0.

validator/tests/unit3/test_client_request_handlers/test_block_handlers.py
```diff
@@ -389,6 +389,199 @@ def test_block_list_paginated_with_head (self):
...
+            {header: {block_num: 0 ...}, header_signature: 'B-0' ...},
+
+        Expects to find:
+            - a status of OK
+            - a head_id of 'B-2', the latest
+            - a paging response showing all 5 resources returned
+            - a list of blocks with 5 items
+            - the items are instances of Block
+            - the first item has a header_signature of 'B-0'
+            - the last item has a header_signature of 'B-2'
```
agunde406: Again should mention the order of B-long and B-longest

validator/sawtooth_validator/state/client_handlers.py
```diff
@@ -382,6 +385,103 @@ def id_by_index(index, resources):
...
+                return 0
+            handler = value_handlers[0]
+
+            value_a, value_b = handler.get_sort_values(resource_a, resource_b)
+
+            if value_a < value_b:
+                return handler.xform_result(-1)
+            if value_a > value_b:
+                return handler.xform_result(1)
+            return sorter(resource_a, resource_b, value_handlers[1:])
```
aludvik: Beware, python does not optimize tail recursion: http://neopythonic.blogspot.com.au/2009/04/tail-recursion-elimination.html

So this is limited to lists of length ~1000. If I understand this right though, that is probably fine since it is unlikely there will be >1000 sort keys.

It could be a security issue though... A malicious actor could query a bunch of validators with requests that have a bunch of sort keys and bring them down. Is there a hard limit somewhere before this on the number of sort keys?

validator/tests/unit3/test_client_request_handlers/test_block_handlers.py
```diff
@@ -389,6 +389,199 @@ def test_block_list_paginated_with_head (self):
...
+        Queries the default mock block store with two added blocks:
+            {header: {block_num: 4 ...}, header_signature: 'B-long' ...},
+            {header: {block_num: 3 ...}, header_signature: 'B-longest' ...},
+            {header: {block_num: 2 ...}, header_signature: 'B-2' ...},
+            {header: {block_num: 1 ...}, header_signature: 'B-1' ...},
+            {header: {block_num: 0 ...}, header_signature: 'B-0' ...},
+
+        Expects to find:
+            - a status of OK
+            - a head_id of 'B-2', the latest
```
agunde406: Wrong head_id, should be B-long

validator/tests/unit3/test_client_request_handlers/test_state_handlers.py
```diff
@@ -438,6 +438,81 @@ def test_state_list_paginated_with_address (self):
...
+            {'a': b'3', 'b': b'5', 'c': b'7'}
+
+        Expects to find:
+            - a status of OK
+            - a head_id of 'B-2', the latest
+            - a paging response showing all 3 resources returned
+            - a list of leaves with 3 items
+            - the items are instances of Leaf
+            - the first item has a header_signature of 'B-0'
+            - the last item has a header_signature of 'B-2'
```
agunde406: I think this should mentioned the checks on state, not the Blocks.

validator/tests/unit3/test_client_request_handlers/test_state_handlers.py
```diff
@@ -438,6 +438,81 @@ def test_state_list_paginated_with_address (self):
...
```
agunde406: Same as above

validator/sawtooth_validator/state/client_handlers.py
```diff
@@ -382,6 +385,103 @@ def id_by_index(index, resources):
...
+            list: The sorted list of resources
+        """
+        if not request.sorting:
+            return resources
+
+        value_handlers = [cls._ValueHandler(controls, fail_enum, header_proto)
+                          for controls in request.sorting]
+
+        def sorter(resource_a, resource_b, value_handlers=value_handlers):
+            if not value_handlers:
```
aludvik: This threw me off because I didn't know an empty list was false-y in python. Is this good practice? Is it better than `if len(value_handlers) < 1:`?

validator/tests/unit3/test_client_request_handlers/test_txn_handlers.py
```diff
@@ -526,6 +526,276 @@ def test_txn_list_paginated_with_head (self):
...
+            {header_signature: 'B-0', ...}
+
+        Expects to find:
+            - a status of OK
+            - a head_id of 'B-2', the latest
+            - a paging response showing all 3 resources returned
+            - a list of transactions with 3 items
+            - the items are instances of Transaction
+            - the first item has a header_signature of 'B-0'
+            - the last item has a header_signature of 'B-2'
```
agunde406: Should mention the transactions, not the blocks.

validator/tests/unit3/test_client_request_handlers/test_txn_handlers.py
```diff
@@ -526,6 +526,276 @@ def test_txn_list_paginated_with_head (self):
...
```
agunde406: Should mention the transactions, not the blocks.

validator/tests/unit3/test_client_request_handlers/test_txn_handlers.py
```diff
@@ -526,6 +526,276 @@ def test_txn_list_paginated_with_head (self):
...
```
agunde406: Should mention the transactions, not the blocks.

validator/tests/unit3/test_client_request_handlers/test_txn_handlers.py
```diff
@@ -526,6 +526,276 @@ def test_txn_list_paginated_with_head (self):
...
```
agunde406: Should mention the transactions, not the blocks.

validator/tests/unit3/test_client_request_handlers/test_txn_handlers.py
```diff
@@ -526,6 +526,276 @@ def test_txn_list_paginated_with_head (self):
...
```
agunde406: Should mention the transactions, not the blocks.

validator/tests/unit3/test_client_request_handlers/test_txn_handlers.py
```diff
@@ -526,6 +526,276 @@ def test_txn_list_paginated_with_head (self):
...
+            {header_signature: 'B-0', ...}
+
+        Expects to find:
+            - a status of OK
+            - a head_id of 'B-2', the latest
+            - a paging response showing all 5 resources returned
+            - a list of transactions with 5 items
+            - the items are instances of Transaction
+            - the first item has a header_signature of 'B-0'
+            - the last item has a header_signature of 'B-2'
```
agunde406: Should mention the transactions, not the blocks. (t-long and t-longest)

rest_api/sawtooth_rest_api/exceptions.py
```diff
@@ -189,6 +189,14 @@ class PagingInvalid(_ApiError):
                "out of range.")
 
 
+class SortInvalid(_ApiError):
+    api_code = 57
+    status_code = 400
+    title = 'Invalid Sort Query'
+    message = ("The sort request failed as written. Some of the keys "
+               "specified are not be valid.")
```
agunde406: remove "be"

validator/sawtooth_validator/state/client_handlers.py
peterschwarz: You can probably handle this by using `functools.reduce` over the value_handlers.

peterschwarz: Though, you should still probably put a limit on the number of value_handlers.

validator/sawtooth_validator/state/client_handlers.py
delventhalz: According to this stack overflow, it is "quite" pythonic:
https://stackoverflow.com/questions/53513/best-way-to-check-if-a-list-is-empty

Not sure if there is a precedent with sawtooth's python code I should be observing, but perhaps Andi, Boyd, or Sawn could weigh in?

validator/sawtooth_validator/state/client_handlers.py
delventhalz: No, there isn't. You would be limited to keys actually present in the resource, and the second the values don't match the sorter will ignore the rest, but you could spam a key with an equal value over and over again. Something like `[{keys: ['family_name']}, {keys: ['family_name']}, ...]`.

What if the list of handlers became a _set_ of handers? 
```
value_handlers = set(cls._ValueHandler(controls, fail_enum, header_proto)
                          for controls in request.sorting)
```

At that point the limit is however many unique valid keys with non-unique values you can reference. A number which is probably not so big.

delventhalz: SortControls are unhashable. The basic idea should still work, but I won't be able to use `set`

validator/sawtooth_validator/state/client_handlers.py
delventhalz: Adam, I deleted your original comment about `not []` to check for an empty list because GitHub was showing duplicates of it for some reason.

Anyway, [the Stack Overflow](https://stackoverflow.com/questions/53513/best-way-to-check-if-a-list-is-empty) says it is _"quite pythonic"_. Happy to switch it if we are checking for empty lists in some other way elsewhere in sawtooth. Do Andi, Boyd, or Shawn have an opinion?

delventhalz:
Adam, the `-` indicates that the sort for _that key_ should be in descending order. So if you did `sort=-transaction_ids.length,signer_pubkey`, it would organize first by whichever batch had the most transactions _(biggest to smallest, descending)_. For any batches that had the same number of transactions, it would sort those by their pubkeys in alphabetical order _(a to z, ascending)_.

I will clarify this in the documentation.

validator/tests/unit3/test_client_request_handlers/test_block_handlers.py
delventhalz: Updated, not sure why Github is not reflecting that.

validator/tests/unit3/test_client_request_handlers/test_block_handlers.py
delventhalz: Updated.

validator/sawtooth_validator/state/client_handlers.py
```diff
@@ -382,6 +385,120 @@ def id_by_index(index, resources):
...
+        """
+        added = set()
+        handlers = []
+
+        for controls in request.sorting:
+            control_bytes = controls.SerializeToString()
+            if control_bytes not in added:
+                added.add(control_bytes)
+                handlers.append(
+                    cls._ValueHandler(controls, fail_enum, header_proto))
```
delventhalz: Added this method to address Adam's concern about spamming `SortControls`. Duplicate controls will be dropped before instantiating `_ValueHandlers` for them.

validator/sawtooth_validator/state/client_handlers.py
delventhalz: Added a fix below.

validator/sawtooth_validator/state/client_handlers.py
```diff
@@ -382,6 +385,118 @@ def id_by_index(index, resources):
...
+        """
+        if not request.sorting:
+            return resources
+
+        value_handlers = cls._get_handler_set(request, fail_enum, header_proto)
+
+        def sorter(resource_a, resource_b):
+            for handler in value_handlers:
+                val_a, val_b = handler.get_sort_values(resource_a, resource_b)
+
```
delventhalz: The sorter function is also no longer recursive.

rest_api/sawtooth_rest_api/route_handlers.py
```diff
@@ -499,6 +503,12 @@ def _check_status_errors(proto, content, error_traps=None):
         except AttributeError:
             pass
 
+        try:
+            if content.status == proto.INVALID_SORT:
+                raise errors.SortInvalid()
+        except AttributeError:
+            pass
```
vaporos: Stylistically, I'd put comments everywhere except/pass is used so the intent is clear.  "# ok if status field not on content" or similar.  Better might be explicitly checking for it and not relying on exceptions.

delventhalz: There is a comment on the first time I do this a few lines above. I could add a comment every time if you think it's important.

I'd also be open to a different approach. I have never been happy with this solution, and have gone over many alternatives in my head, but they all seem imperfect.

PR #509, opened 2017-05-22
Xo update

nick-drozd:
The xo spec hasn't been written yet, but the xo TP test is intended to show what will go into the spec. Note that the xo message factory also contains information that will go into the spec, since it determines the format of xo state data.

This PR makes several changes to xo message formats: 1) the format of transactions has been changed from `(name, action, space)` to `(action, name, space)` (the latter being how transactions are submitted at the CLI); 2) xo addresses are like intkey addresses -- the first six characters of the hash of the family name plus the last 64 characters of the hash of the game name; and 3) the data stored at an address is now a JSON string in the following format: `{"game_name": "board,state,player_1,player_2"}`.

The Python xo implementation has been rewritten to more or less match the documentation, and a few errors have been fixed in the JS implementation.

### DIFF
--- a/bin/run_tests
+++ b/bin/run_tests
@@ -174,6 +174,9 @@ main() {
                 deployment)
                     test_deployment
                     ;;
+                xo)
+                    test_xo
+                    ;;
                 *)
                     echo "Module '$MODULE' not found."
                     ;;
@@ -286,6 +289,19 @@ test_integration() {
 
 test_deployment() {
     run_docker_test systemd_services -s integration_test
+}
+
+test_xo() {
+    # xo python
+    run_docker_test ./sdk/examples/xo_python/tests/tp-xo-python.yaml -s validator
+    copy_coverage .coverage.tp-xo-python
+    run_docker_test xo-python-smoke -s integration_test
+    copy_coverage .coverage.xo-smoke
+    # xo js
+    run_docker_test ./sdk/examples/xo_python/tests/tp-xo-javascript.yaml -s validator
+    copy_coverage .coverage.tp-xo-javascript
+    run_docker_test xo-javascript-smoke -s integration_test
+    copy_coverage .coverage.xo-javascript-smoke
 }
 
 copy_coverage() {
--- a/bin/run_tests
+++ b/bin/run_tests
@@ -276,8 +276,8 @@ test_python_sdk() {
     copy_coverage .coverage.workload
     run_docker_test xo-python-smoke -s integration_test
     copy_coverage .coverage.xo-python-smoke
-    run_docker_test two_families -s integration_test
-    copy_coverage .coverage.two_families
+    run_docker_test two-families -s integration_test
+    copy_coverage .coverage.two-families
 }
 
 test_integration() {
--- a/bin/run_tests
+++ b/bin/run_tests
@@ -271,8 +271,8 @@ test_python_sdk() {
     copy_coverage .coverage.intkey-python-smoke
     run_docker_test workload -s integration_test
     copy_coverage .coverage.workload
-    run_docker_test xo-smoke -s integration_test
-    copy_coverage .coverage.xo-smoke
+    run_docker_test xo-python-smoke -s integration_test
+    copy_coverage .coverage.xo-python-smoke
     run_docker_test two_families -s integration_test
     copy_coverage .coverage.two_families
 }
--- a/bin/run_tests
+++ b/bin/run_tests
@@ -302,6 +302,9 @@ test_xo() {
     copy_coverage .coverage.tp-xo-javascript
     run_docker_test xo-javascript-smoke -s integration_test
     copy_coverage .coverage.xo-javascript-smoke
+    # all languages
+    run_docker_test xo-all-lang-smoke
+    copy_coverage .coverage.xo-all-lang-smoke
 }
 
 copy_coverage() {
--- a/sdk/examples/xo_python/sawtooth_xo/processor/handler.py
+++ b/sdk/examples/xo_python/sawtooth_xo/processor/handler.py
@@ -141,6 +141,12 @@ def _validate_name(name):
     if not name:
         raise InvalidTransaction('Name is required')
 
+    bad_chars = ',', '|'
+
+    if any([bad_char in name for bad_char in bad_chars]):
+        raise InvalidTransaction(
+            'Name cannot contain "," or "|"')
+
 
 def _validate_action_and_space(action, space):
     if not action:
--- a/sdk/examples/xo_python/sawtooth_xo/processor/handler.py
+++ b/sdk/examples/xo_python/sawtooth_xo/processor/handler.py
@@ -25,15 +25,28 @@ LOGGER = logging.getLogger(__name__)
 
 
 # namespace
-def hash_name(name):
+def _hash_name(name):
     return hashlib.sha512(name.encode('utf-8')).hexdigest()
 
 FAMILY_NAME = 'xo'
-XO_NAMESPACE = hash_name(FAMILY_NAME)[:6]
+XO_NAMESPACE = _hash_name(FAMILY_NAME)[:6]
 
 
 def make_xo_address(name):
-    return XO_NAMESPACE + hash_name(name)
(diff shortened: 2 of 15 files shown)


dplumb94:
3) should be: 

> the data stored at an address is now a JSON string in the following format: {"**game_name**": "board,state,player_1,player_2"}.

boydjohnson:
The change to storing json objects in state doesn't seem like it will work. So when/if a hash collision happens, there will be two keys in the json object and since json objects are unordered (http://www.rfc-editor.org/rfc/rfc7159.txt) we can't be certain the order of bytes stored in state. This will cause two different validators to get a merkle root that is different for state changes that are the same. bson is a binary format that is somewhat like json and says that bson "documents" are ordered key:value pairs. http://bsonspec.org/spec.html

sdk/examples/xo_python/sawtooth_xo/processor/handler.py
```diff
@@ -214,66 +217,157 @@ def _store_game_data(state_store, name, board, state, player_1, player_2):
...
+def _split_up_board(board):
+    rows = _get_rows(board)
+    columns = _get_columns(rows)
+    diagonals = _get_diagonals(rows)
+
+    return rows, columns, diagonals
+
+
+def _is_win(board, mark):
+    '''
```
dplumb94: Very clever, but I'm wondering if we should simplify this? Since the goal of XO is to show an example of how a simple transaction processor is written, extra complexity in the game logic may be counterproductive. We would also have to implement this in the other languages.

nick-drozd: To me, the use of having a simple example like xo is to show how a transaction processor handles transactions and state data. The basic model that both xo and intkey follow is: get the transaction data, get the state data, do something, put the result back into state. From the perspective of showing off the system, it's relatively uninteresting what happens in the "do something" step, and the current documentation says so. It might even be more interesting if the implementations do differ, so long as they all produce the same output.

nick-drozd:
@boydjohnson raises a good point, so it's back the drawing board with the state format. I've pulled that commit, but the rest of this PR is unaffected.

nick-drozd:
Okay, hopefully this is the last update. @vaporos suggested that we use the following state format: `a_game,board,state,p1,p2|b_game,board,state,p1,p2|...`. Although it's a little unwieldy to implement, it has the benefit of not requiring any libraries or getting involved with any JSON baggage. On the other hand, this format has the disadvantage of potentially causing confusion regarding the use of csv in both the transaction format and the state format -- one might think that they are related, when really they can be completely independent choices.

Given this format, game names can't contain `,` or `|`. (Really, this restriction only applies to `|`, as an implementation would typically screen out names containing `,` when trying to decode the transaction.)

sdk/examples/xo_python/sawtooth_xo/xo_client.py
```diff
@@ -68,26 +68,38 @@ def take(self, name, space):
...
 
     def show(self, name):
-        address = make_xo_address(name)
-
-        result = self._send_request("state/{}".format(address))
-
         try:
-            return base64.b64decode(yaml.safe_load(result)["data"])
-
+            return self.list()[name]
```
aludvik: This downloads all of XO state to show one game, rather than construct the address to retrieve only one game, which is inefficient. I think it would be better to show how to query for only one address here.

bin/run_tests
```diff
@@ -302,6 +302,9 @@ test_xo() {
     copy_coverage .coverage.tp-xo-javascript
     run_docker_test xo-javascript-smoke -s integration_test
     copy_coverage .coverage.xo-javascript-smoke
+    # all languages
+    run_docker_test xo-all-lang-smoke
```
aludvik: I think this could just be called "xo-all-lang".

bin/run_tests
```diff
@@ -288,6 +291,19 @@ test_deployment() {
     run_docker_test systemd_services -s integration_test
 }
 
+test_xo() {
```
nick-drozd: If changes are made to the whole xo family (for instance, changing the namespace), it's convenient to be able to run all the xo tests at once.

The main `run_tests` module list, however, doesn't include the xo module:
```
MODULE_LIST="
    signing
    cli
    rest_api
    poet
    config
    validator
    go_sdk
    java_sdk
    javascript_sdk
    python_sdk
    integration
"
```
Someone working on xo can run `run_tests -m xo`, but nobody else needs to worry about it. The PoET unit tests can already be run in this way.

nick-drozd:
@aludvik I changed `two_families.yaml` to `two-families.yaml` to match the rest of the yaml files, and I changed `xo-smoke.yaml` to `xo-python-smoke.yaml` to match the rest of the smoke test yaml file. Neither of these changes are interesting enough to warrant separate commits, so I changed them when I changed the actual test files.

I changed the transaction format because it's confusing to have `take game 5` at the command line get changed into `game,take,5` when the transaction is sent.

Xo Javascript was organized somewhat haphazardly, so I moved things around to make it easier to change what needed to be changed. Think of it as a very mild refactor.

The rest client changes make it easier to verify state, and since the xo smoke test was updated to verify state, I included those changes in that commit. I've updated that commit slightly to make the change from `_base_url` to `url` clearer (it's convenient to be able to tell what url a client is pointed at).

sdk/examples/xo_python/sawtooth_xo/xo_client.py
```diff
@@ -30,8 +30,21 @@
 from sawtooth_xo.xo_exceptions import XoException
 
 
-def _sha512(data):
-    return hashlib.sha512(data).hexdigest()
+# namespace
+def _hash_name(name):
+    return hashlib.sha512(name.encode('utf-8')).hexdigest()
+
+FAMILY_NAME = 'xo'
```
vaporos: Hard-coding FAMILY_NAME and XO_NAMESPACE would make it more difficult to re-use this code to run xo games against another namespace.  In previous iterations, games could be played using an arbitrary namespace.

sdk/examples/xo_python/sawtooth_xo/xo_client.py
```diff
@@ -49,45 +62,47 @@ def __init__(self, base_url, keyfile):
...
-        return self._send_xo_txn(name, "create")
+        return self._send_xo_txn("create", name)
 
     def take(self, name, space):
-        return self._send_xo_txn(name, "take", space)
+        return self._send_xo_txn("take", name, space)
 
     def list(self):
-        xo_prefix = self._get_prefix()
+        state = yaml.safe_load(
```
vaporos: Should handle invalid YAML coming out of self._send_request() here.

sdk/examples/xo_python/sawtooth_xo/xo_client.py
```diff
@@ -164,3 +179,7 @@ def _create_batch_list(self, transactions):
             header_signature=signature
         )
         return BatchList(batches=[batch])
+
+
+def _sha512(data):
```
vaporos: Is this used?

sdk/examples/xo_python/sawtooth_xo/xo_message_factory.py
```diff
@@ -13,82 +13,110 @@
...
+
+
 class XoMessageFactory:
     def __init__(self, private=None, public=None):
         self._factory = MessageFactory(
-            encoding="csv-utf8",
-            family_name="xo",
-            family_version="1.0",
-            namespace="",
+            encoding='csv-utf8',
```
vaporos: Many unnecessary quote or whitespace changes in this block.

sdk/examples/xo_python/tests/test_tp_xo.py
```diff
@@ -13,85 +13,320 @@
...
+        self.send_transaction('blorp', 'invalid-action', 8)
+
+        self.expect_invalid()
+
+    def test_no_name(self):
+        self.send_transaction('take', '', 4)
+
+        self.expect_invalid()
+
+    @skip('JS throws InternalError instead of InvalidTransaction')
```
vaporos: Let's exclude the test instead of skipping it, and add a JIRA for adding it.

PR #510, opened 2017-05-22
Fixes to bring core/simulator into alignment with SGX

jjason:
In testing SGX implementation, I found inconsistencies in the core and simulator code that needed to be brought into alignment so that the SGX enclave would work.


jjason:
Rebase before merge

PR #512, opened 2017-05-23
Add Java XO Transaction Processor

dplumb94:


sdk/examples/xo_java/XoHandler.java
```diff
@@ -0,0 +1,447 @@
...
+      return new HashMap<>();
+    }
+  }
+
+  /**
+   * Helper function to store state data.
+   */
+  private void storeGameData(String gameName, String[] gameData, State stateStore)
+      throws InternalError, InvalidTransactionException {
+    //TODO: Store as an array rather than map
```
dplumb94: Methods involving JSON will have to be changed from maps to arrays to ensure a consistent ordering of game names that hash to the same address.

sdk/examples/xo_java/XoHandler.java
```diff
@@ -0,0 +1,445 @@
...
+  public Collection<String> getNameSpaces() {
+    ArrayList<String> namespaces = new ArrayList<>();
+    namespaces.add(this.xoNameSpace);
+    return namespaces;
+  }
+
+  @Override
+  public void apply(TpProcessRequest transactionRequest, State stateStore)
+      throws InvalidTransactionException, InternalError {
+    String[] transactionData = getUnpackedTransaction(transactionRequest);
```
peterschwarz: Instead of this being an area, this should probably just be a package-private class in this file, like so:

```
class TransactionData {
    final String action;
    final String gameName;
    final String space;

    TransactionData(String action, String gameName, String space) {
        this.action = action;
        this.gameName = gameName;
        this.space = space;
    }
}
```

Those fields can just be read directly, no need for getters, as their finals and private.

sdk/examples/xo_java/XoHandler.java
```diff
@@ -0,0 +1,445 @@
...
+
+    if (gameName.equals("")) {
+      throw new InvalidTransactionException("Name is required");
+    }
+
+    if (action.equals("")) {
+      throw new InvalidTransactionException("Action is required");
+    }
+
+    String[] stateData = getStateData(gameName, stateStore);
```
peterschwarz: Similar to the above.

sdk/examples/xo_java/XoHandler.java
```diff
@@ -0,0 +1,445 @@
...
+    String board = stateData[0];
+    String state = stateData[1];
+    String player1 = stateData[2];
+    String player2 = stateData[3];
+
+    String signer = getSignerPubkey(transactionRequest);
+    String[] updatedGameData = playXo(
```
peterschwarz: Similar to the above

sdk/examples/xo_java/XoHandler.java
```diff
@@ -0,0 +1,445 @@
...
+    }
+  }
+
+  /**
+   * Helper function to retrieve game name, action, and space from transaction request.
+   */
+  private String[] getUnpackedTransaction(TpProcessRequest transactionRequest) {
+    String[] transactionData = {"", "", ""};
+    String[] payload =  csvDecodeData(transactionRequest.getPayload().toStringUtf8());
+    System.arraycopy(payload, 0, transactionData, 0, payload.length);
```
peterschwarz: This `arraycopy` is unnecessary, as the underlying split is creating a new array every time. Likewise, if you create the data object class, as mentioned above, you can return an new instance of one of those here.

sdk/examples/xo_java/XoHandler.java
```diff
@@ -0,0 +1,458 @@
...
+    String address = makeGameAddress(transactionData.gameName);
+    // stateStore.get() returns a list.
+    // If no data has been stored yet at the given address, it will be empty.
+    String stateEntry = stateStore.get(
+        Collections.singletonList(address)
+    ).get(address).toStringUtf8();
+
+    GameData stateData = getStateData(stateEntry, transactionData.gameName);
+
+    GameData updatedGameData = playXo(
```
peterschwarz: You could just pass the objects to the function - no need to unwrap them here.

PR #513, opened 2017-05-23
Implement capability to use a validator config file

agunde406:
This commit adds a ValidatorConfig object, an example
validator config file, and updates the validator cli to
check for a config file and combine it with any command
line options set.

Important to note the public-uri is no longer "required" to
be set on the command line, but if it is not set by either
the command line or the config file the validator will
shutdown.

### DIFF
--- a/validator/packaging/validator.toml.example
+++ b/validator/packaging/validator.toml.example
@@ -1,0 +1,35 @@
+#
+# Sawtooth Lake -- Validator Configuration
+#
+
+# This file should exist in the defined config directory and allows
+# validators to be configured without the need for command line options.
+
+# The following is a possible example.
+
+# Bind is used to set the network and component endpoints. It should be a list
+# of strings in the format "option:endpoint", where the options are currently
+# network and component.
+bind = [
+  "network:tcp://127.0.0.1:8800",
+  "component:tcp://127.0.0.1:40000"
+]
+
+# The type of peering approach the validator should take. Choices are 'static'
+# which only attempts to peer with candidates provided with the peers option,
+# and 'dynamic' which will do topology buildouts. If 'dynamic' is provided,
+# any static peers will be processed first, prior to the topology buildout
+# starting.
+peering = "static"
+
+# Advertised network endpoint URL.
+endpoint = "tcp://127.0.0.1:8800"
+
+# Uri(s) to connect to in order to initially connect to the validator network,
+# in the format tcp://hostname:port. This is not needed in static peering mode
+# and defaults to None.
+seeds = ["tcp://127.0.0.1:8801"]
+
+# A list of peers to attempt to connect to in the format tcp://hostname:port.
+# It defaults to None.
+peers = ["tcp://127.0.0.1:8801"]
--- a/validator/sawtooth_validator/config/validator.py
+++ b/validator/sawtooth_validator/config/validator.py
@@ -1,0 +1,179 @@
+# Copyright 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+
+import collections
+import logging
+import os
+
+import toml
+
+from sawtooth_validator.exceptions import LocalConfigurationError
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+def load_default_validator_config():
+    return ValidatorConfig(
+        bind_network='tcp://127.0.0.1:8800',
+        bind_component='tcp://127.0.0.1:40000',
+        endpoint=None,
+        peering='static')
+
+
+def load_toml_validator_config(filename):
+    """Returns a ValidatorConfig created by loading a TOML file from the
+    filesystem.
+    """
+    if not os.path.exists(filename):
+        LOGGER.info(
+            "Skipping validator config loading from non-existent config file:"
+            " %s", filename)
+        return ValidatorConfig()
+
+    LOGGER.info("Loading validator information from config: %s", filename)
+
+    try:
+        with open(filename) as fd:
+            raw_config = fd.read()
+    except IOError as e:
+        raise LocalConfigurationError(
+            "Unable to load validator configuration file: {}".format(str(e)))
+
+    toml_config = toml.loads(raw_config)
+
+    invalid_keys = set(toml_config.keys()).difference(
+        ['bind', 'endpoint', 'peering', 'seeds', 'peers'])
+    if len(invalid_keys) > 0:
+        raise LocalConfigurationError(
+            "Invalid keys in validator config: "
+            "{}".format(", ".join(sorted(list(invalid_keys)))))
+    bind_network = None
+    bind_component = None
+    for bind in toml_config.get("bind"):
+        if "network" in bind:
+            bind_network = bind[bind.find(":")+1:]
+        if "component" in bind:
+            bind_component = bind[bind.find(":")+1:]
+
+    config = ValidatorConfig(
+         bind_network=bind_network,
+         bind_component=bind_component,
+         endpoint=toml_config.get("endpoint", None),
+         peering=toml_config.get("peering", None),
+         seeds=toml_config.get("seeds", None),
+         peers=toml_config.get("peers", None),
+    )
+
+    return config
+
+
+def merge_validator_config(configs):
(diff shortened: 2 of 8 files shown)


validator/packaging/validator.toml.example
```diff
@@ -0,0 +1,17 @@
...
+#
+
+# This file should exist in the defined config directory and allows
+# validators to be configured without the need for command line options.
+# It is important to note that public_uri must be set by either the config or
+# the command line.
+
+# The following is a possible example.
+
+network_endpoint = "tcp://127.0.0.1:8800"
```
peterschwarz: Each of these settings should be documented

validator/sawtooth_validator/config/validator.py
```diff
@@ -0,0 +1,187 @@
...
+        if config.component_endpoint is not None:
+            component_endpoint = config.component_endpoint
+        if config.public_uri is not None:
+            public_uri = config.public_uri
+        if config.peering is not None:
+            peering = config.peering
+        if config.join is not None:
+            join = config.join
+        if config.peers is not None:
+            peers = config.peers
```
aludvik: This might be cleaner in a loop. Might be good to make the list of options part of ValidatorConfig too. Something like:
```
merged = ValidatorConfig()
for config in reversed(configs):
  for field in config.list_fields():
    if config.getattr(field):
      merged.setattr(field, config.getattr(field))
```

validator/packaging/validator.toml.example
aludvik: Aren't they already documented in the validator CLI?. Can we instead reference the cli, explain how to go between the two, and remove the comments from this file so there isn't duplicate documentation?  Eg., "convert remove leading '--' and convert remaining '-' to '_'"

validator/sawtooth_validator/config/validator.py
```diff
@@ -0,0 +1,187 @@
...
+import toml
+
+from sawtooth_validator.exceptions import LocalConfigurationError
+from sawtooth_validator.config.path import get_default_path_config
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+def load_toml_validator_config(filename):
```
aludvik: Would it be clearer if the top level methods were instead class methods of  ValidatorConfig?

```
load_toml_validator_config -> ValidatorConfig.from_file()
merge_path_config -> ValidatorConfig.merge()
+ ValidatorConfig.from_opts()
```

If the default behaviour of from_file() is then to load from default_path_config.config_dir, then you have:
```
def load_validator_config(opts, config_dir=None):
  opt_conf = ValidatorConfig.from_opts(opts)
  file_conf = ValidatorConfig.from_file(config_dir)
  return ValidatorConfig.merge(opt_conf, file_conf)
```

validator/packaging/validator.toml.example
peterschwarz: Was thinking about this file in this context: https://github.com/hyperledger/sawtooth-core/blob/8f1ea3a22f0fcf1ccbe994c05c3b4469edaab026/validator/packaging/path.toml.example

validator/sawtooth_validator/config/validator.py
agunde406: The format of this file is based off of path.py and the PathConfig object.

vaporos: IMO, the current method is better, as it keeps the data separate from the load mechanism.

from_file() with no arguments is singleton-like, which is extremely poor in that it provides a methodology to take control away from the top-level application -- creating technical debt to be incurred later.  Better to force passing the correct path as determined by the top-level application logic.

Since this module need not know about application-level logic, from_opts() functionality is best left elsewhere.

If putting merge in the ValidatorConfig class, one could see it acting like set's union() and acting on self and others but still returning a new config.

validator/sawtooth_validator/server/cli.py
```diff
@@ -194,17 +200,22 @@ def main(args=sys.argv[1:]):
                            human_readable_name='Log'):
         init_errors = True
 
+    if validator_config.public_uri is None:
```
vaporos: Let's get rid of the need for specifying public_uri in the common case -- there is no reason we have to specify this unless we are binding to multiple interfaces via 0.0.0.0 or behind NAT.  If network_endpoint is set to tcp://0.0.0.0:port and public_uri is not been set, then we should exit.  In all other cases, public_uri should default to network_endpoint.

validator/sawtooth_validator/config/validator.py
```diff
@@ -0,0 +1,187 @@
...
+         network_endpoint=network_endpoint,
+         component_endpoint=component_endpoint,
+         public_uri=public_uri,
+         peering=peering,
+         join=join,
+         peers=peers
+    )
+
+
+def load_validator_config(opts, config_dir=None):
```
vaporos: If I read this correctly, this relies on opts (which came from argparse) having the same fields as ValidatorConfig.

A better method would be to explicitly create a ValidatorConfig() explicitly mapping opts fields to a ValidatorConfig and then merging together the two config objects.  This will produce a more obvious error should one of them change and not the other.

ValidatorConfig(network_endpoint=opts.network_endpoint, ...)

I would also move this function to cli.py since nothing else should know about argparse options.

validator/packaging/validator.toml.example
```diff
@@ -0,0 +1,36 @@
+#
```
vaporos: Maybe we should have a unit test which makes sure the values here reflect the actual default values to prevent them from drifting auto-of-sync.

The defaults are probably scattered around various modules currently; we should probably move them to config/validator.py (load_default_validator_config?) and merge them in as well.

validator/sawtooth_validator/server/cli.py
jsmitchell: This also needs to handle the case of network_endpoint specifying interface names (e.g. tcp://eth0:8800).

validator/packaging/validator.toml.example
agunde406: Unit test Added and load_default_validator_config added to config/validator.

validator/sawtooth_validator/config/validator.py
agunde406: I disagree with the above, this makes it harder to see what is part of the multiple configs. ValdiatorConfig already has a way to get a dictionary of the fields.

agunde406:
Updated to include the above requested changes, except for changing the --join option as a better name could not be decided on.

validator/tests/unit3/test_config/tests.py
```diff
@@ -135,3 +137,93 @@ def test_path_config_invalid_setting_in_file(self):
...
             os.environ.update(orig_environ)
             shutil.rmtree(directory)
+
+
+class TestValidatorConfig(unittest.TestCase):
+
+    def test_validator_config_defaults(self):
+        """Tests the default validator configuration when no other configs.
+        The defaults should be as follows:
+            - network_endpoint = "tcp://127.0.0.1:8800"
```
jsmitchell: Still have the prior names here.

validator/sawtooth_validator/server/cli.py
```diff
@@ -137,6 +138,26 @@ def _split_comma_append_args(arg_list):
...
+    default_validator_config = load_default_validator_config()
+    conf_file = os.path.join(config_dir, 'validator.toml')
+
+    toml_config = load_toml_validator_config(conf_file)
+
+    return merge_validator_config(
+        configs=[first_config, toml_config, default_validator_config])
+
+
+def create_valdiator_config(opts):
```
jsmitchell: s/valdiator/validator/

validator/sawtooth_validator/server/cli.py
```diff
@@ -157,6 +178,14 @@ def main(args=sys.argv[1:]):
         LOGGER.error(str(local_config_err))
         sys.exit(1)
 
+    try:
+        opts_config = create_valdiator_config(opts)
```
jsmitchell: valdiator

validator/sawtooth_validator/server/cli.py
```diff
@@ -137,6 +138,26 @@ def _split_comma_append_args(arg_list):
...
+
+    toml_config = load_toml_validator_config(conf_file)
+
+    return merge_validator_config(
+        configs=[first_config, toml_config, default_validator_config])
+
+
+def create_valdiator_config(opts):
+    return ValidatorConfig(
+        bind_network=opts.network_endpoint,
```
jsmitchell: Are you going to change the command line args, or will that be a separate story?

agunde406: I am working on it now, but it will be in a separate PR because of the amount of docker file changes necessary.

validator/packaging/validator.toml.example
```diff
@@ -0,0 +1,35 @@
...
+
+# This file should exist in the defined config directory and allows
+# validators to be configured without the need for command line options.
+
+# The following is a possible example.
+
+# Bind is used to set the network and component endpoints. It should be a list
+# of strings in the format "option:endpoint", where the options are currently
+# network and component.
+    bind = [
```
vaporos: Should remove the indentation on these settings.

PR #520, opened 2017-05-24
Basic CORS Support

peterschwarz:
Basic CORS support is required if a user is going to query a running
REST API from another server (e.g. from a single-page-app that is
submitting transactions, but hosted elsewhere).

This is to support the OMI Summer Lab.  A task to add configuration options to the REST API cli tool has been created in [STL-313](https://jira.hyperledger.org/browse/STL-313)

### DIFF
--- a/rest_api/sawtooth_rest_api/route_handlers.py
+++ b/rest_api/sawtooth_rest_api/route_handlers.py
@@ -127,6 +127,7 @@ class RouteHandler(object):
             link = link.replace('batch_status', 'batches')
 
         return self._wrap_response(
+            request,
             data=data,
             metadata={'link': link},
             status=status)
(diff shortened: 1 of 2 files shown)


rest_api/sawtooth_rest_api/route_handlers.py
```diff
@@ -505,17 +511,27 @@ def _check_status_errors(proto, content, error_traps=None):
...
         envelope = metadata or {}
 
         if data is not None:
             envelope['data'] = data
 
+        headers = None
+        if 'Origin' in request.headers:
+            headers = {
+                'Access-Control-Allow-Origin': request.headers['Origin'],
+                "Access-Control-Allow-Methods": "GET,POST",
```
delventhalz: This could be refactored into a method on the RouteHandler, or maybe a separate utility class, and then called here and in `rest_api.py`.

peterschwarz: Refactored

PR #523, opened 2017-05-24
Make attestation verification report public key configurable via config settings

jjason:
This change updates the validator registry transaction processor so that it will first look in the configuration settings on the blockchain for a value associated with the setting `sawtooth.poet.report_public_key_pem`.  If a value is not found or if the value does not parse to a public key, the validator registry transaction processor will reject the transaction.

To test this locally, start up a validator as such (the instructions assume you are running inside Vagrant VM):
```
sawtooth admin keygen validator
sawtooth config genesis --key ~/sawtooth/keys/validator.priv -o config-genesis.batch
sawtooth config proposal create -k ~/sawtooth/keys/validator.priv sawtooth.consensus.algorithm=poet sawtooth.poet.report_public_key_pem="$(cat /project/sawtooth-core/consensus/poet/simulator/packaging/simulator_rk_pub.pem)" -o config.batch
poet genesis -k ~/sawtooth/keys/validator.priv -o poet_genesis.batch
sawtooth admin genesis config-genesis.batch config.batch poet_genesis.batch
validator -vv --public-uri http://localhost:8080
```
NOTE: order is important in the `sawtooth admin genesis` command as the proposal to create `sawtooth.poet.report_public_key` (`config.batch`) has to appear before the validator registry transaction for the genesis validator (`poet_genesis.batch`) as the validator registry TP needs to be able to fetch the value when verifying the VR transaction.

Start up a config TP:
`tp_config -vv tcp://localhost:40000`

Start up a validator registry TP:
`tp_validator_registry -vv tcp://localhost:40000`

All three should start up successfully.  You can play around by setting `sawtooth.poet.report_public_key_pem` to a key that will result in the signature of the attestation verification report being invalid.  For example, `ssh-keygen -t rsa` and then use the `.pub` file from it in the above set of commands when bringing up the validator.


consensus/poet/cli/sawtooth_poet_cli/genesis.py
```diff
@@ -135,15 +135,21 @@ def do_genesis(args):
     # Create a transaction header and transaction for the validator
     # registry update amd then hand it off to the batch publisher to
     # send out.
-    addresses = [validator_entry_address, VALIDATOR_MAP_ADDRESS]
+    output_addresses = [validator_entry_address, VALIDATOR_MAP_ADDRESS]
+    input_addresses = \
```
jsmitchell: This is nice. cc @peterschwarz

consensus/poet/families/sawtooth_validator_registry/validator_reg_message_factory.py
```diff
@@ -287,3 +289,21 @@ def create_set_request_validator_map(self):
     def create_set_response_validator_map(self):
         addresses = [self._key_to_address("validator_map")]
         return self._factory.create_set_response(addresses)
+
```
jsmitchell: It looks like you accidentally changed the mode of this file to 755.

consensus/poet/families/sawtooth_validator_registry/validator_reg_message_factory.py
```diff
@@ -287,3 +289,21 @@ def create_set_request_validator_map(self):
...
+    def create_get_request_report_key_pem(self):
+        return \
+            self._factory.create_get_request(
+                ['000000a87cb5eafdcca6a87ccc804f5546a'
```
jsmitchell: Should this be in a config file?

consensus/poet/families/sawtooth_validator_registry/validator_registry/processor/handler.py
```diff
@@ -45,6 +46,62 @@
...
+
+def _config_short_hash(byte_str):
+    # Computes the SHA 256 hash and truncates to be the length
+    # of an address part (see _config_key_to_address for information on
+    return hashlib.sha256(byte_str).hexdigest()[:_CONFIG_ADDRESS_PART_SIZE]
+
+_CONFIG_ADDRESS_PADDING = _config_short_hash(byte_str=b'')
+
+
+def _config_key_to_address(key):
```
jsmitchell: @peterschwarz it would feel better if this were a concern of config code that was imported here

consensus/poet/simulator/simulator_rk_pub.pem
```diff
@@ -0,0 +1,9 @@
+-----BEGIN PUBLIC KEY-----
```
jsmitchell: If this needs to be checked in, can we provide some comments (ideally in the file, but also maybe next to it) about what this is and how it was generated?

jsmitchell: Alternatively, could this be a setting in a local config file for simulator?

consensus/poet/families/sawtooth_validator_registry/validator_reg_message_factory.py
jjason: Crud.  I don't know how I did that.....

consensus/poet/families/sawtooth_validator_registry/validator_registry/processor/handler.py
jjason: @jsmitchell I had a quick Slack chat with @peterschwarz regarding that we should consider future story to bring config address, etc.., stuff out to somewhere common.

consensus/poet/simulator/simulator_rk_pub.pem
jjason: I went back and forth on whether or not to check this in.  We can discuss what the proper thing to do is as I have the same dilemma with regard to the IAS report public key.  We need to distribute it (so that when bringing up the validator network we can put it on the blockchain in config) and so was going to put it in repo so that we can package it up.  Same goes for the simulator report public key, although not as important as the validator registry TP will fall back to this key anyway.

consensus/poet/families/sawtooth_validator_registry/validator_reg_message_factory.py
jjason: IDK.  This is only used for running tests.

jsmitchell: how can one tell that this is test code?

jjason: I only figured it was test code based upon looking at how it was used and by grep-ing the code for any import of the code.  This file could be moved to `consensus/poet/families/tests`.

consensus/poet/families/sawtooth_validator_registry/validator_reg_message_factory.py
jjason: Okay.  I think I got the file mode fixed.  I had originally edited that file in Windows, which does't play nice with file mode.  I don't know if that is what caused it.

consensus/poet/families/sawtooth_validator_registry/validator_registry/processor/handler.py
jjason: For this particular file, because it is a TP didn't want to import anything outside the SDK or PoET packages.

consensus/poet/families/sawtooth_validator_registry/validator_reg_message_factory.py
jsmitchell: @agunde406 what do you think?

agunde406: Yes, this is only meant to be used in testing. It needed a lot of it to be hardcoded to be able to run the tp tests. I think it would make sense to move it into the tests directory.

consensus/poet/families/sawtooth_validator_registry/validator_registry/processor/handler.py
peterschwarz: We may want to consider putting some convienence methods for settings in the SDK's, as it's is possible that this sort of use will be common.

jjason: @peterschwarz As part of this PR or future story? I vote for future story as I think this is larger than just in the SDK.  The PoET block publisher also uses it and so would be nice if it was a common area for both TPs and the core.

peterschwarz: Absolutely a future story.

jjason:
Pushed new PR.  Move PEM file and added to setup.py packaging.  Not certain if integration (poet-smoke and dynamic-network) will pass.  We'll see...

consensus/poet/families/sawtooth_validator_registry/validator_reg_message_factory.py
jjason: Moved file.

jjason:
Let's see if tests pass now =)

jjason:
I think PR is finally ready.

consensus/poet/families/tests/validator_reg_message_factory.py
```diff
@@ -287,3 +289,34 @@ def create_set_request_validator_map(self):
...
+            setting.entries.extend([entry])
+
+        data = setting.SerializeToString()
+        return self._factory.create_get_response(
+            {'000000a87cb5eafdcca6a87ccc804f5546a'
+             'b8e97a7d614626e4500e3b0c44298fc1c14': data})
+
+    def create_get_response_simulator_report_key_pem(self, pem=None):
+        pem = '''-----BEGIN PUBLIC KEY-----
+MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEArMvzZi8GT+lI9KeZiInn
```
agunde406: This pem has a bad indention.

jjason: The indentation is because it is a multi-line string and if I indent, it will insert spaces after the newline causing the PEM to not parse correctly into a valid public key.

jsmitchell:
Tested locally.

consensus/poet/families/sawtooth_validator_registry/validator_registry/processor/handler.py
jsmitchell: @peterschwarz can you create that story?

consensus/poet/families/tests/validator_reg_message_factory.py
jjason: The other option was to do what I had done previously, which was embed the newlines in the string, but then that caused the lines to extend beyond 80.  I chose the multi-line string instead of breaking the individual lines up in the middle.

agunde406: Ah Okay.

jjason:
Quick rebase for sanity check before merge

consensus/poet/families/sawtooth_validator_registry/validator_registry/processor/handler.py
peterschwarz: Created [STL-324](https://jira.hyperledger.org/browse/STL-324) for adding these utilities.

PR #526, opened 2017-05-25
Add docker file to support sphinx builds

cianx:


aludvik:
https://github.com/hyperledger/sawtooth-core/blob/master/ci/sawtooth-build-docs

cianx:
@aludvik Nope, I did not find that one when I looked.

PR #528, opened 2017-05-25
Fix XO create wait option

dplumb94:
Previously the --wait cli option for create was parsed but nothing was done with it. Now, when this argument is parsed the cli will pause until the transaction has been committed to the block or an error is returned.

Note: _Currently when the XO transaction processor finds that a transaction is invalid (such as in the case of a "Game already exists" error) the Rest API status remains as 'PENDING'. This means that the cli will pause until the specified timeout has elapsed, or until the batch cache is cleared (in the case of the --wait wait option being passed with no argument). This should be fixed in the future when invalid transactions are given an 'INVALID' status in the Rest API._

### DIFF
--- a/sdk/examples/xo_python/sawtooth_xo/xo_client.py
+++ b/sdk/examples/xo_python/sawtooth_xo/xo_client.py
@@ -48,8 +48,8 @@ class XoClient:
 
         self._public_key = signing.generate_pubkey(self._private_key)
 
-    def create(self, name):
-        return self._send_xo_txn(name, "create")
+    def create(self, name, wait=None):
+        return self._send_xo_txn(name, "create", wait=wait)
 
     def take(self, name, space):
         return self._send_xo_txn(name, "take", space)
@@ -79,6 +79,15 @@ class XoClient:
 
         except BaseException:
             return None
+
+    def _get_status(self, batch_id, wait):
+        try:
+            result = self._send_request(
+                'batch_status?id={}&wait={}'.format(batch_id, wait)
+            )
+            return yaml.safe_load(result)["data"]
+        except BaseException as err:
+            raise XoException(err)
 
     def _get_prefix(self):
         return _sha512('xo'.encode('utf-8'))[0:6]
@@ -110,8 +119,7 @@ class XoClient:
 
         return result.text
 
-    def _send_xo_txn(self, name, action, space=""):
-
+    def _send_xo_txn(self, name, action, space="", wait=None):
         # Serialization is just a delimited utf-8 encoded string
         payload = ",".join([name, action, str(space)]).encode()
 
@@ -140,13 +148,40 @@ class XoClient:
         )
 
         batch_list = self._create_batch_list([transaction])
+        batch_id = batch_list.batches[0].header_signature
 
-        result = self._send_request(
+        if wait and wait > 0:
+            wait_time = 0
+            start_time = time.time()
+            self._send_request(
+                "batches", batch_list.SerializeToString(),
+                'application/octet-stream'
+            )
+            while wait_time < wait:
+                status = self._get_status(
+                    batch_id,
+                    wait - int(wait_time)
+                )
+                wait_time = time.time() - start_time
+
+                if status[batch_id] == 'COMMITTED':
+                    return 'Game created in {:.6} sec'.format(wait_time)
+                elif status[batch_id] == 'INVALID':
+                    return ('Error: You chose an invalid game name. '
+                            'Try again with a different name')
+                elif status[batch_id] == 'UNKNOWN':
+                    return ('Error: Something went wrong with your game. '
+                            'Try again with a different name.')
+                # Wait a moment so as not to hammer the Rest Api
+                time.sleep(0.2)
+
+            return ('Timed out while waiting for game to be created. '
+                    'It may need to be resubmitted.')
+
+        return self._send_request(
             "batches", batch_list.SerializeToString(),
             'application/octet-stream'
         )
-
-        return result
 
     def _create_batch_list(self, transactions):
         transaction_signatures = [t.header_signature for t in transactions]
--- a/sdk/examples/xo_python/sawtooth_xo/xo_cli.py
+++ b/sdk/examples/xo_python/sawtooth_xo/xo_cli.py
@@ -81,9 +81,10 @@ def add_create_parser(subparsers, parent_parser):
 
     parser.add_argument(
         '--wait',
-        action='store_true',
-        default=False,
-        help='wait for this commit before exiting')
+        nargs='?',
+        const=sys.maxsize,
+        type=int,
+        help='wait for game to commit, set an integer to specify a timeout')
 
 
 def add_init_parser(subparsers, parent_parser):
@@ -272,7 +273,12 @@ def do_create(args, config):
 
     client = XoClient(base_url=url,
                       keyfile=key_file)
-    response = client.create(name)
+
+    if args.wait and args.wait > 0:
+        response = client.create(name, wait=args.wait)
+    else:
+        response = client.create(name)
+
     print("Response: {}".format(response))
 
 


sdk/examples/xo_python/sawtooth_xo/xo_client.py
```diff
@@ -140,13 +163,37 @@ def _send_xo_txn(self, name, action, space=""):
...
-        )
-
-        return result
+        batch_ids = [b.header_signature for b in batch_list.batches]
+
+        if wait and wait > 0:
+            wait_time = 0
+            start_time = time.time()
+            self._send_request(
+                "batches", batch_list.SerializeToString(),
```
delventhalz: This request should be sent with the wait parameter. The simplest way to do this is swap `"batches"` for `"batches?wait={}".format(wait)`.

A somewhat more robust implementation would be to refactor `XoClient._send_request` to take a dict of query parameters (or just arbitrary `**kwargs`), and then parse those into the url properly (`?` before the first query, `&` between subsequent queries, `=` between key and value).

If you _don't_ send this request with a wait query, you will get a response immediately, and you'll be hammering the REST API every 0.2 seconds.

sdk/examples/xo_python/sawtooth_xo/xo_client.py
```diff
@@ -80,6 +82,28 @@ def show(self, name):
...
+                headers = {'Content-Type': 'application/octet-stream'}
+            else:
+                batch_ids = json.dumps(batch_ids).encode()
+                headers = {'Content-Type': 'application/json'}
+            headers['Content-Length'] = '%d' % len(batch_ids)
+
+            result = requests.post(url, data=batch_ids, headers=headers)
+            code, status = (result.status_code, result.json()["data"])
+
+            if code == 200 or code == 201 or code == 202:
```
delventhalz: Even though you are sending it as a `POST`, any request to `/batch_status` is considered a `GET` in spirit, and will only return a `200` (or an error code).

_(A `POST /batches`, which actually submits something new, would return a `201` or `202`, but not a `200`)_

sdk/examples/xo_python/sawtooth_xo/xo_client.py
```diff
@@ -80,6 +82,28 @@ def show(self, name):
...
+            url = "http://{}/{}".format(self._base_url, 'batch_status')
+
+            if isinstance(batch_ids, bytes):
+                headers = {'Content-Type': 'application/octet-stream'}
+            else:
+                batch_ids = json.dumps(batch_ids).encode()
+                headers = {'Content-Type': 'application/json'}
+            headers['Content-Length'] = '%d' % len(batch_ids)
+
+            result = requests.post(url, data=batch_ids, headers=headers)
```
delventhalz: If you are only every checking the status of one batch, this would be better formatted as a `GET /batch_status?id={batch id}&wait={wait time}`. The option to `POST /batch_status` is only there as a workaround to allow checking on any number of ids.  Since URLs are limited to ~2048 characters, and our ids are 128 characters long, you won't be able to check on more than ~15 ids with a `GET` request. If you don't need to check on more than that though, a `POST` is overkill.

sdk/examples/xo_python/sawtooth_xo/xo_client.py
```diff
@@ -80,6 +82,28 @@ def show(self, name):
         except BaseException:
             return None
 
+    def _get_statuses(self, batch_ids):
+        try:
+            url = "http://{}/{}".format(self._base_url, 'batch_status')
+
+            if isinstance(batch_ids, bytes):
+                headers = {'Content-Type': 'application/octet-stream'}
```
delventhalz: `application/octet-stream` is only for a `POST /batches`. A `POST /batch_status` will _only_ accept `application/json`. So even if somebody submitted `batch_ids` as bytes, you would only get back a `400` error.

sdk/examples/xo_python/sawtooth_xo/xo_client.py
```diff
@@ -140,13 +163,37 @@ def _send_xo_txn(self, name, action, space=""):
...
+
+        if wait and wait > 0:
+            wait_time = 0
+            start_time = time.time()
+            self._send_request(
+                "batches", batch_list.SerializeToString(),
+                'application/octet-stream'
+            )
+            while wait_time < wait:
+                statuses = self._get_statuses(batch_ids)
```
delventhalz: Same thing here. `XoClient._get_statuses` needs to be able to take a wait parameter and put it into the URL as a query.

sdk/examples/xo_python/sawtooth_xo/xo_client.py
```diff
@@ -140,13 +163,37 @@ def _send_xo_txn(self, name, action, space=""):
...
+                wait_time = time.time() - start_time
+
+                if all(s == 'COMMITTED' for s in statuses.values()):
+                    return 'Batches committed in {:.6} sec'.format(wait_time)
```
delventhalz: Do we want to expose the concept of _"batches"_ to the casual `xo` user? Should this say something like _"Game created in . . ."_? Is reporting back how long at all it took actually important here?

sdk/examples/xo_python/sawtooth_xo/xo_client.py
```diff
@@ -140,13 +163,37 @@ def _send_xo_txn(self, name, action, space=""):
...
+                elif any(s == 'INVALID' for s in statuses.values()):
+                    return 'Error: Invalid batch'
+                elif any(s == 'UNKNOWN' for s in statuses.values()):
+                    return 'Error: Unknown batch'
```
delventhalz: I would like these errors to be more informative, and `xo` specific. The concepts of _invalid_ or _unknown_ batches are possibly meaningless to the person using `xo`. I would probably say something along the lines of _"Something went wrong while creating your games. Try again later."_ Or whatever makes sense in an `xo` context.

sdk/examples/xo_python/sawtooth_xo/xo_client.py
```diff
@@ -140,13 +163,37 @@ def _send_xo_txn(self, name, action, space=""):
...
+
+                # Wait a moment so as not to hammer the Rest Api
+                time.sleep(0.2)
+
+            for batch_id, status in statuses.items():
+                print('{:128.128}  {:8.8}'.format(batch_id, status))
```
delventhalz: Does it make sense to print out the batch statuses here? Is that what an `xo` user will want to see?

sdk/examples/xo_python/sawtooth_xo/xo_client.py
```diff
@@ -140,13 +163,37 @@ def _send_xo_txn(self, name, action, space=""):
...
+            return 'Wait timed out! Some batches have not been committed...'
```
delventhalz: _"Timed out while waiting for game to be created. It may need to be resubmitted."_

delventhalz:
@agunde406 @dplumb94 The `wait` query will inevitably time out (usually after 300s at the longest). The _Sawtooth CLI_ implementation that Darian is borrowing from is designed allow arbitrarily long wait times (`--wait 10000`), or even indefinite waits (`--wait`).

In order to do that, you need to resubmit the request every time it times out, which is what most of this complication is from. That said, you should still use the `wait` query, because otherwise you are just blasting out HTTP requests every ~0.2s, instead of every ~300s. More details are in my comments above.

sdk/examples/xo_python/sawtooth_xo/xo_client.py
```diff
@@ -140,14 +149,41 @@ def _send_xo_txn(self, name, action, space=""):
         )
 
         batch_list = self._create_batch_list([transaction])
-
-        result = self._send_request(
+        batch_ids = [b.header_signature for b in batch_list.batches]
```
delventhalz: -> `batch_id = batch_list.batches[0].header_signature`

sdk/examples/xo_python/sawtooth_xo/xo_client.py
```diff
@@ -140,14 +149,41 @@ def _send_xo_txn(self, name, action, space=""):
...
+        if wait and wait > 0:
+            wait_time = 0
+            start_time = time.time()
+            self._send_request(
+                "batches", batch_list.SerializeToString(),
+                'application/octet-stream'
+            )
+            while wait_time < wait:
+                status = self._get_status(
+                    batch_ids[0],
```
delventhalz: -> `batch_id,`

sdk/examples/xo_python/sawtooth_xo/xo_client.py
```diff
@@ -140,14 +149,41 @@ def _send_xo_txn(self, name, action, space=""):
...
+                    wait - int(wait_time)
+                )
+                wait_time = time.time() - start_time
+
+                if all(s == 'COMMITTED' for s in status.values()):
```
delventhalz: -> `if status[batch_id] == 'COMMITTED`

sdk/examples/xo_python/sawtooth_xo/xo_client.py
```diff
@@ -140,14 +149,41 @@ def _send_xo_txn(self, name, action, space=""):
...
+                    return 'Game created in {:.6} sec'.format(wait_time)
+                elif any(s == 'INVALID' for s in status.values()):
```
delventhalz: -> `elif status[batch_id] == 'INVALID`

sdk/examples/xo_python/sawtooth_xo/xo_client.py
```diff
@@ -140,14 +149,41 @@ def _send_xo_txn(self, name, action, space=""):
...
+                    return ('Error: You chose an invalid game name. '
+                            'Try again with a different name')
+                elif any(s == 'UNKNOWN' for s in status.values()):
```
delventhalz: -> `elif status[batch_id] == 'UNKNOWN`

sdk/examples/xo_python/sawtooth_xo/xo_client.py
```diff
@@ -19,6 +19,7 @@
 import requests
 import yaml
 
+
```
delventhalz: Does this pass lint? I think there should only be one line break between imports.

PR #529, opened 2017-05-25
Implement capability to use a rest api config file

agunde406:
This commit adds a RestApiConfig object, an example
rest_api config file, and updates the rest_api to check
for a config file and combine it with any command line
options set.

### DIFF
--- a/rest_api/packaging/rest_api.toml.example
+++ b/rest_api/packaging/rest_api.toml.example
@@ -1,0 +1,12 @@
+#
+# Sawtooth Lake -- REST API Configuration
+#
+
+# The port and host  for the api to run on
+#   bind = ["127.0.0.1:8800"]
+
+# The url to connect to a running Validator
+#   connect = "tcp://localhost:40000"
+
+# Seconds to wait for a validator response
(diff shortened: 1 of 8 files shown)


rest_api/packaging/rest_api.toml.example
```diff
@@ -0,0 +1,15 @@
+#
+# Sawtooth Lake -- REST API Configuration
+#
+
+# The port for the api to run on
+#   port = 8080
```
vaporos: The port/host configuration here is significantly different than the *_endpoint configuration options in the validator, but both are what to listen on.  We should probably make them consistent.

rest_api/packaging/rest_api.toml.example
```diff
@@ -0,0 +1,15 @@
...
+
+# The host for the api to run on
+#   host = "127.0.0.1"
+
+# The url to connect to a running Validator
+#   stream_url = "tcp://localhost:40000"
```
vaporos: We should standardize on what we call this across all components (TPs, rest api, etc.) connecting to the component endpoint.

PR #530, opened 2017-05-26
Fix the comment of the nonce field

feihujiang:

The same transaction is not clearly defined. Changed it.

### DIFF
--- a/protos/transaction.proto
+++ b/protos/transaction.proto
@@ -39,8 +39,8 @@ message TransactionHeader {
     // what addresses the transaction processor is allowed to read from.
     repeated string inputs = 5;
 
-    // A random string that will allow the same transaction to be submitted
-    // several times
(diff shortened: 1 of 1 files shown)


protos/transaction.proto
```diff
@@ -39,8 +39,8 @@ message TransactionHeader {
     // what addresses the transaction processor is allowed to read from.
     repeated string inputs = 5;
 
-    // A random string that will allow the same transaction to be submitted
-    // several times
+    // A random string that will allow transactions have the same fields
```
jsmitchell: How about: // A random string that provides uniqueness for transactions with otherwise identical fields.

feihujiang: Yea, your recommendation is very good.

feihujiang:
@jsmitchell I changed as your recommendation. Please take a look.

PR #532, opened 2017-05-26
Change privkey format

dcmiddle:
Move from wif to hex

### DIFF
--- a/cli/tests/test_config.py
+++ b/cli/tests/test_config.py
@@ -25,7 +25,7 @@ from sawtooth_cli.protobuf.batch_pb2 import BatchList
 from sawtooth_cli import config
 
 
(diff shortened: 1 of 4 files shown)


vaporos:
We should consider backwards compatibility (users might have keys from sawtooth keygen or sawtooth admin keygen existing in their environments) and compatibility with the javascript SDK.

aludvik:
My opinion is that we move to hex for simplicity and easier compatibility across languages. We are already encoding pubkeys and signatures to hex strings, so I am not sure why we would want to keep private keys encoded differently.

In response to @vaporos's comment, the JavaScript SDK assumes the key is in hex, as does the Go SDK.

dcmiddle:
Fair point on doing a friendlier change for compatibility. This was kind of knee jerk. I will take this down for now and create a proper jira story.

PR #533, opened 2017-05-26
Go SDK Tweaks

aludvik:


sdk/go/src/sawtooth_sdk/client/encoder.go
```diff
@@ -243,6 +246,12 @@ func Decode(encoded, encoding string) []byte {
 			panic("Couldn't decode")
 		}
 		return bytes
+	case "base64":
+		bytes, err := base64.StdEncoding.DecodeString(encoded)
+		if err != nil {
+			panic("Couldn't decode")
```
vaporos: This Decode() func seems to do the same thing as base64.StdEncoding.DecodeString() only panic instead of returning an error.  Seems like we should instead just use the libraries directly so clients can handle the errors.

sdk/go/src/sawtooth_sdk/messaging/connection.go
```diff
@@ -76,7 +76,6 @@ func NewConnection(context *zmq.Context, t zmq.Type, uri string) (*Connection, e
 
 	identity := GenerateId()
 	socket.SetIdentity(identity)
-	logger.Debug("Socket Identity set to ", identity)
```
ojalatodd: General question: Why did you remove the Debug level logger?

sdk/go/src/sawtooth_sdk/client/encoder.go
aludvik: Good point, I'll change it.

sdk/go/src/sawtooth_sdk/messaging/connection.go
aludvik: For anyone using the SDK, it generates a lot of logging output unrelated to what they are trying to use it for. The alternative would be to put more work into the sdk/logging package so that log levels can be set per-package or per-file.

aludvik:
Dropped the base64 commit and rename commit and added a commit to change the encode/decode methods. See commit message for rationale.

sdk/go/src/sawtooth_sdk/client/encoder.go
```diff
@@ -224,28 +224,20 @@ func ParseBatches(b []byte) ([]*Batch, error) {
...
-// Encode converts the given byte slice to a string using the given encoding.
-func Encode(key []byte, encoding string) string {
-	switch encoding {
-	case "hex":
-		return strings.ToLower(hex.EncodeToString(key))
-	default:
-		panic("Unsupported encoding: " + encoding)
-	}
+// MustEncode converts the given byte slice to a lower-case hex-encoded string.
+// Panics if encoding failes.
```
agunde406: failes -> fails

PR #537, opened 2017-05-26
STL-26 Update the context manager to pass unit tests.

boydjohnson:
The schedulers -- the SerialScheduler in master, and the ParallelScheduler in the parallel-scheduler branch -- pass state information to the TransactionExecutor along with each transaction. The TransactionExecutor calls create_context on the context manager, passing in the inputs and outputs from the transaction as well as the context ids for transactions that this transaction depends on. Before this PR, each context holds all addresses and values (bytes) that has been an input up to that point as well as addresses and values that has been an output. This works in the serial case. In the parallel case, distinctions need to be made between whether the address was set in that context or was in that context and merely read from. Certain implementation decisions were made and I'll explain them here:
* Both create context and squash use breadth-first search. For create context, it is used to look back through the graph of contexts to find the values for addresses that are inputs for that context's transaction. For squash, the same, but at that point, if an address was never set in any context, it is never accessed.
* The SerialScheduler drops a lot of information, so that when the contexts need to be combined to make a new state hash or new context, only the context manager has the necessary information to combine contexts correctly and produce the correct state hash. The ParallelScheduler holds on to more information, where it would be possible and maybe desirable for the scheduler implementations to give the context manager all of the contexts that need to be combined at the end of processing, with an ordering that would be known to the scheduler.


validator/sawtooth_validator/execution/context_manager.py
```diff
@@ -79,89 +85,220 @@ def session_id(self):
...
+            address_values (list of tuple): The tuple is string, bytes of the
+                address and value.
+        """
+
+        with self._lock:
+            for add, val in address_values:
+                self._state[add] = _ContextFuture(address=add, result=val)
+
+    def set_from_tree(self, address_value_dict):
+        """Called in _ContextWriter to set the future with the value from
```
aludvik: Is _ContextWriter the only thing allowed to call this method?

validator/sawtooth_validator/execution/context_manager.py
```diff
@@ -209,46 +346,83 @@ def create_context(self, state_hash, base_contexts, inputs, outputs):
...
-            prior_state_results[k] = value
-
-        context.set_futures(prior_state_results)
+
+        contexts_in_chain = deque()
+        contexts_in_chain.extend(base_contexts)
+        reads = list(addresses_to_find)
+        address_values = []
+        context_ids_already_searched = []
+        context_ids_already_searched.extend(base_contexts)
```
aludvik: Why use extend and not `context_ids_already_searched = base_contexts[:]`?

validator/sawtooth_validator/execution/context_manager.py
```diff
@@ -421,48 +607,120 @@ def run(self):
...
 
 
 class _ContextFuture(object):
-    def __init__(self, address, wait_for_tree=False):
+    """Controls access to bytes set in the _result variable. The booleans
+     that are flipped in set_result, based on whether the value is being set
+     from the merkle tree or a direct set on the context manager are needed
+     to later determine whether the value was set in that context or was
+     looked up as a new address location from the merkle tree and then only
+     read from, not set.
```
aludvik: Can you rephrase the above paragraph? I am struggling to understand it.

validator/sawtooth_validator/execution/context_manager.py
```diff
@@ -421,48 +607,120 @@ def run(self):
...
+
+    In any context the lifecycle of a _ContextFuture can be several paths:
```
aludvik: In any context the _life_ of a _ContextFuture can _follow_ several paths:

aludvik: In general, what determines which path it takes? The diagrams below are good, but are sparse on explanation.

validator/sawtooth_validator/execution/context_manager.py
```diff
@@ -421,48 +607,120 @@ def run(self):
...
-                or if it is set from prior state.
+            from_tree (bool): Whether the value is being set by a read from
+                the merkle tree.
 
         Returns:
             None
         """
+
+        if self._read_only:
+            return
```
aludvik: Should this throw an error?

validator/sawtooth_validator/execution/context_manager.py
```diff
@@ -421,48 +607,120 @@ def run(self):
...
+
         with self._condition:
-            if self._wait_for_tree and not from_tree:
-                self._condition.wait_for(lambda: self._tree_has_set)
-            self._result = result
-            self._result_is_set = True
-            if self._wait_for_tree and from_tree:
-                self._tree_has_set = True
+            if self._read_only:
```
aludvik: Why check this twice? Should it only be checked within the condition?

validator/sawtooth_validator/execution/context_manager.py
```diff
@@ -421,48 +607,120 @@ def run(self):
...
+                return
+            if from_tree:
+                if self._result_set_in_context:
+                    # since the result has been set in the context, don't
+                    # overwrite the value with the value from the merkle tree.
+                    pass
+                else:
```
aludvik: `if not self._result_set_in_context:`

validator/sawtooth_validator/execution/context_manager.py
```diff
@@ -79,89 +85,220 @@ def session_id(self):
...
+        with self._lock:
+            return self._contains(item)
+
+    def _get(self, address):
+        value = None
+        if self._contains(address):
+            value = self._state[address].result()
+        return value
+
+    def _get_and_set(self, address):
```
peterschwarz: What is the difference between the `_get` and `_get_and_set`?

aludvik: _get_and_set uses _contains_and_set to check instead of _contains

validator/sawtooth_validator/execution/context_manager.py
```diff
@@ -61,13 +62,18 @@ def __init__(self, state_hash, read_list, write_list, base_context_ids):
...
         self._state_hash = state_hash
 
-        self._read_list = set(read_list)
-        self._write_list = set(write_list)
+        self._read_list = list(read_list)
+        self._write_list = list(write_list)
 
         self._state = {}
-        self.base_context_ids = base_context_ids
+        self._lock = Lock()
```
aludvik: I updated the docstring assuming it is thread-safe now.

validator/sawtooth_validator/execution/context_manager.py
aludvik: Left this as is since I didn't have a good reason to change it.

validator/sawtooth_validator/execution/context_manager.py
aludvik: Don't understand the context manager enough to rewrite the above yet.

validator/sawtooth_validator/execution/context_manager.py
aludvik: Don't understand the context manager enough to rewrite the above yet.

validator/sawtooth_validator/execution/context_manager.py
aludvik: I think there should be a separate read-only `_ContextFuture` that gets created, rather than transforming a read-write instance to be read-only, which is why this is here.

validator/sawtooth_validator/execution/context_manager.py
```diff
@@ -61,13 +62,18 @@ def __init__(self, state_hash, read_list, write_list, base_context_ids):
             base_context_ids (list of str): Context ids of contexts that this
                 context is based off of.
         """
+
         self._state_hash = state_hash
 
-        self._read_list = set(read_list)
-        self._write_list = set(write_list)
+        self._read_list = list(read_list)
+        self._write_list = list(write_list)
```
peterschwarz: If those items are passed in via a generator or a list comprehension, this coerces them to be an explicit list.  If it's already a list, its effectively a no-op.

PR #539, opened 2017-05-30
Update REST API to handle proxying properly

delventhalz:
Updates the REST API to handle proxy urls when building a response `link`. It will now look for an `X-Forwarded-Host` header, and an `X-Forwarded-Path` header, and prefer those to the default host/path.

Also includes a new integration test which will set up an Apache 2 Basic Auth proxy in front of the REST API, and confirm it can be properly authorized, and sends back correct links.

### DIFF
--- a/bin/build_all
+++ b/bin/build_all
@@ -102,6 +102,8 @@ main() {
     # Start in project directory
     cd $top_dir
 
+    build_external
+
     if [[ $BUILD_MODE == $DEBS ]]
     then
         build_debs
@@ -194,6 +196,10 @@ docker_build() {
         --build-arg http_proxy=$http_proxy \
         --build-arg HTTPS_PROXY=$HTTPS_PROXY \
         --build-arg HTTP_PROXY=$HTTP_PROXY
+}
+
+build_external() {
+    docker_build docker/apache-basic_auth_proxy docker/ apache-basic_auth_proxy
 }
 
 build_debs() {
--- a/bin/run_tests
+++ b/bin/run_tests
@@ -280,6 +280,8 @@ test_python_sdk() {
 test_integration() {
     run_docker_test dynamic-network -s integration_test --timeout 600
     copy_coverage .coverage.dynamic-network
+    run_docker_test basic-auth-proxy -s integration_test
+    copy_coverage .coverage.basic-auth-proxy
 }
 
 test_deployment() {
--- a/docker/apache-basic_auth_proxy
+++ b/docker/apache-basic_auth_proxy
@@ -1,0 +1,74 @@
+# Copyright 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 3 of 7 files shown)


delventhalz:
# To Test This PR

This PR adds support for Basic Auth proxies. In order to test it, you'll need to build one. These are the instructions for setting up a simple authed Apache proxy in vagrant.

## Install Apache

```
% sudo apt-get update
% sudo apt-get install -y apache2
% sudo a2enmod headers
% sudo a2enmod proxy_http
```

## Set up a password file

Go to http://www.htaccesstools.com/htpasswd-generator/ and generate a `.htpdpassword` file from a username and password of your choice.
```
% vi /tmp/.password
```

Copy and paste the contents of the textbox into this file, save, and exit.

## Set up the site configuration

```
% sudo vi /etc/apache2/sites-enabled/000-default.conf
```

Edit the file to look like this:
```
<VirtualHost *:80>
        ServerAdmin sawtooth@sawtooth
        DocumentRoot /var/www/html

        ErrorLog ${APACHE_LOG_DIR}/error.log
        CustomLog ${APACHE_LOG_DIR}/access.log combined

        <Location />
                Options Indexes FollowSymLinks
                AllowOverride None
                AuthType Basic
                AuthName "Enter password"
                AuthUserFile "/tmp/.password"
                Require user sawtooth
                Require all denied
        </Location>
</VirtualHost>

ProxyPass /sawtooth http://localhost:8080
ProxyPassReverse /sawtooth http://localhost:8080
RequestHeader set X-Forwarded-Path "/sawtooth"
```

## Start and/or restart Apache

```
% sudo apachectl start
```

```
% sudo apachectl restart
```

## Make sure everything is set up properly

Start up a validator and the REST API normally. When you query the REST API, you should get back some block data as JSON:
```
% curl localhost:8080/blocks
```

You should also be able to get a `401` by querying the proxy without authorization:
```
% curl localhost/sawtooth/blocks
```

And finally, you should be able to get the same JSON by querying the proxy with authorization:
```
% curl localhost/sawtooth/blocks -u {USERNAME}:{PASSWORD}
```

When you send a request to the REST API it should format the `"link"` response parameter relative to the _proxy_ not its own url. So when you get a JSON response back to the authed proxy url, heck the link parameter.

It should start with:
```
"http://localhost/sawtooth/blocks?head=...
```

It should _not_ start with:
```
"http://localhost:8080/blocks?head=...
```

## What to do if you have permissioning issues around port 80

Sometimes vagrant seems to get angry when you try to run the proxy on port `80`. You can switch it to port `8000` by changing two files.
```
% sudo vi /etc/apache2/ports.conf
```

Switch `Listen 80` -> `Listen 8000`.

```
% sudo vi /etc/apache2/sites-enabled/000-default.conf
```

Switch `<VirtualHost *:80>` -> `<VirtualHost *:8000>`

Then:
```
% sudo apachectl restart
```

Afterwards the above example should work fine, but the proxy url will be `localhost:8000/sawtooth` instead of `localhost/sawtooth`.

PR #544, opened 2017-05-31
Rename Rest Api command line arguments

agunde406:
stream_url -> connect
port and host -> bind

This PR is build on top of and is waiting on #529.

### DIFF
--- a/rest_api/sawtooth_rest_api/rest_api.py
+++ b/rest_api/sawtooth_rest_api/rest_api.py
@@ -41,10 +41,9 @@ def parse_args(args):
     """Parse command line flags added to `rest_api` command.
     """
     parser = argparse.ArgumentParser()
-    parser.add_argument('--port',
-                        help='The port for the api to run on')
-    parser.add_argument('--host',
-                        help='The host for the api to run on')
+    parser.add_argument('-B', '--bind',
+                        help='The host and port for the api to run on.',
+                        action='append')
     parser.add_argument('-C', '--connect',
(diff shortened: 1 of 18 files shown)


rest_api/sawtooth_rest_api/rest_api.py
```diff
@@ -157,11 +175,12 @@ def main():
             log_configuration(log_dir=log_dir, name="sawtooth_rest_api")
         init_console_logging(verbose_level=opts.verbose)
 
+        host, port = rest_api_config.bind[0].split(":")
```
peterschwarz: There doesn't appear to be anything that validates that these two parts are there.  The `int(port)` will fail with an Error, that wont reflect back on the fact that port is missing or wrong.

rest_api/sawtooth_rest_api/rest_api.py
```diff
@@ -157,11 +175,19 @@ def main():
             log_configuration(log_dir=log_dir, name="sawtooth_rest_api")
         init_console_logging(verbose_level=opts.verbose)
 
+        try:
+            host, port = rest_api_config.bind[0].split(":")
+            port = int(port)
+        except ValueError as e:
+            LOGGER.error("Error parsing host and port from %s: %s",
```
peterschwarz: This should be a `print`.  Also, the message should say what the format should be:
`"Unable to parse binding %s: Must be '<host>:<port>'"

agunde406:
Rebased

PR #549, opened 2017-06-01
Update REST API Architecture doc to include sort query

delventhalz:

### DIFF
--- a/docs/source/architecture/rest_api.rst
+++ b/docs/source/architecture/rest_api.rst
@@ -174,15 +174,16 @@ Query Parameters
 
 Many routes support query parameters to help specify how a request to the
 validator should be formed. Not every endpoint supports every query, and some
-endpoints have their own parameters specific to just to them. Those
-endpoint-specific queries are not listed here.
+endpoints have their own parameters specific to just to them. Any queries
+specific to a single endpoint are not listed here.
 
 .. list-table::
    :widths: 8, 72
 
    * - **head**
-     - The id of the block to use as the chain head. This is particularly useful
-       to request older versions of state *(defaults to the latest chain head)*.
+     - The id of the block to use as the chain head. This is particularly
+       useful to request older versions of state *(defaults to the latest chain
+       head)*.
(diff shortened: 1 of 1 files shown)


docs/source/architecture/rest_api.rst
```diff
@@ -192,11 +193,18 @@ endpoint-specific queries are not listed here.
    * - **max**
      - For paging, specifies the id or index of the last resource to fetch. It
        would be used instead of *min*, not in the same query.
+   * - **sort**
+     - For endpoints that fetch lists of resources, specifies a key or keys to
+       sort the list by. Sorts can be modified with a simple but powerful
+       syntax: comma-separate multiple keys; dot-notate nested keys; use a
```
peterschwarz: Based on the previous sentence part, this sounds like it is describing the syntax, so using "comma-separated", "dot-notated" adjectives makes more sense. The second set of syntax items sounds like instructions on available operations. Suggest changing to something like the following (note the order about starting with keys, and working up to what can be done with them):

Sorts can be modified with the following rules:
1.  keys follow dot notation. For example, `header.signer_pubkey`
2. `header` may be omitted for nested header keys
3.  append `.length` to sort by length of a field
4. multiple keys may be comma-separated
5. Use of a minus sign to indicate descending order.  E.g. `-transaction_ids.length`

delventhalz: Don't think I can use a list within a list-table unfortunately. Could reorder.

docs/source/architecture/rest_api.rst
```diff
@@ -192,11 +193,19 @@ endpoint-specific queries are not listed here.
    * - **max**
      - For paging, specifies the id or index of the last resource to fetch. It
        would be used instead of *min*, not in the same query.
+   * - **sort**
+     - For endpoints that fetch lists of resources, specifies a key or keys to
+       sort the list by. These key orts can be modified with a few simple
```
peterschwarz: orts -> sorts

PR #550, opened 2017-06-01
Remove validator address from functions that don't need it

jjason:
This change supports the removal of multi-tenancy support in the PoET enclave.


jjason:
Rebase before merge for sanity check

PR #551, opened 2017-06-01
Move the Curve ZMQ keys into the validator config file. 

agunde406:
If the network key pair is not defined or if there is no validator config file, the network will default to being unsecure. 

This PR is build on top of #541

### DIFF
--- a/validator/packaging/validator.toml.example
+++ b/validator/packaging/validator.toml.example
@@ -33,3 +33,10 @@ seeds = ["tcp://127.0.0.1:8801"]
 # A list of peers to attempt to connect to in the format tcp://hostname:port.
 # It defaults to None.
 peers = ["tcp://127.0.0.1:8801"]
+
+# A Curve ZMQ key pair are used to create a secured network based on side-band
+# sharing of a single network key pair to all participating nodes.
+# Note if the config file does not exist or these are not set, the network
+# will default to being insecure.
+network_public_key = b'wFMwoOt>yFqI/ek.G[tfMMILHWw#vXB[Sv}>l>i)',
+network_private_key = b'r&oJ5aQDj4+V]p2:Lz70Eu0x#m%IwzBdP(}&hWM*'
--- a/validator/sawtooth_validator/server/cli.py
+++ b/validator/sawtooth_validator/server/cli.py
@@ -259,6 +259,12 @@ def main(args=sys.argv[1:]):
     if "tcp://" not in bind_component:
         bind_component = "tcp://" + bind_component
 
+    if validator_config.network_public_key is None or \
+            validator_config.network_private_key is None:
+        LOGGER.warning("Network key pair is not configured, Network "
+                       "communications between validators will not be "
+                       "authenticated or encrypted.")
+
     validator = Validator(bind_network,
                           bind_component,
                           endpoint,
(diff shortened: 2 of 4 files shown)


peterschwarz:
This looks like it should actually be two separate PR's

agunde406:
Thats what a bad rebase looks like, fixing

agunde406:
Fixed

validator/packaging/validator.toml.example
```diff
@@ -33,3 +33,10 @@ seeds = ["tcp://127.0.0.1:8801"]
 # A list of peers to attempt to connect to in the format tcp://hostname:port.
 # It defaults to None.
 peers = ["tcp://127.0.0.1:8801"]
+
+# Curve ZMQ key pair are used to create a secured network based on side-band
```
jsmitchell: "A Curve ZMQ key pair is used"

validator/packaging/validator.toml.example
```diff
@@ -33,3 +33,10 @@ seeds = ["tcp://127.0.0.1:8801"]
...
+# sharing of a single network key pair to all participating nodes.
+# Note if the config file does not exist or theses are not set, the network
```
jsmitchell: s/theses/these/

validator/packaging/validator.toml.example
```diff
@@ -33,3 +33,10 @@ seeds = ["tcp://127.0.0.1:8801"]
...
+# will default to being unsecure.
```
jsmitchell: insecure

validator/sawtooth_validator/server/cli.py
```diff
@@ -259,6 +259,11 @@ def main(args=sys.argv[1:]):
     if "tcp://" not in bind_component:
         bind_component = "tcp://" + bind_component
 
+    if validator_config.network_public_key is None or \
+            validator_config.network_private_key is None:
+        LOGGER.warning("Server key pair is not configured, The network is not "
```
jsmitchell: Network key pair

validator/sawtooth_validator/server/cli.py
```diff
@@ -259,6 +259,11 @@ def main(args=sys.argv[1:]):
...
+                       "secure.")
```
jsmitchell: Instead of "The network is not secure", let's say something like "Network communications between validators will not be authenticated or encrypted."

validator/packaging/validator.toml.example
```diff
@@ -33,3 +33,10 @@ seeds = ["tcp://127.0.0.1:8801"]
 # A list of peers to attempt to connect to in the format tcp://hostname:port.
 # It defaults to None.
 peers = ["tcp://127.0.0.1:8801"]
+
+# A Curve ZMQ key pair are used to create a secured network based on side-band
```
jsmitchell: Pair is singular. So you can say "A Curve ZMQ key pair is used" or "Curve ZMQ keys are used".

jsmitchell: I've decided this is irrelevant pedantry.

jjason: If we rule out irrelevant pedantry in PR comments, what value add do I bring to the table?

jsmitchell: @jjason the only pedantry I've seen from you is c++ pedantry -- something I highly approve of.

PR #557, opened 2017-06-01
Provide data directory to PoET enclave on initialization

jjason:


jjason:
@rbuysse @peterschwarz  Thanks for quick review

PR #558, opened 2017-06-01
Document EVM Spec and Client

aludvik:


docs/source/transaction_family_specifications/sawtooth_burrow_evm_family.rst
```diff
@@ -0,0 +1,269 @@
...
+
+        // The yellow paper also includes a signature in the transaction, but this
+        // is already included as part of the Sawtooth transaction so we don't
+        // duplicate it here.
+
+        // "An unlimited size byte array specifying the EVM-code for the account
+        // initialisation procedure, formally T_i."
+        //
+        // This should only be set if this is a contract creation transaction.
+        bytes init = 7;
```
peterschwarz: Something to think about here is to use a `oneof` field between `init` and `data`

docs/source/transaction_family_specifications/sawtooth_burrow_evm_family.rst
```diff
@@ -0,0 +1,269 @@
...
+        - The `gas_limit` field in the transaction payload for the `gas`
+          argument.
+
+    d. The resulting output from the EVM shall be stored in the `code` field of
+       the newly created account.
+    e. The sender account and the newly created account shall be pushed to
+       state.
+
+5. If an error occurs while the EVM is executing, the transaction is invalid.
+6. If a transaction is invalid, no updates to state will be made.
```
peterschwarz: Are you referring to state within the contract here?  If not, this is an unnecessary step, since no invalid transactions apply changes to state.

docs/source/transaction_family_specifications/sawtooth_burrow_evm_family.rst
aludvik: I agree that would be a good idea, but it isn't relevant to this PR which is to document the decisions​ that have already been made.

docs/source/transaction_family_specifications/sawtooth_burrow_evm_family.rst
aludvik: Hmm, I see what you mean. I wrote this with the ethereum model in mind. In the ethereum model, changes to state can be made even if a transaction fails to commit, for example if an out of gas exception occurs. I wanted to clarify that this is not he case here. Although, in that case, we may not want to consider a transaction invalid in that case within the context of sawtooth... I need to think about that some more...

aludvik:
Made changes suggested by @peterschwarz.

docs/source/transaction_family_specifications/sawtooth_burrow_evm_family.rst
```diff
@@ -0,0 +1,268 @@
...
+    a. If the sender address does not exist yet, an Externally Owned Account
+       shall be created at the address.
+    b. If the sender address does exist, a Contract Account shall be created at
+       a new address derived from the sender account as described above. The
+       sender account’s nonce shall then be incremented.
+    c. If the `init` field is set in the transaction payload, the EVM will be
+       called using:
+
+        - The sender account for the `caller` argument
+        - The sender account for the `callee` argument
```
ojalatodd: Did you mean 'receiver' instead of 'sender'?

aludvik:
Made changes based on @ojalatodd's feedback

docs/source/cli/seth.rst
```diff
@@ -0,0 +1,186 @@
...
+output obtained by running seth and its various subcommands with the `-h` flag.
+
+See the final section for a brief explanation of how to generate contract
+initialization data and how to format function calls using the Ethereum
+Contract ABI.
+
+seth
+----
+
+.. code-block:: console
```
aludvik: Added this

docs/source/transaction_family_specifications/sawtooth_burrow_evm_family.rst
```diff
@@ -0,0 +1,269 @@
...
+the Hyperledger Sawtooth framework using the Sawtooth Go SDK.
+
+The primary problems to solve in order to integrate the Burrow EVM into
+Sawtooth are:
+
+1. Define and implement an efficient mapping between EVM World State addresses
+   and Sawtooth Global State addresses.
+2. Define and implement an efficient method for maintaining accounts and
+   account storage in Sawtooth Global State.
+3. Define and implement an “EVM-Wrapper” at the TP level for handling
```
aludvik: Done

PR #518, opened 2017-05-24
 test cases for PoetBlockPublisher finalize_block()

nconde:
Signed-off-by: nconde <naissa.conde@intel.com>

PR #561, opened 2017-06-02
Revert " test cases for PoetBlockPublisher finalize_block()"

nconde:
Reverts hyperledger/sawtooth-core#518


jsmitchell:
Should we hold off on this until we get the master fix in on the validator argument changes?

jsmitchell:
It looks like these tests passed in @delventhalz's PR. So, I'd suggest we wait until master passes and then close this revert.

PR #564, opened 2017-06-02
Xo smoke

nick-drozd:


integration/sawtooth_integration/tests/integration_tools.py
```diff
@@ -12,16 +12,161 @@
...
+                ('Unable to connect to "{}": '
+                 'make sure URL is correct').format(self.url))
+
+    @staticmethod
+    def _format_queries(queries):
+        queries = {k: v for k, v in queries.items() if v is not None}
+        return queries if queries else ''
+
+
+class XoClient(RestClient):
```
peterschwarz: Is this class used across tests? Seems like it would be very integration test specific

nick-drozd: It is. Previously I had used the rest client from `cli/sawtooth_cli/rest_client.py`, but @vaporos didn't want the tests to import from there. Currently there are tests that import the cli rest client (which should be replaced with this rest client) and tests that could be simplified by using this rest client.

PR #569, opened 2017-06-02
Xo state format

nick-drozd:
This PR is based on the xo-smoke PR (#564)

### DIFF
--- a/integration/sawtooth_integration/docker/xo-all-lang-smoke.yaml
+++ b/integration/sawtooth_integration/docker/xo-all-lang-smoke.yaml
@@ -1,0 +1,100 @@
+# Copyright 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


integration/sawtooth_integration/docker/xo-all-lang-smoke.yaml
```diff
@@ -0,0 +1,100 @@
...
+  rest_api:
+    image: sawtooth-rest_api:$ISOLATION_ID
+    volumes:
+      - $SAWTOOTH_CORE:/project/sawtooth-core
+    expose:
+      - 40000
+      - 8080
+    depends_on:
+     - validator
+    command: rest_api -v --stream-url tcp://validator:40000 --host rest_api
```
agunde406: This will need to be updated to match the new rest api command line that was just merged.

sdk/examples/xo_python/sawtooth_xo/processor/handler.py
```diff
@@ -62,6 +62,9 @@ def apply(self, transaction, state_store):
         if name == "":
             raise InvalidTransaction("Name is required")
 
+        if '|' in name:
+            raise InvalidTransaction('Name cannot contain "|"')
```
dplumb94: Do we need to check for `","` as well?

nick-drozd: Because commas are used in the payload serialization, I don't think it's possible to get a comma into the name. If the payload doesn't contain exactly two commas, it should be rejected by `name, action, space = transaction.payload.decode().split(",")`

PR #571, opened 2017-06-03
Validate the transaction_ids field

feihujiang:

Validate the transaction_ids field  contains a list of transaction header_signatures and must be the same order as the transactions field in a batch.

### DIFF
--- a/validator/sawtooth_validator/gossip/structure_verifier.py
+++ b/validator/sawtooth_validator/gossip/structure_verifier.py
@@ -1,0 +1,153 @@
+# Copyright 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+
+import logging
+# pylint: disable=import-error,no-name-in-module
+# needed for google.protobuf import
+from google.protobuf.message import DecodeError
+
+from sawtooth_validator.protobuf import client_pb2
+from sawtooth_validator.protobuf.batch_pb2 import BatchHeader
+from sawtooth_validator.protobuf.batch_pb2 import Batch
+from sawtooth_validator.protobuf.block_pb2 import Block
+from sawtooth_validator.protobuf import network_pb2
+from sawtooth_validator.protobuf.network_pb2 import GossipBlockResponse
+from sawtooth_validator.protobuf.network_pb2 import GossipBatchResponse
+from sawtooth_validator.networking.dispatch import Handler
+from sawtooth_validator.networking.dispatch import HandlerResult
+from sawtooth_validator.networking.dispatch import HandlerStatus
+from sawtooth_validator.protobuf.validator_pb2 import Message
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+def is_valid_block(block):
+    if not all(map(is_valid_batch, block.batches)):
+        return False
+
+    return True
+
+
+def is_valid_batch(batch):
+    # batch structure verification
+    header = BatchHeader()
+    header.ParseFromString(batch.header)
+
+    # validate the transaction_ids field in batch header contains a list of
+    # transaction header_signatures and must be the same order as the
+    # transactions field
+    if len(batch.transactions) > len(header.transaction_ids):
+        LOGGER.debug("Batch has extra transactions. Dropping batch: %s",
+                     batch.header_signature)
+        return False
+    elif len(batch.transactions) < len(header.transaction_ids):
+        LOGGER.debug("Batch lacks transactions. Dropping batch: %s",
+                     batch.header_signature)
+        return False
+
+    for header_txn_id, txn in zip(header.transaction_ids, batch.transactions):
+        if header_txn_id != txn.header_signature:
+            LOGGER.debug("The header.transaction_ids does not match the "
+                         "order of transactions in the batch: %s txn: %s",
+                         batch.header_signature, header_txn_id)
+            return False
+
+    return True
+
+
+class GossipHandlerStructureVerifier(Handler):
+    def handle(self, connection_id, message_content):
+        gossip_message = network_pb2.GossipMessage()
+        gossip_message.ParseFromString(message_content)
+        if gossip_message.content_type == "BLOCK":
+            block = Block()
+            block.ParseFromString(gossip_message.content)
+            if not is_valid_block(block):
+                LOGGER.debug("block's batches structure is invalid: %s",
+                             block.header_signature)
+                return HandlerResult(status=HandlerStatus.DROP)
+
+            LOGGER.debug("block passes batch structure verification %s",
+                         block.header_signature)
+            return HandlerResult(status=HandlerStatus.PASS)
+        elif gossip_message.content_type == "BATCH":
+            batch = Batch()
+            batch.ParseFromString(gossip_message.content)
+            if not is_valid_batch(batch):
+                LOGGER.debug("batch structure is invalid: %s",
+                             batch.header_signature)
+                return HandlerResult(status=HandlerStatus.DROP)
+
+            LOGGER.debug("batch passes structure verification %s",
+                         batch.header_signature)
+            return HandlerResult(status=HandlerStatus.PASS)
+
+        return HandlerResult(status=HandlerStatus.PASS)
+
+
+class GossipBlockResponseStructureVerifier(Handler):
+    def handle(self, connection_id, message_content):
+        block_response_message = GossipBlockResponse()
+        block_response_message.ParseFromString(message_content)
+
+        block = Block()
+        block.ParseFromString(block_response_message.content)
+        if not is_valid_block(block):
+            LOGGER.debug("requested block's batches structure is invalid: %s",
+                         block.header_signature)
+            return HandlerResult(status=HandlerStatus.DROP)
+
+        LOGGER.debug("requested block passes batch structure "
+                     "verification %s", block.header_signature)
+        return HandlerResult(status=HandlerStatus.PASS)
+
+
+class GossipBatchResponseStructureVerifier(Handler):
+    def handle(self, connection_id, message_content):
+        batch_response_message = GossipBatchResponse()
+        batch_response_message.ParseFromString(message_content)
+
+        batch = Batch()
+        batch.ParseFromString(batch_response_message.content)
+        if not is_valid_batch(batch):
+            LOGGER.debug("requested batch's structure is invalid: %s",
+                         batch.header_signature)
+            return HandlerResult(status=HandlerStatus.DROP)
+
+        LOGGER.debug("requested batch passes structure verification %s",
+                     batch.header_signature)
+        return HandlerResult(status=HandlerStatus.PASS)
+
+
+class BatchListStructureVerifier(Handler):
+    def handle(self, connection_id, message_content):
+        response_proto = client_pb2.ClientBatchSubmitResponse
+
+        def make_response(out_status):
+            return HandlerResult(
+                status=HandlerStatus.RETURN,
+                message_out=response_proto(status=out_status),
+                message_type=Message.CLIENT_BATCH_SUBMIT_RESPONSE)
+        try:
+            request = client_pb2.ClientBatchSubmitRequest()
+            request.ParseFromString(message_content)
+        except DecodeError:
+            return make_response(response_proto.INTERNAL_ERROR)
+
+        if not all(map(is_valid_batch, request.batches)):
+            return make_response(response_proto.INVALID_BATCH)
+
+        return HandlerResult(status=HandlerStatus.PASS)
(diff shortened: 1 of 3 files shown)


validator/tests/unit3/test_message_validation/tests.py
```diff
@@ -203,6 +208,12 @@ def test_invalid_batch(self):
         valid = verifier.validate_batch(batch)
         self.assertFalse(valid)
 
+    def test_invalid_transaction_ids_batch(self):
+        batch_list = self._create_batches(1, 1, valid_transaction_ids=false)
```
peterschwarz: `false` -> `False`

cixce:
Our policy is that blocks are not required to be complete on arrival, If the Transaction list is incomplete the completer would retreive the missing transactions from the network and add them to the block. Since transactions are broadcast around the network when submitted, validators should have all the transactions for a block prior to the blocks arrival. Transmitting blocks with out all of the transactions present would be a bandwidth optimization. We have not enabled this yet but @agunde406 has implemented and completed the logic in the Completer to make this work.  This PR would force all block transitted on the network to be formally complete (containing all referenced transactions in the correct order)

@vaporos @jsmitchell Thoughts?

validator/tests/unit3/test_message_validation/tests.py
feihujiang: Fixed it.

feihujiang:
@cixce This validates each batch, not the block. And as I know, a batch contains the complete transactions, because one reason is that the transactions in a batch are submitted at the same time. This PR doesn't break the policy.  If I am wrong, please point it out.
And [The document](http://intelledger.github.io/architecture/transactions_and_batches.html) has the sentence that  "The transaction_ids field contains a list of Transaction header_signatures and must be the same order as the transactions field."
So I think it is needed to validate the transaction_ids field of each batch.

jsmitchell:
This seems to be overloading the functionality of the signature verifier to me.

Here's where the completer is applying rules regarding block structure:

https://github.com/hyperledger/sawtooth-core/blob/master/validator/sawtooth_validator/journal/completer.py#L144

If we want to more cleanly segregate enforcement of structural rules as a gating handler, I suggest we either rename signature verifier, or add another handler that deals with structure.

@agunde406 thoughts?

jsmitchell:
@cixce this change looks to check the transaction_id list in the batch header against the transaction list in the batch. It doesn't deal with block structure, unless the block includes raw batches (in which case verify_batch() is called). The effect this would have is that either batch messages or block messages containing batches which don't have transaction agreement between their header and body would be dropped.

agunde406:
I agree that this is overloading the original intent for the signature verifier. This check could be added to the completer or a new handler as @jsmitchell suggested.

feihujiang:
As documented, transactions must be wrapped in a Batch, and must contain the public key of the batch signer in the batcher_pubkey field. So a transaction is only wrapped in a batch, a batch must contain the complete transactions that the transaction_id list in the batch header.
   I'll add a new handler that checks the transaction_id list in the batch header against the transaction list in the batch. Batches which don't contain the complete transactions that the transaction_id list in the batch header would be dropped. 
   What's your point?

feihujiang:
I updated this PR, please have a look. Thanks.

validator/sawtooth_validator/gossip/structure_verifier.py
```diff
@@ -0,0 +1,174 @@
...
+    header.ParseFromString(batch.header)
+
+    # validate the transaction_ids field in batch header contains a list of
+    # transaction header_signatures and must be the same order as the
+    # transactions field
+    total = len(batch.transactions)
+
+    if total > len(header.transaction_ids):
+        LOGGER.debug("Batch has extra transactions.Dropping batch: %s",
+                     batch.header_signature)
```
agunde406: Missing space, transactions.Dropping -> transactions. Dropping

validator/sawtooth_validator/gossip/structure_verifier.py
```diff
@@ -0,0 +1,174 @@
...
+        return False
+    elif total < len(header.transaction_ids):
+        LOGGER.debug("Batch lacks transactions.Dropping batch: %s",
```
agunde406: Missing space, transactions.Dropping -> transactions. Dropping

validator/sawtooth_validator/gossip/structure_verifier.py
```diff
@@ -0,0 +1,174 @@
...
+
+    index = 0
+    while index < total:
+        txn = batch.transactions[index]
+
+        if header.transaction_ids[index] != txn.header_signature:
+            LOGGER.debug("transaction_ids field does not contains a list of "
+                         "transaction header_signatures or not the same "
+                         "order as the transactions for batch: %s txn: %s",
+                         batch.header_signature,
```
agunde406: This should be reworded, Something like:
The header.transaction_ids does not match the order of transactions in the batch.

validator/sawtooth_validator/gossip/structure_verifier.py
feihujiang: Done

validator/sawtooth_validator/gossip/structure_verifier.py
feihujiang: Done

jsmitchell:
@feihujiang could you force push this branch again? something seems amiss.

feihujiang:
@jsmitchell OK, I forced push this branch again. Something seems amiss, what does this mean?

jsmitchell:
@feihujiang we were just having some weirdness with the kickoffs of CI builds. No worries.

validator/sawtooth_validator/gossip/structure_verifier.py
```diff
@@ -0,0 +1,172 @@
...
+def validate_block_batches(block):
+    return validate_batches(block.batches)
+
+
+def validate_batches(batches):
+    for batch in batches:
+        if not validate_batch(batch):
+            return False
+
+    return True
```
jjason: Can you do this instead?

`return all(map(validate_batch, batches))`

validator/sawtooth_validator/gossip/structure_verifier.py
```diff
@@ -0,0 +1,172 @@
...
+        LOGGER.debug("Batch has extra transactions. Dropping batch: %s",
+                     batch.header_signature)
+        return False
+    elif total < len(header.transaction_ids):
+        LOGGER.debug("Batch lacks transactions. Dropping batch: %s",
+                     batch.header_signature)
+        return False
+
+    index = 0
+    while index < total:
```
jjason: You can replace this with:

```
for index, txn in enumerate(batch.transactions):
    if header.transaction_ids[index] != txn.header_signature:
```
This way, the loop index is done for you automatically.

or maybe even cleaner:
```
for txn_id, txn in zip(header.transaction_ids, batch.trancsactions):
    if txn_id != txn.header_signature:
```
This way, you don't have to worry about index at all.

validator/sawtooth_validator/gossip/structure_verifier.py
```diff
@@ -0,0 +1,172 @@
...
+class GossipHandlerStructureVerifier(Handler):
+
+    def handle(self, connection_id, message_content):
+        gossip_message = network_pb2.GossipMessage()
+        gossip_message.ParseFromString(message_content)
+        if gossip_message.content_type == "BLOCK":
+            block = Block()
+            block.ParseFromString(gossip_message.content)
+            status = validate_block_batches(block)
+            if status is True:
```
jjason: `if validate_block_batches(block):`

validator/sawtooth_validator/gossip/structure_verifier.py
```diff
@@ -0,0 +1,172 @@
...
+                return HandlerResult(status=HandlerStatus.PASS)
+
+            LOGGER.debug("block's batches structure is invalid: %s",
+                         block.header_signature)
+            return HandlerResult(status=HandlerStatus.DROP)
+        elif gossip_message.content_type == "BATCH":
+            batch = Batch()
+            batch.ParseFromString(gossip_message.content)
+            status = validate_batch(batch)
+            if status is True:
```
jjason: `if validate_batch(batch):`

validator/sawtooth_validator/gossip/structure_verifier.py
```diff
@@ -0,0 +1,172 @@
...
+class GossipBatchResponseStructureVerifier(Handler):
+    def handle(self, connection_id, message_content):
+        batch_response_message = GossipBatchResponse()
+        batch_response_message.ParseFromString(message_content)
+
+        batch = Batch()
+        batch.ParseFromString(batch_response_message.content)
+        status = validate_batch(batch)
+
+        if status is True:
```
jjason: `if validate_batch(batch):`

validator/sawtooth_validator/gossip/structure_verifier.py
```diff
@@ -0,0 +1,172 @@
...
+            LOGGER.debug("requested batch passes structure verification %s",
+                         batch.header_signature)
+            return HandlerResult(status=HandlerStatus.PASS)
+        else:
```
jjason: No need for `else:`

validator/sawtooth_validator/gossip/structure_verifier.py
```diff
@@ -0,0 +1,172 @@
...
+class GossipBlockResponseStructureVerifier(Handler):
+    def handle(self, connection_id, message_content):
+        block_response_message = GossipBlockResponse()
+        block_response_message.ParseFromString(message_content)
+
+        block = Block()
+        block.ParseFromString(block_response_message.content)
+        status = validate_block_batches(block)
+
+        if status is True:
```
jjason: `if validate_block_batches(block):`

validator/sawtooth_validator/gossip/structure_verifier.py
```diff
@@ -0,0 +1,172 @@
...
+            return HandlerResult(status=HandlerStatus.DROP)
+        elif gossip_message.content_type == "BATCH":
+            batch = Batch()
+            batch.ParseFromString(gossip_message.content)
+            status = validate_batch(batch)
+            if status is True:
+                LOGGER.debug("batch passes structure verification %s",
+                             batch.header_signature)
+                return HandlerResult(status=HandlerStatus.PASS)
+            else:
```
jjason: No need for `else:`

validator/sawtooth_validator/gossip/structure_verifier.py
```diff
@@ -0,0 +1,172 @@
...
+
+        block = Block()
+        block.ParseFromString(block_response_message.content)
+        status = validate_block_batches(block)
+
+        if status is True:
+            LOGGER.debug("requested block passes batch structure "
+                         "verification %s", block.header_signature)
+            return HandlerResult(status=HandlerStatus.PASS)
+        else:
```
jjason: No need for `else:`

validator/sawtooth_validator/gossip/structure_verifier.py
peterschwarz: While that looks pretty, the existing way is fail-fast.

feihujiang:
@jjason I applied the changes from your review,  please have a look. Thanks.

validator/sawtooth_validator/gossip/structure_verifier.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+def is_valid_block(block):
+    if not all(map(is_valid_batch, block.batches)):
+        return False
+
+    return True
```
jjason: I personally think that this is cleaner:
`return all(map(is_valid_batch, block.batches))` or as @delventhalz recommended:
`return all(is_valid_batch(batch) for batch in block.batches)` but won't let that hold up the review.

validator/sawtooth_validator/gossip/structure_verifier.py
```diff
@@ -0,0 +1,153 @@
...
+    elif len(batch.transactions) < len(header.transaction_ids):
+        LOGGER.debug("Batch lacks transactions. Dropping batch: %s",
+                     batch.header_signature)
+        return False
+
+    for header_txn_id, txn in zip(header.transaction_ids, batch.transactions):
+        if header_txn_id != txn.header_signature:
+            LOGGER.debug("The header.transaction_ids does not match the "
+                         "order of transactions in the batch: %s txn: %s",
+                         batch.header_signature, txn.header_signature)
```
jjason: You need to replace `batch.header_signature` with `header_txn_id`

validator/sawtooth_validator/gossip/structure_verifier.py
```diff
@@ -0,0 +1,153 @@
...
+                             block.header_signature)
+                return HandlerResult(status=HandlerStatus.PASS)
+
+            LOGGER.debug("block's batches structure is invalid: %s",
+                         block.header_signature)
+            return HandlerResult(status=HandlerStatus.DROP)
+        elif gossip_message.content_type == "BATCH":
+            batch = Batch()
+            batch.ParseFromString(gossip_message.content)
+            if not is_valid_batch(batch):
```
jjason: This is going to be a really annoyingly small nit, and you can ignore, but above you check `if is_valid_block(...)` and here you check `if not is_valid_batch(...)`.  From a correctness standpoint, this is not a problem, but from a consistency standpoint, I think it would be better (not just for my anal-retentive ways, but for others looking at the code) if both conditionals either checked for the "success" case or the "failure" case.  If you choose to address this, I don't have much preference either way, except that the handlers below seems to look for the "failure" case in the conditional

validator/sawtooth_validator/gossip/structure_verifier.py
feihujiang: Done.

validator/sawtooth_validator/gossip/structure_verifier.py
feihujiang: Yes, I think so.

validator/sawtooth_validator/gossip/structure_verifier.py
feihujiang: Yes, a consistency is better for others looking at the code.

PR #545, opened 2017-05-31
Add Basic Auth support to Sawtooth CLI

delventhalz:
# To Test This PR

This PR adds support for Basic Auth proxies. In order to test it, you'll need to build one. These are the instructions for setting up a simple authed Apache proxy in vagrant.

## Install Apache

```
% sudo apt-get update
% sudo apt-get install -y apache2
% sudo a2enmod headers
% sudo a2enmod proxy_http
```

## Set up a password file
```
% vi /tmp/.password
```

Edit the file to look like this:
```
sawtooth:$apr1$VHIPQ4Wi$k7PbIPqcfVp1NVKWhkIbS0

```

This saves the password `"sawtooth"` for the user `sawtooth`. You can generate new files like this at the site: http://www.htaccesstools.com/htpasswd-generator/. However, the config file below only allows the user `sawtooth`, so that would need to be modified as well.

## Set up the site configuration

```
% sudo vi /etc/apache2/sites-enabled/000-default.conf
```

Edit the file to look like this:
```
<VirtualHost *:80>
        ServerAdmin sawtooth@sawtooth
        DocumentRoot /var/www/html

        ErrorLog ${APACHE_LOG_DIR}/error.log
        CustomLog ${APACHE_LOG_DIR}/access.log combined

        <Location />
                Options Indexes FollowSymLinks
                AllowOverride None
                AuthType Basic
                AuthName "Enter password"
                AuthUserFile "/tmp/.password"
                Require user sawtooth
                Require all denied
        </Location>
</VirtualHost>

ProxyPass /sawtooth http://localhost:8080
ProxyPassReverse /sawtooth http://localhost:8080
RequestHeader set X-Forwarded-Path "/sawtooth"
```

## Start and/or restart Apache

```
% sudo apachectl start
```

```
% sudo apachectl restart
```

## Make sure everything is set up properly

Start up a validator and the REST API normally. When you query the REST API, you should get back some block data as JSON:
```
% curl localhost:8080/blocks
```

You should also be able to get a `401` by querying the proxy without authorization:
```
% curl localhost/sawtooth/blocks
```

And finally, you should be able to get the same JSON by querying the proxy with authorization:
```
% curl localhost/sawtooth/blocks -u sawtooth:sawtooth
```

## Test the sawtooth CLI

With everything working above, you should be able to:
```
sawtooth blocks list --url http://localhost/sawtooth -u sawtooth:sawtooth
```

## What to do if you have permissioning issues around port 80

[...]

PR #573, opened 2017-06-04
Fix the help of the submit's url argument

feihujiang:


peterschwarz:
This will need to be rebased.

delventhalz:
This typo was fixed by #545, which centralized the url parameter in the `base_http_parser`, which doesn't have this typo. Closing the PR.

PR #574, opened 2017-06-04
Add the batch-status command

feihujiang:

### DIFF
--- a/cli/sawtooth_cli/batch_status.py
+++ b/cli/sawtooth_cli/batch_status.py
@@ -1,0 +1,76 @@
+# Copyright 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the 'License');
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


cli/sawtooth_cli/batch_status.py
```diff
@@ -0,0 +1,63 @@
...
+
+        Args:
+            subparsers: Add parsers to this subparser object
+            parent_parser: The parent argparse.ArgumentParser object
+    """
+    epilog = '''details:
+        Fetches the committed statuses for a single batch.
+    '''
+    parser = subparsers.add_parser(
+        'batch_status', epilog=epilog,
```
peterschwarz: Change this to `batch-status`.  Operating under a no-underscores-in-command-args policy.

cli/sawtooth_cli/batch_status.py
```diff
@@ -0,0 +1,63 @@
...
+
+    parser.add_argument(
+        '--wait',
+        nargs='?',
+        const=maxsize,
+        type=int,
+        help='wait for batches to commit, set an integer to specify a timeout')
+
+    parser.add_argument(
+        'batch_id',
```
peterschwarz: likewise, change this to `batch-id`

cli/sawtooth_cli/batch_status.py
feihujiang: Changed it.

cli/sawtooth_cli/batch_status.py
feihujiang: this is the position argument, should use the batch_id, not the batch-id.

feihujiang:
@peterschwarz I updated this PR, please take a look. Thanks.

PR #576, opened 2017-06-05
Let the admin cmd has the same style

feihujiang:


peterschwarz:
The smoke tests/integration tests verify this, but the unit test does not.  We should properly exercise this in a future PR.

jsmitchell:
@feihujiang Could you change the commit message to "Changes sawtooth cli subparser name to subcommand" or somesuch?

feihujiang:
@jsmitchell I changed it according to your advice, please take a look.

PR #582, opened 2017-06-05
Wait for expected block num

peterschwarz:
Intkey smoke occasionally fails tests due to block commits occasionally
taking longer than 1 second.  Wait for the expected block number before
verifying the state to mitigate this issue.

### DIFF
--- a/integration/sawtooth_integration/tests/test_intkey_smoke.py
+++ b/integration/sawtooth_integration/tests/test_intkey_smoke.py
@@ -89,7 +89,10 @@ class TestIntkeySmoke(unittest.TestCase):
         batch = IntkeyMessageFactory().create_batch(batch)
         LOGGER.info('Posting batch')
         _post_batch(batch)
-        time.sleep(1)
+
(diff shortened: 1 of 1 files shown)


integration/sawtooth_integration/tests/test_intkey_smoke.py
```diff
@@ -89,7 +89,10 @@ def post_and_verify(self, batch, how_many_updates):
         batch = IntkeyMessageFactory().create_batch(batch)
         LOGGER.info('Posting batch')
         _post_batch(batch)
-        time.sleep(1)
+
+        while (how_many_updates + 1) > _get_block_num():
```
vaporos: If you waited for batch status to change instead, could we avoid this sleep(1) in most runs?

If _get_block_num() stops changing (due to something being broken), does this produce a reasonable test failure due to test runner timeouts?

peterschwarz: It should timeout if it stops working.

We can't check for failures until the rest API can report invalid batches.

PR #583, opened 2017-06-06
Fix truncated committed status for the submit cli

feihujiang:

When the status of a batch is COMMITTED, the printing output is COMMITTE. Fixed it.

### DIFF
--- a/cli/sawtooth_cli/submit.py
+++ b/cli/sawtooth_cli/submit.py
@@ -78,7 +78,7 @@ def do_submit(args):
 
         print('Wait timed out! Some batches have not yet been committed...')
         for batch_id, status in statuses.items():
-            print('{:128.128}  {:8.8}'.format(batch_id, status))
+            print('{:128.128}  {:10.10}'.format(batch_id, status))
         exit(1)
 
 


cli/sawtooth_cli/submit.py
```diff
@@ -78,7 +78,7 @@ def do_submit(args):
 
         print('Wait timed out! Some batches have not yet been committed...')
         for batch_id, status in statuses.items():
-            print('{:128.128}  {:8.8}'.format(batch_id, status))
+            print('{:128.128}  {:10.10}'.format(batch_id, status))
```
jsmitchell: @delventhalz what are the different possible status values here? It's not clear to me why we would be applying any formatting rules to this.

peterschwarz: They are `PENDING`, `COMMITTED`, `INVALID` and `UNKNOWN`.  The longest being 9 characters, so I'd agree we can get rid of that truncation.

jsmitchell: @feihujiang just change this to {}

delventhalz: It's a pattern used elsewhere in the CLI, printing out data in a fixed size table including when the data has variable print lengths. In this case it's the second of just two columns, and the first column should always be 128 character unless something weird happens, so the pattern is less applicable, but I think `{10:10}` is fine.

PR #584, opened 2017-06-06
Add Merkle Address validity test to MessageFactory

cianx:


jsmitchell:
Can you split the message factory change into a commit separate from the xo changes?

PR #585, opened 2017-06-06
Allow Executor to recover after a TP Crash.

agunde406:
This PR currently only works for python transaction processor. If solution is approved handling of TpPings will need to be added to all other sdks. 

A monitor socket is listening for disconnect events and notifies the executor that it should check all connected transaction processors to see if they are still connected. This is done by sending a TpPing response and waiting for a TpPingResponse is returned. If a transaction processor is found to not be connected, it should be removed from processors and all pending futures are set to an internal error (causing them to be rescheduled on a new processor).

The reasons that this ping message is needed, is that the zmq disconnect event only returns the following information: remote endpoint, event int, and a FD. I was not able to correlate a FD with a connection id for the transaction processor. Also a rest_api starts up/disconnects it creates the same type of connect/disconnect event as a transaction processor. 

Open issues I am looking for feed back on:
Currently, while waiting for the future to be set or a timeout to occur the executor thread is locked up waiting to know if a transaction processor should be removed. For example if the timeout is set to 10 seconds (the current timeout), and a processor crashes, it will take 10 seconds to continue processing.

### DIFF
--- a/validator/sawtooth_validator/execution/executor.py
+++ b/validator/sawtooth_validator/execution/executor.py
@@ -27,6 +27,7 @@ from sawtooth_validator.execution.scheduler_serial import SerialScheduler
 from sawtooth_validator.execution import processor_iterator
 from sawtooth_validator.networking.future import FutureResult
 from sawtooth_validator.networking.future import FutureTimeoutError
+from sawtooth_validator.networking.interconnect import ThreadsafeDict
 
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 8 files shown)


validator/sawtooth_validator/execution/executor.py
```diff
@@ -79,12 +82,12 @@ def _future_done_callback(self, request, result):
         """
         req = processor_pb2.TpProcessRequest()
         req.ParseFromString(request)
-
         response = processor_pb2.TpProcessResponse()
         response.ParseFromString(result.content)
         if response.status == processor_pb2.TpProcessResponse.OK:
             self._scheduler.set_transaction_execution_result(
                 req.signature, True, req.context_id)
+            del self._open_futures[req.connection_id][req.signature]
```
peterschwarz: Shouldn't this happen for all TP responses?

agunde406: Yes, I think so, I will updated shortly.

agunde406:
Yes, I wanted to make sure the approach was approved before touching all the sdks. @peterschwarz

peterschwarz:
👍

PR #592, opened 2017-06-07
Fix txn not listed in transaction_ids field is accepted and applied

feihujiang:

Transactions in the batch, but not listed by the transaction_ids in the batch header are still accepted and applied. This is a bug, this PR fixes this bug.


feihujiang:
The bug this PR references is fixed by #571. Close this PR.

PR #593, opened 2017-06-07
Fix missing python3 netifaces dependencies in the CI docs build.

cianx:
Removes this error:
  File "/project/sawtooth-core/bin/validator", line 34, in <module>
    from sawtooth_validator.server.cli import main
  File "/project/sawtooth-core/validator/sawtooth_validator/server/cli.py", line 20, in <module>
    import netifaces
ImportError: No module named 'netifaces'

### DIFF
--- a/ci/sawtooth-build-docs
+++ b/ci/sawtooth-build-docs
@@ -52,6 +52,7 @@ RUN echo "deb http://repo.sawtooth.me/ubuntu/nightly xenial universe" >> /etc/ap
     python3-grpcio=1.1.3-1 \
     python3-lmdb=0.92-1 \
     python3-multidict=2.1.4-1 \
+    python3-netifaces=0.10.4-0.1build2 \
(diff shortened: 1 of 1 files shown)


ci/sawtooth-build-docs
```diff
@@ -52,6 +52,7 @@ RUN echo "deb http://repo.sawtooth.me/ubuntu/nightly xenial universe" >> /etc/ap
     python3-grpcio=1.1.3-1 \
     python3-lmdb=0.92-1 \
     python3-multidict=2.1.4-1 \
+    python3-netifaces=0.10.4-0.1build2 \
```
jsmitchell: Do we require this specific version?

agunde406: The version is specified in all other places this dependency is added since we are specifying a version for all dependency. So the answer to your question is no, but should be there to be consistent.

PR #595, opened 2017-06-07
Fix intermittent 503 error from auth proxy integration test and clean up logs

delventhalz:
Test will now wait for both the REST API and Apache proxy to be ready before running any tests. Also includes some changes to the error logs to improve clarity and readability.

### DIFF
--- a/integration/sawtooth_integration/tests/test_basic_auth_proxy.py
+++ b/integration/sawtooth_integration/tests/test_basic_auth_proxy.py
@@ -68,16 +68,16 @@ class TestBasicAuth(unittest.TestCase):
         try:
             response = urlopen(request, context=context)
         except HTTPError as e:
+            LOGGER.error('An error occured while requesting "%s"', url)
+
             try:
                 error = json.loads(e.file.read().decode())['error']
+                fail_msg = 'REST API Error: {} - {}:'.format(
+                    error['code'], error['title'])
             except json.decoder.JSONDecodeError:
-                raise e
+                fail_msg = 'HTTP Error: {} - {}'.format(e.code, e.msg)
 
-            LOGGER.error('REST API Error: {} - {}:'.format(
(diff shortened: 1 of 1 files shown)


integration/sawtooth_integration/tests/test_basic_auth_proxy.py
```diff
@@ -68,16 +68,17 @@ def _assert_valid_authed_request(self, url):
...
             except json.decoder.JSONDecodeError:
-                raise e
-
-            LOGGER.error('REST API Error: {} - {}:'.format(
-                error['code'], error['title']))
-            LOGGER.error(error['message'])
+                error_string = 'HTTP Error: {} - {}'.format(e.code, e.msg)
 
-            raise e
+            LOGGER.error(error_string)
```
peterschwarz: Does this need to be logged? Won't the fail message get printed on failure?

delventhalz: Excellent question. I do not know enough about the details of what is getting logged where and how it is important.

My thinking was that the logs _might_ be getting stored somewhere different than the test output. If in _all_ cases you will get both the error logs and the test output, then it is redundant and I will remove it.

integration/sawtooth_integration/tests/test_basic_auth_proxy.py
```diff
@@ -68,16 +68,17 @@ def _assert_valid_authed_request(self, url):
         try:
             response = urlopen(request, context=context)
         except HTTPError as e:
+            LOGGER.error('An error occured while requesting {}'.format(url))
```
jjason: When using `LOGGER.error` you should use the printf-style formatting:
`LOGGER.error('An error occurred while requesting %s', url)`
At least pylint always seems to complain when I don't ;)

PR #604, opened 2017-06-08
Fix some issues causing test failures

jsmitchell:


validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -414,7 +426,10 @@ def shutdown(self):
                     tasks.remove(task)
             time.sleep(.2)
         if self._event_loop is not None:
-            self._event_loop.call_soon_threadsafe(self._event_loop.stop)
+            try:
+                self._event_loop.call_soon_threadsafe(self._event_loop.stop)
+            except RuntimeError:
```
cianx: Can we add a comment on why we eat this exception?

jsmitchell: Fixed

PR #605, opened 2017-06-09
Code Refactoring for signature verifier module

feihujiang:

### DIFF
--- a/validator/sawtooth_validator/gossip/signature_verifier.py
+++ b/validator/sawtooth_validator/gossip/signature_verifier.py
@@ -39,87 +39,75 @@ from sawtooth_validator.protobuf.validator_pb2 import Message
 LOGGER = logging.getLogger(__name__)
 
 
-def validate_block(block):
+def is_valid_block(block):
     # validate block signature
     header = BlockHeader()
     header.ParseFromString(block.header)
-    valid = signing.verify(block.header,
-                           block.header_signature,
-                           header.signer_pubkey)
+
+    if not signing.verify(block.header,
+                          block.header_signature,
+                          header.signer_pubkey):
+        LOGGER.debug("block failed signature validation: %s",
+                     block.header_signature)
+        return False
 
     # validate all batches in block. These are not all batches in the
     # batch_ids stored in the block header, only those sent with the block.
-    total = len(block.batches)
-    index = 0
-    while valid and index < total:
-        valid = validate_batch(block.batches[index])
-        index += 1
+    if not all(map(is_valid_batch, block.batches)):
+        return False
 
-    return valid
+    return True
 
 
-def validate_batches(batches):
-    valid = False
-    for batch in batches:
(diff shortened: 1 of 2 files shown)


validator/sawtooth_validator/gossip/signature_verifier.py
```diff
@@ -47,15 +47,13 @@ def validate_block(block):
                            block.header_signature,
                            header.signer_pubkey)
 
+    if not valid:
+        LOGGER.debug("block failed signature validation: %s",
+                     block.header_signature)
+
```
jjason: Why bother checking the batches if you have already discerned that the block header signature is invalid?  Why not:
```
if not signing.verify(...):
    LOGGER.debug(...)
    return False
```

validator/sawtooth_validator/gossip/signature_verifier.py
```diff
@@ -78,25 +76,24 @@ def validate_batch(batch):
...
-                LOGGER.debug("txn batcher pubkey does not match signer"
-                             "pubkey for batch: %s txn: %s",
-                             batch.header_signature,
-                             txn.header_signature)
-                valid = False
-        index += 1
-
-    return valid
+        if not valid:
+            return False
```
jjason: Or `transaction_is_valid`.  I agree with @delventhalz - I like functions that return booleans to read kind of like English

jjason: You can, _I think_, also replace the `for` loop with:

```
if not all(map(validate_transaction, batch.transactions)):
    return False
```

Since `map` returns an iterable and `all` is supposed to short-circuit, you get the same result.

Same goes for `validate_batches`

delventhalz: Generator comrehensions are lazy, so you could also use one of those, which might be a little more pythonic:
```
if not all(is_valid_transaction(t) for t in batch.transactions):
    return False
```

In either case, I think you could reasonably remove the function `validate_batches` entirely, and replace its usage with one of these.

peterschwarz: Map is lazy, too - it just returns an iterator - but that does look more "pythonic"

validator/sawtooth_validator/gossip/signature_verifier.py
peterschwarz: Agreed this should just return False in this case

validator/sawtooth_validator/gossip/signature_verifier.py
feihujiang: Done

validator/sawtooth_validator/gossip/signature_verifier.py
feihujiang: Done

feihujiang:
I applied the changes from the reviewers, please have a look. Thanks.

PR #606, opened 2017-06-09
Report Errors in JS SDK Transaction Processors

peterschwarz:
Report the error and extended information back to the validator via the new fields on TpProcessRequest. 

Additionally, add a timeout to the State `get` and `set` methods.

### DIFF
--- a/sdk/javascript/processor/exceptions.js
+++ b/sdk/javascript/processor/exceptions.js
@@ -17,21 +17,67 @@
 
 'use strict'
 
-class InvalidTransaction extends Error {
-  constructor (message = '') {
+class _TransactionProcessorError extends Error {
+  constructor (message = '', extendedData = null) {
     super(message)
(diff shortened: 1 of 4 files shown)


sdk/javascript/processor/exceptions.js
```diff
@@ -17,16 +17,31 @@
 
 'use strict'
 
-class InvalidTransaction extends Error {
-  constructor (message = '') {
+class _TransactionProcessorError extends Error {
+  constructor (message = '', extendedData = null) {
```
delventhalz: Although `extended_data` makes sense in the proto definition, why would the data be "extended" for the person throwing this error? Since it is the only data being included, why not just call it `data`?

peterschwarz: The data is optional, extended application data.  It should be marked as such here.  Added a doc string to reflect this fact.

PR #609, opened 2017-06-09
Send messages and data from Python SDK on TP errors

delventhalz:
Transaction Processors developed with the Python SDK, which reject a
transaction with the InvalidTransaction or InternalError exceptions,
will now automatically send the error messages back to the validator,
as well as optional byte-encoded data.

### DIFF
--- a/sdk/python/sawtooth_sdk/processor/exceptions.py
+++ b/sdk/python/sawtooth_sdk/processor/exceptions.py
@@ -14,9 +14,25 @@
 # ------------------------------------------------------------------------------
 
 
-class InvalidTransaction(Exception):
+class _TpResponseError(Exception):
+    """Parent class for errors that will be parsed and sent to a validator.
+
+    Args:
+        message (str): Standard error message to be logged or sent back
+        extended_data (bytes, optional): Byte-encoded data to be parsed later
+            by the app developer. Opaque to the validator and Sawtooth.
+    """
+    def __init__(self, message, extended_data=None):
+        super().__init__(message)
+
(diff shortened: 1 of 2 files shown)


sdk/python/sawtooth_sdk/processor/exceptions.py
```diff
@@ -14,9 +14,25 @@
...
-class InvalidTransaction(Exception):
+class _TpResponseError(Exception):
+    """Parent class for errors that will be parsed and sent to a validator.
+
+    Args:
+        message (str): Standard error message to be logged or sent back
+        data (bytes, optional): Byte-encoded data to be parsed later by the
+            app developer. Opaque to the validator and Sawtooth.
+    """
+    def __init__(self, message, data=None):
```
peterschwarz: Suggest renaming this to `extended_data` to match the JS SDK.

delventhalz: >:-(

sdk/python/sawtooth_sdk/processor/exceptions.py
```diff
@@ -14,9 +14,26 @@
...
+    Args:
+        message (str): Standard error message to be logged or sent back
+        extended_data (bytes, optional): Byte-encoded data to be parsed later
+            by the app developer. Opaque to the validator and Sawtooth.
+    """
+    def __init__(self, message, extended_data=None):
+        super().__init__(message)
+
+        if extended_data is not None:
+            assert isinstance(extended_data, bytes),\
```
peterschwarz: Check for this and raise a `TypeError`

PR #611, opened 2017-06-09
Implement signing key-based client batch submit permissioning

agunde406:
The signing keys that are allowed to submit batches will be
stored in the validator setting sawtooth.validator.allowed_signing_key.
A batch signed by an unknown signing key will cause the batch to be
dropped. If no signing keys are set, all signing keys are accepted.

### DIFF
--- a/integration/sawtooth_integration/tests/test_permission.py
+++ b/integration/sawtooth_integration/tests/test_permission.py
@@ -1,0 +1,254 @@
+# Copyright 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+
+import logging
+import unittest
+import traceback
+import tempfile
+import os
+import subprocess
+import time
+from sawtooth_cli.main import main
+import sys
+from io import StringIO
+
+from sawtooth_integration.tests.integration_tools import wait_for_rest_apis
+
+LOGGER = logging.getLogger(__name__)
+LOGGER.addHandler(logging.StreamHandler())
+LOGGER.setLevel(logging.DEBUG)
+
+TEST_WIF = '5Jq6nhPbVjgi9vTUuK7e2W81VT5dpQR7qPweYJZPVJKNzSornyv'
+TEST_PUBKEY = \
+    '033775c26a68a3872f03314ccd080b8d8ec828572469737c7d3aa467f853a069d5'
+
+
+class TestPermission(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        wait_for_rest_apis(['rest_api:8080'])
+
+    def setUp(self):
+        self._temp_dir = tempfile.mkdtemp()
+        self._data_dir = os.path.join(self._temp_dir, 'data')
+        os.makedirs(self._data_dir)
+
+        # create a wif key for signing
+        self._wif_file = os.path.join(self._temp_dir, 'test.priv')
+        with open(self._wif_file, 'wb') as wif:
+            wif.write(TEST_WIF.encode())
+
+    def _run(self, args):
+        try:
+            LOGGER.debug("Running %s", " ".join(args))
+            proc = subprocess.run(args,
+                                  stdout=subprocess.PIPE,
+                                  stderr=subprocess.PIPE,
+                                  check=True)
+            LOGGER.debug(proc.stdout.decode())
+        except subprocess.CalledProcessError as err:
+            LOGGER.debug(err)
+
+    def _read_from_stdout(self, cmd, args):
+        # Retrieve string of setting contents for comparison to input settings
+        backup = sys.stdout  # backup the environment
+        sys.stdout = StringIO()  # Capture the output of next statement
+
+        main(cmd, args)
+        settings = sys.stdout.getvalue()  # release the output and store
+        sys.stdout.close()
+        # Restore the environment
+        sys.stdout = backup
+        return settings
+
+    def _try_to_get_output(self, cmd, args, required):
+        # Temporary function until --wait is added to sawtooth config
+        for i in range(10):
+            settings = self._read_from_stdout(cmd, args)
+            if required in settings:
+                return settings
+            time.sleep(1)
+        self.fail("Changes were not committed in 5 seconds.")
+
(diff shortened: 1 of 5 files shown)


validator/sawtooth_validator/gossip/permission_verifier.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+def check_batches_signing_keys(batches, allowed_pubkeys):
+    for batch in batches:
+        if not check_batch_signing_key(batch, allowed_pubkeys):
+            return False
+    return True
```
jjason: As mentioned in a previous review, this could be shortened to:
`return all(check_batch_signing_key(batch, allowed_pubkeys) for batch in batches)`

validator/sawtooth_validator/gossip/permission_verifier.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def handle(self, connection_id, message_content):
+        response_proto = client_pb2.ClientBatchSubmitResponse
+        allowed_pubkeys = None
+        try:
+            state_root = self.current_root_func()
+            config_view = \
+                self.config_view_factory.create_config_view(state_root)
+            allowed_pubkeys = config_view.get_setting(
+                "sawtooth.validator.allowed_signing_key")
```
jjason: A small nit, but if this is can have more than one key, would probably name `sawtooth.validator.allowed_signing_keys`

validator/sawtooth_validator/gossip/permission_verifier.py
```diff
@@ -0,0 +1,85 @@
...
+        except AttributeError:
+            LOGGER.debug("Chain head not yet set")
+
+        def make_response(out_status):
+            return HandlerResult(
+                status=HandlerStatus.RETURN,
+                message_out=response_proto(status=out_status),
+                message_type=Message.CLIENT_BATCH_SUBMIT_RESPONSE)
+        try:
+            status = True
```
aludvik: This is kind of picky, but something other than `status` would be clearer. Maybe `allowed`?

validator/sawtooth_validator/gossip/permission_verifier.py
```diff
@@ -0,0 +1,85 @@
...
+            if allowed_pubkeys is not None:
+                request = client_pb2.ClientBatchSubmitRequest()
+                request.ParseFromString(message_content)
+                status = check_batches_signing_keys(
+                    request.batches, allowed_pubkeys)
+        except DecodeError:
+            return make_response(response_proto.INTERNAL_ERROR)
+
+        if status is not True:
```
aludvik: `if not allowed:`?

validator/sawtooth_validator/gossip/permission_verifier.py
```diff
@@ -0,0 +1,85 @@
...
+def check_batches_signing_keys(batches, allowed_pubkeys):
+    for batch in batches:
+        if not check_batch_signing_key(batch, allowed_pubkeys):
+            return False
+    return True
+
+
+def check_batch_signing_key(batch, allowed_pubkeys):
+    # validate batch signature
+    if allowed_pubkeys is not None:
```
aludvik: Line 74 below is already checking whether `allowed_pubkey` is None. It is probably only necessary to check that in one place.

validator/sawtooth_validator/gossip/permission_verifier.py
```diff
@@ -0,0 +1,85 @@
...
+        try:
+            state_root = self.current_root_func()
+            config_view = \
+                self.config_view_factory.create_config_view(state_root)
+            allowed_pubkeys = config_view.get_setting(
+                "sawtooth.validator.allowed_signing_key")
+        except AttributeError:
+            LOGGER.debug("Chain head not yet set")
+
+        def make_response(out_status):
```
aludvik: Can this be moved outside of the handle function? Or is there a reason to define it inside of the handle function?

validator/sawtooth_validator/gossip/permission_verifier.py
peterschwarz: This doesn't match the setting key in the tests in the next commit.

agunde406:
Added more test cases. Currently that tests require sleeps to wait for block to processed. Theses should be change to have the sawtooth config wait when this added to the command. A story has been created to add the --wait to sawtooth config https://jira.hyperledger.org/browse/STL-442

agunde406:
Working on fixing rebase issues and rewriting the tests to match @delventhalz changes to test_config_smoke in #633.

validator/sawtooth_validator/gossip/permission_verifier.py
delventhalz: What is the reasoning behind _not_ defining helper functions within other functions? 

In this case `make_response` only makes sense in the context of the `handle` method, as it only abstracts away a common pattern used in `handle`, and references a variable in `handle`'s scope.

validator/sawtooth_validator/gossip/permission_verifier.py
```diff
@@ -0,0 +1,81 @@
...
+            return HandlerResult(
+                status=HandlerStatus.RETURN,
+                message_out=response_proto(status=out_status),
+                message_type=Message.CLIENT_BATCH_SUBMIT_RESPONSE)
+        try:
+            allowed = True
+            if allowed_pubkeys is not None:
+                request = client_pb2.ClientBatchSubmitRequest()
+                request.ParseFromString(message_content)
+                allowed = check_batches_signing_keys(
```
delventhalz: You could get rid of `allowed` entirely if you replaced this line with:
```
if not check_batches_signing_keys(...):
    return make_response(response_proto.INVALID_BATCH)
```

validator/sawtooth_validator/gossip/permission_verifier.py
```diff
@@ -0,0 +1,81 @@
...
+
+LOGGER = logging.getLogger(__name__)
+
+
+def check_batches_signing_keys(batches, allowed_pubkeys):
+    return all(check_batch_signing_key(batch, allowed_pubkeys) for batch in
+               batches)
+
+
+def check_batch_signing_key(batch, allowed_pubkeys):
```
delventhalz: Might be better named `is_batch_signer_authorized`

validator/sawtooth_validator/gossip/permission_verifier.py
aludvik: https://realpython.com/blog/python/inner-functions-what-are-they-good-for/

integration/sawtooth_integration/tests/test_permission.py
```diff
@@ -0,0 +1,254 @@
...
+            time.sleep(1)
+        self.fail("Changes were not committed in 5 seconds.")
+
+    def _assert_line_starts_match(self, raw_output, expected_line_starts):
+        output_lines = raw_output.split('\n')
+
+        for output, expected in zip(output_lines, expected_line_starts):
+            self.assertTrue(
+                output.startswith(expected),
+                'List output "{}" did not match "{}"'.format(expected, output))
```
delventhalz: I think this is my fault, but `output` and `expected` should be switched in the format args.

PR #612, opened 2017-06-09
Update to fix code that will trigger error in new version of PyLint

jjason:


jjason:
Argh!!!!  Curses!!!!  Forgot that id() is a built-in function.

jjason:
Better? @peterschwarz @vaporos

PR #613, opened 2017-06-10
fix the typo in the entrypoint for rest_api docker container

ksachdeva:
There is a typo in the entrypoint for rest_api docker container. It should be --bind and not -bind


vaporos:
The commit message is missing the required Signed-off-by line.  This can generally be fixed with "git commit --amend -s" and re-pushing to github.

ksachdeva:
Hi @vaporos,

I have amended the commit message as required by you.

Regards & thanks
Kapil

PR #614, opened 2017-06-10
Lint

nick-drozd:
This PR unpins the pylint version and fixes all resulting lint errors but one. That last error is fixed in https://github.com/hyperledger/sawtooth-core/pull/612, which needs to be merged before this one.

Notes:

1) This PR doesn't extend `run_lint` coverage, so there are still files not being linted.

2) The current version of pylint sometimes triggers a recursion error and spits out a big ugly recursion error stack trace. This makes reading the output very annoying, but it doesn't cause `run_lint` to fail.

### DIFF
--- a/bin/protogen
+++ b/bin/protogen
@@ -34,7 +34,9 @@ JOIN = os.path.join
 TOP_DIR = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
 
 
-def main(args=sys.argv[1:]):
+def main(args=None):
+    if args is None:
+        args = sys.argv[1:]
     # 1. Generate and distribute top-level protos
     proto_dir = JOIN(TOP_DIR, "protos")
 
--- a/cli/sawtooth_cli/exceptions.py
+++ b/cli/sawtooth_cli/exceptions.py
@@ -15,5 +15,4 @@
 
 
 class CliException(Exception):
-    def __init__(self, msg):
-        super(CliException, self).__init__(msg)
+    pass
--- a/rest_api/sawtooth_rest_api/config.py
+++ b/rest_api/sawtooth_rest_api/config.py
@@ -124,4 +124,4 @@ class RestApiConfig:
         ])
 
     def to_toml_string(self):
-        return toml.dumps(self.to_dict()).strip().split('\n')
+        return str(toml.dumps(self.to_dict())).strip().split('\n')
--- a/rest_api/sawtooth_rest_api/config.py
+++ b/rest_api/sawtooth_rest_api/config.py
@@ -54,7 +54,7 @@ def load_toml_rest_api_config(filename):
 
     invalid_keys = set(toml_config.keys()).difference(
         ['bind', 'connect', 'timeout'])
-    if len(invalid_keys) > 0:
+    if invalid_keys:
         raise RestApiConfigurationError(
             "Invalid keys in rest api config: {}".format(
                 ", ".join(sorted(list(invalid_keys)))))
--- a/validator/sawtooth_validator/database/lmdb_nolock_database.py
+++ b/validator/sawtooth_validator/database/lmdb_nolock_database.py
@@ -54,6 +54,7 @@ class LMDBNoLockDatabase(database.Database):
                                       create=create,
                                       lock=True)
 
+    # pylint: disable=no-value-for-parameter
     def __len__(self):
         with self._lmdb.begin() as txn:
             return txn.stat()['entries']
--- a/validator/sawtooth_validator/journal/block_store.py
+++ b/validator/sawtooth_validator/journal/block_store.py
@@ -132,8 +132,8 @@ class BlockStore(MutableMapping):
         """
         if not starting_block:
             return _BlockPredecessorIterator(self, start_block=self.chain_head)
-        else:
-            return _BlockPredecessorIterator(self, start_block=starting_block)
+
+        return _BlockPredecessorIterator(self, start_block=starting_block)
 
     def wait_for_batch_commits(self, batch_ids=None, timeout=None):
         """Waits for a set of batch ids to be committed to the block chain,
--- a/validator/sawtooth_validator/journal/chain.py
+++ b/validator/sawtooth_validator/journal/chain.py
@@ -155,7 +155,7 @@ class BlockValidator(object):
         return True
 
     def _verify_block_batches(self, blkw, committed_txn):
-        if len(blkw.block.batches) > 0:
+        if blkw.block.batches:
 
             prev_state = self._get_previous_block_root_state_hash(blkw)
             scheduler = self._executor.create_scheduler(
--- a/validator/sawtooth_validator/journal/chain.py
+++ b/validator/sawtooth_validator/journal/chain.py
@@ -106,9 +106,8 @@ class BlockValidator(object):
     def _get_previous_block_root_state_hash(self, blkw):
         if blkw.previous_block_id == NULL_BLOCK_IDENTIFIER:
             return INIT_ROOT_KEY
-        else:
-            prev_blkw = self._block_cache[blkw.previous_block_id]
-            return prev_blkw.state_root_hash
+
+        return self._block_cache[blkw.previous_block_id].state_root_hash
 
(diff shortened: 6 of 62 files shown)


bin/protogen
```diff
@@ -34,7 +34,8 @@ JOIN = os.path.join
 TOP_DIR = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
 
 
-def main(args=sys.argv[1:]):
+def main(args=None):
+    args = sys.argv[1:] if args is None else args
```
vaporos: Avoid using conditional expressions as replacements for simplistic flow control statements.  This is better:

```
if args is None:
    args = sys.argv[1:]
```

This is much clearer and technically correct (doesn't re-assign args when it not None).  It is also fewer tokens to read total and fewer tokens-per-line.

consensus/poet/core/sawtooth_poet/poet_consensus/poet_config_view.py
```diff
@@ -139,6 +139,7 @@ def enclave_module_name(self):
                     name='sawtooth.poet.enclave_module_name',
                     value_type=str,
                     default_value=PoetConfigView._ENCLAVE_MODULE_NAME_,
+                    # pylint: disable=len-as-condition
```
vaporos: Could cast to bool here instead of the pylint comment.

validator/sawtooth_validator/state/client_handlers.py
```diff
@@ -82,7 +82,7 @@ def __init__(self, request_proto, response_proto, response_type,
         self._block_store = block_store
         self._batch_cache = batch_cache
 
-    def handle(self, identity, message_content):
+    def handle(self, connection_id, message_content):
```
aludvik: Need to update the docstring below.

rest_api/sawtooth_rest_api/config.py
```diff
@@ -124,4 +124,4 @@ def to_dict(self):
         ])
 
     def to_toml_string(self):
-        return toml.dumps(self.to_dict()).strip().split('\n')
+        return str(toml.dumps(self.to_dict())).strip().split('\n')
```
aludvik: Can you include some notes in your commit message about why this was necessary?

validator/sawtooth_validator/journal/block_store.py
```diff
@@ -130,10 +130,9 @@ def get_predecessor_iter(self, starting_block=None):
         Returns:
             An iterator
         """
-        if not starting_block:
-            return _BlockPredecessorIterator(self, start_block=self.chain_head)
-        else:
-            return _BlockPredecessorIterator(self, start_block=starting_block)
+        start_block = starting_block if starting_block else self.chain_head
```
aludvik: IMO, introducing another variable like this adds an unnecessary level of indirection. I prefer something like the following:
```
if not starting_block:
    return _BlockPredecessorIterator(self, start_block=self.chain_head)

return _BlockPredecessorIterator(self, start_block=starting_block)
```

validator/sawtooth_validator/journal/chain.py
```diff
@@ -105,10 +105,12 @@ def __init__(self,
...
     def _get_previous_block_root_state_hash(self, blkw):
         if blkw.previous_block_id == NULL_BLOCK_IDENTIFIER:
-            return INIT_ROOT_KEY
+            state_hash = INIT_ROOT_KEY
         else:
             prev_blkw = self._block_cache[blkw.previous_block_id]
-            return prev_blkw.state_root_hash
+            state_hash = prev_blkw.state_root_hash
+
+        return state_hash
```
aludvik: IMO this is better:
```
if blkw.previous_block_id == NULL_BLOCK_IDENTIFIER:
    return INIT_ROOT_KEY

return self._block_cache[blkw.previous_block_id].state_root_hash
```

ditto for the remainder of this commit.

validator/sawtooth_validator/journal/consensus/dev_mode/dev_mode_consensus.py
```diff
@@ -140,6 +140,7 @@ def finalize_block(self, block_header):
 class BlockVerifier(BlockVerifierInterface):
     """DevMode BlockVerifier implementation
     """
+    # pylint: disable=useless-super-delegation
```
aludvik: Rather than disable these, can we update the Interface classes to have concrete constructors and then just delete the `__init__`  methods here?

validator/sawtooth_validator/database/lmdb_nolock_database.py
```diff
@@ -54,6 +54,7 @@ def __init__(self, filename, flag):
                                       create=create,
                                       lock=True)
 
+    # pylint: disable=no-value-for-parameter
```
aludvik: Can this be resolved instead of disabling it?

validator/sawtooth_validator/journal/block_store.py
jjason: What about:
```
if not starting_block:
    starting_block = self.chain_head

return _BlockPredecessorIterator(self, start_block=starting_block)
```

jjason: I'll retract my comment....after thinking about it more, I would go with @aludvik recommendation so we don't give lint the opportunity to complain about re-assigning the `starting_block` parameter in a future version.

consensus/poet/core/sawtooth_poet/poet_consensus/poet_config_view.py
```diff
@@ -139,7 +139,8 @@ def enclave_module_name(self):
                     name='sawtooth.poet.enclave_module_name',
                     value_type=str,
                     default_value=PoetConfigView._ENCLAVE_MODULE_NAME_,
-                    validate_function=lambda value: len(value) > 0)
+                    # function should return true if value is nonempty
+                    validate_function=lambda value: value)
```
nick-drozd: Ha, you thought it would be that easy?
```
************* Module sawtooth_poet.poet_consensus.poet_config_view
W:143,38: Lambda may not be necessary (unnecessary-lambda)
```

validator/sawtooth_validator/database/lmdb_nolock_database.py
nick-drozd: Not in any obvious way. The thing works as far as I can tell, so the error might be spurious. In any case, I've alerted @jsmitchell, who can take further action if need be.

validator/sawtooth_validator/journal/consensus/dev_mode/dev_mode_consensus.py
nick-drozd: All of the obvious ways of changing this result in further pylint errors (pylint seems to disagree with this design in general). Nontrivial changes are outside the scope of this PR. I've alerted @jjason.

rest_api/sawtooth_rest_api/config.py
nick-drozd: Added

validator/sawtooth_validator/state/client_handlers.py
nick-drozd: Updated

validator/sawtooth_validator/journal/consensus/dev_mode/dev_mode_consensus.py
nick-drozd: (Actually, @cianx made it, but my general point remains.)

cli/sawtooth_cli/exceptions.py
```diff
@@ -15,5 +15,4 @@
 
 
 class CliException(Exception):
-    def __init__(self, msg):
-        super(CliException, self).__init__(msg)
+    pass
```
nick-drozd: I reworded the relevant commit to reflect this.

validator/sawtooth_validator/database/lmdb_nolock_database.py
jsmitchell: I guess you could try `return txn.stat(self._lmdb)['entries']` to see if it works.

PR #615, opened 2017-06-10
Add state address enforcement to the ContextManager

boydjohnson:
The ContextManager now requires 70 lowercase hex character
addresses, or will raise AuthorizationException.

STL-403


boydjohnson:
A concern: #584 added address enforcement in test code but with a possibly less efficient technique. We should probably be using the same technique in the ContextManager and test code for consistency and maintainability reasons. @cianx @jsmitchell

PR #616, opened 2017-06-11
Add intkey cli test

nick-drozd:
Signed-off-by: Nick Drozd <nicholasx.drozd@intel.com>

### DIFF
--- a/integration/sawtooth_integration/tests/test_intkey_cli.py
+++ b/integration/sawtooth_integration/tests/test_intkey_cli.py
@@ -1,0 +1,78 @@
+# Copyright 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+
+import unittest
+import time
+import logging
+import subprocess
+import shlex
+
+from sawtooth_integration.tests.integration_tools \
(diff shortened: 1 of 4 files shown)


integration/sawtooth_integration/tests/test_intkey_cli.py
```diff
@@ -0,0 +1,65 @@
...
+            _execute('intkey create_batch -v -c {} -o intkey.batch'.format(
+                BATCH_COUNT))
+            _execute('intkey load -v -f intkey.batch -U {}'.format(URL))
+            time.sleep(WAIT_TIME)
+
+        _execute('sawtooth block list --url {}'.format(URL))
+
+        # if the cli commands are working, expect at least
+        # MINIMUM_BLOCK_COUNT blocks to have been created
+        self.assertGreaterEqual(
```
peterschwarz: Is this deterministic?

integration/sawtooth_integration/tests/test_intkey_cli.py
```diff
@@ -0,0 +1,65 @@
...
+from sawtooth_integration.tests.integration_tools \
+    import wait_for_rest_apis, RestClient
+
+
+LOGGER = logging.getLogger(__name__)
+LOGGER.setLevel(logging.INFO)
+
+
+# the appropriateness of these parameters depends on
+# the rate at which blocks are published
```
ojalatodd: Very vague comment - could you be more specific? If not, at least a nod to how to determine the numbers you want. Not going to block on this though.

integration/sawtooth_integration/tests/test_intkey_cli.py
nick-drozd: No. There's no way to verify the expected state, since the batches are created randomly. Another option would be to just run the commands without checking anything, but then it would be possible that the commands failed to do anything. The compromise is to verify that at least a few blocks were published, even if we can't say exactly how many.

I'm open to better ideas on how to test this effectively.

integration/sawtooth_integration/tests/test_intkey_cli.py
nick-drozd: I clarified that comment and also added a general explanation of the test.

PR #620, opened 2017-06-12
Skip state root calculation if no batches

aludvik:
Handle the case where there are no batches.

### DIFF
--- a/validator/sawtooth_validator/execution/scheduler_serial.py
+++ b/validator/sawtooth_validator/execution/scheduler_serial.py
@@ -186,11 +186,14 @@ class SerialScheduler(Scheduler):
         return state_hash
 
     def _calculate_state_root_if_not_already_done(self):
-        last_txn_signature = self._last_in_batch[-1]
-        batch_id = self._txn_to_batch[last_txn_signature]
-        required_state_hash = self._required_state_hashes.get(
-            batch_id)
         if not self._already_calculated:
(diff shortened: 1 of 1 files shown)


validator/sawtooth_validator/execution/scheduler_serial.py
```diff
@@ -186,11 +186,14 @@ def _compute_merkle_root(self, required_state_root):
         return state_hash
 
     def _calculate_state_root_if_not_already_done(self):
-        last_txn_signature = self._last_in_batch[-1]
-        batch_id = self._txn_to_batch[last_txn_signature]
-        required_state_hash = self._required_state_hashes.get(
-            batch_id)
         if not self._already_calculated:
+            if not len(self._last_in_batch) > 0:
```
vaporos: To follow along with the new lint PR, this should probably be "if not self._last_in_batch:"

dcmiddle:
I don't have the context for this. Could you please add a little more about what this is dealing with?

aludvik:
This PR is in response to a bug reported on rocket chat: https://chat.hyperledger.org/channel/sawtooth?msg=tTSYaaFXh6FWdjWe9

validator/sawtooth_validator/execution/scheduler_serial.py
aludvik: okay

PR #625, opened 2017-06-13
[WIP]Add append and remove methods to config transaction family

h4xr:
The PR adds an append and remove method to config blockchain family.
_append_setting_entry allows us to add new values to an existing key
_remove_setting_entry allows us to remove a provided value from an existing key if the provided value exists in the key.

JIRA Issue: STL-261

### DIFF
--- a/families/config/sawtooth_config/processor/handler.py
+++ b/families/config/sawtooth_config/processor/handler.py
@@ -318,6 +318,41 @@ def _append_setting_entry(state, key, value):
                     key, old_value, value)
 
 
+def _remove_setting_entry(state, key, value):
+    address = _make_config_key(key)
+    setting = _get_setting_entry(state, address)
+
+    old_value, old_entry_index = _get_old_entry(setting, key)
+    if old_entry_index is None:
+        LOGGER.warning(
+            'Unable to remove value %s from a non-existent key %s', value, key)
+        raise InternalError(
+            'Unable to remove value from non-existent key {}'.format(key))
+    old_value = old_value.split(',')
+    old_value.remove(value)
+    updated_value = ','.join(old_value)
+    setting.entries[old_entry_index].value = updated_value
+
+    try:
+        addresses = list(state.set(
+            [StateEntry(address=address,
+                        data=setting.SerializeToString())],
+            timeout=STATE_TIMEOUT_SEC))
+    except FutureTimeoutError:
+        LOGGER.warning(
+            'Timeout occured on state.set([%s, <value>])', address)
+        raise InternalError('Unable to set {}'.format(key))
+
+    if len(addresses) != 1:
+        LOGGER.warning(
+            'Failed to save value on address %s', address)
+        raise InternalError(
+            'Unable to save config value {}'.format(key))
+    if setting != 'sawtooth.config.vote.proposals':
+        LOGGER.info('Value %s removed from key %s',
+                    value, key)
+
+
 def _get_setting_entry(state, address):
     setting = Setting()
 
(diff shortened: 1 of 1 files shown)


families/config/sawtooth_config/processor/handler.py
```diff
@@ -246,18 +246,54 @@ def _get_config_value(state, key, default_value=None):
...
+        LOGGER.warning(
+            'Failed to save value on address %s', address)
+        raise InternalError(
+            'Unable to save config value {}'.format(key))
+    if setting != 'sawtooth.config.vote.proposals':
+        LOGGER.info('Config setting %s changed from %s to %s',
+                    key, old_value, value)
+
+
+def _append_setting_entry(state, key, value):
```
jjason: I don't see where this ever gets called.  Is that another story?

families/config/sawtooth_config/processor/handler.py
```diff
@@ -246,18 +246,54 @@ def _get_config_value(state, key, default_value=None):
...
+    address = _make_config_key(key)
+    setting = _get_setting_entry(state, address)
+
+    old_value, old_entry_index = _get_old_entry(setting, key)
     if old_entry_index is not None:
+        value = value + "," + old_value
         setting.entries[old_entry_index].value = value
     else:
         setting.entries.add(key=key, value=value)
```
jjason: Besides, prepending `new_value,` to the existing value if the key already exists, this function looks the same as `_set_config_value`.  Would recommend practicing DRY and refactoring common code.

families/config/sawtooth_config/processor/handler.py
```diff
@@ -318,6 +318,41 @@ def _append_setting_entry(state, key, value):
                     key, old_value, value)
 
 
+def _remove_setting_entry(state, key, value):
```
jjason: I don't see this being called anywhere.   Did I missing something or is that for a future story?

jjason: This function removes a _value_ from a setting.  It doesn't actually remove the setting.  Should it remove the setting if the only or last value is removed or should this function continue doing what it is doing (i.e., when removing the last value the setting is an empty string) and then a new remove setting is created?

families/config/sawtooth_config/processor/handler.py
```diff
@@ -318,6 +318,41 @@ def _append_setting_entry(state, key, value):
...
+
+    if len(addresses) != 1:
+        LOGGER.warning(
+            'Failed to save value on address %s', address)
+        raise InternalError(
+            'Unable to save config value {}'.format(key))
+    if setting != 'sawtooth.config.vote.proposals':
+        LOGGER.info('Value %s removed from key %s',
+                    value, key)
+
```
jjason: This function also seems to share code with `_set_config_value` and `_append_setting_entry`.  Better to refactor and share common code than repeating it.

families/config/sawtooth_config/processor/handler.py
h4xr: jjason, yes, currently, these are just added to implement the functionality. Still have to add the calls to the functions.

h4xr: Thanks, will modify this one to remove the setting if it's empty

jsmitchell:
hi @h4xr - are you planning on continuing with this PR? Otherwise I will probably mark it closed.

h4xr:
Hello,
Sorry for a long inactivity, got occupied with some work. Will be working on this PR this weekend. Hoping to complete it by Monday :)

h4xr:
Hi, I am unable to get some free time to complete the PR and hence closing this PR for now. If anyone will like to work on this one, please feel free to do so.
Hope to work along in future with the awesome team.

PR #627, opened 2017-06-13
Add TpPing handling to the Java, Javascript and Go sdks

agunde406:
This PR Also includes some fix ups to intkey javascript around handling a value of 0.

This PR is built on top of #585.

### DIFF
--- a/sdk/go/src/sawtooth_sdk/processor/processor.go
+++ b/sdk/go/src/sawtooth_sdk/processor/processor.go
@@ -200,7 +200,8 @@ func receiveValidator(ids map[string]string, validator, workers *messaging.Conne
 	corrId := msg.GetCorrelationId()
 
 	// If this is a new request, put in on the work queue
-	if t == validator_pb2.Message_TP_PROCESS_REQUEST {
+	switch t {
+	case validator_pb2.Message_TP_PROCESS_REQUEST:
 		select {
 		case queue <- msg:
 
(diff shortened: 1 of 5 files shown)


sdk/go/src/sawtooth_sdk/processor/processor.go
```diff
@@ -224,6 +225,19 @@ func receiveValidator(ids map[string]string, validator, workers *messaging.Conne
 			}
 		}
 		return
+	case validator_pb2.Message_TP_PING:
+		data, err := proto.Marshal(&processor_pb2.TpPingResponse{
+			Status: processor_pb2.TpPingResponse_OK,
+		})
+		if err != nil {
+			logger.Errorf(
+				"Failed to notify validator the request is denied: %v", err,
```
aludvik: Update this error message to something like "Failed to respond to ping: %v"

sdk/java/src/main/java/sawtooth/sdk/processor/TransactionProcessor.java
```diff
@@ -205,6 +206,19 @@ public void run() {
       if (!this.handlers.isEmpty()) {
         this.currentMessage = this.stream.receive();
         if (this.currentMessage != null) {
+          if (this.currentMessage.getMessageType() == Message.MessageType.TP_PING) {
```
peterschwarz: Should there be a similar check after this to only handle `TP_PROCESS_REQUEST` messages?

PR #628, opened 2017-06-13
New signing keys

rbuysse:


jjason:
Stupid question.  Is there away to do this so that when keys change (I don't know how often they change) you don't have to edit all of these files?  Is there a way to download the key?  If not, can we put the key(s) in a well-known file and then you only have to change it(them) in one place?

rbuysse:
@jjason They shouldn't change again, but that is something I will investigate.

PR #630, opened 2017-06-13
Rename Config Family To Settings

peterschwarz:
This PR renames `sawtooth_config` to `sawtooth_settings` and all related materials.

* `ConfigView` -> `SettingsView` and the like
* `sawtooth.config.*` settings to `sawtooth.settings.*`
* Protobuf messages for the settings txn family are updated from `Config*` to `Setting*`
* Updated the docs

Note that the `sawtooth config` cli tool has been unchanged, as there is a future task to reorganize and separate out the config subcommand into its own command.


ci/ns-control
```diff
@@ -10,7 +10,7 @@ Package: sawtooth
 Version: @VERSION@
 Maintainer: Hyperledger Sawtooth <distributedledger@intel.com>
 # Pre-Depends: <comma-separated list of packages>
-Depends: python3-sawtooth-cli, python3-sawtooth-config, python3-sawtooth-intkey, python3-sawtooth-manage, python3-sawtooth-poet-cli, python3-sawtooth-poet-common, python3-sawtooth-poet-core, python3-sawtooth-poet-families, python3-sawtooth-poet-simulator, python3-sawtooth-rest-api, python3-sawtooth-sdk, python3-sawtooth-signing, python3-sawtooth-validator, python3-sawtooth-xo
+Depends: python3-sawtooth-cli, python3-sawtooth-settings, python3-sawtooth-intkey, python3-sawtooth-manage, python3-sawtooth-poet-cli, python3-sawtooth-poet-common, python3-sawtooth-poet-core, python3-sawtooth-poet-families, python3-sawtooth-poet-simulator, python3-sawtooth-rest-api, python3-sawtooth-sdk, python3-sawtooth-signing, python3-sawtooth-validator, python3-sawtooth-xo
```
rbuysse: Can you update this so it's in alpha order?

families/supplychain/protos/application.proto
```diff
@@ -16,7 +16,7 @@
 syntax = "proto3";
 
 option java_multiple_files = true;
-option java_package = "sawtooth.config.protobuf";
+option java_package = "sawtooth.supplychain.protobuf";
```
rbuysse: Should this be settings?

families/supplychain/protos/record.proto
```diff
@@ -16,7 +16,7 @@
 syntax = "proto3";
 
 option java_multiple_files = true;
-option java_package = "sawtooth.config.protobuf";
+option java_package = "sawtooth.supplychain.protobuf";
```
rbuysse: Should this be settings?

families/supplychain/protos/application.proto
peterschwarz: This was a copy-paste error - it should `supplychain` like the other protos in this family

families/supplychain/protos/record.proto
peterschwarz: This was a copy-paste error - it should `supplychain` like the other protos in this family

peterschwarz:
Rebased with the signing key changes as well.

consensus/poet/core/sawtooth_poet/poet_consensus/poet_settings_view.py
```diff
@@ -249,7 +249,7 @@ def signup_commit_maximum_delay(self):
                 self._get_config_setting(
                     name='sawtooth.poet.signup_commit_maximum_delay',
                     value_type=int,
-                    default_value=PoetConfigView._SIGNUP_COMMIT_MAXIMUM_DELAY_,
+                    default_value=self._SIGNUP_COMMIT_MAXIMUM_DELAY_,
```
jjason: Aside from maybe preventing lint-ing error with respect to 79-column limit, why are some of these `PoetSettingsView` and some are `self`?

I would recommend change all to `self` or change none.

docs/source/app_developers_guide/ubuntu.rst
```diff
@@ -257,34 +257,34 @@ the intkey transaction processor's output is shown below:
...
+-----------------------------------------
 
-To start the config family transaction processor, run the following commands
+To start the settings family transaction processor, run the following commands
 from the Linux CLI:
 
 .. code-block:: console
 
-  $ tp_config
+  $ tp_settings tcp://localhost:4004
```
aludvik: The endpoint is not needed, since that is the default.

docs/source/transaction_family_specifications/settings_transaction_family.rst
```diff
@@ -60,18 +60,18 @@ sawtooth.validator.max_transactions_per_block 1000
...
-| sawtooth.config.vote.authorized_keys     | List of public keys allowed to vote.                       |
-+------------------------------------------+------------------------------------------------------------+
-| sawtooth.config.vote.approval_threshold  | Percentage of keys required for a proposal to be accepted. |
-+------------------------------------------+------------------------------------------------------------+
-| sawtooth.config.vote.proposals           | A list of proposals to make settings changes (see note)    |
-+------------------------------------------+------------------------------------------------------------+
++-------------------------------------------+------------------------------------------------------------+
+| Setting (Settings)                        | Value Description                                          |
++===========================================+============================================================+
+| sawtooth.settings.vote.authorized_keys    | List of public keys allowed to vote.                       |
```
aludvik: Why are these changes not in the previous commit?

consensus/poet/core/sawtooth_poet/poet_consensus/poet_settings_view.py
```diff
@@ -16,12 +16,12 @@
 import math
```
jjason: I thought that our normal approach to renaming that also caused changes to the file was to split this into two commits - one commit was simply renaming the files.  The second commit made any fix-ups to the file.  My understanding was even though it meant that the first commit pulled in isolation would result in a broken system, it was cleaner and clearer to separate out the renaming from the updates.

peterschwarz: I prefer having commits that work, such that no commit causes a broken system.  Renaming these files is renaming the package.  I could see breaking the commit between renaming the class in the file, then the package name itself, but crafting that commit might be quite difficult at this point.

jjason: Having "atomic" commits would be my preference as well.  I would like a policy established on which way we go on this so that these type of moves/renames are done the same way all of the time.  When I moved the PoET consensus code out from the validator package into its own package, I originally followed the approach that moving + fixing were in a single commit so that the commit in isolation didn't break anything.  However, at the time it was requested that I change the commit to split it into a change/rename commit and then a fixup commit.

PR #632, opened 2017-06-14
Check whether block/batch has duplicate batches/transactions

feihujiang:

Check whether block has duplicate batches and check whether batch has duplicate transactions. A block which has duplicate batches would be dropped. A batch which has duplicate transactions would be dropped.


peterschwarz:
@cianx is this duplicating functionality from Journal?

PR #634, opened 2017-06-14
Context Manager namespace enforcement and namespace support

boydjohnson:
This PR enforces that inputs and outputs are 0-70 lowercase hex characters, or fails the transaction. It also makes it so that addresses under a namespace can be correctly found using 'get' on the context manager. Prior to this PR, it was possible to set an address that was partially specified in the outputs, but it wasn't possible to 'get' the address and always find the correct value for that context.

STL-121

### DIFF
--- a/validator/sawtooth_validator/execution/context_manager.py
+++ b/validator/sawtooth_validator/execution/context_manager.py
@@ -375,6 +375,10 @@ class ContextManager(object):
 
         addresses_to_find = list(inputs)
 
+        address_values, reads = self._find_address_values_in_chain(
+            base_contexts=base_contexts,
+            addresses_to_find=addresses_to_find)
+
         context = StateContext(
             state_hash=state_hash,
             read_list=inputs,
@@ -388,6 +392,29 @@ class ContextManager(object):
                 "Basing a new context off of context ids {} "
                 "that are not in context manager".format(
                     contexts_asked_not_found))
+
+        context.create_initial(address_values)
+
+        self._contexts[context.session_id] = context
+
+        if reads:
+            context.create_prefetch(reads)
+            self._address_queue.put_nowait(
+                (context.session_id, state_hash, reads))
+        return context.session_id
+
+    def _find_address_values_in_chain(self, base_contexts, addresses_to_find):
+        """Breadth first search through the chain of contexts searching for
+        the bytes values at the addresses in addresses_to_find.
+
+        Args:
+            base_contexts (list of str): The context ids to start with.
+            addresses_to_find (list of str): Addresses to find values in the
+                chain of contexts.
+
+        Returns:
+            tuple of found address_values and still not found addresses
+        """
 
         contexts_in_chain = deque()
         contexts_in_chain.extend(base_contexts)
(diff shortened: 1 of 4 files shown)


validator/sawtooth_validator/execution/context_manager.py
```diff
@@ -320,6 +320,8 @@ def __init__(self, database, state_delta_store):
 
         self._address_regex = re.compile('^[0-9a-f]{70}$')
 
+        self._namespace_regex = re.compile('^[0-9a-f]{0,70}$')
```
jsmitchell: I think you want to only match even numbers of characters for valid namespaces -- something like ([0-9a-f][0-9a-f]){0,35}$

jjason: @boydjohnson thought, "hmmm....I have a problem, I think I'll use a regular expression to fix it"....now he has two problems :)

or `([0-9a-f]{2}){0,35}`

Reading regular expressions reminds me of the old obfuscated C coding contests.

boydjohnson: Thanks. I'll update.

boydjohnson: Haha! @jjason

boydjohnson: Just to make sure, @jsmitchell , can you provide the reasoning on why namespaces must be even number of characters?

jsmitchell: A full address is 70 hex characters which represents 35 bytes. Each byte is an address path element in the radix addressing scheme. So, a given non-leaf node in the tree can have up to 2^8 pointers to children. I have always thought of a namespace as a proper sub-tree, but maybe this is unnecessary?

boydjohnson: With respect to 'Each byte is an address path element in the radix addressing scheme". That is how it is currently, but I thought that STL-149 would be changing that, potentially. Or would it be that at minimum the address path element would be 1 byte and potentially more.

I'm trying to think of a situation where it might make sense for the input or output to reference a non-subtree. It would be odd, because a client couldn't query it. If it make sense to you to stick with the even number of characters in the namespace requirement I'll change it.

jsmitchell: Even with the STL-149 changes, a byte will still be the atomic unit of address pathing. There will be cases where more than one byte is in the pointer, but never less than one byte. My preference is for enforcing an even number of characters.

boydjohnson: Cool. I'll do that.

boydjohnson:
Here's that PR. #638

validator/tests/test_context_manager/tests.py
```diff
@@ -211,6 +211,182 @@ def test_address_enforcement(self):
...
                                     {invalid_address2: b'2'}])
 
+    def test_exception_on_invalid_input(self):
+        """Tests that invalid inputs raise an exception. Tested with invalid
+        characters, odd number of characters, and too long namespace;
+
+        Notes:
+            1) Assert that inputs with a namespace with an odd number of
+               characters raises a CreateContextException.
+            2) Assert that inputs with a 71 character namespace raises a
```
ojalatodd: Grammar - subj./verb agreement for plural vs. singular. Suggestion:
"Assert that inputs with a 71 character namespace raise a..."

validator/tests/test_context_manager/tests.py
```diff
@@ -211,6 +211,182 @@ def test_address_enforcement(self):
...
+               CreateContextException.
+            3) Assert that inputs with a namespace with several invalid
```
ojalatodd: Same as above. s/raises/raise

boydjohnson:
@ojalatodd I fixed the subject/verb agreement problems, but did so while rebasing so I still only have 3 commits.

PR #635, opened 2017-06-14
Initial draft of the SupplyChain rest api spec.

cianx:
Defines domain specific entity query and retreival for state
managed by the SupplyChain Transaction Processor.


families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,377 @@
+swagger: '2.0'
+
+info:
+  version: "0.8.0"
+  title: Sawtooth Supplychain REST API
+  description: |
+    _This HTTP pragmatic REST API is built on top of Sawtooth Lake's existing
+    ZMQ/Protobuf infrastructure, simplifying SupplyChain clients interaction
```
jsmitchell: s/clients/clients'/

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,377 @@
...
+    with the blockchain by exposing endpoints that use common HTTP(S)/JSON
+    standards. The SupplyChain REST API provides secure means for clients to
```
jsmitchell: s/provides secure/provides a secure/

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,377 @@
...
+    access SupplyChain records held in global state in a easy manner. The
+    rest api provides retrieval using natural keys of the the records.
```
jsmitchell: be consistent with style on REST API

jsmitchell: I think you mean a comma after records, not a period.

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,377 @@
...
+    provides indexing and query capabilities for Agents, Applications, and
+    Records stored with in._
```
jsmitchell: s/with in/within/

jsmitchell: be consistent with style on Records/records

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,377 @@
...
+
+paths:
+  /agents/:
+    get:
+      summary: Fetches the data for the current set of agents in state
+      description: >
+        Fetches a paginated list of leaves for the current state, or relative
```
jsmitchell: It doesn't seem like the domain specific representation should know anything about addresses or the fact that things are stored in a tree underneath.

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,377 @@
...
+                $ref: "#/definitions/Link"
+              paging:
+                $ref: "#/definitions/Paging"
+        400:
+          $ref: "#/responses/400BadRequest"
+        500:
+          $ref: "#/responses/500ServerError"
+        503:
+          $ref: "#/responses/503ServiceUnavailable"
+  /applications/{record_id}:
```
jsmitchell: Should this be {application_id}?

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,377 @@
...
+    get:
+      summary: Fetches the data for the current set of agents in state
+      description: >
+        Fetches a paginated list of leaves for the current state, or relative
+        to a particular head block. Using the `address` filter parameter, will
+        narrow the list to any leaves that have an address beginning with the
+        characters specified.
+      parameters:
+        - $ref: "#/parameters/head"
+        - name: address
```
jsmitchell: shouldn't this be some agent natural key?

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,377 @@
...
+    type: string
+    required: true
+    description: Agent identifier, the public key the agent uses.
+  record_id:
+    name: record_id
+    in: path
+    type: string
+    required: true
+    description: "Record identifier, The serial number or natural identifier
+     of the item(material) being tracked."
```
delventhalz: Switch to:
```
Record identifier, the serial number or natural identifier
of the item (material) being tracked.
```

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,377 @@
...
+  head:
```
delventhalz: Should be removed if the `head` parameter is not being used with this API.

families/supplychain/restapi/openapi.yaml
cianx: Applications are stored by record id. I have added text to the descriptions to clarify this.

cianx: Applications are stored as a list under record id. I have add description text to make it clear.

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,377 @@
...
+                $ref: "#/definitions/Link"
+              paging:
+                $ref: "#/definitions/Paging"
+        400:
+          $ref: "#/responses/400BadRequest"
+        500:
+          $ref: "#/responses/500ServerError"
+        503:
+          $ref: "#/responses/503ServiceUnavailable"
+  /agents/{agent_id}:
```
cianx: These are added, github did not detect the change...

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,398 @@
...
+    _This HTTP pragmatic REST API is built on top of Sawtooth Lake's existing
+    ZMQ/Protobuf infrastructure, simplifying SupplyChain client interactions
+    with the blockchain by exposing endpoints that use common HTTP(S)/JSON
+    standards. The SupplyChain REST API provides a secure means for clients to
+    access SupplyChain objects held in global state in a easy manner.
+
+    The REST API provides retrieval using natural keys. Natural keys are unique
+    keys that identify the objects in the real world.
+    (https://en.wikipedia.org/wiki/Natural_key) This provides abstraction of
+    the underlying implementation and achitecture of the Sawtooth blockchain,
```
jsmitchell: s/achitecture/architecture/

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,398 @@
...
+        500:
+          $ref: "#/responses/500ServerError"
+        503:
+          $ref: "#/responses/503ServiceUnavailable"
+  /applications:
+    get:
+      summary: Fetches the data for the current set of Applications
+      description: >
+        Fetches a paginated list of all Applications. Using the `applicant`
+        parameter, will narrow the list to any Applications that have a
```
jsmitchell: remove comma

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,398 @@
...
+        matching value in their applicant field. Using the `identifier` filter
+        parameter, will narrow the list to any Applications that have an
```
jsmitchell: remove comma

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,398 @@
...
+        identifier containing the specified string. Using the `state`
+        parameter, will narrow the list to any Applications that have a
```
jsmitchell: remove comma

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,398 @@
...
+        matching value in their state field.
+
+      parameters:
+        - name: applicant
+          in: query
+          type: string
+          description: The public key of the applicatant Agent to filter by
```
jsmitchell: s/applicatant/applicant/

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,398 @@
...
+          $ref: "#/responses/503ServiceUnavailable"
+  /applications/{record_id}:
+    parameters:
+      - $ref: "#/parameters/record_id"
+    get:
+      summary: Fetches a particular Application.
+      description: >
+        Fetches a paginated list of Applications for a record. Applications are
+        identified by the identifier of the Record they are associated with.
+        Using the `applicant` parameter, will narrow the list to any
```
jsmitchell: remove comma

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,398 @@
...
+        Applications that have a matching value in their applicant field. Using
+        the `state` parameter, will narrow the list to any Applications that
```
jsmitchell: remove comma

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,398 @@
...
+        500:
+          $ref: "#/responses/500ServerError"
+        503:
+          $ref: "#/responses/503ServiceUnavailable"
+  /records:
+    get:
+      summary: Fetches the data for the current set of Records
+      description: >
+        Fetches a paginated list of all Records. Using the `identifier` filter
+        parameter, will narrow the list to any Records that have an identifiers
```
jsmitchell: remove comma

jsmitchell: s/identifiers/identifier/

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,398 @@
...
+        containing the specified string.
+
+        Records  are identified by their serial number or other marking on the
```
jsmitchell: extra space between 'Records' and 'are'

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,398 @@
+swagger: '2.0'
+
+info:
+  version: "0.8.0"
+  title: Sawtooth Supplychain REST API
+  description: |
+    _This HTTP pragmatic REST API is built on top of Sawtooth Lake's existing
+    ZMQ/Protobuf infrastructure, simplifying SupplyChain client interactions
```
peterschwarz: Should this be against the Sawtooth REST API?  I'm not sure we are promoting client applications directly against the ZMQ interface, just yet (there's no easy SDK method for this).

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,398 @@
...
+
+      parameters:
+        - name: applicant
+          in: query
+          type: string
+          description: The public key of the applicatant Agent to filter by
+        - name: identifier
+          in: query
+          type: string
+          description: A partial identifier to filter Applications by.
```
peterschwarz: Do partial identifiers still work as filters in this case (and the others), or is this partial idea a hold-over from cloning this file from the base Sawtooth REST API spec?

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,398 @@
...
+    name: max
+    in: query
+    type: string
+    description: Id or index to end paging (inclusive)
+  sort:
+    name: sort
+    in: query
+    type: string
+    description: |
+      Comma-separated keys to sort a list by:
```
peterschwarz: Many of these sort parameters don't make sense for this API - they shouldn't be referencing headers and signing keys, necessarily, but fields on the objects returned by the call.

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,398 @@
...
+    required: true
+    description: Agent identifier, the public key the agent uses.
+  record_id:
+    name: record_id
+    in: path
+    type: string
+    required: true
+    description: "Record identifier, the serial number or natural identifier
+      of the item (material) being tracked."
+  wait:
```
peterschwarz: Wait isn't used by any other call, so this should be removed.

families/supplychain/restapi/openapi.yaml
cianx: The filter will allow reterival of applications open against a subset of records.

peterschwarz: Are ID's capable of defining subsets by partial ID's?

families/supplychain/restapi/openapi.yaml
cianx: We can implement it either way for this iteration. As long as we keep our data requests small additional data handling probably won't be an issue. The future will probably have this component become a state export client, so it can build and maintain it's own indexes for more efficient access and reporting.

families/supplychain/restapi/openapi.yaml
jsmitchell: Still says applicantant

families/supplychain/restapi/openapi.yaml
delventhalz: If this is getting built against the ZMQ/Protobuf interface, it will require significant rewriting of existing functionality. The correct answer is probably to refactor the REST API so it's key components are exportable in some sort of REST API SDK, or to have this API submit HTTP requests to a REST API instead of using ZMQ/Protobuf directly.

I don't hate the idea of a REST API SDK, but that is a sprint or two worth of work, so I feel like building this on top of the existing REST API is the correct answer here.

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,378 @@
...
+      applicant:
+        type: string
+        example: 02d260a46457a064733153e09840c322bee1dff34445d7d49e19e60abd18fd0758
+      type:
+        type: string
+        example: OWNER
+        enum:
+          - OWNER
+          - CUSTODIAN
+      status:
```
delventhalz: This appears to be the same value as the `state` parameter above. This value should only have one name. Status seems more meaningful to me.

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,378 @@
...
+        type: string
+        example: OWNER
```
delventhalz: This example is not in the `enum` below. Typo?

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,378 @@
...
+        enum:
+          - OPEN
+          - CANCELED
+          - REJECTED
+          - ACCEPTED
+      terms:
+        type: string
+        example: Text describing the terms of the transfer application.
```
delventhalz: This reads like a description, not an example.

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,378 @@
...
+        type: string
+        example: https://api.sawtooth.com/agents?min=0&count=1000
+      next:
+        type: string
+        example: https://api.sawtooth.com/agents?min=2000&count=1000
+  Error:
+    properties:
+      code:
+        type: integer
+        example: 34
```
cianx: https://github.com/hyperledger/sawtooth-core/blob/master/rest_api/sawtooth_rest_api/openapi.yaml#L494

cianx: I don't understand the linkage of those exception codes to the user of the RestAPI. How would some one calling the Sawtooth RestAPI know to go look in some python file for the error code they were returned?

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,378 @@
...
+        filter parameter, will narrow the list to any Agents that have an
+        identifier or name respectively containing the specified string.
+
+        Agents are identified by their public keys.
+      parameters:
+        - name: identifier
+          in: query
+          type: string
+          description: A partial public_key to filter Agents by
+        - name: name
```
cianx: Partial name searches are useful, the description specifies contains not starts with filtering. I have expanded the description to specify case insensitivity filtering. so `?name=john` would get `"John Smith"`, `"Smith, John"`, `"Jimmy Johnson"`, `"Billy John Jasper Johanson"`, etc

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,378 @@
...
+    get:
+      summary: Fetches the data for the current set of Applications
+      description: >
+        Fetches a paginated list of all Applications. Using the `applicant`
+        parameter will narrow the list to any Applications that have a
+        matching value in their applicant field. Using the `identifier` filter
+        parameter will narrow the list to any Applications that have an
+        identifier containing the specified string. Using the `state`
+        parameter will narrow the list to any Applications that have a
+        matching value in their state field.
```
cianx: Corrected to be status field, so it matches the Application proto definition.

families/supplychain/restapi/openapi.yaml
delventhalz: I'm not convinced on the use case (why do I want everyone named John?), but if you are going to implement it as described, I would clarify that it is case-insensitive.

families/supplychain/restapi/openapi.yaml
delventhalz: My comment is on the entire example error, not just the error code. The examples should all be something that might plausibly get returned. If you use an existing error code, for an error that does not apply to your API, then we know this example will never actually be returned. I was suggesting using an existing error your API would be likely to use.

As to your other question, a _user_ would not be interested in an error code any more than they are interested in a `404` status code, but a _developer_ might be, especially if they plan to write code around it. I don't expect them to read the python source code, but the official documentation here:

http://intelledger.github.io/rest_api/error_codes.html

delventhalz: Note also that a key part of this scheme is that _the error code is fixed_. It is a contract I have established with the developer that any given error code will _always_ refer to the same error and _never_ change. This is part of why I am sensitive to any Sawtooth documentation stating that an error code means something different than it does, even if technically it is a different API. There are enough integers that we don't need to have any collisions here.

cianx: agreed, I need to have a corresponding error code document. I need to find a home for it in he doc directory.

cianx: I will add the error api document to my supply chain documention update later this week.

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,377 @@
...
+                $ref: "#/definitions/Link"
+              paging:
+                $ref: "#/definitions/Paging"
+        400:
+          $ref: "#/responses/400BadRequest"
+        500:
+          $ref: "#/responses/500ServerError"
+        503:
+          $ref: "#/responses/503ServiceUnavailable"
+  /records/{record_id}:
```
cianx: Updated github did not detect the change.

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,378 @@
...
+          $ref: "#/responses/500ServerError"
+        503:
+          $ref: "#/responses/503ServiceUnavailable"
+  /applications/{record_id}:
+    parameters:
+      - $ref: "#/parameters/record_id"
+    get:
+      summary: Fetches a particular Application.
+      description: >
+        Fetches a paginated list of Applications for a record. Applications are
```
cianx: Updated this to be /resource/{record_identifier}/applications

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,378 @@
...
+        parameter will narrow the list to any Records that have an identifier
+        containing the specified string.
+
+        Records are identified by their serial number or other marking on the
+        physical asset.
+      parameters:
+        - name: identifier
+          in: query
+          type: string
+          description: A partial identifier to filter Records by
```
cianx: They could be anything, we are expecting the common case to serial numbers which could be a arbitrary identifier or a "smart number", if it is a "smart number" then a partial filter can be useful

families/supplychain/restapi/openapi.yaml
```diff
@@ -0,0 +1,352 @@
...
+          $ref: "#/responses/404NotFound"
+        500:
+          $ref: "#/responses/500ServerError"
+        503:
+          $ref: "#/responses/503ServiceUnavailable"
+  /records/{record_id}/applications:
+    parameters:
+      - $ref: "#/parameters/record_id"
+    get:
+      summary: Fetches a particular Application.
```
peterschwarz: The summary is singular, but the description is plural.

families/supplychain/restapi/openapi.yaml
delventhalz: So, to be clear, I wasn't asking for full error documentation. I think that is premature. I was asking that just this one particular example be a valid error that the API might send back. It's not the end of the world, but I prefer examples to be something you might actually see in the wild whenever possible.

The reason I think this is inherently an _invalid_ example is because `34: Agent Not Found`, defines an error code that already exists in the main REST API and means something else. While technically you could do this, I would strongly object to redefining error codes in any official Sawtooth APIs, because it could cause confusion. I only suggested you use `54: Invalid Paging Query` because it's already defined, and I imagine the SupplyChain API will reuse it.

Anyway. I'm not requesting changes on this. Just wanted to make sure my concern was understood.

PR #638, opened 2017-06-15
Correct input, output namespace checking

boydjohnson:
As a consequence of another PR review, @aludvik and I realized that this checking for if an address was under any namespace in the inputs or outputs for a Context wasn't correct. This PR fixes the checking and provides a unit test.

### DIFF
--- a/validator/sawtooth_validator/execution/context_manager.py
+++ b/validator/sawtooth_validator/execution/context_manager.py
@@ -282,7 +282,7 @@ class StateContext(object):
             AuthorizationException
         """
 
-        if not any([o in address for o in self._write_list]):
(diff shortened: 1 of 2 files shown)


validator/sawtooth_validator/execution/context_manager.py
```diff
@@ -282,7 +282,7 @@ def _validate_write(self, address):
             AuthorizationException
         """
 
-        if not any([o in address for o in self._write_list]):
+        if not any([address.startswith(ns) for ns in self._write_list]):
```
peterschwarz: You can drop the brackets around the generator here - make it lazier

boydjohnson: Ok. I'll do so.

PR #640, opened 2017-06-15
Added help for proxy setup in Docker to App Dev Guide

ojalatodd:
Section on configuring Docker to work with proxy servers
added to App Dev Guide, for Windows, macOS and Linux.


jjason:
When I was getting docker working on Linux, I also had to edit `/etc/resolv.conf` so that DNS resolution worked properly.  Have we verified that everything works with the instructions provided without needing to edit this file as well?  Or will that be a different PR/story?

PR #648, opened 2017-06-16
Fix Xo wait response message

dplumb94:
Currently using the `--wait` option with `xo create` causes a different response than all other commands. This PR fixes this inconsistency.

_note: A future PR should add more descriptive responses for all Xo transactions._

### DIFF
--- a/sdk/examples/xo_python/sawtooth_xo/xo_client.py
+++ b/sdk/examples/xo_python/sawtooth_xo/xo_client.py
@@ -179,7 +179,7 @@ class XoClient:
         if wait and wait > 0:
             wait_time = 0
             start_time = time.time()
-            self._send_request(
+            response = self._send_request(
                 "batches", batch_list.SerializeToString(),
                 'application/octet-stream',
                 auth_user=auth_user,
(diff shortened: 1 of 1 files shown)


sdk/examples/xo_python/sawtooth_xo/xo_client.py
```diff
@@ -179,34 +179,22 @@ def _send_xo_txn(self, name, action, space="", wait=None,
...
-            self._send_request(
+            response = self._send_request(
                 "batches", batch_list.SerializeToString(),
                 'application/octet-stream',
                 auth_user=auth_user,
                 auth_password=auth_password
             )
             while wait_time < wait:
-                status = self._get_status(
+                self._get_status(
```
delventhalz: Without checking the status, this will just keep looping until the wait time expires.

delventhalz:
@boydjohnson @dplumb94 These are interesting questions we should probably sort out.

The future PR @dplumb94 references, is refactoring the output from _all_ xo commands to be something more friendly than raw JSON. STL-185 seems to conflate that with the wait parameter which doesn't make much sense to me.

That said, I think it's a good call that the `take` command, and any other command that submits a batch, should be able to `--wait`. There either needs to be a JIRA task for that, or it should be added to this PR.

sdk/examples/xo_python/sawtooth_xo/xo_client.py
```diff
@@ -194,19 +194,10 @@ def _send_xo_txn(self, name, action, space="", wait=None,
...
-                            'Try again with a different name')
-                elif status[batch_id] == 'UNKNOWN':
-                    return ('Error: Something went wrong with your game. '
-                            'Try again with a different name.')
-                # Wait a moment so as not to hammer the Rest Api
-                time.sleep(0.2)
-
-            return ('Timed out while waiting for game to be created. '
-                    'It may need to be resubmitted.')
+                if status[batch_id] in ['COMMITTED', 'INVALID', 'UNKNOWN']:
```
delventhalz: This would also work:
```
if status[batch_id] != 'PENDING`
```

PR #649, opened 2017-06-16
Refactor wait functionality to use Observer pattern

delventhalz:
This PR adds a `notification` module to the validator, allowing notifications to be sent from one component to another using a generic abstracted interface. As a test of its capabilities, the existing *wait for commit* client behavior has been refactored to use this interface.

### DIFF
--- a/validator/sawtooth_validator/journal/block_store.py
+++ b/validator/sawtooth_validator/journal/block_store.py
@@ -13,12 +13,11 @@
 # limitations under the License.
 # ------------------------------------------------------------------------------
 
-from time import time
-from threading import Condition
 # pylint: disable=no-name-in-module
 from collections.abc import MutableMapping
 
 from sawtooth_validator.exceptions import PossibleForkDetectedError
+from sawtooth_validator.state.notification import Observable
 from sawtooth_validator.journal.block_wrapper import NULL_BLOCK_IDENTIFIER
 from sawtooth_validator.journal.block_wrapper import BlockStatus
 from sawtooth_validator.journal.block_wrapper import BlockWrapper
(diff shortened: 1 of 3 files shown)


validator/sawtooth_validator/state/notification.py
```diff
@@ -0,0 +1,89 @@
...
+        """Notifies all observers, sending optional data if provided.
+
+        Args:
+            data (optional): Any extra data that the observer may need
+        """
+        for observer in self._observers.copy():
+            observer.notify(self, data)
+
+
+def _get_class_name(obj, default='Object'):
```
peterschwarz: Any reason to not just use the built-in `type` function?  E.g. `type(obj)`

delventhalz: Well this returns a string name like `'Observer'` instead of `<class 'Observer'>`. If there is a better way to do that, I'm happy to do it. Given my previous Python experience, `object.__class__.__name__` within a nested try/except block seemed like it was probably the simplest option.

validator/sawtooth_validator/journal/block_store.py
```diff
@@ -77,6 +76,13 @@ def __str__(self):
             out.append(str(value))
         return ','.join(out)
 
+    def set_batch(self, add_pairs, del_keys=None):
```
cianx: This should be private and should be named some that is not likely to get confused with Sawtooth Batches. perhaps "_commit_to_block_store()"

delventhalz: Good call.

delventhalz:
This code will all be thrown out in favor of a more specific BatchTracker. Closing this PR.

PR #650, opened 2017-06-16
Add PoET SGX Enclave Code to Repository

jjason:
This PR brings the PoET SGX code into the sawtooth-core repository.  It also makes any changes necessary to account for the change in directory structure from its original location.
The first commit is simply adding all of the files.  The remaining commits clean up build, path, lint, issues with moving the code to the new location.

If you want to build the code, please contact me directly and I can give you information on how to do so.  It is slightly involved and you need additional pieces of information that are not part of the code.

If you want to test the code there is a much higher bar as you need not only additional packages on top of what is needed to build, but you also need hardware with SGX capabilities.


consensus/poet/sgx/sawtooth_poet_sgx/libpoet_bridge/enclave.cpp
```diff
@@ -0,0 +1,890 @@
...
+        // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
+        void Enclave::LoadEnclave()
+        {
+            if (!this->enclaveId) {
+                /* Enclave id, used in communicating with enclave */
+                Enclave::QuerySgxStatus();
+
+                sgx_launch_token_t token = { 0 };
+                int flags = SGX_DEBUG_FLAG;
+//              #ifdef _DEBUG
```
peterschwarz: Commented code

consensus/poet/sgx/sawtooth_poet_sgx/libpoet_bridge/utils.h
```diff
@@ -0,0 +1,107 @@
...
+            const char* inEncodedBuffer
+            )
+        {
+            std::string dataBuffer = base64_decode(inEncodedBuffer);
+
+            ThrowIf<ValueError>(
+                dataBuffer.length() != sizeof(T),
+                "Data size mismatch");
+
+            memcpy(outBinaryData, dataBuffer.data(), dataBuffer.length());
```
peterschwarz: Should you zero-out this `outBinaryData`? Similar to line 47?

jjason:
@dan For some reason, the link between Jenkins and github seems bonked.  The third build passed =)  I did not add any of the SGX tests to the `run_tests` script so I don't _think_ they should be run.

consensus/poet/sgx/sawtooth_poet_sgx/libpoet_bridge/enclave.cpp
jjason: That's old cruft.  Let me see what that flag is.

consensus/poet/sgx/sawtooth_poet_sgx/libpoet_bridge/utils.h
jjason: Let me check

consensus/poet/sgx/sawtooth_poet_sgx/libpoet_bridge/enclave.cpp
jjason: Fixed.

consensus/poet/sgx/sawtooth_poet_sgx/libpoet_bridge/utils.h
jjason: Removed this function as it is never used.

PR #653, opened 2017-06-19
Implement state handling for parallel scheduler

boydjohnson:
As of this PR, all scheduler unit tests pass and poet-smoke with a heterogeneous scheduler network passes often, but not always. STL-488 and STL-476 will hopefully improve that.

Note: Any transaction that uses dependencies will not work with the parallel scheduler at this point. https://jira.hyperledger.org/browse/STL-499 When STL-499 is complete, dependencies will work with the parallel scheduler.

### DIFF
--- a/validator/sawtooth_validator/execution/scheduler_parallel.py
+++ b/validator/sawtooth_validator/execution/scheduler_parallel.py
@@ -265,13 +265,15 @@ class TransactionExecutionResult:
 
 
 class ParallelScheduler(Scheduler):
-    def __init__(self, squash_handler, first_state_hash):
+    def __init__(self, squash_handler, first_state_hash, always_persist):
         self._squash = squash_handler
         self._first_state_hash = first_state_hash
         self._last_state_hash = first_state_hash
         self._condition = Condition()
         self._predecessor_tree = PredecessorTree()
         self._txn_predecessors = {}
+
+        self._always_persist = always_persist
 
         # Transaction identifiers which have been scheduled.  Stored as a list,
         # since order is important; SchedulerIterator instances, for example,
@@ -482,7 +484,7 @@ class ParallelScheduler(Scheduler):
                 contexts = self._get_contexts_for_squash(batch_signature)
                 state_hash = self._squash(self._first_state_hash,
                                           contexts,
-                                          persist=False,
+                                          persist=self._always_persist,
                                           clean_up=True)
             return BatchExecutionResult(is_valid=True, state_hash=state_hash)
 
--- a/validator/sawtooth_validator/execution/scheduler_parallel.py
+++ b/validator/sawtooth_validator/execution/scheduler_parallel.py
@@ -249,9 +249,9 @@ class PredecessorTree:
 
 class TransactionExecutionResult:
     def __init__(self, is_valid, context_id=None, state_hash=None):
-        if context_id is not None and state_hash is not None:
+        if is_valid and context_id is None:
             raise ValueError(
-                "context_id and state_hash are exclusive arguments")
+                "There must be a context_id for valid transactions")
         if not is_valid and context_id is not None:
             raise ValueError(
                 "context_id must be None for invalid transactions")
@@ -285,6 +285,9 @@ class ParallelScheduler(Scheduler):
         # All batches in their natural order (the order they were added to
         # the scheduler.
         self._batches = []
+        # The batches that have state hashes added in add_batch, used in
+        # Block validation.
+        self._batches_with_state_hash = {}
 
         # Indexes to find a batch quickly
         self._batches_by_id = {}
@@ -329,6 +332,10 @@ class ParallelScheduler(Scheduler):
             self._batches_by_id[batch.header_signature] = batch
             for txn in batch.transactions:
                 self._batches_by_txn_id[txn.header_signature] = batch
+
+            if state_hash is not None:
+                b_id = batch.header_signature
+                self._batches_with_state_hash[b_id] = state_hash
 
             # For dependency handling: First, we determine our dependencies
             # based on the current state of the predecessor tree.  Second,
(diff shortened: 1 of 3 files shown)


validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -377,34 +381,90 @@ def _get_batch_index(self, batch):
...
+        if not self._is_valid_batch(batch):
+            return False
+        index_of_next = self._batches.index(batch) + 1
+        for later_batch in self._batches[index_of_next:]:
+            if self._is_valid_batch(later_batch):
+                return False
+        return True
+
+    def _get_contexts(self, batch_signature):
+        batch = self._batches_by_id[batch_signature]
```
boydjohnson: Working on making this clearer.

validator/tests/test_scheduler/tests.py
```diff
@@ -2069,3 +2096,10 @@ def node_to_string(node, indent=2):
         string += node_to_string(child_node, indent + 2)
 
     return string
+
```
boydjohnson: This was added through a bad diff edit. Removing.

validator/tests/test_scheduler/tests.py
```diff
@@ -43,7 +43,55 @@
...
+    def _setup_parallel_scheduler(self):
+        context_manager = self._context_manager
+        squash_handler = context_manager.get_squash_handler()
+        first_state_root = context_manager.get_first_root()
+        scheduler = ParallelScheduler(squash_handler,
+                                      first_state_root)
+        return context_manager, scheduler
+
+    def test_parallel_simple_scheduler_test(self):
```
nick-drozd: Is there a more compact way of describing these four tests? It's just "run each of these two tests against each of these two schedulers", but the code doesn't make that obvious.

boydjohnson: Yeah, and what makes it more complicated is there is a yaml file in test_scheduler/data that goes along with each test. I made the yaml filename part of the test method name, so 'test_parallel_simple_scheduler_test' is the parallel scheduler being run against the 'simple_scheduler_test.yaml' file. Do you have any thoughts on how to make it clearer?

boydjohnson:
@agunde406 I verified by modifying the test_workload.yaml test, and that fails with a parallel scheduler on this branch. I will work to solve this.

boydjohnson:
@nick-drozd I think your comment got lost in a force push. Do you have ideas for making those tests clearer? The 'test_parallel_simple_scheduler_test' tests the parallel scheduler against the simple_scheduler_test.yaml file.

nick-drozd:
It might be time to think about reorganizing the test file, which will, after this PR, be by far the longest test file:
```
ubuntu@ubuntu-xenial:/project/sawtooth-core$ find . -iname test*.py -exec wc -l {} \; | sort -nr | head -n 20
2091 ./validator/tests/test_scheduler/tests.py
1166 ./validator/tests/test_journal/tests.py
1084 ./validator/tests/test_context_manager/tests.py
888 ./consensus/poet/families/tests/test_tp_validator_registry.py
880 ./consensus/poet/core/tests/test_consensus/test_poet_block_publisher.py
863 ./consensus/poet/core/tests/test_consensus/test_consensus_state.py
858 ./validator/tests/test_client_request_handlers/test_txn_handlers.py
823 ./rest_api/tests/unit/test_state_requests.py
819 ./consensus/poet/common/tests/test_sgx_structs/tests.py
805 ./rest_api/tests/unit/test_txn_requests.py
783 ./rest_api/tests/unit/test_block_requests.py
783 ./rest_api/tests/unit/test_batch_requests.py
658 ./validator/tests/test_client_request_handlers/test_state_handlers.py
643 ./validator/tests/test_client_request_handlers/test_block_handlers.py
642 ./consensus/poet/core/tests/test_consensus/test_poet_fork_resolver.py
619 ./validator/tests/test_client_request_handlers/test_batch_handlers.py
606 ./validator/tests/test_state_delta_processor/tests.py
590 ./consensus/poet/core/tests/test_consensus/test_poet_block_verifier.py
528 ./consensus/poet/core/tests/test_consensus/test_poet_key_state_store.py
453 ./consensus/poet/core/tests/test_consensus/test_poet_settings_view.py
```

nick-drozd:
@boydjohnson It looks like the general pattern is there is the actual test (say, `_finalize_blah`) which then gets a different wrapper for each scheduler (`test_serial_finalize_blah` and `test_parallel_finalize_blah`). To buy a little more compactness, you could combine the wrappers (assuming they don't collide):
```
def test_finalize_blah(self):
    parallel = _set_up_parallel_scheduler()
    _finalize_blah(parallel)
    serial = _set_up_serial_scheduler()
    _finalize_blah(serial)
```
In other words, you would run a single test against both schedulers.

Alternatively, you could set up the test suite in such a way that it runs all the tests against the serial scheduler and then runs all the tests against the parallel scheduler.

validator/tests/test_scheduler/tests.py
```diff
@@ -43,7 +43,56 @@
...
+    def _setup_parallel_scheduler(self):
+        context_manager = self._context_manager
+        squash_handler = context_manager.get_squash_handler()
+        first_state_root = context_manager.get_first_root()
+        scheduler = ParallelScheduler(squash_handler,
+                                      first_state_root,
+                                      always_persist=False)
+        return context_manager, scheduler
+
+    def test_parallel_simple_scheduler_test(self):
```
peterschwarz: Docstring these tests

boydjohnson: @peterschwarz I am going to put the parallel and serial tests under one test method, and then I'll put the docstring on the test method.

boydjohnson:
@agunde406 and I figured out it was the use of explicit dependencies in 'intkey workload' that caused the problems. STL-499 is the story about implementing support for explicit dependencies in the parallel scheduler.

vaporos:
I'd rather see more separation between serial and parallel tests (I like the test suite idea or something similar), but reorganizing the tests would be better done in a separate dedicated PR, and maybe even potentially after this is merged into master to avoid conflict/merge work between the branches.

boydjohnson:
@nick-drozd Since the tests in master have test_serial_blah as the pattern which fits with the parallel-scheduler branch which has test_serial_blah and test_parallel_blah, I will make a task in jira to be done after the parallel-scheduler branch is merged into master. @peterschwarz I will also add docstrings to the test methods.

PR #656, opened 2017-06-19
Send back transaction failure information via the Java SDK

peterschwarz:
Send back the message and optional extended data on the `TpProcessResponse`.


sdk/java/src/main/java/sawtooth/sdk/processor/TransactionProcessor.java
```diff
@@ -140,31 +139,29 @@ private static void process(Message message, Stream stream,
...
+        }
       } catch (InternalError ie) {
         logger.log(Level.WARNING, "State Exception!: " + ie.toString());
-        TpProcessResponse response = TpProcessResponse.newBuilder()
-                .setStatus(TpProcessResponse.Status.INTERNAL_ERROR)
-                .build();
-        stream.sendBack(Message.MessageType.TP_PROCESS_RESPONSE,
-                message.getCorrelationId(),
-                response.toByteString());
+        builder.setStatus(TpProcessResponse.Status.INVALID_TRANSACTION);
```
boydjohnson: Shouldn't this be INTERNAL_ERROR? Invalid_transaction is in the block above.

PR #658, opened 2017-06-19
modified test_wait_timer.py to allow faster testing without sleeps

nconde:


jjason:
Please clean up first commit by removing commented out code and prints, then rebase and re-push PR.

nconde:
Those were fixed in the second commit

jjason:
@nconde Instead of being removed in the second commit, I believe that the group "standard" is to not allow them in the first commit.

nconde:
@jjason fixed

PR #644, opened 2017-06-16
Print the version information for the sawtooth

feihujiang:
Signed-off-by: feihujiang <jiangfeihu@huawei.com>

Add /version endpoint for REST API, and add sawtooth version cli.

PR #659, opened 2017-06-20
Add a version option for cli

feihujiang:

Ref #644
Implement a "--version, -V" argument for the sawtooth command. And if this way is accepted, I'll make the same for all installed commands.

### DIFF
--- a/cli/sawtooth_cli/main.py
+++ b/cli/sawtooth_cli/main.py
@@ -20,6 +20,7 @@ import logging
 import os
 import traceback
 import sys
+import pkg_resources
 
 from colorlog import ColoredFormatter
 
@@ -82,6 +83,18 @@ def create_parent_parser(prog_name):
         '-v', '--verbose',
         action='count',
         help='enable more verbose output')
+
+    try:
+        version = pkg_resources.get_distribution('sawtooth-cli').version
+    except pkg_resources.DistributionNotFound:
+        version = 'UNKNOWN'
+
+    parent_parser.add_argument(
+        '-V', '--version',
+        action='version',
+        version='sawtooth-cli (Hyperledger Sawtooth) version {}'
+                .format(version),
+        help='print version information')
 
     return parent_parser
 
--- a/cli/setup.py
+++ b/cli/setup.py
@@ -15,7 +15,6 @@
 
 from __future__ import print_function
 
-import os
 import subprocess
 
 from setuptools import setup, find_packages


feihujiang:
@vaporos Please review this PR, thanks.

cli/sawtooth_cli/main.py
```diff
@@ -83,6 +84,18 @@ def create_parent_parser(prog_name):
         action='count',
         help='enable more verbose output')
 
+    try:
+        version = pkg_resources.get_distribution('sawtooth-cli').version
+    except pkg_resources.DistributionNotFound:
+        print("please install sawtooth-cli with setup.py")
```
vaporos: Let's just set "version = 'UNKNOWN'" here instead of making it an error.

cli/sawtooth_cli/main.py
```diff
@@ -83,6 +84,18 @@ def create_parent_parser(prog_name):
...
+        return
+
+    parent_parser.add_argument(
+        '-V', '--version',
+        action='version',
+        version='%(prog)s {}'.format(version),
```
vaporos: Let's do '%(prog)s (Hyperledger Sawtooth) {}'.  This is somewhat similar to some GNU utilities like grep, sed, etc.

I may be better to do 'sawtooth-cli (Hyperledger Sawtooth) {}' to make it clear that it's the sawtooth-cli's version number explicitly.  This is a bit confusing with the using %(prog)s since it is 'sawtooth' which is ambiguous.

cli/sawtooth_cli/main.py
```diff
@@ -83,6 +84,18 @@ def create_parent_parser(prog_name):
...
+        help='print the version number of the executable')
```
vaporos: Maybe 'print version information'

vaporos:
Exception thrown when just running sawtooth in vagrant:

```ubuntu@ubuntu-xenial:/project/sawtooth-core$ ./bin/sawtooth
please install sawtooth-cli with setup.py
  File "/project/sawtooth-core/cli/sawtooth_cli/main.py", line 160, in main_wrapper
    main()
  File "/project/sawtooth-core/cli/sawtooth_cli/main.py", line 125, in main
    parser = create_parser(prog_name)
  File "/project/sawtooth-core/cli/sawtooth_cli/main.py", line 107, in create_parser
  File "/usr/lib/python3.5/argparse.py", line 1658, in __init__
    self._add_container_actions(parent)
  File "/usr/lib/python3.5/argparse.py", line 1400, in _add_container_actions
    for group in container._action_groups:
AttributeError: 'NoneType' object has no attribute '_action_groups'
```

cli/sawtooth_cli/main.py
```diff
@@ -83,6 +84,18 @@ def create_parent_parser(prog_name):
...
 
+    try:
+        version = pkg_resources.get_distribution('sawtooth-cli').version
+    except pkg_resources.DistributionNotFound:
+        print("please install sawtooth-cli with setup.py")
+        return
+
+    parent_parser.add_argument(
+        '-V', '--version',
+        action='version',
```
boydjohnson: I think you want action='store_true' as the -V flag is a boolean. https://docs.python.org/3.5/howto/argparse.html.

cli/sawtooth_cli/main.py
feihujiang: Done

cli/sawtooth_cli/main.py
feihujiang: Done

cli/sawtooth_cli/main.py
feihujiang: No, I want to use ```parser.add_argument('--version', action='version', version='...')``` on https://docs.python.org/3.5/library/argparse.html, The version= keyword argument prints version information and exits when invoke.

feihujiang:
@vaporos @boydjohnson I updated this PR, please take a look. Thanks.

boydjohnson:
I verified locally that sawtooth -V prints out the version in the file that was added.

feihujiang:
@vaporos Could you please review this PR, thanks.

vaporos:
@feihujiang what is the reason for switching away from "pkg_resources.get_distribution('sawtooth-cli').version"?  That seemed to be a pretty clean approach.

cli/sawtooth_cli/version.py
```diff
@@ -0,0 +1 @@
+VERSION = """0.8.6.dev1"""
```
vaporos: This should maybe be """UNKNOWN""".  But a bigger problem here is that because this file will be overwritten by setup.py, it will cause the git repository to be dirty -- which causes different output from 'git describe', which is used by the get_version script.

feihujiang: The reason for switching away from "pkg_resources.get_distribution('sawtooth-cli').version" is that I want to have an explicit version file. But, yes, you are right. This makes the problem.  I'll switch back.

feihujiang:
@vaporos @peterschwarz @boydjohnson  I updated this PR, please take a look.

PR #660, opened 2017-06-20
Revised and added to Using Sawtooth with Docker

ojalatodd:
Revised Using Sawtooth With Docker to follow new outline,
and added additional information and sub-sections.

### DIFF
--- a/docs/source/app_developers_guide/docker.rst
+++ b/docs/source/app_developers_guide/docker.rst
@@ -33,8 +33,8 @@ transaction processors. The second validator's transaction processors are not
 depicted.
 
 
-Installing Docker Engine and Docker Compose
-===========================================
+Install Docker Engine And Docker Compose
+========================================
 
 Windows
 -------
@@ -70,10 +70,10 @@ On Linux, follow these steps:
   Linux distributions often ship with older versions of Docker.
 
 
-Installation
-============
+Environment Setup
+=================
 
-Download the Docker Compose File
+Download The Docker Compose File
 --------------------------------
 
 A Docker Compose file is provided which defines the process for constructing a
@@ -123,8 +123,8 @@ See the instructions for proxy configuration in
 `Control and configure Docker with Systemd <https://docs.docker.com/engine/admin/systemd/#httphttps-proxy>`_.
 
 
-Environment Startup
--------------------
+Starting Sawtooth
+-----------------
 
 To start up the environment, perform the following tasks:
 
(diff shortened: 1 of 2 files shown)


docs/source/app_developers_guide/docker.rst
```diff
@@ -490,3 +461,234 @@ look similar to this:
...
+  % docker exec -it sawtooth-validator-default bash
+
+Then run the following commands from the validator container:
+
+.. code-block:: console
+
+  $ sawtooth config proposal create \
+    --url http://rest_api:8080 \
+    --key /root/.sawtooth/keys/my_key.priv \
+    sawtooth.validator.transaction_families='[{"family": "intkey", "version": "1.0", "encoding": "application/protobuf"}, {"family":"sawtooth_settings", "version":"1.0", "encoding":"application/protobuf"}]'
```
boydjohnson: For intkey, it should be "application/cbor".

jsmitchell: This probably needs to be fixed in a few places in the docs. This caught somebody with a very strange error in rocketchat over the weekend. The log message from executor is "failing transaction %s of type(name=intkey, ...) since it isn't required in the configuration.

ojalatodd: Fixed in both Ubuntu and Docker versions of the tutorial.

ojalatodd:
Note: until the docker containers are updated, you will run into problems testing this document. 
The container **sawtooth-tp_config-default** was changed to **sawtooth-tp_settings-default**, but the docker containers published have not been updated.

ojalatodd:
Note 2: Also, the **intkey** command on the docker images does not yet include the --count option, which was recently added: 
intkey create_batch --count 10 --key-count 5

docs/source/app_developers_guide/docker.rst
```diff
@@ -490,3 +461,234 @@ look similar to this:
...
+
+.. code-block:: console
+
+  $ docker exec -it sawtooth-validator-default bash
+
+To see which components are running, run this command from the container:
+
+.. code-block:: console
+
+  $ ps -ef | grep python | grep -v grep
```
aludvik: With docker, the command you give as the entrypoint for an image will be started with PID 1 within the container and PPID 0. A better command for this would be something like `ps --pid 1 fw`. That will just show whatever the entrypoint for the container was, which should be what they want to see.

ojalatodd: Awesome. Changing now.

PR #655, opened 2017-06-19
Add xo transaction family specification

nick-drozd:
Signed-off-by: Nick Drozd <nicholasx.drozd@intel.com>

PR #663, opened 2017-06-20
Rewrite xo python handler to match documentation

nick-drozd:
This PR updates the xo python handler to match the transaction family tutorial and the xo transaction family spec (https://github.com/hyperledger/sawtooth-core/pull/655).

### DIFF
--- a/sdk/examples/xo_python/sawtooth_xo/processor/handler.py
+++ b/sdk/examples/xo_python/sawtooth_xo/processor/handler.py
@@ -50,27 +50,13 @@ class XoTransactionHandler:
         name, action, space, signer = _unpack_transaction(transaction)
 
         # 2. Retrieve the game data from state storage
-
         board, state, player1, player2, game_list = \
             _get_state_data(state_store, name)
 
         # 3. Validate the game data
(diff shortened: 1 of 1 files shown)


sdk/examples/xo_python/sawtooth_xo/processor/handler.py
```diff
@@ -47,164 +47,235 @@ def namespaces(self):
...
+            raise InvalidTransaction(
+                "Not this player's turn: {}".format(signer[:6]))
+
+        if board[space - 1] != '-':
+            raise InvalidTransaction(
+                'Invalid Action: space {} already taken'.format(space))
+
+
+def _make_xo_address(name):
+    return '5b7349' + hashlib.sha512(name.encode('utf-8')).hexdigest()[:64]
```
agunde406: Is there a reason to not use self.namespace_prefix here?

nick-drozd: Passing in a namespace as an argument to the handler seems to me like an obsolete feature that can be removed, so I didn't want to get it involved. That said, if it's important, I can add it in to my changes.

PR #664, opened 2017-06-20
Refactor existing wait/status functionality with new BatchTracker

delventhalz:
BatchTracker utilizes the observer pattern to collect notifications from elsewhere in the validator about the progress of submitted batches. Other validator components call the appropriate notify method at points where a status change has occurred.
    
All batch status and wait-for-commit client code, including unit tests, has been refactored to query this new BatchTracker, which can be queried for the current statuses of batches, or send out a notification itself when all specified batches are no longer PENDING.

### DIFF
--- a/rest_api/tests/unit/test_submit_requests.py
+++ b/rest_api/tests/unit/test_submit_requests.py
@@ -172,7 +172,7 @@ class PostBatchTests(BaseApiTest):
             - a link property that ends in '/batches?id=a'
         """
         batches = Mocks.make_batches('a')
-        statuses = {'a': self.status.COMMITTED}
+        statuses = {'a': client_pb2.COMMITTED}
         self.stream.preset_response(batch_statuses=statuses)
 
         request = await self.post_batches(batches, wait=True)
@@ -204,7 +204,7 @@ class PostBatchTests(BaseApiTest):
             - a data property matching the batch statuses received
         """
         batches = Mocks.make_batches('pending')
-        statuses = {'pending': self.status.PENDING}
+        statuses = {'pending': client_pb2.PENDING}
         self.stream.preset_response(batch_statuses=statuses)
 
         request = await self.post_batches(batches, wait=True)
@@ -245,7 +245,7 @@ class BatchStatusTests(BaseApiTest):
             - a link property that ends in '/batch_status?id=pending'
             - a data property matching the batch statuses received
         """
-        statuses = {'pending': self.status.PENDING}
+        statuses = {'pending': client_pb2.PENDING}
         self.stream.preset_response(batch_statuses=statuses)
 
         response = await self.get_assert_200('/batch_status?id=pending')
@@ -303,7 +303,7 @@ class BatchStatusTests(BaseApiTest):
             - a link property that ends in '/batch_status?id=pending&wait'
             - a data property matching the batch statuses received
         """
-        statuses = {'pending': self.status.COMMITTED}
+        statuses = {'pending': client_pb2.COMMITTED}
         self.stream.preset_response(batch_statuses=statuses)
 
         response = await self.get_assert_200('/batch_status?id=pending&wait')
@@ -334,9 +334,9 @@ class BatchStatusTests(BaseApiTest):
             - a data property matching the batch statuses received
         """
         statuses = {
-            'committed': self.status.COMMITTED,
-            'unknown': self.status.UNKNOWN,
-            'bad': self.status.UNKNOWN}
+            'committed': client_pb2.COMMITTED,
+            'unknown': client_pb2.UNKNOWN,
+            'bad': client_pb2.UNKNOWN}
         self.stream.preset_response(batch_statuses=statuses)
 
         response = await self.get_assert_200(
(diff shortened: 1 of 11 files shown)


validator/sawtooth_validator/state/batch_tracker.py
```diff
@@ -0,0 +1,147 @@
...
+
+        self._lock = RLock()
+        self._observers = {}
+
+    def notify_committed(self, blocks):
+        """Adds new batches to the committed list, and removes them from the
+        pending list if present.
+
+        Args:
+            blocks (str): The Blocks being committed
```
agunde406: I don't think that blocks is a string in this context. Should this be a list of blocks?

validator/tests/test_client_request_handlers/mocks.py
```diff
@@ -143,18 +132,20 @@ def make_db_and_store(size=3, start='a'):
...
     The three batch ids in the store will be:
         * 'b-0', 'b-1', b-2'
-    And the two batch ids in the cache will be:
-        * 'b-2', 'b-3'
     """
-    store = MockBlockStore(size=size)
-    cache = MockBatchCache(start=str(size-1))
-    return store, cache
+    tracker = BatchTracker()
+    tracker.notify_pending('b-3')
```
peterschwarz: This pending batch is no longer mentioned in the dock string.  Is it still valid?

delventhalz: Meant to replace the docstring with the contents of the batch cache, with one with the contents of the tracker. It has been added back in.

rest_api/tests/unit/test_submit_requests.py
```diff
@@ -172,7 +172,7 @@ class PostBatchTests(BaseApiTest):
             - a link property that ends in '/batches?id=a'
         """
         batches = Mocks.make_batches('a')
-        statuses = {'a': self.status.COMMITTED}
+        statuses = {'a': client_pb2.COMMITTED}
```
jsmitchell: doesn't this need to be client_pb2.BatchStatus.COMMITTED?

validator/sawtooth_validator/journal/block_store.py
```diff
@@ -161,13 +151,11 @@ def _build_add_block_ops(self, blkw):
...
         out = []
         blk_id = blkw.identifier
-        with self._commit_condition:
-            out.append((blk_id, blkw.block.SerializeToString()))
-            for batch in blkw.batches:
-                out.append((batch.header_signature, blk_id))
-                for txn in batch.transactions:
-                    out.append((txn.header_signature, blk_id))
-            self._commit_condition.notify_all()
+        out.append((blk_id, blkw.block.SerializeToString()))
```
jsmitchell: How this works is not clear to me. It seems like the returned list is a list of tuples which might be blocks or batches or transactions. Are we not differentiating between these different types from a notification perspective?

validator/sawtooth_validator/state/batch_tracker.py
```diff
@@ -0,0 +1,147 @@
...
+    only be relied on for batches submitted locally, and is not persisted
+    after restart.
+
+    When a batch moves from one component to another, the appropriate notify
+    method should be called in the appropriate component:
+        * notify_committed
+        * notify_pending
+    """
+    def __init__(self):
+        self._committed = set()
```
jsmitchell: What prevents these structures from growing in an unbounded way?

rest_api/tests/unit/test_submit_requests.py
delventhalz: No, the way Protobuf's referenced enums is a little confusing. You reference the wrapping element, not the name of the enum. This enum is at the root level of the proto file, so it is just `client_pb2.COMMITTED`.

(Confusingly, I am about to wrap this enum in a _message_ named `BatchStatus`, and then it _will be_ `client_pb2.BatchStatus.COMMITTED`, but that will be in a future PR).

validator/sawtooth_validator/journal/block_store.py
delventhalz: Other than the `_commit_condition`, this code is not mine. I believe you are correct that the tuples include blocks, batches, and txns, and the wait condition should never have been placed here. It's replacement (`_notify_observers`), is not going elsewhere.

validator/sawtooth_validator/state/batch_tracker.py
delventhalz: They are unbounded. `_committed` will include the `header_signature` of every block in the blockchain, and `_pending` will include every submitted batch _not_ in the blockchain.

In a future PR, a timed cache for invalid batches will be included, which means every submitted batch will end up in either `_committed` or `_invalid` and the invalid ones will expire. However, until that PR is merged, one could spam invalid batches and clog up the `_pending` list. I will make `_pending` a timed cache until that next PR comes in.

delventhalz:
A significant refactor has been pushed to this PR, removing the local cache of committed batches from the BatchTracker, and instead querying the BlockStore directly.

PR #667, opened 2017-06-21
Correct BlockPublisher test for empty blocks to use the CandidateBloc…

cianx:
…k state.

### DIFF
--- a/validator/sawtooth_validator/journal/publisher.py
+++ b/validator/sawtooth_validator/journal/publisher.py
@@ -72,11 +72,15 @@ class _CandidateBlock(object):
     def previous_block_id(self):
         return self._block_builder.previous_block_id
 
+    def has_pending_batches(self):
+        return len(self._pending_batches) != 0
+
     @property
     def last_batch(self):
         if self._pending_batches:
             return self._pending_batches[-1]
-        return None
+        raise ValueError('last_batch called on an empty block.'
+                         'Empty block publishing is not supported.')
 
     @property
     def can_add_batch(self):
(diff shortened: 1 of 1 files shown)


cianx:
Addresses the root cause of this issue, 
```
```
[19:07:46.282 CRITICAL publisher] on_check_publish_block exception.                                                   
[19:07:46.282 ERROR    publisher] None is not in list                                                                 
  File "/project/sawtooth-core/validator/sawtooth_validator/journal/publisher.py", line 505, in on_check_publish_block
    last_batch_index = self._pending_batches.index(last_batch)                                                        
ValueError: None is not in list  
```
Originally seen by @peterschwarz

File: validator/sawtooth_validator/journal/publisher.py [hunk 1]
```diff
@@ -490,8 +494,9 @@ def on_check_publish_block(self, force=False):
                         self._pending_batches:
                     self._build_candidate_block(self._chain_head)
 
-                if self._candidate_block and \
-                        (force or self._pending_batches) and \
+                if self._candidate_block and (
+                            force or
+                            self._candidate_block.has_pending_batches()) and \
```
peterschwarz: Given that if `_pending_batches` is empty, it evaluates to false, this doesn't necessarily fix the cause as to why `last_batch` was being called in the first place.  The error message from last_batch might be slightly more informative, though.

cianx: self._pending_batches and self._candidate_block._pending_batches are different lists and are not guarenteed to have the same content. self._candidate_block.last_batch should never be called on a Candididate block as we don't publish empty blocks.

peterschwarz: The diff did not make that clear.   I think I see now.

peterschwarz: A more informative commit message would have made it easier to understand what the change was.

PR #669, opened 2017-06-22
Replace client wait with wait query parmeters

feihujiang:

Use the ```wait``` query parameter  of ```POST /batches``` endpoint to replace client wait process.

### DIFF
--- a/sdk/examples/xo_python/sawtooth_xo/xo_client.py
+++ b/sdk/examples/xo_python/sawtooth_xo/xo_client.py
@@ -109,7 +109,7 @@ class XoClient:
         return xo_prefix + game_address
 
     def _send_request(
-            self, suffix, data=None,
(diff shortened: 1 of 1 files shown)


delventhalz:
This PR removes functionality. If merged, the XO Client will be unable to `--wait` for times longer than 300s.

After ~300s to REST API will timeout on it's own, regardless of how long you asked it to wait. This is configurable, but sooner or later you will hit limits with how long you can keep an HTTP request open (usually around 300s). The solution implemented here is to make a new request on every timeout, until the `--wait` time is expired, or the Batch is no longer `PENDING`.

Some tweaks to the REST API could be made to better support >300s wait times, but for now at least, this is the recommended way to implement an indefinite wait.

feihujiang:
@vaporos @delventhalz Thanks for your detailed explanation. I got it, I closed this PR.

PR #670, opened 2017-06-22
EVM Nonc'ing

aludvik:
This PR does some refactoring in order to handle nonces cleanly, handles nonces on both the client and processor side, and also correctly restricts code to contract accounts.

As usual, github is showing my commits out of order. Locally, `git log` shows me:
```
eb9f2713 Check and update EVM nonce's correctly
65878a10 Add support for nonces to EVM client
280cac9d Use new address types in EVM Family
28bacd6d Add unit tests for address interop
e0e006b6 Encapsulate EVM and State address interop
d3adbf40 Rename burrow_evm_client package to client
d61d8542 Allow only contract accounts to have code
16177dd2 Refactor contract creation code
```


integration/sawtooth_integration/tests/burrow_evm_intkey_test.go
```diff
@@ -89,8 +106,8 @@ func TestIntkey(t *testing.T) {
 
 }
 
-func exec(t *testing.T, client *burrow_evm_client.Client, priv, to, data []byte) {
-  _, err := client.Exec(priv, to, data, 1000)
+func exec(t *testing.T, c *client.Client, priv, to, data []byte, nonce uint64) {
```
aludvik: I will disagree with you that one-letter parameters are always bad. In this case it is a 3-line function in a test file, so I would argue that it is fine. If this was a much longer function with a lot going on, a more descriptive name would be needed.

To support my claim, I cite the Go code review comments page: https://github.com/golang/go/wiki/CodeReviewComments#variable-names

peterschwarz: See my overall comment about naming.  This line was merely an instance where you were removing the context of a variable for no gain in readability.

families/burrow_evm/src/sawtooth_burrow_evm/client/client.go
```diff
@@ -47,23 +47,29 @@ func New(url string) *Client {
 	}
 }
 
-func (c *Client) Load(priv, init []byte, gas uint64) (*RespBody, error) {
-	address, err := PrivToAddr(priv)
+func (c *Client) Load(priv, init []byte, gas, nonce uint64) (*RespBody, error) {
```
peterschwarz: This is ugly - don't hide the type of gas by implicitly typing it the same is nonce (it has always been a terrible language feature in any language - it saves writing but adds cognitive load on reading).  Be explicit.

families/burrow_evm/src/sawtooth_burrow_evm/client/client.go
```diff
@@ -88,29 +94,35 @@ func (c *Client) Load(priv, init []byte, gas uint64) (*RespBody, error) {
 	return ParseRespBody(resp)
 }
 
-func (c *Client) Exec(priv, to, data []byte, gas uint64) (*RespBody, error) {
-	fromAddr, err := PrivToAddr(priv)
+func (c *Client) Exec(priv, to, data []byte, gas, nonce uint64) (*RespBody, error) {
```
peterschwarz: Same as above.

families/burrow_evm/src/sawtooth_burrow_evm/processor/handler/sawtooth_app_state.go
```diff
@@ -41,36 +41,47 @@ func NewSawtoothAppState(mgr *StateManager) *SawtoothAppState {
 
 // GetAccount retrieves an existing account with the given address. Panics if
 // the account doesn't exist.
-func (s *SawtoothAppState) GetAccount(vmAddress Word256) *Account {
-	logger.Debugf("GetAccount(%v)", vmAddress.Bytes())
+func (s *SawtoothAppState) GetAccount(w Word256) *Account {
```
peterschwarz: Again, one letter name - what is `w` is lost, and not accounted for in your function comment.

PR #671, opened 2017-06-22
Add TP Tutorial for Java SDK

dplumb94:
- Add TP tutorial for Java SDK
- Update Javascript and Python addressing section to conform with XO spec

### DIFF
--- a/docs/source/_templates/sdk_TP_tutorial.rst
+++ b/docs/source/_templates/sdk_TP_tutorial.rst
@@ -1,11 +1,15 @@
 {% set short_lang = 'python' %}
 {% if language == 'JavaScript' %}
     {% set short_lang = 'js' %}
+{% elif language == 'Java' %}
+    {% set short_lang = 'java' %}
 {% endif %}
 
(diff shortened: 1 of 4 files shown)


docs/source/_templates/sdk_TP_tutorial.rst
```diff
@@ -319,6 +398,29 @@ updated state of the game and store it back at the address from which it came.
...
+      Map.Entry<String, ByteString> entry
+          = new AbstractMap.SimpleEntry<>(address, gameByteString);
+
+      Collection<Map.Entry<String, ByteString>> addressValues
+          = Collections.singletonList(entry);
+
+      Collection<String> addresses = stateStore.set(addressValues);
+
+      if (addresses.size() < 1) {
+        throw new InternalError("State Error");
```
peterschwarz: Realizing that this might be a bug in the transaction family itself:  this should probably be an invalid transaction, as it most likely means that that inputs and outputs for the transaction don't match the generated transaction.

PR #672, opened 2017-06-23
REST API: Replace SDK Stream with asyncio Connection

peterschwarz:
Use of the sawtooth python SDK Stream required blocking for the results
of concurrent futures, within the confines of an asyncio loop.  In order
to avoid blocking the loop on waits, a thread pool executer was
introduced to handle these futures.

This commit replaces the concurrent stream with an asyncio-based
Connection class.  This class performs all operations on an asyncio
loop, which is shared with the aiohttp application.  All calls to and
from the validator are handled using coroutines.

This is in support of STL-223

### DIFF
--- a/rest_api/sawtooth_rest_api/messaging.py
+++ b/rest_api/sawtooth_rest_api/messaging.py
@@ -1,0 +1,288 @@
+# Copyright 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


rest_api/sawtooth_rest_api/messaging.py
```diff
@@ -0,0 +1,290 @@
...
+        """
+        poller = Poller()
+        poller.register(self._socket, zmq.POLLIN)
+        self._is_running = True
+
+        while self._is_running:
+            try:
+                events = await poller.poll()
+                if self._socket in dict(events):
+                    zmq_msg = await self._socket.recv_multipart()
```
boydjohnson: Since this is a Dealer socket and we only want the bytes transported, I think self._socket.recv() will work.

peterschwarz: the Pyzmq doc recommends using recv_multipart: https://pyzmq.readthedocs.io/en/v16.0.2/api/zmq.asyncio.html#socket

peterschwarz:
Made some minor tweaks with receiving (no zmq.Poller), and removed the custom logging in favor of aiohttp standard logging.

PR #676, opened 2017-06-23
Send back invalid transaction data through APIs

delventhalz:
This PR expands the BatchTracker to track invalid Batches, which transaction was rejected, and any data associated with the transaction. This data is then sent back through both the ZMQ and REST APIs with a more robust list of status objects, rather than a simple id to status hashmap.

### DIFF
--- a/protos/client.proto
+++ b/protos/client.proto
@@ -30,16 +30,33 @@ message Leaf {
     bytes data = 2;
 }
 
-// Possible commit statuses for a batch submitted to a validator
-//   * COMMITTED - the batch was accepted and has been committed to the chain
-//   * INVALID - the batch failed validation, it should be resubmitted
-//   * PENDING - the batch is still being processed
-//   * UNKNOWN - no status for the batch could be found (possibly invalid)
-enum BatchStatus {
-    COMMITTED = 0;
-    INVALID = 1;
-    PENDING = 2;
-    UNKNOWN = 3;
+// Information about the status of a batch submitted to the validator.
+//
+// Attributes:
+//     batch_id: The id (header_signature) of the batch
+//     status: The committed status of the batch
+//     invalid_transactions: Info for transactions that failed, if any
+//
+// Statuses:
+//     COMMITTED - the batch was accepted and has been committed to the chain
+//     INVALID - the batch failed validation, it should be resubmitted
+//     PENDING - the batch is still being processed
+//     UNKNOWN - no status for the batch could be found (possibly invalid)
+message BatchStatus {
+    enum Status {
+        COMMITTED = 0;
+        INVALID = 1;
+        PENDING = 2;
+        UNKNOWN = 3;
+    }
+    message InvalidTransaction {
+        string transaction_id = 1;
+        string message = 2;
+        bytes extended_data = 3;
+    }
+    string batch_id = 1;
+    Status status = 2;
+    repeated InvalidTransaction invalid_transactions = 3;
 }
 
 // Paging controls to be sent with List requests.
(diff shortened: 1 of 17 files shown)


protos/client.proto
```diff
@@ -30,16 +30,32 @@ message Leaf {
...
+message BatchStatus {
+    enum Status {
+        COMMITTED = 0;
+        INVALID = 1;
+        PENDING = 2;
+        UNKNOWN = 3;
+    }
+    string batch_id = 1;
+    Status status = 2;
+    string invalid_transaction = 3;
```
peterschwarz: This possibly should be a list of invalid transactions, but I can't remember if batch processing fails fast.

rest_api/sawtooth_rest_api/route_handlers.py
```diff
@@ -834,6 +834,18 @@ def _set_wait(self, request, validator_query):
                 validator_query.timeout = int(self._timeout * 0.95)
 
     @staticmethod
+    def _format_statuses(statuses):
+        """Reformat converted BatchStatus dicts: drop empty keys, rename 'id'.
```
peterschwarz: Perhaps a note that this mutates the status list - returning the value makes it appear like it's creating a new list. Or, just drop the return value, since it looks like it should be properly changed without it.

validator/sawtooth_validator/state/batch_tracker.py
```diff
@@ -93,6 +125,28 @@ def get_statuses(self, batch_ids):
         with self._lock:
             return {b: self.get_status(b) for b in batch_ids}
 
+    def get_invalid_txn_info(self, batch_id):
```
peterschwarz: Is this used anywhere in this commit?  Or should it be in a different commit?

rest_api/sawtooth_rest_api/route_handlers.py
```diff
@@ -835,16 +835,17 @@ def _set_wait(self, request, validator_query):
 
     @staticmethod
     def _format_statuses(statuses):
-        """Reformat converted BatchStatus dicts: drop empty keys, rename 'id'.
+        """Reformat converted BatchStatus dicts: rename keys and drop empties.
         """
         for status in statuses:
             status['id'] = status.pop('batch_id')
+            status['invalid_message'] = status.pop('message', '')
```
peterschwarz: Why rename these?  It makes it so the message is only for invalid statuses, and can't be used for any other status.  No reason that valid transactions could return extended data in the future.

validator/sawtooth_validator/state/batch_tracker.py
delventhalz: Nothing uses any invalid txn functionality in this commit. It just adds the functionality. This method is the only way to access that new functionality. It would only make sense to move it to the next commit (where it is used), if the two commits were combined.

rest_api/sawtooth_rest_api/openapi.yaml
```diff
@@ -125,10 +125,9 @@ paths:
     get:
       summary: Fetches the committed statuses for a set of batches
       description: |
-        Fetches an object with a status for each batch requested. The keys of
-        the data object will be the ids of the batch(es), while the values will
-        be a string status with the values `'COMMITTED'`, `'INVALID'`,
-        `'PENDING'`, or `'UNKNOWN'`.
+        Fetches an array of objects with a status and id for each batch
+        requested. There are for possible statuses with string values
```
agunde406: for -> Four

protos/client.proto
```diff
@@ -30,16 +30,32 @@ message Leaf {
...
-//   * COMMITTED - the batch was accepted and has been committed to the chain
-//   * INVALID - the batch failed validation, it should be resubmitted
-//   * PENDING - the batch is still being processed
-//   * UNKNOWN - no status for the batch could be found (possibly invalid)
-enum BatchStatus {
-    COMMITTED = 0;
-    INVALID = 1;
-    PENDING = 2;
-    UNKNOWN = 3;
+// Information about the the status of a batch submitted to the validator.
```
agunde406: Remove extra "the"

protos/client.proto
```diff
@@ -30,16 +30,32 @@ message Leaf {
...
+//
+// Attributes:
+//     batch_id: The id (header_signature) of the batch
+//     status: The committed status of the batch
+//     invalid_transaction: The header_signature of the txn that was rejected
+//     message: An message with additional information for an invalid batch
```
agunde406: An -> A

PR #677, opened 2017-06-26
Use WIF encoding for private keys in Go SDK

aludvik:
This PR depends on PR #670


aludvik:
Split the commit into more focused commits and added the dependency to vagrant.

sdk/go/src/sawtooth_sdk/client/signer.go
```diff
@@ -91,6 +92,22 @@ func SHA256(input []byte) []byte {
...
 }
 
+// -- WIF --
+
+// PrivToWif converts a private key generated to a WIF string
+func PrivToWif(priv []byte) string {
+	extended := append([]byte{0x80}, priv...)
+	checksum := SHA256(SHA256(extended))[:4]
+	extcheck := append(extended, checksum...)
+	return base58.Encode(extcheck)
```
dcmiddle: Is this just translating to base58? You might need checkencode(). I'm not sure off the top of my head if that's sufficient for wif or if there are other requirments.

aludvik:
We are already using https://github.com/btcsuite/btcd which has the same license.  Additionally, https://github.com/tendermint/go-crypto is using that library but is licensed with Apache 2.0 https://github.com/tendermint/go-crypto/blob/master/LICENSE. I think the license is okay based on that.

aludvik:
Used a private key generated from sawtooth keygen for all the tests.

aludvik:
@vaporos, @TomBarnes found that the ISC license is okay.

PR #678, opened 2017-06-26
Change PoET policy logging error messages to informational

jjason:
The consensus state and block publisher error messages when PoET policy will be violated have been demoted from error to informational logging messages.


jjason:
Sanity check rebase before merge

PR #680, opened 2017-06-28
Yaml tester fixes, increase in yaml tests

boydjohnson:
The first three commits make the yaml testing framework better. The yaml test added in the last commit will be improved by a change to the yaml testing framework, in a future PR. It will also be able to be passed based on work done under issue STL-476, coming in a future PR. This PR is against the parallel-scheduler branch.

### DIFF
--- a/validator/tests/test_scheduler/yaml_scheduler_tester.py
+++ b/validator/tests/test_scheduler/yaml_scheduler_tester.py
@@ -151,7 +151,10 @@ class SchedulerTester(object):
         """
         return self._batch_results
 
-    def run_scheduler(self, scheduler, context_manager, validation_state_hash=None):
+    def run_scheduler(self,
+                      scheduler,
+                      context_manager,
+                      validation_state_hash=None):
         """Add all the batches to the scheduler in order and then run through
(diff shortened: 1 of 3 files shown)


validator/tests/test_scheduler/yaml_scheduler_tester.py
```diff
@@ -191,8 +192,14 @@ def run_scheduler(self,
                     txns_to_process.append(txn_info)
                 else:
                     stop = True
-            t_info = txns_to_process.pop()
+
+            if txns_executed_fifo:
```
jjason: If you are going to pop from either end, should you switch `txns_to_process` to a `deque` instead?  I don't have a handle on how large this list might be, but popping from the front is not efficient...and you have chosen that as the default.

boydjohnson: I will switch to a deque. The size of the list would depend on the number of transactions in the yaml file, and the number of independent transactions at any one time for the parallel scheduler. When the serial scheduler is being tested, txns_to_process is always 1.

boydjohnson:
Updated test comments to accurately describe the yaml files after talking with @agunde406 .

PR #681, opened 2017-06-28
Add ability for yaml scheduler tester to calculate state for each transaction

boydjohnson:
This PR depends on #680 and is made against the parallel-scheduler branch. The first commit is unrelated to the second commit, but is a cleanup commit from a prior PR. The second commit implements the ability to assert that input addresses for a transaction have particular values in state.


boydjohnson:
Closing for now. I think the way I designed it, there is an assertion that each state is correct at the first time of running. This may not always be correct.

PR #682, opened 2017-06-29
Add Dockerfile for publishing JS SDK to NPM

delventhalz:
Also fixes some small typos that broke the JS SDK, and adds a prepublish hook to run tests and compile protobufs.

Note that this Docker will only run if you are in a clean tagged version of that Sawtooth repo.

### DIFF
--- a/sdk/javascript/package.json
+++ b/sdk/javascript/package.json
@@ -5,7 +5,8 @@
   "main": "index.js",
   "scripts": {
     "test": "standard && node_modules/mocha/bin/mocha --recursive spec",
-    "compile_protobuf": "node compile_protobuf.js > protobuf/protobuf_bundle.json"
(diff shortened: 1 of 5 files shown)


delventhalz:
I've updated the PR with a README and some metadata. Re-reviews would be appreciated.

sdk/javascript/package.json
```diff
@@ -1,7 +1,10 @@
 {
   "name": "sawtooth-sdk",
   "version": "0.0.0",
-  "description": "An SDK for interacting with the Sawtooth Lake distributed ledger",
+  "description": "An SDK for interacting with the Hyperledger Sawtooth distributed ledger.",
+  "keywords": ["hyperledger", "blockchain", "signing", "crypto", "protobuf"],
+  "homepage": "http://intelledger.github.io/",
```
peterschwarz: Technically, that's the documentation site.  The home page is https://www.hyperledger.org/projects/sawtooth, I think

PR #684, opened 2017-07-01
Add encoding batches and transactions to Python SDK

delventhalz:
Expand Python Client SDK to include some features in JS SDK. Includes ~~signing module for creating keys and signatures, and~~ an encoding module for creating and encoding Batches and Transactions.

### DIFF
--- a/docs/source/_templates/sdk_submit_tutorial.rst
+++ b/docs/source/_templates/sdk_submit_tutorial.rst
@@ -27,7 +27,15 @@ key, and it is fairly simple to generate this using the SDK's *signer* module.
     const privateKey = signer.makePrivateKey()
 
 {% else %}
+
 {# Python 3 code should be the default #}
+
+.. code-block::  python
+
+    from sawtooth_signing.secp256k1_signer import generate_privkey
+
+    private_key = privkey = generate_privkey(privkey_format='bytes')
+
 
 {% endif %}
 
@@ -93,6 +101,23 @@ these Transactions (see below).
 
 {% else %}
 
+.. code-block::  python
+
+    from sawtooth_sdk.client.encoding import TransactionEncoder
+
+    encoder = TransactionEncoder(
+        private_key,
+        # We don't want a batcher pubkey or dependencies for our example,
+        # but this is what setting them might look like:
+        # batcherPubkey='02d260a46457a064733153e09840c322bee1dff34445d7d49e19e60abd18fd0758',
+        # dependencies=['540a6803971d1880ec73a96cb97815a95d374cbad5d865925e5aa0432fcf1931539afe10310c122c5eaae15df61236079abbf4f258889359c4d175516934484a'],
+        payload_encoder=cbor.dumps,
+        family_name='intkey',
+        family_version='1.0',
+        inputs=['1cf1266e282c41be5e4254d8820772c5518a2c5a8c0c7f7eda19594a7eb539453e1ed7'],
+        outputs=['1cf1266e282c41be5e4254d8820772c5518a2c5a8c0c7f7eda19594a7eb539453e1ed7']
+        payload_encoding='application/cbor')
+
 {% endif %}
 
 .. note::
@@ -137,6 +162,18 @@ Optionally, you may pass in header properties in order to override any defaults 
     })
 
 {% else %}
+
+.. code-block::  python
+
+    txn = encoder.create(
+        payload,
+        inputs=['1cf12663ae9d398142a7d84c49b73ba2f667c8d377ceb7832db69b1a416133562ea496'],
+        outputs=['1cf12663ae9d398142a7d84c49b73ba2f667c8d377ceb7832db69b1a416133562ea496'])
+
+    txn2 = encoder.create({
+        'Verb': 'inc',
+        'Name': 'foo',
+        'Value': 1})
 
 {% endif %}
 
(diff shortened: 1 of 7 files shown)


sdk/python/sawtooth_sdk/client/encoding.py
```diff
@@ -0,0 +1,302 @@
...
+
+    Note:
+        Every optional argument can be set or changed later as an attribute.
+    """
+    def __init__(self,
+                 private_key,
+                 payload_encoder=lambda x: x,
+                 batcher_pubkey=None,
+                 dependencies=None,
+                 family_name=None,
```
vaporos: Why not make family_name, family_version, payload_encoding required and omit them from create?

delventhalz: The TransactionEncoder tries to be unopinionated. Most people will probably set those three defaults, and only those three, and not override them. But it is entirely possible that someone might create an encoder and use it with multiple families or what have you.

delventhalz:
@vaporos How likely is it we would change signing algorithms? That would break previous blocks wouldn't it? Also, you could always add an additional optional parameter to `sign` that would take an algorithm name. That could be added anytime, and would be backwards compatible.

I'm a little reticent to add an extraneous extra step, forcing everyone to instantiate a `Signer` object, just in the off chance it might matter. If alternate algorithms are a part of the roadmap, then it makes more sense to me.

dcmiddle:
@delventhalz I'm entertaining the idea of changing to eddsa in a future sawtooth version.  So it wouldn't be changing crypto during runtime.
I'm not sure the trade-off of having the python-sdk use core packages, but I do see some simplicity in just using the same signing code so there's no additional maintenance.

delventhalz:
@dcmiddle So it sounds I should follow @vaporos's prescription and replace the current static `sign` function with a `Signer` parent class, an instantiable `Secp256k1Signer` child class with a `sign` method, and a `get_default_signer` function which will instantiate it?

As for whether or not to integrate with `sawtooth_signing`, I wrote this as a separate module based on the Slack conversation the three of us had about a week ago. The importance of keeping the SDK and validator functionality separate was emphasized. My feeling is that the use cases are actually somewhat different (no need to go through the rigamarole with `Signer` classes internally for the validator for example), and that the repeated code is actually pretty minimal, since they are mostly just calls to third party libraries.

docs/source/_templates/sdk_submit_tutorial.rst
```diff
@@ -144,6 +164,18 @@ Optionally, you may pass in header properties in order to override any defaults
 
 {% else %}
```
ojalatodd: Question (more than a comment) regarding the text above this code: what if an encoder function is **not** set? Would this cause an error, or is there another way to get the transaction payload encoded? If an error will be generated if an encoder function is not set, maybe we should state that it is a requirement to set an encoder method...

delventhalz: The `payload_encoder` is just a client-defined convenience function that takes in arbitrary data and spits out _bytes_. You don't have to use one, but if you don't you must make sure the payload is already encoded as _bytes_ when you pass it to `TransactionEncoder.create`.

In the example in the docs, if you tried to call
```
encoder.create({
        'Verb': 'inc',
        'Name': 'foo',
        'Value': 1})
```
And there were _no payload encoder_ it would throw an error. However the following would work fine:
```
encoder.create(cbor.dumps({
        'Verb': 'inc',
        'Name': 'foo',
        'Value': 1}))
```

sdk/python/sawtooth_sdk/client/encoding.py
```diff
@@ -0,0 +1,302 @@
...
+                       batcher_pubkey=None,
+                       dependencies=None,
+                       family_name=None,
+                       family_version=None,
+                       inputs=None,
+                       nonce=None,
+                       outputs=None,
+                       payload_encoding=None):
+        """Convenience method which creates a single Transaction from a payload
+        and supplied or default TransactionHeadewr values, before wrapping it
```
ojalatodd: Typo:  TransactionHeadewr

sdk/python/sawtooth_sdk/client/encoding.py
```diff
@@ -0,0 +1,302 @@
...
+        self._private_key = private_key
+        self._public_key = signing.get_public_key(private_key)
+
+    def create(self, transactions):
+        """Creates and signs a new Batch message with one or more Transactions.
+        Transactions can be input as Transaction instances or as a serialized
+        TransactionList.
+
+        Args:
+            transactions (bytes or Transaction or list of Transaction):
```
ojalatodd: Change singular to plural here?
s/Transaction/Transactions/

delventhalz: The stuff in parentheses precisely outline the expected type(s) of the arguments. In this case a `list` of `Transaction` instances. You do not pluralize the class name.

delventhalz:
PR has been updated to remove the client signing module and replace its usage with the existing `sawtooth_signing` module. Support for `bytes` has been added to the signing module, but otherwise it is unchanged. Refactoring this module to be more client friendly will be left to someone else

sdk/python/sawtooth_sdk/client/encoding.py
ojalatodd: Roger.

PR #689, opened 2017-07-06
Chain Controller and BlockStore updates

cianx:
1) Update Journal Test BlockTreeManager to not use the Publisher for BlockGeneration
2) Move Block store operation that require multiple reads inside a single LMDB transaction to ensure consistancy.
3) Add unit tests for the BlockStore
4) Add tests to the BlockValidator to reject duplicate batches and transactions.
5) Add early detection of Chain Head update to the BlockValidator.


validator/tests/test_block_store/tests.py
```diff
@@ -24,6 +26,178 @@
...
+
+LOGGER = logging.getLogger(__name__)
+
+
+class BlockStoreTest(unittest.TestCase):
+
+    def setUp(self):
+        self.block_tree_manager = BlockTreeManager()
+
+    def test_chain_head(self):
```
agunde406: Please add Doc strings to all of theses unit tests.

validator/sawtooth_validator/journal/chain_commit_state.py
```diff
@@ -0,0 +1,107 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+
+
+class _CommitCache(object):
+    """Tracks the committment status of a set of identifiers in these
+    identifiers are either explicitly committed, explicity uncommitted, if
+    they fall in to neither of these cases then the fallback is to look in
+    the BlockStore to see if they are there. Explicit committed ids
```
agunde406: Suggest changing first sentence to the following:

Tracks the commit status of a set of identifiers and these
identifiers are either explicitly committed, or explicitly uncommitted. If
they fall in to neither of these cases then the fallback is to look in
the BlockStore to see if they are there.

validator/sawtooth_validator/journal/chain.py
```diff
@@ -40,6 +47,23 @@ class BlockValidationAborted(Exception):
...
 
 
+def look_ahead(iterable):
+    """Pass through all values from the given iterable, augmented by the
+    information if there are more values to come after the current one
+    (True), or if it is the last value (False).
+    """
+    # Get an iterator and pull the first value.
+    it = iter(iterable)
+    last = next(it)
```
peterschwarz: This line will throw a `StopIteration` if `iterable` is empty.

validator/sawtooth_validator/journal/chain_commit_state.py
```diff
@@ -0,0 +1,108 @@
+# Copyright 2016 Intel Corporation
```
peterschwarz: 2017

validator/sawtooth_validator/journal/chain_commit_state.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def __contains__(self, identifier):
+        if identifier in self._committed:
+            return True
+        elif identifier in self._uncommitted:
+            return False
+        return self.block_store_check(identifier)
+
+
+class ChainCommitState(object):
```
peterschwarz: Does this duplicate the functionality of the BatchTracker?  Are they similar enough to serve the each use?

validator/sawtooth_validator/journal/chain_commit_state.py
```diff
@@ -0,0 +1,108 @@
...
+            self._transaction_commit_state.remove(txn.header_signature)
+
+    def has_batch(self, batch_id):
+        return batch_id in self._batch_commit_state
+
+    def has_transaction(self, txn_id):
+        return txn_id in self._transaction_commit_state
+
+
+class TransactionCommitState(_CommitCache):
```
peterschwarz: Why is this one a `_CommitCache` and the other has a `_CommitCache`.  Seems like the both should be composition or this one should be called a `Cache`

validator/sawtooth_validator/journal/chain_commit_state.py
cianx: The BatchTrack and the ChainCommitState provide different functionality. The BatchTracker appears to be attepting to provide notifications when a batch appears in the BlockStore. The ChainCommitState is focused on simulating an alternate version of the the BlockStore.

validator/sawtooth_validator/journal/chain.py
cianx: That is correct behavior, it exits the generator with no iterations performed.

PR #690, opened 2017-07-06
Update wait timer expiration in SGX enclave to match simulator.

jjason:


jjason:
Realized when chatting with Peter, James, and Boyd that the wait timer expiration values in the simulator and SGX versions are different.

PR #691, opened 2017-07-06
Implement transaction replay with new state in parallel scheduler

boydjohnson:
When a batch fails, txns that depend implicitly due to inputs and outputs on txns in that batch must be replayed with new state.

Several yaml based tests are introduced that test this capability.


nick-drozd:
I strongly suggest breaking out the predecessor tree tests and the yaml tests into separate files, as in this commit: https://github.com/nick-drozd/sawtooth-core/commit/bc5572cafbac3e9356163ee66d506a40613b6220. It's not how obvious how the tests should be further organized (I have some opinions, but not strong ones), but separate files is a good start in any case.

boydjohnson:
@nick-drozd I will look into adding that commit to break the test file up.

validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -488,19 +489,88 @@ def get_batch_execution_result(self, batch_signature):
...
+            possible_successor)
+        in_different_batch = not self._is_in_same_batch(txn_id,
+                                                        possible_successor)
+        has_not_been_seen = possible_successor not in already_seen
+
+        return is_successor and in_different_batch and has_not_been_seen
+
+    def _remove_subsequent_result_because_of_batch_failure(self, sig):
+        """Remove transactions from scheduled and txn_results for
+         successors of txns in a batch that failed for another transaction.
```
agunde406: I suggest rewording this. It is unclear what you mean by "for another transaction". I also suggest adding that they will be rescheduled instead of just "This gives the replay ability."

validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -544,49 +617,78 @@ def _txn_is_in_valid_batch(self, txn_id):
...
+    def _predecessor_not_in_chain(self,
+                                  prior_txn_id,
+                                  chain):
         """
 
         Args:
             prior_txn_id (str): The predecessor's txn header_signature.
-            predecessors (list): The predecessors of the txn that the
-                prior_txn_id is a predecessor of.
+            predecessors (list): The txn_ids that are already in the graph.
```
agunde406: Should be chain. I am not sure what chain means here.

validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -544,49 +617,78 @@ def _txn_is_in_valid_batch(self, txn_id):
...
 
         Returns:
             (bool): The prior_txn_id is not a predecessor of a predecessor.
```
agunde406: This should be updated to match the purpose of the function.

validator/tests/test_scheduler/tests.py
```diff
@@ -162,10 +213,158 @@ def test_serial_batch_fails_valid_txn(self):
             context_manager=context_manager,
             name='batch_fails_valid_txn.yaml')
 
+    def test_parallel_complex_batches_multiple_failures(self):
+        """Tests the parallel scheduler against the
+        test_scheduler/data/complex_batches_multiple_failures.yaml file.
+
+        Notes:
+            This yaml file has txns that depend on txns in several batches
+            and several failed batches.
```
agunde406: Please add some more description of what this test is doing (along with the subsequent tests)

validator/tests/test_scheduler/tests.py
```diff
@@ -162,10 +213,158 @@ def test_serial_batch_fails_valid_txn(self):
...
+            context_manager=context_manager,
+            name='enclosing_writer_fails.yaml')
+
+    def test_parallel_heterogeneous_workload(self):
+        """Tests the parallel scheduler against the
+        test_scheduler/data/heterogeneous_workload.yaml file.
+
+        Notes:
+            This yaml file has namespaced addresses and several failed batches
+        """
```
agunde406: Please add some more description of what this test is doing (along with the subsequent tests)

boydjohnson:
@nick-drozd @agunde406 I updated the PR based on your comments.

validator/tests/test_scheduler/test_schedulers_with_yaml.py
```diff
@@ -0,0 +1,472 @@
...
+        self.assert_batch_validity(
+            defined_batch_results_dict,
+            batch_results)
+
+        for t_id in txns_to_assert_state:
+            state_and_txn_context = txns_to_assert_state.get(t_id)
+            if state_and_txn_context is not None:
+                txn_context, state_found, state_assert = state_and_txn_context
+
+                self.assertEquals(state_found, state_assert,
```
nick-drozd: `assertEquals` is deprecated in favor of `assertEqual` (https://docs.python.org/3.3/library/unittest.html#deprecated-aliases)

There are a few instances of this. It's not specific to this PR, but they might as well be cleaned up now.

PR #692, opened 2017-07-06
Add permissioned EVM support and overhaul design

aludvik:
In anticipation of this transaction family becoming more complex, I have overhauled the design to be more resilient to change and easier to follow.

Using this new design, I added support for EVM account permissions. For an explanation of how permissions work, do `seth permissions -h`.


aludvik:
Removed some stray comments

aludvik:
Rebased post name change

bin/run_tests
```diff
@@ -310,7 +310,7 @@ test_python_sdk() {
 
 test_burrow_evm() {
   run_docker_test ./families/burrow_evm/tests/unit_burrow_evm.yaml
-  run_docker_test burrow_evm_intkey_test
+  # run_docker_test burrow_evm_intkey_test
```
peterschwarz: This commit seems unnecessary in the context of this PR, since they are re-enabled later.

peterschwarz:
My comments were marked as out of date, but please take a look at them and reply.

families/burrow_evm/src/protos/evm.proto
```diff
@@ -32,17 +34,73 @@ message EvmStateAccount {
...
+      CREATE_EXTERNAL_ACCOUNT = 1;
+      CREATE_CONTRACT_ACCOUNT = 2;
+      MESSAGE_CALL = 3;
+    }
+    TransactionType transaction_type = 1;
+
+    // To eliminate the need for two deserialization steps, all types of
+    // transactions are included as fields, but only the field indicated by the
+    // transaction type should be set.
+    CreateExternalAccountTxn create_external_account = 2;
```
aludvik: I looked into using the oneof feature, but it creates a wrapper and relies on type casting/switching to work in generated code which I felt was a bad design. I chose to follow this technique instead: https://developers.google.com/protocol-buffers/docs/techniques#union The required/optional keywords are omitted in my code because they are deprecated in version 3.

peterschwarz: Ah, ok, fair enough.

I'm amused by this quote, though: 

> All that said, the reason that this functionality is not included in the Protocol Buffer library is because we have never had a use for it inside Google.

families/burrow_evm/src/sawtooth_burrow_evm/cli/show.go
```diff
@@ -95,35 +60,50 @@ func (s *Show) Run() (err error) {
...
+	if len(acct.GetAddress()) == 0 {
 		fmt.Println("Account does not exist")
 		return
 	}
 
+	addr := hex.EncodeToString(acct.GetAddress())
+	code := hex.EncodeToString(acct.GetCode())
+
 	fmt.Printf(`
 Address: %v
```
dplumb94: This address is being printed out with trailing 0's (which makes it a 35 byte addr). This gives us a malformed address if it is copied into another seth command.

aludvik:
Removed trailing 0s per @dplumb94's comment

aludvik:
Added a commit to treat EVM errors as an InvalidTransactionError. For example, if the EVM runs out of gas, that is a problem with the transaction, not the EVM.

PR #693, opened 2017-07-07
Ensure that PoET enclave always has key state restored properly.

jjason:
The PoET block publisher has been updated so that the PoET key pair is always restored from the key state store.

### DIFF
--- a/consensus/poet/core/sawtooth_poet/poet_consensus/poet_block_publisher.py
+++ b/consensus/poet/core/sawtooth_poet/poet_consensus/poet_block_publisher.py
@@ -261,9 +261,9 @@ class PoetBlockPublisher(BlockPublisherInterface):
             pass
 
         # If we don't have a validator registry entry, then check the active
-        # key.  If we don't have one, then we need to sign up.
-        # If we do have one, then our validator registry entry has not
-        # percolated through the system, so nothing to to but wait.
+        # key.  If we don't have one, then we need to sign up.  If we do have
+        # one, then our validator registry entry has not percolated through the
+        # system, so nothing to to but wait.
         active_poet_public_key = self._poet_key_state_store.active_key
         if validator_info is None:
             if active_poet_public_key is None:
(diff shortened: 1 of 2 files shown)


jjason:
This PR fixes a problem when a validator is started up and it already has an entry in the validator registry and its local PoET key state store has the corresponding PoET public key marked as active.

jjason:
Note - I still need to think a little more about how to handle when everything goes completely off of the rails.

consensus/poet/core/sawtooth_poet/poet_consensus/poet_block_publisher.py
```diff
@@ -278,34 +278,34 @@ def initialize_block(self, block_header):
...
-            # the active PoET public key.
-            poet_key_state = \
-                self._poet_key_state_store[
-                    validator_info.signup_info.poet_public_key]
-
-            active_poet_public_key = \
-                SignupInfo.unseal_signup_data(
-                    poet_enclave_module=poet_enclave_module,
-                    sealed_signup_data=poet_key_state.sealed_signup_data)
+            active_poet_public_key = validator_info.signup_info.poet_public_key
```
peterschwarz: Do we need to set this value, since we are going to set it again anyway?

jjason: Yes.  We need to change it because on line 291, the key is used to retrieve the key state.  The assignment on line 293 is the one that is not necessarily required, but I put it there so that I can do the assert just to make sure that I didn't screw something up in development.

jjason:
Updated in hopes of fixing the problem Peter sees when blowing away the data directory.  In this update, if the PoET block publisher sees that the PoET public key in the validator registry does not result in a valid key state, it will sign up again.

jjason:
Crap....I broke a unit test on Jenkins....I need to see what is different locally.

PR #694, opened 2017-07-07
fixing a bug in the client.py file of supplychain. The bug was a simp…

benoitrazet:
…le conversion missing from float to int for timestamp, which triggered a typechecking error from protobuf

### DIFF
--- a/families/supplychain/python/sawtooth_supplychain/client.py
+++ b/families/supplychain/python/sawtooth_supplychain/client.py
@@ -184,9 +184,8 @@ class SupplyChainClient(object):
         except BaseException:
             return None
 
-    def application_accept(self, record_identifier,
-                           applicant, application_type,
-                           timestamp=None, wait=None):
+    def application_accept(self, record_identifier, applicant_public_key,
+                           application_type, timestamp=None, wait=None):
         outputs = [Addressing.application_address(record_identifier),
                    Addressing.record_address(record_identifier)]
         inputs = outputs + [Addressing.agent_address(self.public_key),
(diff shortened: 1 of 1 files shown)


families/supplychain/python/sawtooth_supplychain/client.py
```diff
@@ -195,9 +194,9 @@ def application_accept(self, record_identifier,
             SupplyChainPayload.APPLICATION_ACCEPT,
             ApplicationAcceptPayload(
                 record_identifier=record_identifier,
-                applicant=applicant,
+                applicant=applicant_public_key,
                 type=application_type,
-                timestamp=timestamp or time.time()),
+                timestamp=timestamp or int(time.time())),
```
peterschwarz: There is no discussion the spec as to what granularity this timestamp should be.  Is it millis or seconds?  If the former you should probably multiply this by 1000 before calling `int`

benoitrazet: Based on similar type conversions present in the code, it looks like currently the granularity is seconds. Also, since the timestamp is an optional argument to the function, any design choice for the granularity (milliseconds or seconds) would have to be consistent across usages or use cases.

benoitrazet: I have no preference, it should be easy to multiply by 1000 everywhere it is required.

peterschwarz: @cianx should update the spec with the details about what this should be.  I think seconds is fine, if that's what is in the rest.

PR #696, opened 2017-07-07
STL-479 IAS Client unit tests

aludvik:


jjason:
Instead of using real http.server and creating your own mock server, you might want to investigate using `unittest.mock.patch`.  @peterschwarz and @cianx both are experts on how to patch classes in unit tests.

aludvik:
I looked into `unittest.mock.path` and it probably would have been easier to use than this. My understanding is that it would require patching `requests`. Since this module is just a thin wrapper for requests, I like using the full HTTP server here as it accomplishes the same goal but if `requests` interface ever changes, this test will catch the change.

utility/ias_client/tests/unit/test_ias_client.py
```diff
@@ -0,0 +1,89 @@
...
+    @classmethod
+    def setUpClass(cls):
+        cls.mock_server = mock_ias_server.create(URL)
+        cls.mock_server.start("up")
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.mock_server.stop()
+
+    def test_up(self):
```
agunde406: Please add Docstrings for the tests.

aludvik:
Added docstrings to tests

PR #700, opened 2017-07-10
Isolate open futures per executor thread

peterschwarz:
By Isolating the dictionary of open futures to a given executor thread,
an issue where the transaction is being processed by one executor thread
(as part of one schedule), as well as by another thread (due to it being
part of a claimed block on another validator).  Prior to this fix, the
dictionary of open futures was shared across threads, but the sharing
would result in an uncaught exception in the above scenario.


agunde406:
Tested locally with two TP. When a one crashes, the other one can still recover.

PR #703, opened 2017-07-11
Add sgx poet builds to build_debs

rbuysse:


bin/build_debs
```diff
@@ -103,6 +103,23 @@ main() {
     # Group 5: Depends on 1-4
     build_pkg "consensus/poet/cli" $pypath
 
+    # Group 6: PoET
```
aludvik: Does this package depend on any of the other packages being built?

PR #709, opened 2017-07-12
Fix typos and inaccuracies in Supply Chain documentation

delventhalz:
This PR fixes spelling/grammar errors and misaligned code blocks in the Supply Chain Transaction Family document. It does not alter the wording or flow of this document.

It also addresses some inaccurate and unclear docstrings in the Supply Chain proto-files. These have been reworded somewhat more heavily than the Transaction Family document.

This PR does not alter any functionality.

### DIFF
--- a/docs/source/examples/supplychain/supplychain_transaction_family.rst
+++ b/docs/source/examples/supplychain/supplychain_transaction_family.rst
@@ -1,39 +1,39 @@
-***************************************
+*******************************
 Supply Chain Transaction Family
-***************************************
+*******************************
 
 Overview
 =========
 
-The Supplychain Transaction Processor (SCTP) provides a mechanism for
+The Supply Chain Transaction Processor (SCTP) provides a mechanism for
 storing records of asset ownership. The SCTP tracks records, which correlate
-to a real world item. Every Record has an owner and a custodaian, which can
+to a real world item. Every Record has an owner and a custodian, which can
 be the same Agent. The owner is the Agent the legally owns the good. The
-custodian is the Agent that is in possesion of the good, which may be a
+custodian is the Agent that is in possession of the good, which may be a
 warehousing service, transportation service, etc. Agents are actors in the
-system that are can manupulate the Records by updating them or transfering
+system that can manipulate the Records by updating them or transferring
 their role to other Agents. Agents are identified by their public key and
 authenticated by the signature on the transactions they submit. Transfer of
-roles (Owner, Custodian) betweeen Agents is done via an Application process,
-Where the Agent wishing to assume the role Applies for the new role and the
-Agent in the role currently can either accept or reject that applicaiton.
+roles (Owner, Custodian) between Agents is done via an Application process,
+where the Agent wishing to assume the role applies for the new role and the
+Agent in the role currently can either accept or reject that Application.
 
-This is an proof of concept Transaction Processor that is actively being
-developed. As such it has not been thoughly tested and there will be
-protentially breaking changes made to it.
+This is a proof of concept Transaction Processor that is actively being
+developed. As such it has not been thoroughly tested and there will potentially
+be breaking changes made to it.
 
 State
 =====
 
 This section describes in detail how SCTP objects are stored and addressed in
-the global state. All objects in state are stored as encoded protobufs. Each
-type of object below is stored in containers at nodes in the Merkle Trie. The
+the global state. All objects in state are stored as encoded Protobufs. Each
+type of object below is stored in containers at nodes in the Merkle trie. The
 address of the Merkle nodes is detailed below in the addressing section. The
 entry at the nodes is a container object that allows for multiple items to be
 stored at that node if there is an address collision. All items stored in state
-are include their natural key. When an address collision occurs the item stored
-at the address are searched to find the item with the natural key that matches
-the key being searched for.
+include their natural key. When an address collision occurs the items stored at
+the address are searched to find the item with the natural key that matches the
+key being searched for.
 
 SCTP stores the following data objects:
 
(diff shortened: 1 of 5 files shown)


docs/source/examples/supplychain/supplychain_transaction_family.rst
```diff
@@ -43,61 +43,121 @@ Agent
 Agents represent owners or custodian of goods in the system. All agents must
 be registered in the system before they can interact with records.
 
-.. literalinclude:: ../../../../families/supplychain/protos/agent.proto
```
jsmitchell: Hmm. I prefer the include model. I understand the issue with the offsets. Seems to me that the least evil option is to do the include without the lines stripped.

delventhalz: You mean embed the entirety of every proto-file including the Intel header? You could probably cut out the header without worrying about the offsets too much since that probably isn't changing.

BUT, the Agent Transactions sections would have to be completely rewritten somehow, as every example is a piece the same file. I did not want to include any rewrites in the scope of this PR, so this is one of the reasons I decided against `literalinclude`. I suppose I could front-load the section with the entirety of the `payload.proto` file, and delete the code examples from all the subsections. That seems not great though.

jsmitchell: blargh, everything is terrible

delventhalz: Indeed

cianx: It would seem more obvious when reviewing the documents generated output to spot an incorrect offset in an include than a line in the explicit protobuf definition that is of sync with the protobuf definition.

docs/source/examples/supplychain/supplychain_transaction_family.rst
```diff
@@ -1,39 +1,39 @@
...
+be breaking changes made to it.
 
 State
 =====
 
 This section describes in detail how SCTP objects are stored and addressed in
-the global state. All objects in state are stored as encoded protobufs. Each
-type of object below is stored in containers at nodes in the Merkle Trie. The
+the global state. All objects in state are stored as encoded Protobufs. Each
+type of object below is stored in containers at nodes in the Merkle trie. The
```
cianx: Isn't "Merkle Trie" a proper noun?

docs/source/examples/supplychain/supplychain_transaction_family.rst
```diff
@@ -1,39 +1,39 @@
...
+roles (Owner, Custodian) between Agents is done via an Application process,
+where the Agent wishing to assume the role applies for the new role and the
+Agent in the role currently can either accept or reject that Application.
 
-This is an proof of concept Transaction Processor that is actively being
-developed. As such it has not been thoughly tested and there will be
-protentially breaking changes made to it.
+This is a proof of concept Transaction Processor that is actively being
+developed. As such it has not been thoroughly tested and there will potentially
+be breaking changes made to it.
```
cianx: Suprised @peterschwarz didn't request this paragraph be removed.

peterschwarz: Good point.  Remove it.

peterschwarz: In a broader sense, we could use the version of the TP to denote this.  By already marking them as 1.0, but saying they'll change is a bit inconsistent.

docs/source/examples/supplychain/supplychain_transaction_family.rst
delventhalz: "Trie" is just German for "tree" I believe. Most sources online use the spelling _"Merkle tree"_, but the few I found that use trie, capitalize it _"Merkle trie"_. I did not find any pages where it was spelled "Merkle Trie".

delventhalz:
@cianx Probably true if anyone checks. In this case there were offset errors which have gone unnoticed for at least a few weeks. It's also only obvious if you build the docs. You cannot tell from the source code.

I'm also not sure including the exact source code is important. People can always look that up if they need it. A doc should have enough example code to explain how to use a component. Minor inaccuracies may not impair that use case. And regardless of whether it is in the example code, or elsewhere in the doc, you will always have the headache of making sure engineers update documentation when they change functionality.

docs/source/examples/supplychain/supplychain_transaction_family.rst
delventhalz: Probably true if anyone checks. In this case there were offset errors which have gone unnoticed for at least a few weeks. It's also only obvious if you build the docs. You cannot tell from the source code.

I'm also not sure including the exact source code is important. People can always look that up if they need it. A doc should have enough example code to explain how to use a component. Minor inaccuracies may not impair that use case. And regardless of whether it is in the example code, or elsewhere in the doc, you will always have the headache of making sure engineers update documentation when they change functionality.

docs/source/examples/supplychain/supplychain_transaction_family.rst
delventhalz: I am keeping the scope for this document strictly to correctly spelling and grammar errors. If I start rewriting it, that quickly becomes a JIRA Task.

docs/source/examples/supplychain/supplychain_transaction_family.rst
nick-drozd: > Note: The name comes from reTRIEval and is pronounced, "tree." See the historical note.
>There are many who, in an attempt to verbally distinguish a trie from the more general tree and in contrast to typical English pronunciation (of which calorie, eerie, and reverie are examples, but die, lie, and pie are not), pronounce the name "try".
>...
>Historical Note
>On 31 July 2008 Edward Fredkin wrote
>As defined by me, nearly 50 years ago, it is properly pronounced "tree" as in the word "retrieval". At least that was my intent when I gave it the name "Trie". The idea behind the name was to combine reference to both the structure (a tree structure) and a major purpose (data storage and retrieval).

http://www.webcitation.org/5pqUULy24

delventhalz: This was unexpected. Spelling still stands though.

cianx: If you want the sentence to refer to the algorithm and follow the algorithm naming capitailization, can you change the sentence to read "in a Merkle trie". Otherwise the intention of this sentence is to refer to the actual component implementation in Sawtooth not the algorithm and therefore should follow standard proper naming conventions.

docs/source/examples/supplychain/supplychain_transaction_family.rst
cianx: Bad document review process is not a reason to create more work. 
There are not guarentees that the differences from the souce to the document will be minor. 
If we want to people to look the code up directly then the documents should just have a link to the code directly and not include a copy of it.

PR #719, opened 2017-07-19
Handle dependencies in the serial scheduler

boydjohnson:


boydjohnson:
The changes I made to the YamlSchedulerTester have it calculate the transaction validity from the dependencies, and then set this information on the scheduler, bypassing the scheduler's ability to determine dependencies.

PR #720, opened 2017-07-19
STL-461 Java SDK transaction signing

ghadishayban:
*Includes*

- wif reading
- public key derivation from the private key
- transaction signing using the 'compact signature' scheme

*Does not cover*
- payload verification

This has been verified by signing a payload in Java, then verifying it in Python.

I had to exclude a transitive scrypt lib inclusion, as it had a signed but invalid jar.

### DIFF
--- a/sdk/java/pom.xml
+++ b/sdk/java/pom.xml
@@ -78,5 +78,16 @@
             <artifactId>protobuf-java</artifactId>
             <version>3.2.0</version>
         </dependency>
+        <dependency>
+            <groupId>org.bitcoinj</groupId>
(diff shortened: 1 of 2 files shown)


ghadishayban:
The bitcoinj library is Apache licensed, and has some MIT dependencies.

peterschwarz:
The current issue with the build looks like it has to do with the source and target versions specified in the maven `pom.xml`.  Currently, the sawtooth java sdk and examples are targeting 1.7, but bitcoinj targets 1.8.

ghadishayban:
This is passing and commits properly squashed; ready4review.

PR #730, opened 2017-07-31
Fix monotonic counter use to bind to signup data

dcmiddle:
Per PoET spec, the monotonic counter must be part of the sealed signup data.
It ensures that only a single enclave can produce wait timers for a validator
registration (embodied by a signup data).

Trying to track the MCID and other items in enclave state does not preclude
certain errors. Instead it is best to pass the signup data into the two calls
which rely on the poet keys and hardware counter for security
(i.e. ecall_CreateWaitTimer and ecall_createWaitCertificate).

Whenever a wait timer is created we increment the counter.

Whenever a wait certificate is requested we check if the counter in the wait
timer matches the value currently in the monotonic counter.

Whenever a wait certificate is created we increment the counter.

Changes:
Add the counter ID to the signup data structure.
Remove reliance on poet state object.
Add counter tracking behavior to wait timer and certificate functions.

Related:
STL-611 opened to add cleanup functionality to the validator related to MCID release.
STL-641 opened to add graceful handling during block creation in the event of corrupted signup data.
STL-642 opened to add HW counter simulation to PoET Simulator and update unit tests accordingly.


consensus/poet/sgx/sawtooth_poet_sgx/libpoet_enclave/poet_enclave.cpp
```diff
@@ -392,6 +394,13 @@ poet_err_t ecall_CreateSignupData(
             ret,
             "Failed to create PoET public/private key pair");
 
+        // Create the monotonic counter bound to the keypair
+        ret =
+            sgx_create_monotonic_counter(
+                &validatorSignupData.counterId,
+                &initialCounterValue);
+        sp::ThrowSgxError(ret, "Failed to create monotonic counter.");
```
cianx: I don't see where counters are destroyed,  this code will exaust the supply of system monotonic counters once the number of ecall_CreateSignupData  is greater then the number of counters.

dcmiddle: That's a good point. I have a separate story in Jira to deal with that, but I could roll it into this PR.
Currently counters are destroyed as a consequence of ecall_Terminate which needs to be fixed.
The approach I have in mind is the validator makes a decision when signup data is no longer useful to keep. It should call an enclave method that will release the counter, then the validator needs to clean up its local database of signups. That's a little dicey due to fork handling.

dcmiddle: After tugging at the thread I have unraveled a little further than I originally intended to. I'm now going to get rid of poet_State entirely. I'll update the call pattern such that calls to create the Timer and the WaitCert both need the signupdata. SignupData will effectively become the enclave state.

consensus/poet/simulator/sawtooth_poet_simulator/poet_enclave_simulator/poet_enclave_simulator.py
```diff
@@ -343,78 +339,72 @@ def create_wait_timer(cls,
...
             # Several criteria need to be met before we can create a wait
             # certificate:
-            # 1. We have an active timer
-            # 2. The caller's wait timer is the active wait timer.  We are not
-            #    going to rely the objects, being the same, but will compute
-            #    a signature over the object and verify that the signatures
-            #    are the same.
-            # 3. The active timer has expired
-            # 4. The active timer has not timed out
+            # 1. This signup data was used to sign this timer.
```
cianx: Think you still have to have an active timer.

consensus/poet/simulator/sawtooth_poet_simulator/poet_enclave_simulator/poet_enclave_simulator.py
```diff
@@ -343,78 +339,72 @@ def create_wait_timer(cls,
...
+            #    i.e. the key sealed / unsealed by the TEE signed this
+            #    wait timer.
+            # 2. This timer has expired
+            # 3. This timer has not timed out
+            #
+            # In a TEE implementation we would check HW counter agreement.
+            # We can't usefully simulate a HW counter though.
+            # i.e. wait_timer.counter_value == signup_data.counter.value
```
cianx: That does not look right, the counter value is not stored on the signup data.

dcmiddle: its very pseudo pseudo-code. I'm trying to say briefly that your reference the counter from the signup data and reference the value from that counter. It's a hint for the future should we decide to simulate the counter. Right now I don't think we can do that without miring in simulation challenges (e.g. counters survive reboots and file system changes).

consensus/poet/core/tests/test_consensus/test_wait_certificate.py
```diff
@@ -164,49 +173,60 @@ def test_create_after_wait_timer_timed_out(self):
         with self.assertRaises(ValueError):
             WaitCertificate.create_wait_certificate(
                 poet_enclave_module=self.poet_enclave_module,
+                sealed_signup_data=signup_info.sealed_signup_data,
                 wait_timer=wt,
                 block_hash="Reader's Digest")
 
-    def test_create_with_wrong_wait_timer(self):
+    def test_create_with_wrong_signup_data(self):
```
dcmiddle: The thing is I took MC out of the simulator. While I think these tests are meant to be applied to both, in practice they seem to only be applied to simulator.

consensus/poet/simulator/sawtooth_poet_simulator/poet_enclave_simulator/poet_enclave_simulator.py
dcmiddle: The previous code used a private variable to track an "active wait timer". I took that out (like removing poet_state). The updated comment was meant to remove the allusion to that old private variable. This function still requires a wait timer to be passed in. The concept of whether or not it is "active" is spread across checks like signature validity and time.

consensus/poet/core/tests/test_consensus/test_wait_certificate.py
cianx: I would have thought we would want the simulator to move closer to mirroring the functionality of the enclave not further away. Regardless of whether an MC simulation could survive a reboot, the interesting runtime features of being able to invalidate wait timers is a trival implementation and llows for those expect features to be tested on both implementations. 

These tests should be run on both the enclave and the simulator.

consensus/poet/simulator/sawtooth_poet_simulator/poet_enclave_simulator/poet_enclave_simulator.py
cianx: Don't believe you need a counter that survives reboots and fs changes for it to be useful in simulation. It needs to provide the ability to invalidate wait timers. A counter stored in the simulator generated sign up data would be sufficient, which would probably satisified the previous conditions.

dcmiddle: I created STL-642 to add simulated counters to the poet simulator and update the core tests to apply equally to both simulator and sgx implementations.

PR #738, opened 2017-08-04
Add PyFormance InfluxDB Reporter

dplumb94:
Currently the statistics reporter in the PyFormance library is broken for Python 3. I fixed it by swapping an `iteritems()` call for `items()`, and fixing the encoding required for the Python 3 urllib. I will also be creating a PR on the PyFormance library so this change should eventually be implemented in the library itself, which will allow us to delete this code.

### DIFF
--- a/validator/sawtooth_validator/stats/__init__.py
+++ b/validator/sawtooth_validator/stats/__init__.py
@@ -1,0 +1,18 @@
+# Copyright 2014 Omer Gertel
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------
+
+def InfluxReporter(*args, **kwargs):
+    from .influx import InfluxReporter as cls
+    return cls(*args, **kwargs)
--- a/validator/sawtooth_validator/stats/influx.py
+++ b/validator/sawtooth_validator/stats/influx.py
@@ -1,0 +1,109 @@
+# Copyright 2014 Omer Gertel
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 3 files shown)


validator/sawtooth_validator/stats/reporter.py
```diff
@@ -0,0 +1,48 @@
...
+        self._stopped.set()
+
+    def _loop(self):
+        while not self._stopped.is_set():
+            try:
+                self.report_now(self.registry)
+            except:
+                pass
+            time.sleep(self.reporting_interval)
+        # self._stopped.clear()
```
boydjohnson: Commented out code.

dplumb94: This code came directly from the [PyFormance library](https://github.com/omergertel/pyformance/blob/master/pyformance/reporters/reporter.py) which includes this commented out section. I tried to make the smallest set of changes possible to get it functioning (a released version wouldn't contain this code, we would just import the library). Do you still think it makes sense to clean this up?

_Note: I erroneously set the base branch to master for this PR, but I just corrected it to the metrics branch._

validator/sawtooth_validator/stats/__init__.py
```diff
@@ -0,0 +1,7 @@
+def OpenTSDBReporter(*args, **kwargs):
```
peterschwarz: Is this also taken from the pyformance library?  If so, I would put the appropriate comments with license and copyright notices in the file.

Otherwise, it needs the Intel copyright header

validator/sawtooth_validator/stats/influx.py
```diff
@@ -0,0 +1,94 @@
+# -*- coding: utf-8 -*-
```
peterschwarz: This also needs a header that refers to the pyformance license and copyrights.

validator/sawtooth_validator/stats/reporter.py
```diff
@@ -0,0 +1,48 @@
+import time
```
peterschwarz: License header

protos/validator.proto
```diff
@@ -129,7 +129,7 @@ message Message {
     // content.
     MessageType message_type = 1;
 
-    // The identifier used to coorelate response messages to their related
+    // The identifier used to correlate response messages to their related
```
peterschwarz: You might have branched off of master, so I'm not sure if this commit will cause some weirdness

validator/sawtooth_validator/stats/reporter.py
boydjohnson: No, I think you are right.

boydjohnson:
What does anyone think of pip installing from @dplumb94's fork of the Pyformance library, so that we don't include someone else's code in our repo?

dplumb94:
@boydjohnson I like this idea. How would we go about doing that?

boydjohnson:
@dplumb94 If you have the changes in a fork of omergertel/pyformance, say dplumb94/pyformance you could `pip3 install git+https://github.com/dplumb94/pyformance`. You can also specify a branch like `pip3 install git+https://github.com/dplumb94/pyformance/tree/my-changes`.

PR #743, opened 2017-08-08
Support YAML format for logging config

aludvik:
TOML does not support enough data types to support the SysLog handler type in
python logging. YAML does.

### DIFF
--- a/sdk/python/sawtooth_sdk/client/config.py
+++ b/sdk/python/sawtooth_sdk/client/config.py
@@ -15,6 +15,7 @@
 import os
 import sys
 import toml
+import yaml
 
 
 def get_config_dir():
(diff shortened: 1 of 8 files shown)


sdk/python/sawtooth_sdk/client/config.py
```diff
@@ -82,12 +83,21 @@ def _get_log_config(filename=None):
         log_config (dict): The dictionary to pass to logging.config.dictConfig
     """
     if filename is not None:
+        try:
```
vaporos: This would split a filename like "validator.logging.file.yaml" into [validator, logging, file, yaml] and use 'logging' string as the ext.  Better to just do filename.endswith(".yaml") check.

PR #747, opened 2017-08-09
Initial Identity Transaction Processor.  

agunde406:
This PR includes the first pass a an identity transaction processor that allows you to set roles and policies.  It also includes unit tests and extends sawtooth cli to be able to set roles and policies from the command line. This is the first step needed to be able to implement on-chain permissioning.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,7 @@
 /core/sawtooth_core.egg-info/
 /families/settings/sawtooth_settings/protobuf/
 /families/supplychain/python/sawtooth_supplychain/protobuf
+/families/identity/sawtooth_identity/protobuf
 /families/**/__pycache__/
 
 /coverage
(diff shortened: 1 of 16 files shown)


.gitignore
```diff
@@ -29,6 +29,7 @@
 /core/sawtooth_core.egg-info/
 /families/settings/sawtooth_settings/protobuf/
 /families/supplychain/python/sawtooth_supplychain/protobuf
+/families/identity/python/sawtooth_identity/protobuf
```
boydjohnson: According to how the protogen script has been modified, it appears `/python` shouldn't be there.

cli/sawtooth_cli/identity.py
```diff
@@ -13,14 +13,16 @@
 # limitations under the License.
 # ------------------------------------------------------------------------------
 
+from base64 import b64decode
```
aludvik: These changes should be squashed into the "Add identity to sawtooth cli." commit

peterschwarz:
This should follow a similar model to the settings family where the state values, which are used by the validator, should be in the base `protos` directory.  The transactions which are, in effect, a reference implementation for updating these values can be in the `families/identity/proto`.

agunde406:
protos updated

PR #748, opened 2017-08-10
Invalidate block on cache miss

peterschwarz:
During fork resolution, when finding a common ancestor, a miss to the block cache will cause an exception.  While the callback is notified that block validation has failed, the blocks in the new chain are not marked as invalid.  This commit corrects the issue.

Additionally, it removes the calling of the callback from `_find_common_height`, since it raise an error that is caught by its calling function that results in the callback being executed.

### DIFF
--- a/validator/sawtooth_validator/journal/chain.py
+++ b/validator/sawtooth_validator/journal/chain.py
@@ -284,11 +284,10 @@ class BlockValidator(object):
                         self._block_cache[
                             new_blkw.previous_block_id]
                 except KeyError:
-                    LOGGER.debug("Block rejected due missing" +
+                    LOGGER.debug("Block rejected due to missing" +
                                  " predecessor: %s", new_blkw)
                     for b in new_chain:
                         b.status = BlockStatus.Invalid
-                    self._done_cb(False, self._result)
                     raise BlockValidationAborted()
         elif new_blkw.block_num < cur_blkw.block_num:
             # current chain is longer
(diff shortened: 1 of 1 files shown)


dcmiddle:
I don't recall the guarantees on the block_cache, but I see there's no `try` block around its use with the current chain / current block wrapper. 
e.g. line 301 & line 335
https://github.com/hyperledger/sawtooth-core/pull/748/files#diff-c9a7607e1eb461eddb7b4d1708f012ffR301
https://github.com/hyperledger/sawtooth-core/pull/748/files#diff-c9a7607e1eb461eddb7b4d1708f012ffR335

validator/sawtooth_validator/journal/chain.py
```diff
@@ -320,9 +319,16 @@ def _find_common_ancestor(self, new_blkw, cur_blkw, new_chain, cur_chain):
...
             new_chain.append(new_blkw)
-            new_blkw = \
-                self._block_cache[
-                    new_blkw.previous_block_id]
+            try:
+                new_blkw = \
+                    self._block_cache[
+                        new_blkw.previous_block_id]
+            except KeyError:
+                LOGGER.debug("Block rejected due missing" +
```
jsmitchell: s/due missing/due to missing/

dcmiddle: that's copied from similar block above

peterschwarz: Fixed this in both places.

peterschwarz:
Given that the cache will look things up from the block store first, before deciding there's a miss, it should be fine for the current chain (which, by definition, is in our block store).

PR #752, opened 2017-08-15
Add initial Rust signing SDK and secp2561k support

vaporos:
This implements the majority of signing library support, with the
exception of key generation.


vaporos:
Yes, doc comments before merging to master would be good.

PR #753, opened 2017-08-15
Add YAML smallbank playlist generation

peterschwarz:
Add generation of YAML playlists for the use of generating smallbank
transactions. The number of accounts should be specified, as well as the
number of subsequent transactions to operate on those accounts.

This commit generates all the values in memory before writing.


peterschwarz:
In a future PR, I will create playlist subcommands for `generate` and `read` (or whatever the transaction creation function would be called).

perf/sawtooth/sawtooth_workload/src/playlist.rs
```diff
@@ -0,0 +1,326 @@
...
+            self.current_transaction += 1;
+
+            Some(payload)
+        } else {
+            None
+        }
+    }
+}
+
+impl From<smallbank::SmallbankTransactionPayload> for Yaml {
```
vaporos: cool

PR #756, opened 2017-08-16
Rename Burrow-EVM to Seth

aludvik:


aludvik:
Only 2 of the 4 transactions actually interact with the EVM: CreateContractAccount and MessageCall.

PR #762, opened 2017-08-18
Add Supplychain REST API 

peterschwarz:
Add the REST API for the supplychain family.  This API is backed by a postgres database, which is populated via the supplychain state delta subscriber.  It includes a modification to the API spec, simplifying the paging model.  This makes querying pages with filtering far easier to do.

Additionally, change the use of NULL's for `end_block_num` in the state export tables to using 2^63-1 (MAX_U64, effectively), as this makes the queries much easier for paging and values at particular block number.


agunde406:
Tested Locally

PR #763, opened 2017-08-18
Create Transaction Receipt Store and Handlers

aludvik:
This PR adds a receipt store object, transaction receipt protobuf messages, message handlers, and tests that they all function. The receipt store will be used once the transaction execution results are returned from the executor.

### DIFF
--- a/protos/txn_receipt.proto
+++ b/protos/txn_receipt.proto
@@ -1,0 +1,60 @@
+// Copyright 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// -----------------------------------------------------------------------------
+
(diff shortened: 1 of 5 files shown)


protos/txn_receipt.proto
```diff
@@ -0,0 +1,60 @@
...
+option java_package = "sawtooth.sdk.protobuf";
+option go_package = "txn_receipt_pb2";
+
+import "state_delta.proto";
+import "events.proto";
+
+message TransactionReceipt {
+  message Data {
+    // Interpretation left to transaction family
+    string data_type = 1;
```
peterschwarz: What's the data type for?  We don't do this sort of thing for any other piece of opaque data.

aludvik: It is a hint for how to deserialize the data. It is functionally the same as "payload_encoding" in the transaction header: https://github.com/hyperledger/sawtooth-core/blob/master/protos/transaction.proto#L53

peterschwarz: That's really more for routing the transaction to the correct TP.  State Data, for example, has no hints.

vaporos: As @aludvik stated, it's a hint for deserialization.  Without this hint, the serialization method would presumably be tightly coupled to the transaction type (family, version).  The hint allows the list to contain data serialized using different methods (for example, different protobuf messages).

As it does not seem to be strictly necessary, I'd be fine with keeping it as-is or a simplified version of just "repeated bytes data".

With respect to payload_encoding, I think we are considering removing it from the routing.

peterschwarz: I don't see how it is that any different then state encoding being tightly coupled to the transaction type spec, but I'm not going to argue against any more.

PR #764, opened 2017-08-21
Implement Transactor Permissioning

agunde406:
This includes transactor permissioning checks when receiving a batch from a client and on block validation.  The permission verifier uses identity view to check the corresponding roles stored in state. If there is not role stored, the default policy is used. If the default policy is not stored, permit all transactors.


validator/sawtooth_validator/gossip/permission_verifier.py
```diff
@@ -13,13 +13,16 @@
 # limitations under the License.
 # ------------------------------------------------------------------------------
 import logging
-# pylint: disable=import-error,no-name-in-module
-# needed for google.protobuf import
+
+# pylint: disable=no-name-in-module,import-error
```
boydjohnson: It might have been a pylint change, but this isn't necessary any more. I tested it locally without the disable pylint comment and it showed no lint.

validator/sawtooth_validator/gossip/permission_verifier.py
```diff
@@ -28,46 +29,163 @@
...
+                        LOGGER.debug(
+                            "Transaction Signer: %s is not permitted.",
+                            header.signer_pubkey)
+                        return False
+        return True
+
+    def _allowed(self, public_key, policy):
+        for entry in policy.entries:
+            if entry.type == Policy.PERMIT_KEY:
+                if public_key == entry.key or "*" in entry.key:
```
peterschwarz: `"*" in entry.key` will hold true for any string with a `*` in it.  Can  we specify partials? I.e. `"<partial key>*"`.  If that's the case, we need to do more of a match.  if not, we can probably just do an `==` here.

validator/sawtooth_validator/journal/chain.py
```diff
@@ -220,6 +222,15 @@ def _verify_block_batches(self, blkw):
                 return False
         return True
 
+    def _validate_permissions(self, blkw):
```
peterschwarz: Could you add a little doc to this?  What's the context that this is validating permissions for the block.

validator/sawtooth_validator/journal/journal.py
```diff
@@ -234,6 +237,9 @@ def get_current_root(self):
     def get_block_store(self):
         return self._block_store
 
+    def add_permission_verifier(self, permission_verifier):
+        self._permission_verifier = permission_verifier
```
peterschwarz: This feels a little off given that the `_permission_verify` can be `None`for a period of time, from the Journal's perspective.  Clearly, this is due to the circular dependencies between the two.  It requires ordering of calls out side of the journal.  Can the PermissionVerifier be initialized with the current state root in another way?  I.e. not via the Journal.

Minimally, the `_init_subprocesses` function should through an error if it is `None`, before creating the subprocesses, to ensure that the journal isn't created without on.

validator/sawtooth_validator/gossip/permission_verifier.py
agunde406: Partial are not supported right now updating.

validator/sawtooth_validator/journal/chain.py
agunde406: Added

agunde406:
rebase before merging

PR #766, opened 2017-08-23
Implement Off-chain Transactor Permissioning. 

agunde406:
This PR is built on top of #764 Implement Transactor Permissioning which implements on-chain transactor permissioning. 

This implements off-chain transactor permissioning by adding the following to validator.toml:
[permissions]
ROLE = POLICY_NAME (filename for the policy)

Where the policy file referenced by the policy_name is stored in the policy_dir defined by the PathConfig. The ValidatorConfig will read the files and pass a permission dictionary with the roles and policy to the PermissionVerifier. These roles will only be checked when a batch is received form a client.


validator/sawtooth_validator/config/validator.py
```diff
@@ -54,21 +56,21 @@ def load_toml_validator_config(filename):
...
             "{}".format(", ".join(sorted(list(invalid_keys)))))
     bind_network = None
     bind_component = None
-    for bind in toml_config.get("bind"):
-        if "network" in bind:
-            bind_network = bind[bind.find(":")+1:]
-        if "component" in bind:
-            bind_component = bind[bind.find(":")+1:]
+    if "bind" in toml_config:
+        for bind in toml_config.get("bind"):
```
peterschwarz: You could return a default of `[]`: 
```
for bind in toml_config.get("bind", []):
    ...
```

PR #776, opened 2017-08-25
Update Identity Transaction Processor to accept transaction from allowed signers

agunde406:
Only those pub keys that are in the setting sawtooth.identity.allowed_keys should be allowed to sign identity transactions. If an identity transaction is received signed by someone else, the transaction is deemed invalid. 

This PR is build on top of #766 Implement Off-Chain Transactor Permissioning. This Pr will be rebased once that PR is merged.

### DIFF
--- a/cli/sawtooth_cli/identity.py
+++ b/cli/sawtooth_cli/identity.py
@@ -37,6 +37,7 @@ from sawtooth_cli.protobuf.transaction_pb2 import Transaction
 from sawtooth_cli.protobuf.batch_pb2 import BatchHeader
 from sawtooth_cli.protobuf.batch_pb2 import Batch
 from sawtooth_cli.protobuf.batch_pb2 import BatchList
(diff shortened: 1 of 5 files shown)


cli/sawtooth_cli/identity.py
```diff
@@ -37,6 +37,7 @@
 from sawtooth_cli.protobuf.batch_pb2 import BatchHeader
 from sawtooth_cli.protobuf.batch_pb2 import Batch
 from sawtooth_cli.protobuf.batch_pb2 import BatchList
+from sawtooth_cli.config import _key_to_address
```
peterschwarz: If this is now a public function of the `sawtooth_cli.config` package, it should have a public version called `setting_key_to_address`, so it's clear what that is referring to once imported.

agunde406:
rebasing before merge

PR #790, opened 2017-08-30
Create BlockInfo transaction family

aludvik:


families/block_info/sawtooth_block_info/processor/handler.py
```diff
@@ -0,0 +1,179 @@
...
+        txn = BlockInfoTxn()
+        txn.ParseFromString(transaction.payload)
+
+        # Validate block info included in payload
+        next_block = txn.block
+        if any([next_block.block_num < 0,
+                next_block.previous_block_id == "",
+                next_block.signer_pubkey == "",
+                next_block.header_signature == "",
+                next_block.timestamp < 0]):
```
peterschwarz: This should be `== 0`,  since the default value for numbers is `0`

families/block_info/sawtooth_block_info/processor/handler.py
```diff
@@ -0,0 +1,179 @@
...
+
+            config.target_count = \
+                DEFAULT_TARGET_COUNT if txn.target_count == 0 \
+                else txn.target_count
+
+            config.latest_block = next_block.block_num
+            config.oldest_block = next_block.block_num
+
+            now = time.time()
+            if next_block.timestamp > now + config.sync_tolerance:
```
peterschwarz: This should be `abs(now - next_block.timestamp) > config.sync_tolerance` to account for the peer being slightly behind

PR #796, opened 2017-08-31
Update serial scheduler to fail fast on invalid batch

boydjohnson:
This commit also updates tests and testing framework to not set an
invalid transaction result and expect further transactions within
a batch.

STL-579


boydjohnson:
Build 2 failed on test_two_families. I ran that test 5 times locally and they each passed, so I am rebuilding.

boydjohnson:
I am able to get it to fail intermittently locally by setting the vm on my mac to 2.5 Gbs and a single core. I will rethink this PR in light of that.

PR #797, opened 2017-08-31
Add smallbank docker support

vaporos:


dplumb94:
When we merge this with the metrics branch and we want to collect statistics with the influx/grafana stack, the containers will have to be set to join the `sawtooth_stats` network. This can be set in the docker-compose file.

PR #798, opened 2017-08-31
Restrict smoke tests to appropriate namespaces

nick-drozd:
Currently the intkey smoke and two-family tests verify that the entire
state is initially empty. This will work if the validator is started
with an empty genesis batch, as it is in these tests' compose files:

        sawtooth admin keygen && \
        sawtooth admin genesis && \
        sawtooth-validator -v \
            --endpoint tcp://validator:8800 \
            --bind component:tcp://eth0:4004 \
            --bind network:tcp://eth0:8800 \

However, if the genesis batch is non-empty, then state won't be empty,
and the tests will fail:

        sawtooth admin keygen && \
        sawtooth keygen my_key && \
        sawtooth config genesis -k /root/.sawtooth/keys/my_key.priv && \
        sawtooth admin genesis config-genesis.batch && \
        sawtooth-validator -vv \
          --endpoint tcp://validator:8800 \
          --bind component:tcp://eth0:4004 \
          --bind network:tcp://eth0:8800 \

This commit fixes the tests to only check their specific namespaces so
that the tests will pass regardless of what the genesis batch is.

The fix here is quick and dirty. A more elegant approach would be to refactor both tests using the `RestClient` at `integration/sawtooth_integration/tests/integration_tools.py`.

### DIFF
--- a/integration/sawtooth_integration/docker/seth_intkey_test.yaml
+++ b/integration/sawtooth_integration/docker/seth_intkey_test.yaml
@@ -17,6 +17,17 @@ version: "2.1"
 
 services:
 
+  settings-tp:
+    image: sawtooth-settings-tp:$ISOLATION_ID
+    volumes:
+      - $SAWTOOTH_CORE:/project/sawtooth-core
+    expose:
+      - 4004
+    depends_on:
+      - validator
+    command: settings-tp -vv tcp://validator:4004
+    stop_signal: SIGKILL
+
   seth-tp:
     image: sawtooth-seth-tp:$ISOLATION_ID
     volumes:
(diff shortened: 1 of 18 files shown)


boydjohnson:
http://intelledger.github.io/app_developers_guide/ubuntu.html#validator-start-up-process It seems like `sawtooth config genesis` is a required startup command. So all of our integration tests should have them in the validator startup.

peterschwarz:
`sawtooth config genesis` is required if one is going to do any sort of settings transactions.  It can be skipped, but that could result in anyone hijacking the settings family (by being the only authorized key).

PR #801, opened 2017-09-01
Fix bug when getting and setting to wildcarded and non-wildcarded addresses

boydjohnson:
This fixes the bug that @aludvik and @nick-drozd were both seeing.

### DIFF
--- a/validator/sawtooth_validator/execution/context_manager.py
+++ b/validator/sawtooth_validator/execution/context_manager.py
@@ -499,8 +499,6 @@ class ContextManager(object):
                 # There aren't any more contexts known about.
                 break
             current_context = self._contexts[current_c_id]
-            if not current_context.is_read_only():
-                current_context.make_read_only()
 
             # First, check for addresses that have been deleted.
             deleted_addresses = current_context.get_if_deleted(reads)
@@ -920,10 +918,18 @@ class _ContextFuture(object):
         """
 
         if self._read_only:
+            if not from_tree:
+                LOGGER.warning("Tried to set address %s on a"
+                               " read-only context.",
+                               self.address)
             return
 
         with self._condition:
             if self._read_only:
+                if not from_tree:
+                    LOGGER.warning("Tried to set address %s on a"
+                                   " read-only context.",
+                                   self.address)
                 return
             if from_tree:
                 # If the result has not been set in the context, overwrite the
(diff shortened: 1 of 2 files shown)


validator/tests/test_context_manager/tests.py
```diff
@@ -1033,6 +1033,54 @@ def test_complex_basecontext_squash(self):
                                          "manager calculated merkle hashes "
                                          "are the same")
 
+    def test_wildcarded_inputs_outputs(self):
+        """Tests the context manager with wildcarded inputs and outputs.
+
+        Notes:
+            1. Create a context with a wildcarded input and output and
+               another non-wildarded input and output.
```
aludvik: typo "wildcarded"

validator/tests/test_context_manager/tests.py
```diff
@@ -1033,6 +1033,54 @@ def test_complex_basecontext_squash(self):
...
+            2. Get an address under the wildard and set to both the non-wildcarded
```
aludvik: type wildcard

validator/sawtooth_validator/execution/context_manager.py
```diff
@@ -920,10 +918,14 @@ def set_result(self, result, from_tree=False):
         """
 
         if self._read_only:
+            if not from_tree:
+                LOGGER.warning("Tried to set to a read-only context.")
```
aludvik: Log the set address and/or payload?

validator/tests/test_context_manager/tests.py
```diff
@@ -1033,6 +1033,54 @@ def test_complex_basecontext_squash(self):
...
+               address and an address under the wildcard..
```
nick-drozd: `..` -> `.`

validator/tests/test_context_manager/tests.py
```diff
@@ -1033,6 +1033,54 @@ def test_complex_basecontext_squash(self):
...
+            4. Squash the context and compare to a manually generated state
+               hash.
+        """
+
+        namespaces = [
+            self._create_address('a')[:8],
```
nick-drozd: I guess this is consistent with the rest of the file, but why is `_create_address` called over and over instead of assigning variables? `address_a = self._create_address('a')` or something like that.

boydjohnson: So I guess in general, I think `self._create_address('a')` is as readable and informative as `address_a` and saves lines where `address_a = self._create_address('a')` would be. Line 1049 where you comment is different in that `namespace_a` would have more information than `self._create_address('a')[:8]`.  I will take this under advisement, but don't want to have some of the namespaces and addresses using `_create_address` and some not.

boydjohnson:
@nick-drozd I fixed your typo comment and had a half-way rebuttal to your other comment.

PR #802, opened 2017-09-01
Implement Trust Authorization and Update PermissionVerifier to check "network"

agunde406:


validator/sawtooth_validator/networking/handlers.py
```diff
@@ -40,19 +50,53 @@ def handle(self, connection_id, message_content):
         self._network.update_connection_endpoint(connection_id,
                                                  message.endpoint)
 
-        ack = NetworkAcknowledgement()
-
-        if self._network.allow_inbound_connection():
-            LOGGER.debug("Allowing incoming connection: %s",
-                         connection_id)
-            ack.status = ack.OK
+        # Get what AuthorizationType the network role requries
```
boydjohnson: requries -> requires

validator/sawtooth_validator/networking/handlers.py
```diff
@@ -88,3 +132,74 @@ def handle(self, connection_id, message_content):
...
+                if "network" in roles:
+                    permitted = self._permission_verifier.check_network_role(
+                        request.public_key)
+                if not permitted:
+                    violation = AuthorizationViolation(
+                        violation=AuthorizationViolation.NETWORK)
+                    return HandlerResult(
+                        HandlerStatus.RETURN_AND_CLOSE,
+                        message_out=violation,
+                        message_type=validator_pb2.Message
```
boydjohnson: If splitting on the dot is necessary for line width, can you put in a \ to alert the reader that there is a line break.

boydjohnson:
Hi @agunde406 Could you give a high level overview of what is happening in each of the handlers and why?

validator/sawtooth_validator/networking/handlers.py
agunde406: That would cause the lint error: the backslash is redundant between brackets

agunde406:
Descriptions added to Authorization handlers

protos/authorization.proto
```diff
@@ -0,0 +1,94 @@
...
+// limitations under the License.
+// -----------------------------------------------------------------------------
+
+syntax = "proto3";
+option java_multiple_files = true;
+option java_package = "sawtooth.sdk.protobuf";
+option go_package = "authorization_pb2";
+
+message ConnectionRequest {
+	string endpoint = 1;
```
nick-drozd: Indentation is irregular and uses hard tabs.

protos/authorization.proto
```diff
@@ -0,0 +1,94 @@
...
+  NETWORK = 1;
+}
+
+message ConnectionResponse {
+  enum Status {
+      OK = 0;
+      ERROR = 1;
+  }
+
+	enum RoleType {
```
nick-drozd: Does this enum need to be repeated in all these message definitions?

protos/authorization.proto
```diff
@@ -0,0 +1,94 @@
...
+
+message AuthorizationTrustRequest {
+	enum RoleType {
+		// A shorthand request for asking for all allowed roles.
+		ALL = 0;
+		// Role defining validator to validator communication
+		NETWORK = 1;
+	}
+	// A set of requested RoleTypes
+	repeated RoleType requested_roles = 1;
```
nick-drozd: Since `requested_roles` and `approved_roles` below are both `repeated RoleType`, they would be easier to work with if they were both just called `roles`.

protos/authorization.proto
agunde406: I was having problems making it work when it was defined outside but i think i figured it out and will update

validator/sawtooth_validator/gossip/permission_verifier.py
```diff
@@ -241,6 +245,43 @@ def check_off_chain_transaction_roles(self, transactions):
...
+        else:
+            policy_name = role.policy_name
+        policy = identity_view.get_policy(policy_name)
+        if policy is not None:
+            if not self._allowed(public_key, policy):
+                LOGGER.debug(
+                    "Node is not permitted: %s.",
+                    public_key)
+                return False
+        return True
```
aludvik: Will there be a new method for every role that is created and needs to be checked? A more reusable function might be better. Something like `check_role()` that is generic.

protos/authorization.proto
```diff
@@ -0,0 +1,68 @@
...
+// limitations under the License.
+// -----------------------------------------------------------------------------
+
+syntax = "proto3";
+option java_multiple_files = true;
+option java_package = "sawtooth.sdk.protobuf";
+option go_package = "authorization_pb2";
+
+message ConnectionRequest {
+  string endpoint = 1;
```
peterschwarz: Add a comment on this one.

protos/authorization.proto
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  //Authorization Type required for the authorization procedure
+  enum AuthorizationType {
+    TRUST = 0;
+    CHALLENGE = 1;
+  }
+
+  message RoleEntry {
+    RoleType role = 1;
```
peterschwarz: Add some comments on these.

protos/validator.proto
```diff
@@ -127,6 +127,12 @@ message Message {
         STATE_DELTA_EVENT = 404;
         STATE_DELTA_GET_EVENTS_REQUEST = 405;
         STATE_DELTA_GET_EVENTS_RESPONSE = 406;
+
+        // Message types for Authorization Types
+        CONNECTION_RESPONSE = 600;
```
peterschwarz: Should this be an `AUTHORIZATION_CONNECTION_RESPONSE`?

validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -598,25 +706,72 @@ def add_outbound_connection(self, uri,
...
+            connection,
+            )
+
+    def _authorization_callback(self, request, result,
+                                connection=None,
+                                success_callback=None,
+                                failure_callback=None):
+        if result.message_type == \
+                validator_pb2.Message.AUTHORIZATION_TRUST_RESPONSE:
+            success_callback(connection.connection_id)
```
peterschwarz: Need to check that these callbacks are not `None`

validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -815,6 +1071,35 @@ def send(self, message_type, data, callback=None):
...
+        Returns:
+            future.Future
+        """
+        message = validator_pb2.Message(
+            correlation_id=_generate_id(),
+            content=data,
+            message_type=message_type)
+
+        fut = future.Future(message.correlation_id, message.content,
+                            has_callback=True if callback is not None
```
peterschwarz: We should just pass the callback in to the constructor and get rid of this `has_callback` business.  There are no places where we don't pass the callback in immediately after creating the future.

PR #805, opened 2017-09-02
Implement Batch Injection - Phase 1/3

aludvik:
Add feature to inject batches into the beginning of a block and create a new BlockInfoInjector which utilizes this feature. Also add tests.

This is an incremental PR towards the completion of STL-567. It implements the portion of the task that is required by the Solidity Developer Compatibility epic and unblocks STL-495.

This PR also depends on the context manager fix @boydjohnson has up.

Phase 2 - Injection before and after a batch
Phase 3 - Injection at the end of a block

### DIFF
--- a/integration/sawtooth_integration/tests/test_block_info_injector.py
+++ b/integration/sawtooth_integration/tests/test_block_info_injector.py
@@ -1,0 +1,102 @@
+# Copyright 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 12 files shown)


integration/sawtooth_integration/tests/test_block_info_injector.py
```diff
@@ -0,0 +1,84 @@
...
+
+def get_block_info(block_num):
+    bi = BlockInfo()
+    bi.ParseFromString(get_state(create_block_address(block_num)))
+    return bi
+
+
+class TestBlockInfoInjector(unittest.TestCase):
+    def test_block_info_injector(self):
+        batches = make_batches('abcd')
```
agunde406: Please add a docstring for the test

aludvik:
Added a doc string and additional assertion to `test_block_info_injector.py`

PR #807, opened 2017-09-05
Update serial scheduler to fail fast on invalid batch

boydjohnson:
This commit also updates tests and testing framework to not set an
invalid transaction result and expect further transactions within
a batch.

STL-579

### DIFF
--- a/validator/sawtooth_validator/execution/scheduler_serial.py
+++ b/validator/sawtooth_validator/execution/scheduler_serial.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 # ------------------------------------------------------------------------------
 
-import queue
+from collections import deque
 from threading import Condition
 
 from sawtooth_validator.execution.scheduler import BatchExecutionResult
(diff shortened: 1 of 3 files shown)


validator/sawtooth_validator/execution/scheduler_serial.py
```diff
@@ -154,30 +154,48 @@ def _get_batch_result(self, txn_id):
...
+                # Handle this transaction being invalid based on a
+                # dependency.
+                if any(self._dep_is_known(d) and self._in_invalid_batch(d)
+                        for d in self._get_dependencies(txn)):
+                    self._set_batch_result(
+                        txn.header_signature,
+                        False,
+                        None)
+                    txn = None
+                    continue
```
peterschwarz: Will this continue with transactions that are still in the same batch?

boydjohnson: Yeah, the transactions are read in order from the queue, and if the batch has already been determined to be invalid, they are handled as invalid.

PR #809, opened 2017-09-05
T&T TP

nick-drozd:


families/track_and_trade/nose2.cfg
```diff
@@ -0,0 +1,8 @@
+[unittest]
```
delventhalz: Should this be located at `families/track_and_trade/tests/nose2.cfg`?

peterschwarz: It's in the right place.

PR #816, opened 2017-09-06
Allow out of date warnings in poet

dcmiddle:
Fix attestation verification report checking in the swig code to handle
exceptions raised logging the error rather than throwing a stack trace.

### DIFF
--- a/consensus/poet/sgx/sawtooth_poet_sgx/poet_enclave_sgx/poet_enclave.i
+++ b/consensus/poet/sgx/sawtooth_poet_sgx/poet_enclave_sgx/poet_enclave.i
@@ -243,7 +243,7 @@ def _check_verification_report(verification_report, signature):
     # 2. Does not include a revocation reason.
 
     if 'revocationReason' in verification_report_dict:
-        raise ValueError('AVR indicates the EPID group has been revoked')
+        raise EnvironmentError('AVR indicates the EPID group has been revoked')
 
     # 3. Includes an enclave quote status
 
@@ -255,9 +255,14 @@ def _check_verification_report(verification_report, signature):
     # 4. Enclave quote status should be "OK".
 
     if isv_enclave_quote_status.upper() != 'OK':
-        raise ValueError(
-            'AVR enclave quote status is not OK: {}'.format(
-                isv_enclave_quote_status))
+        # Allow out of date severity issues to pass.
+        if isv_enclave_quote_status.upper() == 'GROUP_OUT_OF_DATE':
+            logger.error('Machine requires update (probably BIOS)'
+                ' for SGX compliance.')
+        else:
+            raise EnvironmentError(
+                'AVR enclave quote status is bad: {}'.format(
+                    isv_enclave_quote_status))
 
     # 5. Includes an enclave quote.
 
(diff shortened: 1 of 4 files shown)


consensus/poet/sgx/sawtooth_poet_sgx/poet_enclave_sgx/poet_enclave.i
```diff
@@ -344,7 +344,12 @@ def create_signup_info(originator_public_key_hash, nonce):
 
         # Check the AVR to make sure it is valid and does not indicate
         # any errors
-        _check_verification_report(verification_report, signature)
+        try:
+            _check_verification_report(verification_report, signature)
+        except ValueError as e:
+            logger.warning("Invalid attestation verification report",
```
vaporos: Isn't this a fatal error, or can we recover from this?  This would ignore it and continue to the next step which doesn't seem right.

Would it be more appropriate to propagate this up to near main() and handle the problem there (either logging as error or printing to stderr, depending on what is most appropriate) prior to exiting?

dcmiddle: My bad. That should have had a `return None`. 
Some of those errors could be transient and worth some retries. Something like the machine's bios being old would not be recoverable though. So then there's a choice, do we say the validator should shut down because it can't publish blocks or should it stay up extending its chain until an admin can address the bios?

vaporos: I'm concerned that the sysadmin would not see the error messages and think that the validator was working, when in fact it would be broken.  Because  of how we have logging setup, using "logger.error()" might mitigate this to some extent, since the error log is low-volume.

When I ran into this, the block publisher sits in a very very tight loop spitting out many lines of logging, so if we stay up we should probably rate limit the amount of logging generated due to this error to some level that won't fill up a smallish log directory.

The "Invalid attestation verification report" logging message is probably not good enough in isolation to indicate the state of the validator to the sysadmin (though it seems possible the str(e) makes it clearer...).  The only way I can think of to do this is setup a machine in this state and run the validator, observe the logging, then tune the logging output at the appropriate points so that what ends up in the error log is fairly obvious as to what is happening.

jsmitchell: I feel like if you are joining a network that is specifying sgx poet, and you think you are starting as a validator, then a fatal error is the right thing to do. Degrading to some observer behavior does not seem correct.

dcmiddle: I'm thinking less about startup and more about a machine that's been running for a while. Do you want the validator to halt 2 months into execution or degrade to observer? Probably still better to halt and make it strongly apparent that the machine needs servicing.

vaporos: Exiting anytime that is not immediately upon startup is completely undesirable.  No other (common) long-lived service acts that way (on purpose anyway).  There are obvious reasons not do that as well, since it will break the REST API completely, offering no opportunity to pass some sort of intelligent status back up to a monitoring service.  (This is also a reason to stay running in the first place.)

dcmiddle: I updated the PR to not attempt to exit on this type of error which I now throw as an environment error.
Instead it just logs as critical. Main intent of this PR now is just properly handling the "out of date" differently than a revocation condition.

PR #821, opened 2017-09-08
Fix scheduler cli with addition of scheduler option to ValidatorConfig creation

boydjohnson:


peterschwarz:
What's the reasoning for switching back the scheduler to all serial?

boydjohnson:
@peterschwarz So due to the bug in the cli, the parallel scheduler wasn't actually being run. Due to a bug in the parallel scheduler when it was run it wouldn't pass this test. This afternoon, I was able to fix the bug in the parallel scheduler, so is can pass the dynamic network test, so switching back to serial probably isn't needed. I will update this PR with my bug-fix for the parallel scheduler.

PR #823, opened 2017-09-08
Add txn execution results to scheduler interface

aludvik:
This is an incremental PR for STL-631

### DIFF
--- a/validator/sawtooth_validator/execution/scheduler.py
+++ b/validator/sawtooth_validator/execution/scheduler.py
@@ -52,8 +52,24 @@ class Scheduler(object, metaclass=ABCMeta):
         raise NotImplementedError()
 
     @abstractmethod
+    def get_transaction_execution_results(self, batch_signature):
+        """Get all TransactionExecutionResults for a batch. If the batch isn't
+        finished or is invalid, returns None.
+
+        Args:
+            batch_signature (str): The signature of the batch, which must match
+                the header_signature field of the Batch.
+
+        Returns:
+            list of :obj:`TransactionExecutionResult`: The results of all
+                transactions executed in the batch.
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
     def set_transaction_execution_result(
-            self, txn_signature, is_valid, context_id):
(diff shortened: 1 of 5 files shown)


validator/sawtooth_validator/execution/scheduler.py
```diff
@@ -52,8 +52,24 @@ def get_batch_execution_result(self, batch_signature):
         raise NotImplementedError()
 
     @abstractmethod
+    def get_transaction_execution_results(self, batch_signature):
+        """Get all TransactionExecutionResults for a batch. If the batch isn't
+        finished or is invalid, returns None.
```
boydjohnson: Even if the Batch is invalid, wouldn't we want to return a list of TransactionExecutionResults, to see which one failed? I would expect only when the batch is incomplete would we want to return None.

validator/sawtooth_validator/execution/scheduler.py
```diff
@@ -222,6 +238,43 @@ def __init__(self, is_valid, state_hash):
         self.state_hash = state_hash
 
 
+class TxnExecutionResult:
+    """The resulting execution data from running the transaction through the
+    executor.
+
+    Attributes:
+        is_valid (bool): True if the batch is valid, False otherwise.
```
boydjohnson: batch -> transaction ?

validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -491,6 +475,18 @@ def get_batch_execution_result(self, batch_signature):
                                           clean_up=True)
             return BatchExecutionResult(is_valid=True, state_hash=state_hash)
 
+    def get_transaction_execution_results(self, batch_signature):
+        with self._condition:
+            # This method calculates the BatchExecutionResult on the fly,
```
boydjohnson: Does this comment apply to this method?

boydjohnson:
@aludvik Looks good, generally. I had a thought. Will anything be calling get_txn_execution_results before the scheduler is complete? Because of the way the parallel scheduler works, until the scheduler is complete it is still possible that a batch of transactions all with results will have the result thrown away and rerun.

PR #825, opened 2017-09-08
Make TimedCaches self-purging, and improve BatchTracker's cache usage

delventhalz:
Adds the concept of purge frequency internally to TimedCaches. Every time a value is set, if the specified purge frequency has elapsed (default 30s), the cache will be purged. This removes the need to manually purge TimedCaches.

Updates all uses of TimedCaches (and BlockCaches) to take advantage of this behavior, and the BatchTracker class to use init parameters instead of global variables to configure the wait times. Also reduces the BatchTracker keep time from 1 hour, to 10 minutes.

### DIFF
--- a/validator/sawtooth_validator/journal/block_cache.py
+++ b/validator/sawtooth_validator/journal/block_cache.py
@@ -20,8 +20,8 @@ class BlockCache(TimedCache):
     """
     A dict like interface to access blocks. Stores BlockState objects.
     """
-    def __init__(self, block_store=None, keep_time=10):
-        super(BlockCache, self).__init__(keep_time)
+    def __init__(self, block_store=None, keep_time=10, purge_frequency=10):
+        super(BlockCache, self).__init__(keep_time, purge_frequency)
         self._block_store = block_store if block_store is not None else {}
 
     def __getitem__(self, key):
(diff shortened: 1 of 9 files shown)


jsmitchell:
I've tested this locally and it works. Since purging of the cache is determined on every insert, we could have a situation where a large number of entries are encached prior to a long period of time with no insert activity. This would result in lots of references being kept in memory for arbitrary durations. On a busy cache, this is not an issue. On a bursty cache, eg. batch submission, this may be an issue. Do we want to switch to some other scheme? Spin up a thread to fire a purge event per duration? @peterschwarz @vaporos

aludvik:
@jsmitchell Would putting a limit on the size of the cache help solve that problem?

peterschwarz:
If we are going to put a size limit on it, we might as well just use a LRU Cache, which already exists in the Python core library.

delventhalz:
Based on the existing code, it seems not important for `TimedCache` to be very precise about when it purges old data. They mostly purges on block updates, which are not predictable (and potentially not all that regular depending on the deployment). That makes me doubt that it is important enough to spin up extra threads to ensure `TimedCaches` get regular semi-precise purges.

Admittedly though, I don't fully understand the purpose they were originally developed for. When I used them in the BatchTracker, I want to be able to say to clients _"you can access invalid batch information for at least ten minutes"_. Longer is fine for this purpose, but less time is not fine. This is why a size-delimited cache does not work for the BatchTracker, and why I wasn't necessarily concerned about a lack of updates making invalid batches stick around indefinitely.

PR #829, opened 2017-09-10
STL 631 - Transaction Execution Results

aludvik:
This PR updates the context manager, executor, and chain controller to enable passing receipts from a transaction processor up to chain observers. It then makes the receipt store observe chain updates and save receipts and enables message handling to allow querying for receipts.


aludvik:
Added comments to commits

validator/sawtooth_validator/journal/chain.py
```diff
@@ -797,7 +797,7 @@ def _set_genesis(self, block):
                            "genesis block: %s", block)
 
     def _make_receipts(self, results):
-        receipts = {}
+        receipts = []
```
agunde406: This change should be squashed with the previous commit

validator/tests/test_context_manager/tests.py
```diff
@@ -183,6 +184,40 @@ def _create_txn_inputs_outputs(self, start=None):
...
+        sets = {addr1: b'1'}
+        events = [Event(
+            event_type="test",
+            attributes=[Event.Attribute(key="test", value="test")],
+            data="test".encode())]
+        deletes = {addr2: None}
+        data = [("test", "test".encode())]
+
+        self.context_manager.set(context_id, [sets])
+        for event in events:
```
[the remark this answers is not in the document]
aludvik: Updated so there is more than one.

validator/sawtooth_validator/journal/chain.py
aludvik: Fixed

PR #830, opened 2017-09-11
Add transaction submission route to T&T Client Server

delventhalz:
To test you must:
1. Create a user by submitting a JSON `POST` request to `localhost:3000/api/users`:
```
{
  "username": <string>,
  "password": <string>,
  "email": <string, email>,
  "publicKey": <string, valid public key>
}
```
2. Save the `authorization` string that is returned.
3. Create a transaction signed by the with the same key
4. Encode in a `TransactionList` and submit as the body in a `POST` request to `localhost:3000/api/transactions`. Set the headers to:
```
"Authorization": <string, the authorization key>
"Content-Type": "application/octet-stream"
```

### DIFF
--- a/families/track_and_trade/server/blockchain/deltas.js
+++ b/families/track_and_trade/server/blockchain/deltas.js
@@ -1,0 +1,88 @@
+/**
+ * Copyright 2017 Intel Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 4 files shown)


delventhalz:
@peterschwarz You submit it in the POST to create the user. The server is agnostic about how you generate it, but once you've created a user account with a public key, it will expect that key to match for submitted transactions.

The next PR will be login/signup which will generate public keys client side used the SDK.

peterschwarz:
The batching pubkey is different from the client pubkey, but needs to be included in the transaction header.  Is it returned to the user on the initial authorization task?

delventhalz:
@peterschwarz Ah. That is a good question. My thought was it would just be a configuration variable, that whoever was deploying the client would set up. I'm not sure sending it back on authorization makes much sense since it will be the same for all users. Perhaps an API route which will send back the publicKey? `GET /api/config` or some such?

families/track_and_trade/server/blockchain/deltas.js
```diff
@@ -0,0 +1,87 @@
...
+const getProtoName = address => {
+  const typePrefix = address.slice(6, 8)
+  if (typePrefix === 'ea') {
+    if (address.slice(-4) === '0000') return 'Property'
+    else return 'PropertyPage'
+  }
+
+  const names = {
+    ae: 'Agent',
+    aa: 'Property',
```
nick-drozd: 'Proposal'

PR #835, opened 2017-09-11
New Yaml scheduler tester method

boydjohnson:
The main difference with the new yaml scheduler method is that it calls finalize after next_transaction.

There are some small fixes to the parallel scheduler: the fix having to do with predecessors was found from writing more yaml tests. There is also a fix to reduce iteration in finding the next available transaction.

### DIFF
--- a/validator/sawtooth_validator/execution/scheduler_parallel.py
+++ b/validator/sawtooth_validator/execution/scheduler_parallel.py
@@ -593,6 +593,13 @@ class ParallelScheduler(Scheduler):
         for predecessor_id in self._txn_predecessors[txn.header_signature]:
             if predecessor_id not in self._txn_results:
                 return True
+            # Since get_initial_state_for_transaction gets context ids not
+            # just from predecessors but also in the case of an enclosing
+            # writer failing, predecessors of that predecessor, this extra
+            # check is needed.
+            for pre_pred_id in self._txn_predecessors[predecessor_id]:
+                if pre_pred_id not in self._txn_results:
+                    return True
 
         return False
 
--- a/validator/sawtooth_validator/execution/scheduler_parallel.py
+++ b/validator/sawtooth_validator/execution/scheduler_parallel.py
@@ -16,7 +16,6 @@
 from ast import literal_eval
 from threading import Condition
 from collections import deque
-from collections import OrderedDict
 
 from sawtooth_validator.protobuf.transaction_pb2 import TransactionHeader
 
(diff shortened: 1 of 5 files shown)


validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -593,6 +593,9 @@ def _has_predecessors(self, txn):
         for predecessor_id in self._txn_predecessors[txn.header_signature]:
             if predecessor_id not in self._txn_results:
                 return True
+            for pre_pred_id in self._txn_predecessors[predecessor_id]:
```
aludvik: Since the reason this change is needed seems fairly complex, a comment explaining why the check is being performed here seems like a good idea.

validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -580,14 +585,8 @@ def set_transaction_execution_result(
...
     def _unscheduled_transactions(self):
-        # shouldn't actually return txn if dependencies were
-        # marked invalid.
-        txns = self._transactions.copy()
 
-        for txn_header in self._scheduled:
-            txns.pop(txn_header, None)
-
-        return list(txns.values())
+        return list(self._txns_available.copy())
```
aludvik: I think the `list()` is redundant.

boydjohnson: Yeah, I'll change that.

validator/sawtooth_validator/execution/scheduler_parallel.py
boydjohnson: Yeah, I am going to do some more testing to see if I can break this with a unit test.

validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -287,7 +286,8 @@ def __init__(self, squash_handler, first_state_hash, always_persist):
         # Transaction results
         self._txn_results = {}
 
-        self._transactions = OrderedDict()
+        self._txns_available = []
```
peterschwarz: This seems to be an addition of more parallel arrays/dictionaries, with respect to knowledge of the state a transaction as it moves through the schedule.  We should consider a refactor that collects the information about the state of a transaction (available, results etc) into a single object and a single store (either an array or map) of the transactions.  Any thing else is just a view of this data.

boydjohnson: That sounds like a good idea.

PR #842, opened 2017-09-12
Remove sleeps from integration tests

RyanLassigBanks:
Remove sleeps from integration tests

### DIFF
--- a/integration/sawtooth_integration/tests/test_dynamic_network.py
+++ b/integration/sawtooth_integration/tests/test_dynamic_network.py
@@ -48,8 +48,8 @@ class TestDynamicNetwork(unittest.TestCase):
             'rounds': 3,
             'start_nodes_per_round': 2,
             'stop_nodes_per_round': 0,
-            'batches': 5,
-            'time_between_batches': 1,
+            'batches': 3,
+            'time_between_batches': 0,
             'poet_kwargs': {
                 'minimum_wait_time': 1.0,
                 'initial_wait_time': 100.0,
--- a/integration/sawtooth_integration/tests/test_dynamic_network.py
+++ b/integration/sawtooth_integration/tests/test_dynamic_network.py
@@ -24,7 +24,7 @@ from sawtooth_intkey.client_cli.intkey_client import IntkeyClient
 
 LOGGER = logging.getLogger(__name__)
 LOGGER.setLevel(logging.INFO)
-WAIT = 300
+WAIT = 120
 
 
 class TestDynamicNetwork(unittest.TestCase):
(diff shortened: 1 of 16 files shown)


integration/sawtooth_integration/tests/test_events_and_receipts.py
```diff
@@ -180,13 +180,17 @@ def assert_unsubscribe_response(self, msg):
             events_pb2.ClientEventsUnsubscribeResponse.OK)
 
 class BatchSubmitter:
-    def __init__(self):
+    def __init__(self, timeout):
         self.batches = []
         self.imf = IntkeyMessageFactory()
+        self.timeout=timeout
```
boydjohnson: This is a small problem. The '=' should have whitespace on both sides. Our test files aren't linted, but it is best not to introduce more lint.

integration/sawtooth_integration/tests/test_poet_smoke.py
```diff
@@ -73,21 +74,18 @@ def test_poet_smoke(self):
 
         LOGGER.info('Sending populate txns')
         self.clients[0].send_txns(populate)
-        time.sleep(TIME_BETWEEN_BATCHES)
```
agunde406: If TIME_BETWEEN_BATCHES is no longer used, it should be deleted above.

integration/sawtooth_integration/tests/test_dynamic_network.py
```diff
@@ -181,8 +181,7 @@ def start_node(self, num,
                 raise subprocess.CalledProcessError(proc.pid, proc.returncode)
 
         self.nodes[num] = processes
-        self.clients[num] = IntkeyClient(
-            NodeController.http_address(num), WAIT)
+        self.clients[num] = IntkeyClient(NodeController.http_address(num))
```
aludvik: You should update the IntkeyClient to take a WAIT parameter instead of removing it. It should use that value as a timeout when it submits batches to the rest api, similar to how it is used in the `_post_batches()` functions.

integration/sawtooth_integration/tests/test_poet_smoke.py
```diff
@@ -127,28 +122,28 @@ def _assert_consensus(self):
...
-        # the most recent nth of the chain, at least 2 blocks
-        return max(
-            2,
-            length // 5)
+    # def calculate_tolerance(self):
+        # length = len(self.list_blocks())
+        # # the most recent nth of the chain, at least 2 blocks
+        # return max(
+            # 2,
+            # length // 5)
```
aludvik: Delete this

PR #844, opened 2017-09-13
STL-459 Seth Transaction Receipts

aludvik:
Would like to get an opinion from @zac how hard it would be to make it possible to wait for a receipts to appear. Hoping to get this merged without that for now though.

### DIFF
--- a/families/seth/src/sawtooth_seth/client/client.go
+++ b/families/seth/src/sawtooth_seth/client/client.go
@@ -189,15 +189,15 @@ func (c *Client) CreateContractAccount(priv []byte, init []byte, perms *EvmPermi
 // to the contract call.
 //
 // Returns the output from the EVM call.
-func (c *Client) MessageCall(priv, to, data []byte, nonce uint64, gas uint64, wait int, chaining_enabled bool) ([]byte, error) {
+func (c *Client) MessageCall(priv, to, data []byte, nonce uint64, gas uint64, wait int, chaining_enabled bool) (string, error) {
 	fromAddr, err := PrivToEvmAddr(priv)
 	if err != nil {
-		return nil, err
+		return "", err
 	}
 
 	toAddr, err := NewEvmAddrFromBytes(to)
 	if err != nil {
-		return nil, err
+		return "", err
 	}
 
 	var addresses []string
@@ -229,14 +229,12 @@ func (c *Client) MessageCall(priv, to, data []byte, nonce uint64, gas uint64, wa
 		},
 	}
 
-	body, err := c.sendTxn(transaction, encoder, wait)
+	txnId, err := c.sendTxn(transaction, encoder, wait)
 	if err != nil {
-		return nil, err
+		return "", err
 	}
-	logger.Debug(body)
 
-	// In the future, this will return EVM output from the transaction
-	return []byte{}, nil
+	return txnId, nil
 }
 
 // SetPermissions updates the permissions of the account at the given address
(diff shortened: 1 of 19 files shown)


families/seth/src/sawtooth_seth/client/client.go
```diff
@@ -189,15 +189,15 @@ func (c *Client) CreateContractAccount(priv []byte, init []byte, perms *EvmPermi
 // to the contract call.
 //
 // Returns the output from the EVM call.
-func (c *Client) MessageCall(priv, to, data []byte, nonce uint64, gas uint64, wait int, chaining_enabled bool) ([]byte, error) {
+func (c *Client) MessageCall(priv, to, data []byte, nonce uint64, gas uint64, wait int, chaining_enabled bool) (string, error) {
 	fromAddr, err := PrivToEvmAddr(priv)
 	if err != nil {
-		return nil, err
+		return "", err
```
peterschwarz: Is there no `nil` string?

validator/sawtooth_validator/server/core.py
```diff
@@ -178,7 +178,6 @@ def __init__(self, bind_network, bind_component, endpoint,
         block_store = BlockStore(block_db)
 
         batch_tracker = BatchTracker(block_store)
-        block_store.add_update_observer(batch_tracker)
```
delventhalz: Is it worth keeping the `add_update_observer` method and the `StoreUpdateObserver` class around if nothing uses it? They should probably be removed.

families/seth/src/sawtooth_seth/client/client.go
aludvik: Go defines `""` as the nil string..

aludvik:
@zac, I believe I addressed all your change requests, with the exception of renaming `/receipts`, which we can do in a later pr.

rest_api/openapi.yaml
```diff
@@ -385,6 +385,76 @@ paths:
...
+                $ref: "#/definitions/Link"
+        400:
+          $ref: "#/responses/400BadRequest"
+        500:
+          $ref: "#/responses/500ServerError"
+        503:
+          $ref: "#/responses/503ServiceUnavailable"
+
+    post:
+      summary: Fetches the committed statuses for a set of batches
```
delventhalz: Should read: `Fetches the receipts for a set of transactions`

rest_api/openapi.yaml
```diff
@@ -385,6 +385,76 @@ paths:
...
+      description: |
+        Identical to `GET /batch_status`, but takes ids of batches as a JSON
+        formatted POST body rather than a query parameter. This allows for many
+        more batches to be checked and should be used for more than 15 ids.
```
delventhalz: Should read:
```
Identical to `GET /receipts`, but takes ids of transactions as a JSON
formatted POST body rather than a query parameter. This allows for many
more receipts to be fetched and should be used with more than 15 ids.
```

rest_api/openapi.yaml
```diff
@@ -385,6 +385,76 @@ paths:
...
+
+        Note that since query information is not encoded in the URL, no `link`
+        will be returned with this query.
```
delventhalz: `this query.` -> `this request.`

This is my typo. But now it is in your PR.

rest_api/openapi.yaml
```diff
@@ -385,6 +385,76 @@ paths:
...
+      consumes:
+        - application/json
+      parameters:
+        - name: Batch Ids
```
delventhalz: `Batch` -> `Transaction`

rest_api/openapi.yaml
```diff
@@ -385,6 +385,76 @@ paths:
...
+          in: body
+          description: A JSON array of batch id strings
```
delventhalz: `batch` -> `transaction`

rest_api/openapi.yaml
```diff
@@ -385,6 +385,76 @@ paths:
...
+              type: string
+              example: 89807bfc9089e37e00d87d97357de14cfbc455cd608438d426a625a30a0da9a31c406983803c4aa27e1f32a3ff61709e8ec4b56abbc553d7d330635b5d27029c
+        - $ref: "#/parameters/wait"
+      responses:
+        200:
+          description: Successfully retrieved statuses
+          schema:
+            properties:
+              data:
+                $ref: "#/definitions/BatchStatuses"
```
delventhalz: -> `$ref: "#/definitions/TransactionReceipt"`

rest_api/openapi.yaml
```diff
@@ -592,6 +662,52 @@ definitions:
         type: string
         format: binary
 
+  TransactionReceipt:
```
delventhalz: This should be `TransactionReceipts` and wrapped in an array, since the route will return many receipts. 

Just replace this line with:
```
  TransactionReceipts:
    type: array
    items:
```
And then increase the indentation of everything else.

(also update the two `$ref` above)

PR #845, opened 2017-09-13
Adjust poet k (blockclaim limit) policy

dcmiddle:
Default was too small for launching small networks.
Larger value reduces risk of validators exhausting block claim allocations
and then being unable to readmit themselves.


dcmiddle:
@amundson it reduces risk because it allows for more time for other validators to join the network. if you burn through 25 or 50 blocks before another validator can join that's where you risk a deadlock.

dcmiddle:
rebased

PR #849, opened 2017-09-13
T&T /agents endpoint and DB testing

nick-drozd:
This PR adds the `/agents` endpoint and adds the server and database to the T&T test.

Two TODOs:

1. The assertion that checks the `/agents` endpoint isn't as rigorous as it could be, as it doesn't assert anything about the records agents report on. The desired TP behavior regarding reporter authorization and transfer of ownership hasn't been thought through very well, so I'd like to put off testing that until we figure out exactly what it should do.

2. Currently transactions are submitted to the rest api. When possible, I'd like to convert the test to submit transactions directly to the server and cut out the rest api entirely if possible.

### DIFF
--- a/families/track_and_trade/server/db/agents.js
+++ b/families/track_and_trade/server/db/agents.js
@@ -1,0 +1,103 @@
+/**
+ * Copyright 2017 Intel Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ----------------------------------------------------------------------------
+ */
+'use strict'
+
+const r = require('rethinkdb')
+
+const db = require('./')
+
+const MAX_BLOCK = Number.MAX_SAFE_INTEGER
+
+const hasCurrentBlock = currentBlock => obj => {
+  return r.and(
+    obj('startBlockNum').le(currentBlock),
+    obj('endBlockNum').gt(currentBlock)
+  )
+}
+
+const getAttribute = attr => obj => obj(attr)
+const getRecordId = getAttribute('recordId')
+const getPublicKey = getAttribute('publicKey')
+const getName = getAttribute('name')
+const getReporters = getAttribute('reporters')
+const getAuthorized = getAttribute('authorized')
+
+const hasPublicKey = key => obj => {
+  return r.eq(
+    key,
+    getPublicKey(obj)
+  )
+}
+
+const getAssociatedAgentId = role => record => record(role).nth(-1)('agentId')
+const getOwnerId = getAssociatedAgentId('owners')
+const getCustodianId = getAssociatedAgentId('custodians')
+
+const isAssociatedWithRecord = association => agent => record => {
+  return r.eq(
+    association(record),
+    getPublicKey(agent)
+  )
+}
+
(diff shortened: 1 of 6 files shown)


delventhalz:
Transaction submission just got merged. Rebase and you'll have a `/transactions` endpoint you can `POST` TransactionList bytes to. Make sure you set up a user and get an auth token, then send the request with the following headers:
```
Authorization: <the auth token from user creation>
Content-Type: 'application/octet-stream
```

families/track_and_trade/server/db/agents.js
```diff
@@ -0,0 +1,82 @@
...
+        agent('publicKey')
+      )
+
+const isRecordOwner = isAssociatedWithRecord(getOwnerId)
+const isRecordCustodian = isAssociatedWithRecord(getCustodianId)
+
+const isReporter = agent => property => {
+  return property('reporters')
+    .map(getPublicKey)
+    .contains(getPublicKey(agent))
```
delventhalz: This could be done in one iteration instead of two using `contains` with a predicate function. That would represent a pretty significant time savings, especially if the agent happens to be one of the first in the list.

More importantly, this function will erroneously return `true` when the Agent _was_ a reporter, but is no longer authorized. You need look at both public key and the authorized boolean.

families/track_and_trade/server/db/agents.js
```diff
@@ -0,0 +1,82 @@
...
+const getAgentsQuery = agents => {
+  return agents
+    .filter(hasMaxBlock)
+    .map(agent => r.expr({
+      'name': agent('name'),
+      'key': agent('publicKey'),
+      'owns': r.table('records')
+        .filter(hasMaxBlock)
+        .filter(isRecordOwner(agent))
+        .map(getRecordId)
```
delventhalz: This line could also be replaced with:
```
  .getField('recordId')
```

That's probably preferable going forward, but don't waste a bunch of time refactoring.

families/track_and_trade/server/db/agents.js
```diff
@@ -0,0 +1,82 @@
...
+        .filter(hasMaxBlock)
+        .filter(isRecordCustodian(agent))
+        .map(getRecordId)
+        .distinct(),
+      'reports': r.table('properties')
+        .filter(hasMaxBlock)
+        .filter(isReporter(agent))
+        .map(property => r.expr({
+          'recordId': getRecordId(property),
+          'property': getName(property)
```
delventhalz: In the specification, only the record id gets returned (just like owns and custodian). That is all the FishNet client needs to build out its views. Something like this might arguably be useful for a future client, but we should avoid redesigning and stick to the specs unless we have an urgent need to deviate.

integration/sawtooth_integration/docker/test_track_and_trade.yaml
```diff
@@ -26,6 +26,29 @@ services:
     command: track-and-trade-tp -vv tcp://validator:4004
     stop_signal: SIGKILL
 
+  rethinkdb:
+    image: rethinkdb
+    expose:
+      - 8080
+      - 28015
+
+  tt-server:
```
delventhalz: I'm not interested in going back and refactoring right now, but going forward we should be consistent about whether we are using `tt` or `tnt` as a prefix for Track and Trade resources. I've been doing one, and you've been doing the other. Do you have an opinion @peterschwarz?

families/track_and_trade/server/db/agents.js
```diff
@@ -0,0 +1,93 @@
...
+    .do(seq => r.branch(
+      seq.isEmpty(),
+      false,
+      getAuthorized(seq.nth(0))
+    ))
+}
+
+const getAgentsQuery = agents => {
+  return agents
+    .filter(hasMaxBlock)
```
peterschwarz: This has to query the block, and then query based on current block.  Otherwise, the query may result in agents that are in an inconsistent state from one block to the other.

integration/sawtooth_integration/docker/test_track_and_trade.yaml
peterschwarz: Personally, I'm inclined to lean towards `tnt`

PR #852, opened 2017-09-14
Add Validator Key Permissioning Doc

agunde406:
This includes the converted design doc for Validator Key Permissioning and how to set the roles. 

I do not think that all of this information belongs in the sysadmin section and would appreciate suggestion on how it should be split up.


docs/source/sysadmin_guide/validator_key_permissioning.rst
```diff
@@ -0,0 +1,336 @@
...
+To see the policy in state, run the following command:
+
+.. code-block:: console
+
+  $ sawtooth identity policy list
+  policy_1:
+    Entries:
+      PERMIT_KEY *
+
+Finally, run the following command to set the role for netowrk to permit all:
```
aludvik: typo, "network"

docs/source/sysadmin_guide/validator_key_permissioning.rst
```diff
@@ -0,0 +1,336 @@
...
+
+    $ sawtooth identity role list
+    transactor: policy_1
+
+
+The following is the suggested high-level role for on-chain validator network
+permissioning.
+
+network
+  If a validator receives a peer requests from a node whose public key is not
```
aludvik: typo "request"

docs/source/sysadmin_guide/validator_key_permissioning.rst
```diff
@@ -0,0 +1,336 @@
...
+
+  $ sawtooth config proposal create sawtooth.identity.allowed_keys=02b2be336a6ada8f96881cd55fd848c10386d99d0a05e1778d2fc1c60c2783c2f4
+
+Once your signer key is stored in the setting, the identity cli can be used to
+set and update roles and policies. Make sure that the identity transaction
+processor and the the rest api are running.
+
+.. code-block:: console
+
+  $ sawtooth identity policy create -h
```
aludvik: Rather than commit the usage here, these should be added to the CLI Reference section and have their usage generated automatically. A link to that page would go here then.

agunde406:
I believe we need both svg and pdf when adding an image so it will be build pdf and html correctly.

PR #855, opened 2017-09-14
Update Python SDK for receipts and events

aludvik:
This depends on #844

### DIFF
--- a/sdk/python/sawtooth_sdk/processor/state.py
+++ b/sdk/python/sawtooth_sdk/processor/state.py
@@ -14,6 +14,7 @@
 # ------------------------------------------------------------------------------
 from sawtooth_sdk.protobuf.validator_pb2 import Message
 from sawtooth_sdk.protobuf import state_context_pb2
+from sawtooth_sdk.protobuf import events_pb2
 from sawtooth_sdk.processor.exceptions import InvalidTransaction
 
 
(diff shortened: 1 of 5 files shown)


sdk/python/sawtooth_sdk/processor/state.py
```diff
@@ -112,3 +113,47 @@ def delete(self, addresses, timeout=None):
...
+        response = state_context_pb2.TpAddReceiptDataResponse()
+        response.ParseFromString(
+            self._stream.send(
+                Message.TP_ADD_RECEIPT_DATA_REQUEST,
+                request).result(timeout).content)
+        if response.status == state_context_pb2.TpAddReceiptDataResponse.ERROR:
+            raise InvalidTransaction(
+                "Failed to add receipt data: {}".format((data_type, data)))
+
+    def add_event(self, event_type, attributes=None, data=None, timeout=None):
```
agunde406: Docstring please!

agunde406:
I also suggest removing the rename of State->Contexts from this PR as there is probably a significant amount of documentation that would need to be reviewed and updated.

aludvik:
Added docstring to `add_event`. Passed on other commits and doc updating work to @RyanLassigBanks

aludvik:
Added tests

sdk/python/sawtooth_sdk/processor/state.py
```diff
@@ -112,3 +113,59 @@ def delete(self, addresses, timeout=None):
...
+        """Add a new event to the execution result for this transaction.
+
+        Args:
+            event_type (str): This is used to subscribe to events. It should be
+                globally unique and describe what, general, has occured.
```
boydjohnson: general -> in general ?

PR #856, opened 2017-09-15
Enforce rate limiting on Ping messages if the connection is not Authorized

agunde406:
Also fix a typo in gossip.

### DIFF
--- a/validator/sawtooth_validator/networking/handlers.py
+++ b/validator/sawtooth_validator/networking/handlers.py
@@ -14,6 +14,7 @@
 # ------------------------------------------------------------------------------
 import logging
 import os
+import time
 import netifaces
 
 import sawtooth_signing as signing
@@ -42,6 +43,7 @@ from sawtooth_validator.protobuf.authorization_pb2 import RoleType
 from sawtooth_validator.protobuf.network_pb2 import NetworkAcknowledgement
 from sawtooth_validator.protobuf.network_pb2 import DisconnectMessage
 from sawtooth_validator.protobuf.network_pb2 import PingRequest
+from sawtooth_validator.journal.timed_cache import TimedCache
 
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


validator/sawtooth_validator/networking/handlers.py
```diff
@@ -177,13 +179,44 @@ def handle(self, connection_id, message_content):
...
+
+            return HandlerResult(
+                HandlerStatus.RETURN,
+                message_out=ack,
+                message_type=validator_pb2.Message.NETWORK_ACK)
+
+        if connection_id in self._last_message:
+            if time.time() - self._last_message[connection_id] < \
+                    self._allowed_frequency:
+                LOGGER.debug("To many Pings in %s seconds before authorization"
```
aludvik: typo, "Too"

validator/sawtooth_validator/networking/handlers.py
```diff
@@ -177,13 +179,45 @@ def handle(self, connection_id, message_content):
...
+
+            if connection_id in self._last_message:
+                del self._last_message[connection_id]
+
+            return HandlerResult(
+                HandlerStatus.RETURN,
+                message_out=ack,
+                message_type=validator_pb2.Message.NETWORK_ACK)
+
+        if connection_id in self._last_message:
```
peterschwarz: So this will automatically consider them unauthed after a second ping if it arrives in to small a window? (Obviously, in the case where they haven't authed)

agunde406: Correct. This is done to keep with the them of being very strict about correct behavior during the authorization period. If you think this is too strict, I could add a counter that allows them to send pings to close together up to a certain limit and then close the connection. Thoughts?

peterschwarz: I think that it's fine, just wanted to double check on the intent

PR #857, opened 2017-09-15
Python SDK rename state context

RyanLassigBanks:


sdk/python/sawtooth_sdk/processor/context.py
```diff
@@ -112,3 +112,7 @@ def delete(self, addresses, timeout=None):
             raise InvalidTransaction(
                 'Tried to delete unauthorized address: {}'.format(addresses))
         return response.addresses
+
+    get = get_state
```
boydjohnson: Were these notes?

boydjohnson: Nevermind. Though this doesn't seem like the greatest idea since `set` is a python builtin.

peterschwarz: Make sure that thesee deprecated methods are documented.

aludvik: @boydjohnson it's not overwriting python's set builtin method any more than it was before when we did `def set()`. This is just a more efficient way to make `set_state` and `set` equivalent.

PR #858, opened 2017-09-15
Implement fail-fast in the parallel scheduler

boydjohnson:
In this PR, I chose to fail-fast only at the point that all prior batches have results. If we didn't do it this way the replay mechanism would have to be more complicated, possibly replaying an entire batch instead of just replaying transactions.

Also the yaml scheduler state assertions must be relaxed because a txn may be replayed and then failed-fast leaving in the yaml scheduler tester the state of the prior txn. So currently what is being asserted in the yaml tests are batch validity, that each address has the correct value at the start of the txn for valid batches, and that the state hashes calculated are correct.


validator/tests/test_scheduler/data/complex_dependency_failures.yaml
```diff
@@ -0,0 +1,551 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# -----------------------------------------------------------------------------
+
+---
+-  #------------------------- Batch 1
```
peterschwarz: Are there any notes about these batches that can make it more informative about what each one is testing, or if they are all related to one test?

validator/tests/test_scheduler/test_schedulers_with_yaml.py
```diff
@@ -434,6 +434,53 @@ def test_deletes(self):
             context_manager=context_manager,
             name='deletes_test.yaml')
 
+    def test_complex_dependency_failures(self):
+        """Tests the schedulers against the
+        test_scheduler/data/deletes_test.yaml file.
```
peterschwarz: Wrong yaml file mentioned

validator/tests/test_scheduler/data/complex_dependency_failures.yaml
boydjohnson: Yes, I could add notes at the top of each yaml file that explain the emphasis of each test.

boydjohnson: Do you have thoughts on how to be more informative?

peterschwarz: At least highlight where the poison pill is - i.e. which batch/txn is bad in the group.

boydjohnson: Yeah, I will put more notes in the yaml about what is tricky or hard for the scheduler.

PR #860, opened 2017-09-15
Implement IdentityObserver and Update PermissionVerifier to use a cache

agunde406:


validator/sawtooth_validator/gossip/permission_verifier.py
```diff
@@ -363,3 +363,52 @@ def handle(self, connection_id, message_content):
...
+        if value == -1 or value is None:
+            if self._identity_view is None:
+                self.update_view(state_root)
+            value = self._identity_view.get_policy(item)
+            self._cache[item] = value
+        return value
+
+    def forked(self):
+        for key in self._cache:
+            self._cache[key] = -1
```
aludvik: Could None be used here instead of -1? Or could you just delete the key from the dictionary?

validator/sawtooth_validator/gossip/identity_observer.py
```diff
@@ -0,0 +1,77 @@
...
+        For "identity_update", the corresponding cache entry will be updated.
+        """
+
+        block_events = BlockEventExtractor(block).extract([
+            EventSubscription(event_type="block_commit")])
+        receipt_events = ReceiptEventExtractor(receipts).extract([
+            EventSubscription(event_type="identity_update")])
+
+        for event in block_events:
+            if event.event_type == "block_commit":
```
aludvik: This check is unnecessary because you are only subscribed to block_commit events (assuming the extractor works correctly..).

PR #863, opened 2017-09-18
Removing sleeps

RyanLassigBanks:


integration/sawtooth_integration/tests/test_dynamic_network.py
```diff
@@ -205,16 +205,12 @@ def stop_node(self, num):
...
-                return
-
-            time.sleep(sleep_time)
-            sleep_time *= growth_rate
+        timeout = ASSERT_CONSENSUS_TIMEOUT + time.time()
+        in_consensus = False
 
-        self.assertTrue(self.in_consensus())
+        while not in_consensus and time.time() < timeout:
+            in_consensus = self.in_consensus()
```
nick-drozd: `in_consensus` includes a call to `sawtooth block list`. If the validators aren't in consensus, how many times will `in_consensus` be called, and how frequently?

integration/sawtooth_integration/tests/test_poet_smoke.py
```diff
@@ -85,17 +76,22 @@ def test_poet_smoke(self):
...
-            time.sleep(TRY_AGAIN_TIME)
-            self._assert_consensus()
+        time_out = time.time() + WAIT
+
+        while time.time() < time_out:
+            try:
+                self._assert_consensus()
+                return
+            except AssertionError:
+                LOGGER.info('Blocks not yet in concesus')
```
vaporos: Typo 'concesus' should be 'consensus'

vaporos: Suggest rewriting this loop so the logging message has more information (ideally, a list of the block ids at that point and the time which has passed).

integration/sawtooth_integration/tests/test_dynamic_network.py
RyanLassigBanks: After running the tests a few times, it looks like in_consensus is called 1 to 6 times per call to assert_consensus. It doesn't seem to be affecting test stability, but we could add some sleeps back in to throttle it, if that's what you're worried about.

integration/sawtooth_integration/tests/test_poet_smoke.py
RyanLassigBanks: Fixed type

RyanLassigBanks: added some logging to communicate how much time has passed. `_assert_consensus` already has logging to list the block ids

PR #868, opened 2017-09-18
Base Rust SDK With Initial ZMQ, Signing and Protobuf Support

peterschwarz:


peterschwarz:
dup of successfully building #871

PR #869, opened 2017-09-19
Update log config example file to have full path

agunde406:


vaporos:
Is the intent to make it easier to start using the file?  If so, suggest also removing the id from the log filename (currently validator-xxxx-debug.log, suggesting validator-debug.log) and changing the filenames in this example file to match as well.

The original reason for including the id in the log file name was to support running multiple validators within the same environment.  This now seems like an edge case, given our use of docker and debian-based vm installs.

PR #870, opened 2017-09-19
Update endpoint authorization log message to include connection_id

agunde406:

This should make it easier to read the logs. Also update log message on authorization completion to show the endpoint  and the corresponding connection id.


agunde406:
@vaporos Okay. It was based on comments @peterschwarz had made when we were trying to debug a peering issue. I updated it to only update the endpoint authorization message.

PR #872, opened 2017-09-19
STL-572 Initial Seth RPC Implementation

aludvik:
Creates the core request handling part of the rpc server and creates empty methods for all the calls that need to be implemented. Implements a few methods as well. Unit tests to follow later, but you can test manually with:

`./bin/build_all -l python -l rust`
`docker-compose -f docker/compose/sawtooth-seth-rpc.yaml`
`curl -d '{"jsonrpc": "2.0", "method": "eth_blockNumber", "id": 1}' -H "Content-Type: application/json" 0.0.0.0:3030`

### DIFF
--- a/families/seth/rpc/src/block.rs
+++ b/families/seth/rpc/src/block.rs
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2017 Intel Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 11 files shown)


families/seth/rpc/src/block.rs
```diff
@@ -16,13 +16,79 @@
...
+use sawtooth_sdk::messaging::zmq_stream::*;
+use sawtooth_sdk::messaging::stream::*;
+
+use sawtooth_sdk::messages::client::{
+    ClientBlockListRequest, ClientBlockListResponse, PagingControls,
+};
+use sawtooth_sdk::messages::block::BlockHeader;
+use sawtooth_sdk::messages::validator::Message_MessageType;
 
 pub fn block_number(_params: Params, mut sender: ZmqMessageSender) -> Result<Value, Error> {
```
peterschwarz: Could you add a quite comment on what this does?  Is it "get a block by block number"?

families/seth/rpc/src/main.rs
```diff
@@ -0,0 +1,123 @@
...
+mod network;
+mod transaction;
+
+use requests::{RequestExecutor, RequestHandler};
+
+const SERVER_THREADS: usize = 3;
+
+fn main() {
+    let arg_matches = clap_app!(intkey =>
+        (version: "1.0")
```
vaporos: You can use a cargo/crate macro version here so it is pulled from Cargo.toml.

PR #873, opened 2017-09-19
Transactor Permissioning Integration Test

boydjohnson:


integration/sawtooth_integration/tests/test_identity.py
```diff
@@ -0,0 +1,399 @@
...
+
+from sawtooth_processor_test.message_factory import MessageFactory
+
+from sawtooth_integration.tests.integration_tools import RestClient
+from sawtooth_integration.tests.integration_tools import wait_for_rest_apis
+
+from sawtooth_sdk.protobuf import batch_pb2
+
+
+class TestTransactorPermissioning(unittest.TestCase):
```
nick-drozd: This file should be named `test_transactor_permissioning.py` to be consistent with the test class and with the associated yaml file.

integration/sawtooth_integration/tests/test_identity.py
```diff
@@ -0,0 +1,399 @@
...
+
+    def setUp(self):
+        self.REST_ENDPOINT = "rest-api:8080"
```
nick-drozd: Only one rest api url is used in this test, right? If so, I would suggest making the url a global constant rather than passing it around as an argument and assigning it as an attribute to all the transactors.

In any case, the `wait_for_rest_apis` call should be separated from the Transactor creation. The usual approach has been to put it in `setUpClass`, since with rest api only has to be waited for once per test.

integration/sawtooth_integration/tests/test_identity.py
```diff
@@ -0,0 +1,399 @@
...
+    def __init__(self, name, rest_endpoint):
+        """
+        Args:
+            name (str): An identifier for this Transactor
+            rest_endpoint (str): The rest api that this Transactor will
+                communicate with.
+        """
+
+        self.name = name
+        self._rest_endpoint = rest_endpoint \
```
nick-drozd: `wait_for_rest_apis` incorrectly expects its argument to not have `http://` and adds it. This should be changed, so test logic shouldn't be built around that behavior. The pattern I've been using is to have to variables, `REST_API = rest-api:8080` and `URL = 'http://' + REST_API`, with `REST_API` being used for `wait_for_rest_apis` and `URL` being used for everything else.

sdk/python/sawtooth_processor_test/message_factory.py
```diff
@@ -176,7 +177,7 @@ def create_tp_process_request(self, payload, inputs, outputs, deps,
             payload=payload,
             signature=signature)
 
-    def create_batch(self, transactions):
+    def create_batch_object(self, transactions):
```
peterschwarz: Why the name change?

integration/sawtooth_integration/tests/test_identity.py
```diff
@@ -0,0 +1,399 @@
...
+        self.mallory = Transactor('mallory', self.REST_ENDPOINT)
+
+        self.walter = Transactor('walter', self.REST_ENDPOINT)
+
+    def test_transactor_permissioning(self):
+        """Test the transactor permissioning system using the Identity
+        transaction family.
+
+        Notes:
+            The test works from general to specific on the transactor
```
peterschwarz: Given that these are basically separate tests and all the users are created in setup, does it makes sense to break these into individual test functions, `test_transactor_broad_permissioning`, `test_transactor_batch_signer_persmissioning`, etc

integration/sawtooth_integration/tests/test_identity.py
```diff
@@ -0,0 +1,399 @@
...
+    Families.IDENTITY: {
+        'encoding': 'application/protobuf',
+        'family_name': 'identity',
+        'family_version': '1.0',
+        'namespace': '00001d'
+    }
+}
+
+
+def INTKEY_PAYLOAD(unique_value):
```
nick-drozd: These are functions, not constants, so they should be named something like `make_intkey_payload`, etc.

integration/sawtooth_integration/tests/test_identity.py
```diff
@@ -0,0 +1,399 @@
...
+
+        return self._factories[family_name].create_transaction(
+            payload=payload,
+            inputs=[address],
+            outputs=[address],
+            deps=[],
+            batcher=batcher)
+
+    def create_batch(self, family_name):
+        transactions = [self.create_txn(family_name) for _ in range(1)]
```
nick-drozd: `range(1)` should be `range(count)` for an argument `count`, or else just make this `[self.create_txn(family_name]`.

integration/sawtooth_integration/tests/test_identity.py
```diff
@@ -0,0 +1,399 @@
...
+        # From local configuration Dave is denied from being a batch_signer,
+        # and Chuck and Mallory are denied from all transactor permissions.
+
+        self.walter.set_pubkey_for_role(
+            'deny_bob_allow_alice_walter',
+            'transactor',
+            permit_keys=[self.alice.public_key, self.walter.public_key],
+            deny_keys=[self.bob.public_key])
+
+        for t, f in [(self.alice, Families.INTKEY),
```
nick-drozd: I'd prefer if `t` and `f` were written out in full (the list pairs can be indented if needed).

nick-drozd: Actually, it might be better to have the assertions expect a variable number of pairs, so the test body wouldn't have the loops at all. Then the assertions would look like
```
self.assert_able_to_send(
    (self.alice, Families.INTKEY),
    (self.walter, Families.INTKEY),
    ...,
)

File: integration/sawtooth_integration/tests/test_identity.py [hunk 9]
```diff
@@ -0,0 +1,399 @@
...
+            permit_keys=[self.alice.public_key, self.dave.public_key],
+            deny_keys=[self.bob.public_key])
+
+        self.assert_able_to_send(self.alice, Families.INTKEY)
+
+        for t, f in [(self.bob, Families.INTKEY), (self.chuck, Families.INTKEY),
+                     (self.mallory, Families.INTKEY), (self.dave, Families.INTKEY)]:
+            self.assert_not_able_to_send(t, f)
+
+    def assert_able_to_send(self, transactor, family):
```
nick-drozd: How does this assertion fail?

nick-drozd: Same for `assert_able_to_send_batch`.

integration/sawtooth_integration/tests/test_identity.py
```diff
@@ -0,0 +1,399 @@
...
+        transactor.send(family)
+
+    def assert_able_to_send_batch(self, transactor, family, transactions):
+        transactor.send(family_name=family, transactions=transactions)
+
+    def assert_not_able_to_send(self, transactor, family):
```
nick-drozd: `Exception` should be more specific if possible.

integration/sawtooth_integration/tests/test_identity.py
boydjohnson: I had worried about the state of the validator at the end of one test affecting another test.

integration/sawtooth_integration/tests/test_scripts/local_transactor_permissioning.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
```
nick-drozd: If this really, truly, absolutely cannot be put into the yaml file, then the test file should announce prominently that it expects this script to be run and what it does and the script should be named the same as the test file and yaml file (but with the `.sh` suffix).

I also don't think it should go into a subdirectory; `/test_scripts` suggests to me that these are scripts that maybe run for every test or something like that.

sdk/python/sawtooth_processor_test/message_factory.py
boydjohnson: I had thought that I needed to get the batch id to get a status, but I am getting the '--wait' from the identity cli change. The change you called out is unnecessary.

integration/sawtooth_integration/tests/test_identity.py
boydjohnson: Updating.

integration/sawtooth_integration/tests/test_identity.py
boydjohnson: Updating.

integration/sawtooth_integration/tests/test_identity.py
boydjohnson: Updating.

integration/sawtooth_integration/tests/test_identity.py
boydjohnson: I will go with the former, with a default variable.

integration/sawtooth_integration/tests/test_scripts/local_transactor_permissioning.sh
boydjohnson: I will do the renaming. I couldn't get it to work from within a bash -c command.

integration/sawtooth_integration/tests/test_identity.py
boydjohnson: Updating.

integration/sawtooth_integration/tests/test_identity.py
boydjohnson: RestClient in integration_tools.py transforms any non 200, 201, or 202 http status into an Exception. https://github.com/hyperledger/sawtooth-core/blob/master/integration/sawtooth_integration/tests/integration_tools.py#L107. Failures from the permission_verifier for transactors get a 400.

integration/sawtooth_integration/tests/test_identity.py
boydjohnson: It fails with an error if the rest api returns a 400 if transactor permissions don't work out.

integration/sawtooth_integration/tests/test_identity.py
nick-drozd: Hmm, that's not great. Is it possible to get the specific code out of the exception? Alternatively, you might mess around with `RestClient` to get it to do something more convenient.

PR #876, opened 2017-09-19
Warn if dispatch queue is too large

peterschwarz:
In order to figure out if the incoming message queue is becoming saturated, log the size of the queue after a there are more than 10 messages backed up.


validator/sawtooth_validator/networking/dispatch.py
```diff
@@ -119,6 +119,11 @@ def dispatch(self, connection, message, connection_id):
                     self._msg_type_handlers[message.message_type])
             )
             self._in_queue.put_nowait(message_id)
+
+            queue_size = self._in_queue.qsize()
+            if queue_size > 10:
+                LOGGER.warning("Dispatch incoming queue size: %s",
```
jsmitchell: Seems like a debug level message

PR #878, opened 2017-09-19
Add /agents/:publicKey endpoint

nick-drozd:


families/track_and_trade/server/db/agents.js
```diff
@@ -95,8 +95,33 @@ const listQuery = queryWithCurrentBlock(currentBlock => {
...
 })
 
+const fetchQuery = publicKey => queryWithCurrentBlock(currentBlock => {
+  return getTable('agents', currentBlock)
+    .filter(hasPublicKey(publicKey))
+    .pluck('name', 'publicKey')
+    .nth(0)
+})
+
+const fetchAuthQuery = publicKey => queryWithCurrentBlock(currentBlock => {
```
delventhalz: A better way to break this up would be to have two queries `agents.fetch` and `users.fetch`, and `api/users` contains the knowledge to make both queries and combine them. This would both keep your code DRY, and better separate concerns. `db/agents` wouldn't have to have any concept of authorization or a 'users' table.

Then `api/users` could combine these queries just by using `Promise.all` to make each separately. This might be _slightly_ less performant than what you have here, but if that is a concern you could also create a utility function in `db/index` that combines multiple queries, returning the results in an array. Then `api/users` would merge the two results.

Anyway. If this is a quick refactor, go for it, but don't let this hold up the PR if it's going to take too long.

families/track_and_trade/server/api/index.js
```diff
@@ -144,6 +143,8 @@ router.get('/info', handle(() => {
     }))
 }))
 
+router.post('/authorization', handleBody(auth.authorize))
```
delventhalz: These endpoints should stay in alphabetical order and should not be rearranged.

PR #880, opened 2017-09-20
Old style exceptions --> new style for Python 3

cclauss:
This PR assures that exceptions are compatible with both Python 2 and Python 3.


vaporos:
These files are from 0.7 and have been removed from the repository entirely, so closing this PR.  Thanks!

PR #881, opened 2017-09-20
Add in missing imports

cclauss:
Added missing imports:
* import getpass
* from  base64 import b64encode


dcmiddle:
@peterschwarz can you take a look at this? Looks like you added the code requiring the imports.

peterschwarz:
Probably need to add these files to the pylint paths.

cclauss:
These are standard library imports.  I would hope that if they are imported then pylint should be able to find them without further hinting.

peterschwarz:
I don't mean the libraries imported, I'm referring to the portion of the
sawtooth source tree.  This sub path isn't part of the `run_lint` script

cclauss:
Closed in favor of #987

PR #882, opened 2017-09-20
Remove unused method _generate_id()

cclauss:
string is used on line 97 but is never imported.

### DIFF
--- a/validator/tests/test_completer/test.py
+++ b/validator/tests/test_completer/test.py
@@ -15,6 +15,7 @@
 import unittest
 import random
 import hashlib
+import string
 import cbor
 
 import sawtooth_signing as signing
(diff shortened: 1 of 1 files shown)


dcmiddle:
@agunde406 did this import get by because we don't use that _generate_id method in the tests?
If not maybe we should cut that method rather than add the import?

agunde406:
Ya it doesn't look like that method is used, it should be removed.

dcmiddle:
Regarding the sign-off. You might find it easier to amend your commit and then rebase. Then you don't need to close this PR like you did with the other contributions.
The below works if you named this remote "upstream".
```git fetch upstream master
git rebase -i upstream/master
git push <yourfork> <thisbranch> -f ```

cclauss:
Closed in favor of #1002

PR #887, opened 2017-09-20
Metrics for Rest API

RyanLassigBanks:


docker/compose/smallbank-local-mounted.yaml
```diff
@@ -69,7 +69,7 @@ services:
       - "8080:8080"
     depends_on:
      - validator
-    command: sawtooth-rest-api -v --connect tcp://validator:4004 --bind rest-api:8080
+    command: sawtooth-rest-api -v --connect tcp://validator:4004 --bind rest-api:8080 --open_tsdb_url http://influxdb:8086 --open_tsdb_db metrics
```
vaporos: Command line arguments should never have underscores, and OpenTSDB seems to be one word.  Use dashes instead - "--opentsdb-url" and "--opentsdb-db".

docker/sawtooth-dev-python
```diff
@@ -86,8 +86,9 @@ RUN echo "deb http://repo.sawtooth.me/ubuntu/ci xenial universe" >> /etc/apt/sou
  && pip3 install \
     pylint \
     bandit \
-    pyformance \
-    coverage --upgrade
+    coverage --upgrade \
+&& pip3  install git+https://github.com/omergertel/pyformance.git \
```
vaporos: We will need to package this as a deb package since this is a runtime dependency.

RyanLassigBanks:
No longer relevant

PR #893, opened 2017-09-21
Add T&T records endpoints

nick-drozd:
The query is a little complex at one point. Is there a cleaner way to do `do` scoping? Is there any other way to assign variables in Reql?

### DIFF
--- a/families/track_and_trade/server/db/records.js
+++ b/families/track_and_trade/server/db/records.js
@@ -1,0 +1,144 @@
+/**
+ * Copyright 2017 Intel Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ----------------------------------------------------------------------------
+ */
+'use strict'
+
+const r = require('rethinkdb')
+
+const db = require('./')
+
+/* Helpers */
+
+const getAttribute = attr => obj => obj(attr)
+const getRecordId = getAttribute('recordId')
+const getName = getAttribute('name')
+const getPublicKey = getAttribute('publicKey')
+const getDataType = getAttribute('dataType')
+const getReporters = getAttribute('reporters')
+const getAuthorization = getAttribute('authorized')
+const getReportedValues = getAttribute('reportedValues')
+
+const getAssociatedAgentId = role => record => record(role).nth(-1)('agentId')
+const getOwnerId = getAssociatedAgentId('owners')
+const getCustodianId = getAssociatedAgentId('custodians')
+
+const hasAttribute = getAttr => attr => obj => r.eq(attr, getAttr(obj))
+const hasName = hasAttribute(getName)
+const hasRecordId = hasAttribute(getRecordId)
+const hasPublicKey = hasAttribute(getPublicKey)
+
+const hasBlock = block => obj => {
+  return r.and(
+    obj('startBlockNum').le(block),
+    obj('endBlockNum').gt(block)
+  )
+}
+
+const queryWithCurrentBlock = query => {
+  return r.table('blocks')
+    .orderBy(r.desc('blockNum'))
+    .nth(0)('blockNum')
+    .do(query)
+}
+
+const getTable = (tableName, block) =>
+      r.table(tableName).filter(hasBlock(block))
+
+const findRecord = recordId => block => {
+  return getTable('records', block)
+    .filter(hasRecordId(recordId))
+    .nth(0)
+}
+
+const findProperty = recordId => propertyName => block => {
+  return getTable('properties', block)
+    .filter(hasRecordId(recordId))
+    .filter(hasName(propertyName))
+    .nth(0)
+}
+
+const getAuthorizedReporterKeys = property => {
+  return getReporters(property)
+    .filter(getAuthorization)
+    .map(getPublicKey)
+    .coerceTo('array')
+}
+
+const getReporter = publicKey => block => {
+  return getTable('agents', block)
+    .filter(hasPublicKey(publicKey))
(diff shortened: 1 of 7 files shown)


delventhalz:
@nick-drozd For the most part, you can save variables normally and REQL will reference them. But `do` let's you keep everything in the same query, so that may or may not work. I'll keep my eye out for any obvious refactors.

integration/sawtooth_integration/tests/test_track_and_trade.py
```diff
@@ -602,6 +607,38 @@ def test_track_and_trade(self):
...
+        log_json(get_record_property)
+
+        self.assertIn('dataType', get_record_property)
+        self.assertIn('name', get_record_property)
+        self.assertIn('recordId', get_record_property)
+        self.assertEqual(get_record_property['dataType'], 'INT')
+        self.assertEqual(get_record_property['name'], 'temperature')
+        self.assertEqual(get_record_property['recordId'], 'fish-456')
+
+        # Timestamps and public keys differ for each run of the test,
```
peterschwarz: I would just assert that the timestamps are set - a minimal test that the api is providing the information.

families/track_and_trade/server/db/records.js
```diff
@@ -0,0 +1,148 @@
...
+
+const getAssociatedAgentId = role => record => record(role).nth(-1)('agentId')
+const getOwnerId = getAssociatedAgentId('owners')
+const getCustodianId = getAssociatedAgentId('custodians')
+
+const hasName = name => obj => r.eq(name, getName(obj))
+const hasRecordId = recordId => obj => r.eq(recordId, getRecordId(obj))
+const hasPublicKey = publicKey => obj => r.eq(publicKey, getPublicKey(obj))
+
+// const hasAttribute = getAttr => attr => obj => r.eq(attr, getAttr(obj))
```
peterschwarz: Remove commented code.

families/track_and_trade/server/db/records.js
```diff
@@ -0,0 +1,148 @@
...
+  })
+}
+
+/* Queries */
+
+const fetchPropertyQuery = (recordId, propertyName) => {
+  return queryWithCurrentBlock(block => {
+    return findRecord(recordId)(block).do(record => {
+      return findProperty(recordId)(propertyName)(block).do(property => {
+        return findReportedValues(recordId)(propertyName)(getDataType(property))(getAuthorizedReporterKeys(property))(block).do(values => {
```
delventhalz: So, what you can do instead of these nested `do` statements, is just make your queries in an array you return. 
```
return queryWithCurrentBlock(block => {
  return [
    findRecord(recordId)(block),
    findProperty(recordId)(propertyName)(block),
    . . .
  ]
}).do(([ record, property, . . . ]) => {
  return r.expr({
    name: getName(property),
    recordId: getRecordId(record),
    . . .
  })
})
```

families/track_and_trade/server/db/records.js
```diff
@@ -0,0 +1,148 @@
...
+    r.eq(dataType, 'LOCATION'), value('locationValue'),
+    'intValue'
+  )
+}
+
+const getUpdate = dataType => reporterKeys => block => value => {
+  return r.expr({
+    'value': getValue(dataType)(value),
+    'timestamp': value('timestamp'),
+    'reporter': getReporter(reporterKeys.nth(value('reporterIndex')))(block)
```
delventhalz: You don't need quotes around the keys in JS objects unless they include special characters like `-` or ` `. JS will convert them to a string automatically.

nick-drozd: I come from languages with a clear and sensible distinction between strings and evaluated expressions, so that kind of thing makes me uncomfortable. If it's equivalent, I would prefer to leave the quotation marks.

nick-drozd:
The `/records` and `/records/:recordId` endpoints differ slightly from what's in the design doc.

1. The doc has
```
      "properties": {
        <property name>: [
          {
            "timestamp": <integer, seconds>,
            "value": <a value matching the property's "type">
          }
        ]
      }
    },
```
whereas these endpoints have
```
test-track-and-trade_1  |             "properties": [
test-track-and-trade_1  |                 {
test-track-and-trade_1  |                     "name": "species",
test-track-and-trade_1  |                     "values": [
test-track-and-trade_1  |                         {
test-track-and-trade_1  |                             "timestamp": 1506020855,
test-track-and-trade_1  |                             "value": "trout"
test-track-and-trade_1  |                         }
test-track-and-trade_1  |                     ]
test-track-and-trade_1  |                 },
```
This is because I couldn't figure out how to the key to evaluate rather than get converted to a string. 

"But haven't you heard the good news?" you say. "ES6 has _computed property keys_. Just wrap the key in brackets and it will evaluate as usual."

I tried that, but Reql mangles the names:
```
test-track-and-trade_1  |         "properties": [
test-track-and-trade_1  |             {
test-track-and-trade_1  |                 "var_169(\"name\")": [
test-track-and-trade_1  |                     {
test-track-and-trade_1  |                         "timestamp": 1506010634,
test-track-and-trade_1  |                         "value": "trout"
test-track-and-trade_1  |                     }
test-track-and-trade_1  |                 ]
test-track-and-trade_1  |             },
```
Anyway, if there's an easy way to fix this, I'll do it, but I don't think it's a huge difference.

families/track_and_trade/server/db/records.js
peterschwarz: You don't need it, but it's a stylistic difference that fine.

peterschwarz:
I think, instead of: 
```
...
'properties': propertyValues
              .map(propertyValue => r.expr({
                'name': getName(propertyValue),
                'values': propertyValue('values').pluck('value', 'timestamp')
              }))
...
```
you could do
```
...
'properties': propertyvalues.reduce((properties, propertyValue) => {
   properties[getName(propertyValue)] = propertyValue('values').pluck('value', 'timestamp')
}).default({})
...
```

delventhalz:
@nick-drozd The pre-ES6 way to use an evaluated key is to do it in two statements. One creating the object, and the other setting the key. In other words:
```
const newObj = (key, value) => {
  const obj = {}
  obj[key] = value
  return obj
}
```

Hideous, I know, but ES6 niceties mostly don't seem to work in ReQL. You should be able to abstract this into a function or some such.

delventhalz:
@nick-drozd Seems to work well, but I would like to see the properties updated to match the spec if possible.

peterschwarz:
I agree with @delventhalz about the properties

PR #897, opened 2017-09-21
Properly copy sets as lists

peterschwarz:
To avoid concurrent modifications of the `all_tasks` set, wrap the call in a list before iterating.


peterschwarz:
Note, this is the proper way to fix the same problem being fixed in #888

PR #899, opened 2017-09-22
Merge smallbank into master

vaporos:


vaporos:
In addition to the previously-reviewed content in the smallbank branch, this includes 3 new commits.

PR #901, opened 2017-09-22
Add Pyformance deb creation to build_ext_debs

RyanLassigBanks:

### DIFF
--- a/bin/build_ext_debs
+++ b/bin/build_ext_debs
@@ -154,6 +154,7 @@ build_python_debs() {
     pytest==2.9.0
     cryptography==1.7.2
     lmdb==0.92
+    git+git://github.com/omergertel/pyformance.git@de7bee8ceddbb1c3ea3b6c28373a5b066b04b3b7
     '
 
     # secp256k1 needs a couple of libraries not readily available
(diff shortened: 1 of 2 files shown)


bin/build_ext_debs
```diff
@@ -132,36 +132,35 @@ build_python_debs() {
...
-    lmdb==0.92
-    '
-
-    # secp256k1 needs a couple of libraries not readily available
-    libsecp256k1_debs='
-    libsecp256k1-0_0.1~20161228-1_amd64.deb
-    libsecp256k1-dev_0.1~20161228-1_amd64.deb
-    '
-
+	pip_pkgs='
```
vaporos: Need to replace tabs with spaces.  Most of this chunk of the diff shouldn't change.

bin/build_ext_debs
```diff
@@ -217,9 +216,22 @@ package_python_deb() {
             --disable-pip-version-check \
             --no-deps $pip_pkg_name
 
-        tarball=$tar_pkg_name*.tar.gz
+		if [[ $pip_pkg_name == *".git"* ]]
```
vaporos: Is this pyformance specific? That might be obvious later.  Maybe more specific match or a comment.

ci/sawtooth-build-debs
```diff
@@ -83,7 +84,8 @@ RUN echo "deb http://repo.sawtooth.me/ubuntu/ci xenial universe" >> /etc/apt/sou
     unzip \
     wget \
  && apt-get clean \
- && rm -rf /var/lib/apt/lists/*
+ && rm -rf /var/lib/apt/lists/* \
+ && pip3 install mock --upgrade
```
vaporos: Can we use the python3-mock package instead (if this exists...)?

bin/build_ext_debs
RyanLassigBanks: Fair point since this is a "quick fix" I've changed it check for "pyformance" instead

ci/sawtooth-build-debs
RyanLassigBanks: I don't think there is one. The reason for this install is because in Python 3.3+ mock is a standard library and the import is slightly different. Something like `import unittest.mock` instead of `import mock`

PR #913, opened 2017-09-22
Add metrics to validator to measure the size

RyanLassigBanks:
of the pending batch queue over time

### DIFF
--- a/validator/sawtooth_validator/journal/publisher.py
+++ b/validator/sawtooth_validator/journal/publisher.py
@@ -335,7 +335,8 @@ class BlockPublisher(object):
                  data_dir,
                  config_dir,
                  permission_verifier,
-                 batch_injector_factory=None):
+                 batch_injector_factory=None,
+                 metrics_registry=None):
         """
         Initialize the BlockPublisher object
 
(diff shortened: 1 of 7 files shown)


validator/sawtooth_validator/server/cli.py
```diff
@@ -291,6 +304,20 @@ def main(args=None):
...
 
+    metrics_reporter = None
+    if validator_config.opentsdb_url:
+        url = urlparse(validator_config.opentsdb_url)
+        proto, db_server, db_port, = url.scheme, url.hostname, url.port
+
+        registry = MetricsRegistry()
+        metrics_reporter = InfluxReporter(
+            registry=registry,
+            reporting_interval=1,
```
jsmitchell: Probably should default to 10 secs.

validator/sawtooth_validator/journal/publisher.py
```diff
@@ -379,6 +383,11 @@ def __init__(self,
         self._permission_verifier = permission_verifier
         self._batch_injector_factory = batch_injector_factory
 
+        # For metric gathering
+        self._pending_batch_gauge = \
+            metrics_registry.gauge('pending_batch_gauge') \
+            if metrics_registry else None
```
peterschwarz: This looks like it will most likely fail later on when there's no metrics registry.  Is there a NO OP metrics registry that can be used when we don't have the OpenSTDB settings in place?

RyanLassigBanks: Fixed add method `_set_gauge` to `BlockPublisher`

PR #914, opened 2017-09-22
Add agents list

peterschwarz:
Adds a list of agents page.  

Still requires the filter form element for typing a filter


families/track_and_trade/client/src/components/tables.js
```diff
@@ -0,0 +1,104 @@
...
+
+  view (vnode) {
+    return [
+      m('table.table',
+        m('thead',
+          m('tr',
+            vnode.attrs.headers.map((header) => m('th', header)))),
+        m('tbody',
+          vnode.attrs.rows.length > 0
+          ? vnode.attrs.rows
```
delventhalz: Ternary operands on their own line should indented per Standard JS.

peterschwarz:
Added truncation for names

PR #917, opened 2017-09-23
Add Fish List

peterschwarz:
Display the fish in a list.

### DIFF
--- a/families/track_and_trade/client/src/views/list_fish.js
+++ b/families/track_and_trade/client/src/views/list_fish.js
@@ -1,0 +1,151 @@
+/**
+ * Copyright 2017 Intel Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ----------------------------------------------------------------------------
+ */
+'use strict'
+
+const m = require('mithril')
+const truncate = require('lodash/truncate')
+const moment = require('moment')
+const {Table, FilterGroup, PagingButtons} = require('../components/tables')
+const api = require('../services/api')
+
+const PAGE_SIZE = 50
+
+const FishList = {
+  oninit (vnode) {
+    vnode.state.records = []
+    vnode.state.filteredRecords = []
+
+    vnode.state.currentPage = 0
+
+    api.get('/records').then((records) => {
+      vnode.state.records = records
+      vnode.state.filteredRecords = records
+    })
+  },
+
+  view (vnode) {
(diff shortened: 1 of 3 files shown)


families/track_and_trade/client/src/views/list_fish.js
```diff
@@ -0,0 +1,147 @@
...
+    return ''
+  }
+}
+
+const _getPropTimeByComparison = (compare) => (record) => {
+  if (!record.updates.properties) {
+    return null
+  }
+
+  return Object.entries(record.updates.properties)
```
delventhalz: Looks like what you want is `Object.values`

families/track_and_trade/client/src/views/list_fish.js
```diff
@@ -0,0 +1,147 @@
...
+
+  return Object.entries(record.updates.properties).reduce(
+    (sum, [key, updates]) => sum + updates.length, 0)
+}
+
+const _formatTimestamp = (sec) => {
+  if (!sec) {
+    sec = Date.now() / 1000
+  }
+  return moment(sec).format('MM/DD/YYYY, h:mm:ss a')
```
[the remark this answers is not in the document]
delventhalz: screen shot 2017-09-25 at 8 35 37 am

The wrong "Added" time seems to be cause by always being passed in as null, and then turned into seconds instead of milliseconds. I am not sure why the "Updated" time is getting the the day right, but erroneously reporting the time as midnight. Looking at the code it seems like you _should_ be correctly getting the latest update time (in seconds), and then turn it into 01/18/1970.

families/track_and_trade/client/src/views/list_fish.js
```diff
@@ -0,0 +1,147 @@
...
+const FishList = {
+  oninit (vnode) {
+    vnode.state.records = []
+    vnode.state.filteredRecords = []
+
+    vnode.state.currentPage = 0
+
+    api.get('/records').then((records) => {
+      vnode.state.records = records
+      vnode.state.filteredRecords = records
```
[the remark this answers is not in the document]
peterschwarz: It's not redundant, since all the filtering is happening on the client side. We need to keep the original list so we can perform the filtering actions.  The initial value is unfiltered, so we set both values to records.

families/track_and_trade/client/src/views/list_fish.js
```diff
@@ -0,0 +1,151 @@
...
+
+const _getPropTimeByComparison = (compare) => (record) => {
+  if (!record.updates.properties) {
+    return null
+  }
+
+  return Object.values(record.updates.properties)
+      .reduce((acc, updates) => acc.concat(updates), [])
+      .reduce((selected, update) =>
+              compare(selected.timestamp, update.timestamp) ? update: selected)
```
delventhalz: Need a space before `:` (StandardJS)

families/track_and_trade/client/src/views/list_fish.js
```diff
@@ -0,0 +1,151 @@
...
+const _getOldestPropUpdateTime =
+  _getPropTimeByComparison((selected, timestamp) => selected > timestamp)
+
+const _countPropUpdates = (record) => {
+  if (!record.updates.properties) {
+    return 0
+  }
+
+  return Object.values(record.updates.properties).reduce(
+    (sum,  updates) => sum + updates.length, 0)
```
delventhalz: Remove extra space before `updates`

PR #918, opened 2017-09-24
Metrics enhancements

vaporos:


rest_api/sawtooth_rest_api/route_handlers.py
```diff
@@ -86,40 +129,43 @@ def __init__(
             link: /batches or /batch_status link for submitted batches
 
         """
+        timer_ctx = self._post_batches_total_time.time()
```
peterschwarz: Should this be a python Context?  What happens if an exception is thrown before the later stop call?

vaporos: Within time(), a new TimerContext() is created and returned.  The Timer instance itself does not keep a reference to the TimerContext, so there is no resource leak as far as I can tell. If an exception is thrown, stop() will not be called, which means this is, in it's currently form, a measurement of happy-path time.

peterschwarz: If stop is not called, is a measurement not actually made?

vaporos: Correct, timer_ctx basically does nothing until stop() is called, other than record it's creation time.

PR #921, opened 2017-09-24
BlockValidator now uses current chain_head when checking a block.

cianx:
Previously the BlockValidator had used the chain_head of when the Block was
queued for validation. This caused wasted block validations, as if a Block
was earlier in the queue was determined to be the chain_head the later
would detect the change after it finished validation and would requeue
the validation. If that requeue was not at the front of the queue then
the same case would happen again.

### DIFF
--- a/validator/sawtooth_validator/journal/chain.py
+++ b/validator/sawtooth_validator/journal/chain.py
@@ -85,7 +85,6 @@ class BlockValidator(object):
                  consensus_module,
                  block_cache,
                  new_block,
-                 chain_head,
                  state_view_factory,
                  done_cb,
                  executor,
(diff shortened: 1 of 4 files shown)


validator/sawtooth_validator/journal/chain.py
```diff
@@ -271,6 +271,8 @@ def validate_block(self, blkw):
                 if valid:
                     valid = consensus.verify_block(blkw)
 
+                # since changes to the chain-head can changes the state of the
```
jsmitchell: s/can changes the/can change the/

validator/sawtooth_validator/journal/chain.py
```diff
@@ -121,7 +119,8 @@ def __init__(self,
         self._chain_commit_state = ChainCommitState(
             self._block_cache.block_store, [])
         self._new_block = new_block
-        self._chain_head = chain_head
+        self._chain_head = None  # set during execution of the of the
```
peterschwarz: This comment should go before `self._chain_head`. In it's current format, the second half looks like it's referring to the next property.

PR #928, opened 2017-09-26
Add timer for interconnect's send response time

vaporos:


validator/sawtooth_validator/networking/future.py
```diff
@@ -82,6 +84,11 @@ def run_callback(self):
     def get_duration(self):
         return self._reconcile_time - self._create_time
 
+    def timer_stop(self):
```
peterschwarz: Where does this get called?

PR #930, opened 2017-09-26
Implement On-chain Validation Rules

agunde406:


validator/sawtooth_validator/journal/validation_rule_enforcer.py
```diff
@@ -0,0 +1,138 @@
...
+            try:
+                rule_type, arguments = rule.split(":")
+            except ValueError:
+                LOGGER.warning("Validation Rule Ignored, not in the correct "
+                               "format: %s",
+                               rule)
+                continue
+
+            rule_type = rule_type.strip()
+            if rule_type == "NofX":
```
peterschwarz: Could you add a comment for what each of these is?

validator/sawtooth_validator/journal/validation_rule_enforcer.py
```diff
@@ -0,0 +1,138 @@
...
+            elif rule_type == "local":
+                valid = self._do_local(
+                    transactions, expected_signer, arguments)
+
+            if not valid:
+                return False
+
+        return valid
+
+    def _do_nofx(self, transactions, arguments):
```
peterschwarz: Also, a doc string of what this rule does would be very helpful.  Hard to review without knowing what these are supposed to be evaluating.

validator/sawtooth_validator/journal/validation_rule_enforcer.py
```diff
@@ -0,0 +1,138 @@
...
+        try:
+            family, num = arguments.split(',')
+            position = int(num.strip())
+        except ValueError:
+            LOGGER.warning("Ignore, XatY requires arguments in the format "
+                           "family,position not %s", arguments)
+            return True
+
+        family = family.strip()
+        txn = transactions[position]
```
peterschwarz: This will throw an exception if the position is outside of the array.

validator/sawtooth_validator/journal/validation_rule_enforcer.py
agunde406: Added

validator/sawtooth_validator/journal/validation_rule_enforcer.py
```diff
@@ -0,0 +1,179 @@
...
+    def validate(self, block, state_root):
+        """
+        Retrieve the validation rules stored in state and check that the
+        block does not violate any of those rules. These rules include:
+
+            NofX: Only N of transaction type X may be included in a block.
+            XatY: A transaction of type X must be in the block at position Y.
+            local: A transaction must be signed by the same key as the block.
+
+        If any setting stored in statem, does not match the required format for
```
boydjohnson: The comma seems like it isn't needed.

boydjohnson: statem -> state

PR #932, opened 2017-09-26
Add basic fish detail screen

peterschwarz:
Add initial fish detail screen, which allows the user to transfer
ownership/custodianship, and update any property for which the user is
an authorized reporter.

Additionally,

- refactors common record-related items into `utils/records.js`
- fixes some linting issues

### DIFF
--- a/families/track_and_trade/client/src/views/add_fish_form.js
+++ b/families/track_and_trade/client/src/views/add_fish_form.js
@@ -21,6 +21,8 @@ const payloads = require('../services/payloads')
 const transactions = require('../services/transactions')
 const {MultiSelect} = require('../components/forms')
 
+const PRECISION = payloads.FLOAT_PRECISION
+
 /**
  * Possible selection options
  */
@@ -76,7 +78,7 @@ const AddFishForm = {
                    min: 0,
                    step: 'any',
                    oninput: m.withAttr('value', (value) => {
-                     vnode.state.lengthInCM = parseFloat(value)
+                     vnode.state.lengthInCM = value
                    }),
                    value: vnode.state.lengthInCM
                  })),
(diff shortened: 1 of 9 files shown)


peterschwarz:
Added the "Finalize" Button

peterschwarz:
Added a little refactoring

peterschwarz:
Rebased, and added an number of commits.  

Remaining work - Accepting/rejecting proposals

families/track_and_trade/client/src/views/fish_detail.js
```diff
@@ -38,6 +38,12 @@ const _labelProperty = (label, value) => [
     m('dd', value))
 ]
 
+const _row = (...cols) =>
+  m('.row',
+    cols
+    .filter((col) => col !== null)
```
delventhalz: FYI, This essentially rewrites an existing function from `components/layout`

delventhalz: Also, you don't have to filter out `null` values. Mithril already ignores those.

families/track_and_trade/client/src/views/add_fish_form.js
```diff
@@ -19,70 +19,21 @@ const m = require('mithril')
 
 const payloads = require('../services/payloads')
 const transactions = require('../services/transactions')
+const {MultiSelect} = require('../components/forms')
+
+const PRECISION = payloads.FLOAT_PRECISION
```
delventhalz: Looks like there are issues parsing user inputs into millionths and back again. 

This is what the output looks like when I enter a value of `1` for lat, long, and temp:
screen shot 2017-09-27 at 10 51 32 am

delventhalz: Checking the database, it looks like these values are being submitted to the blockchain just as `1` (should be 1,000,000).

Similarly, inputting `3.14`, just gets submitted as `3` (should be 3,140,000).

peterschwarz: Pushed a commit to correct this.

peterschwarz:
Rebased

families/track_and_trade/client/src/views/fish_detail.js
peterschwarz: It's a little different - different column type.

PR #934, opened 2017-09-26
Add basic property detail page

delventhalz:
Adds a property detail page which lists a history of value updates, and if properly authorized displays a form to submit new updates. Currently includes a map for location, but no graph widget.

To test, follow the setup instructions in [PR-889](https://github.com/hyperledger/sawtooth-core/pull/889). Then signup a new user, create a new record, and navigate to the property detail url:
```
http://localhost:3000/fish/#!/properties/{RECORD ID}/{PROPERTY NAME}
```

_Screenshots:_
screen shot 2017-09-26 at 2 26 44 pm

screen shot 2017-09-26 at 10 55 19 pm

### DIFF
--- a/families/track_and_trade/client/package.json
+++ b/families/track_and_trade/client/package.json
@@ -21,6 +21,7 @@
   "homepage": "https://github.com/hyperledger/sawtooth-core#readme",
   "dependencies": {
     "bootstrap": "^4.0.0-beta",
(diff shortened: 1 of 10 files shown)


peterschwarz:
Only concern is the licensing of the google map stuff, but otherwise looks good.

PR #938, opened 2017-09-27
Fix EVM contract addressing

dplumb94:
Previously the Seth TP left empty bytes at the beginning of contract account addresses. This caused addressing errors for some types of contracts, specifically contracts that call other contracts and require that contract's address.

This PR fixes the aforementioned addressing issue by left padding the address and taking the last 20 bytes (which is the length of an EVM address). It also adds a test for contract chaining which is a feature that is dependent on this fix.


families/seth/src/sawtooth_seth/processor/handler/sawtooth_app_state.go
```diff
@@ -44,7 +44,8 @@ func NewSawtoothAppState(state *processor.Context) *SawtoothAppState {
 // GetAccount retrieves an existing account with the given address. Returns nil
 // if the account doesn't exist.
 func (s *SawtoothAppState) GetAccount(addr Word256) *Account {
-	vmAddress, err := NewEvmAddrFromBytes(addr.Bytes()[:20])
+	addrBytes := addr.Bytes()[len(addr.Bytes())-20:]
```
aludvik: `len(addr.Bytes())` is constant, so you can just use `Word256Length`: https://github.com/hyperledger/sawtooth-core/blob/master/families/seth/src/burrow/word256/word.go#L32

`vmAddress, err := NewEvmAddrFromBytes(addr.Bytes()[Word256Length-20:]`

families/seth/src/sawtooth_seth/processor/handler/sawtooth_app_state.go
```diff
@@ -216,7 +222,7 @@ func toStateAccount(acct *Account) *EvmStateAccount {
 		return nil
 	}
 	return &EvmStateAccount{
-		Address:     acct.Address.Bytes()[:20],
+		Address:     acct.Address.Bytes()[len(acct.Address.Bytes())-20:],
```
aludvik: `acct.Address.Bytes()[Word256Length-20:]`

PR #941, opened 2017-09-27
Add T&T query parameters

nick-drozd:


nick-drozd:
I redid the queries as per Zac's suggestion. I also fixed the missing value empty list bug.

PR #943, opened 2017-09-27
STL-675 - RPC Account Calls - Part 1/2

aludvik:
Implement the calls the just retrieve state.


families/seth/rpc/src/account.rs
```diff
@@ -17,19 +17,111 @@
...
+    if address.len() < 4 || address.len() % 2 != 0{
+        Err(Error::invalid_params(format!("Invalid storage position: {}", address)))
+    } else {
+        Ok(String::from(&address[2..]))
+    }
+}
+
+pub fn get_balance<T>(params: Params, mut client: ValidatorClient<T>) -> Result<Value, Error> where T: MessageSender {
+    info!("eth_getBalance");
+    let (address, block): (String, String) = match params.parse() {
```
peterschwarz: This might be doable more concisely as

```
let Ok(address, block): ParseResult<(String, String)> = params.parse().map_err(|_| Error::invalid_params("msg"))?;
```

Where in this case, 

```
type ParseResult<T> = Result<T, jsonrpc::Error>;
```

Would have to test...

families/seth/rpc/tests/test_seth_rpc.py
```diff
@@ -33,6 +35,12 @@
 from sawtooth_sdk.protobuf.txn_receipt_pb2 import ClientReceiptGetResponse
 from sawtooth_sdk.protobuf.txn_receipt_pb2 import TransactionReceipt
 from protobuf.seth_pb2 import SethTransactionReceipt
+from protobuf.seth_pb2 import EvmEntry
+from protobuf.seth_pb2 import EvmStateAccount
+from protobuf.seth_pb2 import EvmStorage
+
+import logging
+LOGGER = logging.getLogger(__name__)
```
vaporos: This probably wouldn't pass the linter.

PR #945, opened 2017-09-27
Allow users to accept or reject proposals

peterschwarz:
Additionally:

- Correct temp to parse as float
- Reload data when transactions commit


delventhalz:
Accepting owner/custodian proposals seems to work.

PR #954, opened 2017-09-29
STL 675 - Finish Seth RPC account calls

aludvik:


families/seth/rpc/comp_seth_rpc.yaml
```diff
@@ -30,7 +30,14 @@ services:
       - 4004
     environment:
       RUST_BACKTRACE: 1
-    command: cargo run -- --connect tcp://comp-seth-rpc:4004 --bind 0.0.0.0:3030 -v
+    command: "bash -c \"
+      mkdir -p ~/.sawtooth/ &&
+      cp tests/data/test.wif ~/.sawtooth/test.wif &&
```
vaporos: Can we put 'test' in the filename of this compose file?

PR #959, opened 2017-09-29
UI Fixes

delventhalz:
- Fix tab order in add fish form
- Fix lint
- Prevent state where FishNet would attempt to submit txn without batcher pubkey
- Redirect to fish detail page after creation
- Sort agents list alphabetically
- Sort Fish list by most recently updated
- Make modals submit when enter is pressed
- Truncate Agent public keys

### DIFF
--- a/families/track_and_trade/client/src/views/list_agents.js
+++ b/families/track_and_trade/client/src/views/list_agents.js
@@ -17,6 +17,7 @@
 'use strict'
 
 const m = require('mithril')
+const sortBy = require('lodash/sortBy')
 const truncate = require('lodash/truncate')
 const {Table, FilterGroup, PagingButtons} = require('../components/tables.js')
 const api = require('../services/api')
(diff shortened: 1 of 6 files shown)


families/track_and_trade/client/src/views/list_agents.js
```diff
@@ -17,6 +17,7 @@
 'use strict'
 
 const m = require('mithril')
+const _ = require('lodash')
```
peterschwarz: Using the explicit require will allow the minimization via webpack to produce a much smaller file.

families/track_and_trade/client/src/views/list_fish.js
```diff
@@ -40,6 +40,9 @@ const FishList = {
 
   view (vnode) {
     let publicKey = api.getPublicKey()
+    vnode.state.filteredRecords.sort((a, b) => {
```
peterschwarz: `vnode.state.records` is the source of truth, and `filteredRecords` are just slices of that.  So you can probably get by just sorting once on load.

PR #962, opened 2017-09-29
Add T&T run script

nick-drozd:
The script itself is very simple: it runs docker-compose and then builds the client. 

It runs out of `/families/track_and_trade`; should it be in `/bin` instead?

### DIFF
--- a/docker/compose/sawtooth-track-and-trade.yaml
+++ b/docker/compose/sawtooth-track-and-trade.yaml
@@ -1,0 +1,86 @@
+# Copyright 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


families/track_and_trade/run-tnt.sh
```diff
@@ -0,0 +1,24 @@
+# Copyright 2017 Intel Corporation
```
vaporos: Suggest removing this script and (probably in another PR) adding a client docker container with the client installed.  This will remove the requirement to install npm locally on the host.  It will also allow running docker compose in the foreground, which is useful for watching logs.

docker/compose/sawtooth-track-and-trade.yaml
```diff
@@ -0,0 +1,86 @@
...
+    stop_signal: SIGKILL
+
+  validator:
+    image: sawtooth-validator:latest
+    volumes:
+      - ../../:/project/sawtooth-core
+    expose:
+      - 4004
+      - 8800
+    command: "bash -c \"\
```
vaporos: We should copy the use of the grafana/influxdb setup from the smallbank compose file.

families/track_and_trade/run-tnt.sh
vaporos: Implied here is doing something like "docker exec -it tnt-client bash" to get into the container after it's been run using docker compose.

delventhalz:
@vaporos Given the end-of-month deadline, if this deployment workflow is not acceptable, I think it is best we close this PR, and merge track-and-trade into master. A modified PR can be made against master at a later date.

vaporos:
I'm fine with this being re-submitted against master directly after the track-and-trade merge.

PR #971, opened 2017-10-01
Return logs from eth_getFilterChanges, etc (Part 1 of 2)

peterschwarz:
This commit returns the log events, regardless of filter values.  The
events are transformed into the correct JSON values for viewing, but not
filtered based on topics or address, as specified by calls to
`eth_newFilter` or the filter values provided directly to the method.

### DIFF
--- a/families/seth/rpc/src/calls/logs.rs
+++ b/families/seth/rpc/src/calls/logs.rs
@@ -16,21 +16,35 @@
  */
 
 use std::sync::atomic::{AtomicUsize, Ordering, ATOMIC_USIZE_INIT};
+use std::collections::HashMap;
 
 use jsonrpc_core::{Params, Value, Error};
+use serde_json::Map;
+use protobuf;
 
 use client::{
     ValidatorClient,
     hex_prefix,
+    num_to_hex,
+    bytes_to_hex_str,
 };
 use requests::{RequestHandler};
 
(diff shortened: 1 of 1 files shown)


peterschwarz:
Updated to maintain transaction order.

families/seth/rpc/src/calls/logs.rs
```diff
@@ -174,20 +190,157 @@ pub fn get_filter_logs<T>(params: Params, mut client: ValidatorClient<T>)
...
+/// Returns the logs by the given filter, since the last block id
+fn get_logs_by_filter<T>(mut client: ValidatorClient<T>,
+                         _filter_id: Option<String>,
+                         filter_spec: LogFilterSpec,
+                         last_block_id: String)
+    -> Result<Value, Error> where T: MessageSender
+{
+    debug!("log filter: {:?} since {}", filter_spec, &last_block_id[..8]);
+    let mut paging = PagingControls::new();
+    paging.set_count(100);
```
aludvik: Why 100?

peterschwarz: This is a way to fetch back far enough, with the assumption that this is being called regularly (as it is meant to be polled).  Was a balance between fetching too many blocks.  Will probably change this in a future PR to fetch chunks.  Or to simply store relevant events in map that is updated via subscriptions.

PR #974, opened 2017-10-02
Network Permissioning Integration Test

agunde406:


integration/sawtooth_integration/tests/test_network_permissioning.py
```diff
@@ -0,0 +1,726 @@
...
+    def send(self):
+        name = uuid4().hex[:20]
+        txns = [self._factory.create_transaction(
+            cbor.dumps({'Name': name,
+                        'Verb': 'set',
+                        'Value': 1000}),
+            inputs=[self._namespace + self._factory.sha512(name.encode())[-64:]],
+            outputs=[self._namespace + self._factory.sha512(name.encode())[-64:]],
+            deps=[])]
+        self._rest.send_batches(self._factory.create_batch(txns))
```
peterschwarz: In order to alleviate the test failure, perhaps these need to wait for commit

agunde406: send batch already uses return self._post('/batches?wait', batch_list). I think the issue is with the function that waits until the validators are out of consensus. Boyd said he would look at it tonight.

PR #987, opened 2017-10-09
Add missing imports

cclauss:
#881 with the obligatory signoff

Added missing imports:
* import getpass
* from base64 import b64encode

Signed-off-by: Chris Clauss <cclauss@bluewin.ch>


dcmiddle:
The DCO bot is throwing a false positive. The sign-off is clearly in the commit.
I will merge in spite of this tool false positive.

PR #988, opened 2017-10-09
import logging and define LOGGER

cclauss:
LOGGER is used on lines 49, 52, 55, and 57.  Without this change NameError might be raised at runtime.

Signed-off-by: Chris Clauss cclauss@bluewin.ch


dcmiddle:
The DCO bot is throwing a false positive. The sign-off is clearly in the commit.
I will merge in spite of this tool false positive.

PR #993, opened 2017-10-10
Doc fixes from install test

chenette:
Improve the Docker, Ubuntu, and AWS procedures for installing and running Sawtooth as a result of yesterday's tests (preparation for this week's point release).

Summary of changes:
- Reorganize to clarify the procedures (what steps to perform) and reduce confusion. 
- Specify when to open a new terminal window. 
- Docker: List the container names in the container sections. 
- Ubuntu: Rewrite a redundant "Start the REST API" step (added ps command to check status) and delete the other redundant step.
- Docker & Ubuntu: Correct the output for "Viewing the List of Blocks". 
- Change "intkey" and "IntKey" to IntegerKey unless referring to the actual intkey command. 
- Make heading titles (aka tasks) clear and consistent between procedures.  
- Fix typos, punctuation, capitalization, fonts, EOL spaces, etc. 

Note that this is a light rewrite - there is more work to do in a later pass (for release 1.0).


docs/source/app_developers_guide/ubuntu.rst
```diff
@@ -114,36 +125,35 @@ The following output appears:
...
 
 
-Start Validator
----------------
+Starting the Validator
+----------------------
 
 To start a validator that listens locally on the default ports, run the
-following commands from a Linux terminal:
+following commands from the validator window:
```
dplumb94: I think "from a Linux terminal" was correct here. The terminal doesn't really become a "validator window" until the `sudo -u sawtooth sawtooth-validator -vv` command is run (which starts up the validator in that terminal)

chenette: @dplumb94 - Thanks for this catch!  I was trying a new thing that doesn't quite work - i.e, "naming" the current terminal window in the previous step (see line 106).  Yup, it's confusing.  I'd like to fix it by not mentioning "from a Linux terminal", because they just opened the terminal in the previous step and should have a prompt available.  That is, I propose changing this sentence to: 

     "To start a validator that listens locally on the default ports, run the following commands:"

dplumb94: That sounds like a good idea

PR #1006, opened 2017-10-14
Remove Unused Network Messages and Fields from Network Protos

agunde406:


protos/network.proto
```diff
@@ -18,30 +18,20 @@ option java_multiple_files = true;
 option java_package = "sawtooth.sdk.protobuf";
 option go_package = "network_pb2";
 
-// The connect message from a client to the server
-message ConnectMessage {
```
peterschwarz: Do we lose some sort of a handshake by removing this message?  I.e. can a malicious process connect with  ZMQ sockets directly and cause memory issues by filling up the connections for a longer period of time?

validator/sawtooth_validator/journal/responder.py
```diff
@@ -130,7 +129,7 @@ def handle(self, connection_id, message_content):
 
             block_response = network_pb2.GossipBlockResponse(
                 content=block.get_block().SerializeToString(),
-                node_id=node_id)
+                )
```
peterschwarz: This might be a tad pedantic, but would suggest removing the trailing comma and moving the close paren either to the previous line or the same tab spacing as the opening line

protos/network.proto
peterschwarz: Based on in-person conversation, this has been replaced by ConnectionRequest in the authorization namespace of messages

agunde406: This message has been replaced by ConnectionRequest and is no longer used.

validator/sawtooth_validator/journal/responder.py
agunde406: updated

PR #1010, opened 2017-10-17
Remove encoding from Transaction, TpRegister, and Transaction Processors

agunde406:


protos/transaction.proto
```diff
@@ -47,16 +47,11 @@ message TransactionHeader {
...
     repeated string outputs = 7;
 
-    // The payload encoding correlates to the transaction processor's payload
-    // encoding that this transaction can be processed on,
-    // for example ""application/cbor""
-    string payload_encoding = 8;
-
     //The sha512 hash of the encoded payload
-    string payload_sha512 = 9;
+    string payload_sha512 = 8;
```
vaporos: We should resist renumbering protobuf fields; if we take a 0.8 protobuf and deserialize it for 0.9, it will probably work as long as the remaining fields haven't been renumbered.  We should consider this approach for all protobuf changes.

peterschwarz: I would definitely agree after 1.0.  I could be convinced either way for changes between 0.8 and 1.0, though

agunde406:
Removed the renumbering from the protos

PR #1013, opened 2017-10-18
Update Client Messages, Part 1.

peterschwarz:
This PR updates the client messages by:

- Replacing Leaf with a nested Entry
- Renaming messages to have consistent `Client-` prefixes
- Removes `wait_for_commit` from batch submit, and renames it in Get Batch Status
- Changes type of timeout.


peterschwarz:
Added a commit to delete a TnT integration test

nick-drozd:
There are still a few T&T lines in `run_lint`, if you'd care to remove them.

PR #1022, opened 2017-10-23
basestring was removed from Python 3

cclauss:
This reformulation makes the test against basestring unnecessary (and avoids backslash line termination).

Signed-off-by: cclauss <cclauss@bluewin.ch>


cclauss:
Can someone please explain the Jenkins failure?  I do not understand.

dcmiddle:
@cclauss jenkins went down. Shouldn't be an issue with this PR. We'll get the job re-run.

PR #1025, opened 2017-10-24
Add TransactionHandler Abstract Base Class to Python SDK

agunde406:
The other SDKs have an interface defined for TransactionHandlers, but python did not. This PR adds an Abstract Base Class that all TransactionHandlers should inherit from. It also updates the ValidatorRegistryTransactionHandler to expect context instead of state.

### DIFF
--- a/sdk/python/sawtooth_sdk/processor/handler.py
+++ b/sdk/python/sawtooth_sdk/processor/handler.py
@@ -1,0 +1,61 @@
+# Copyright 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+
+import abc
+
+
+class TransactionHandler(object, metaclass=abc.ABCMeta):
+    """
+    TransactionHandler is the Abstract Base Class that defines the business
+    logic for a new transaction family.
+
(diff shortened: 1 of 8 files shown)


sdk/python/sawtooth_sdk/processor/handler.py
```diff
@@ -0,0 +1,61 @@
...
+    @abc.abstractproperty
+    def family_name(self):
+        """
+        FamilyName should return the name of the transaction family that this
+        handler can process. Eg., "intkey"
+        """
+        pass
+
+    @abc.abstractproperty
+    def family_versions(self):
```
nick-drozd: Is this a string or a list of strings?

agunde406: Good catch. I updated it say it is a list. I think the other SDKs have this as just a string. We should create a Story to update go and javascript to a list and make sure the filtering works correctly.

sdk/python/sawtooth_sdk/processor/handler.py
```diff
@@ -0,0 +1,61 @@
...
+
+import abc
+
+
+class TransactionHandler(object, metaclass=abc.ABCMeta):
+    """
+    TransactionHandler is the Abstract Base Class that defines the business
+    logic for a new transaction family.
+
+    The family_name, family_version, and namespaces properties are
```
nick-drozd: family_versions

sdk/python/sawtooth_sdk/processor/handler.py
```diff
@@ -0,0 +1,61 @@
...
+    used by the processor to route processing requests to the handler.
+    """
+
+    @abc.abstractproperty
+    def family_name(self):
+        """
+        family_Name should return the name of the transaction family that this
```
nick-drozd: family_name

sdk/python/sawtooth_sdk/processor/handler.py
nick-drozd: family_versions

sdk/python/sawtooth_sdk/processor/handler.py
nick-drozd: family_name

PR #1027, opened 2017-10-25
Return AuthorizationException from Context

agunde406:
This allows TransactionHandler to return a better error message if an AuthorizationException occurs. If the transaction handler does not handle the AuthorizationException, the transactor processor will return an InvalidTransaction by default. 

This also fixes the issue where if an AuthorizationException occurs in go, many transaction handlers return an InternalError, when it should return an Invalid Transaction error. This is fixed by returning the error received and relying on the defaults to handle the error correctly.

### DIFF
--- a/sdk/javascript/processor/exceptions.js
+++ b/sdk/javascript/processor/exceptions.js
@@ -84,8 +84,25 @@ class ValidatorConnectionError extends Error {
   }
 }
 
+/**
+ * Thrown when a authorization error occurs
+ */
+class AuthorizationException extends Error {
+  /**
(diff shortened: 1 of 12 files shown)


sdk/javascript/processor/exceptions.js
```diff
@@ -84,8 +84,25 @@ class ValidatorConnectionError extends Error {
   }
 }
 
+/**
+ * Thrown when a authorization error occurs
+ */
+class AuthroizationException extends Error {
```
peterschwarz: s/Authoirzation/Authorization

PR #1034, opened 2017-10-27
Break up client protos

nick-drozd:
This PR breaks up the big `client.proto` file, as per https://jira.hyperledger.org/browse/STL-716


protos/client_event.proto
```diff
@@ -17,7 +17,7 @@ syntax = "proto3";
 
 option java_multiple_files = true;
 option java_package = "sawtooth.sdk.protobuf";
-option go_package = "events_pb2";
+option go_package = "client_event_pb2";
```
agunde406: This file should be split into events.proto and client_event.proto where only the messages prefixed by "Client" should go into client_event.proto.

PR #1035, opened 2017-10-27
Create PoET Liveness Test

aludvik:
The existing PoET Smoke Test is not a sufficient test for gating PRs. This test
simulates the long running network test environment using docker containers and
verifies the validators reach a certain block in consensus within the test
timelimit while ensuring blocks have batches in them.

The PoET settings and validator configuration are identical to the LR1 network
test.

To run the test locally, do:
`./bin/run_docker_test test_poet_liveness`

### DIFF
--- a/bin/run_tests
+++ b/bin/run_tests
@@ -368,6 +368,8 @@ test_integration() {
     copy_coverage .coverage.test_transactor_permissioning
     run_docker_test test_network_permissioning --timeout 800
     copy_coverage .coverage.test_network_permissioning
+    run_docker_test test_poet_liveness
+    copy_coverage .coverage.test_poet_liveness
 }
 
 test_deployment() {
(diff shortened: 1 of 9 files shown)


jsmitchell:
How much time does this add to test execution?

aludvik:
The test takes ~6min to run locally.

aludvik:
Added an extra test to ensure a minimum number of batches are committed during the test.

aludvik:
Added a fix to check_consensus and also added a lag between when the validators reach the target block and when they check consensus, since PoET sometimes has short forks.

dcmiddle:
Ran around 7 min on my Nuc. That's  a pretty expensive addition on a 20 min PR test cycle. Any thoughts on how you could shorten it without losing utility of the test?
I cut down the sleep in the main test loop to 5 seconds but that only got me to 5:35.

aludvik:
This test was originally 15-20 minutes. I already tweaked the PoET settings to make it faster. We could always tweak the settings more (maybe make the boot strap period shorter if anyone knows how?). I think even though it is a longer test, it is immensely valuable as we do not currently have any tests that verify everything works to the extent that this test does.

aludvik:
There are also opportunities to remove and or shorten other tests.

PR #1036, opened 2017-10-27
Correctly clean up outbond connection on heartbeat

agunde406:
This fixes the issues seen where if an outbound connection goes away unexpectedly, the connection is not correctly removed when it fails the heartbeat. This causes a validator to try to send a message to the shutdown connection which causes the validator to block forever.


validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -230,7 +230,9 @@ def _do_heartbeat(self):
                                          " - removing connection.",
                                          self._connection,
                                          self._connection_timeout)
-                            self._ready.clear()
+                            connection_id = hashlib.sha512(
+                                self.connection.encode()).hexdigest()
+                            del self._connections[connection_id]
```
peterschwarz: I would guard this with a `if connection_id in self._connections:` ; we have had issues where this collection get's modified by multiple threads

PR #1040, opened 2017-10-31
Add better locking around temp_endpoints in ConnectionManager

agunde406:

### DIFF
--- a/validator/sawtooth_validator/gossip/gossip.py
+++ b/validator/sawtooth_validator/gossip/gossip.py
@@ -653,11 +653,12 @@ class ConnectionManager(Thread):
                     if endpoint in self._temp_endpoints:
                         del self._temp_endpoints[endpoint]
 
+                    self._temp_endpoints[endpoint] = EndpointInfo(
+                        EndpointStatus.TOPOLOGY,
+                        time.time(),
+                        INITIAL_RETRY_FREQUENCY)
+
                 self._network.add_outbound_connection(endpoint)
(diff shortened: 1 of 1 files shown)


validator/sawtooth_validator/gossip/gossip.py
```diff
@@ -699,10 +700,11 @@ def _attempt_to_peer_with_endpoint(self, endpoint):
             # if the connection uri wasn't found in the network's
             # connections, it raises a KeyError and we need to add
             # a new outbound connection
-            self._temp_endpoints[endpoint] = EndpointInfo(
-                EndpointStatus.PEERING,
-                time.time(),
-                INITIAL_RETRY_FREQUENCY)
+            with self._lock:
```
peterschwarz: Which lock is this?  What else is it guarding?

agunde406: This is the lock is in ConnectionManager, it is guarding self._temp_endpoints but is also used around other things. I saw a "dictionary change size during iteration error " in _check_temp_endpoints which is already wrapped in a lock.

agunde406: Do you have other suggestions about what may cause this error or how to fix it?

peterschwarz: No, I think it's the right call.  I think the lock should be renamed to its specific use.  `_lock` is too generic, and it's only guarding the temp connections collection.

PR #1041, opened 2017-10-31
Initial journal reorganization

aludvik:
This PR reorganizes journal to support the consensus engine architecture changes.

PR #1045, opened 2017-10-31
Instrumented threading

aludvik:
This PR depends on #1041 and hopefully captures and logs all exceptions from threads

### DIFF
--- a/validator/sawtooth_validator/concurrent/threadpool.py
+++ b/validator/sawtooth_validator/concurrent/threadpool.py
@@ -1,0 +1,112 @@
+# Copyright 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+
+import logging
+import multiprocessing
+import time
+import os
+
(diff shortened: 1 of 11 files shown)


validator/sawtooth_validator/concurrent/threadpool.py
```diff
@@ -0,0 +1,114 @@
...
+                        task_name,
+                        run_time)
+
+            # pylint: disable=broad-except
+            except Exception as exc:
+                LOGGER.exception(
+                    '(%s) Unhandled exception during execution of task %s',
+                    self._name,
+                    task)
+                LOGGER.exception('(%s) %s', self._name, exc)
```
peterschwarz: This is what exception does - you don't need to pass it the exception.  You're basically double logging

validator/sawtooth_validator/concurrent/threadpool.py
```diff
@@ -0,0 +1,114 @@
...
+                         self._max_workers)
+                    LOGGER.debug(
+                        '(%s) Executing task %s', self._name, task_details)
+
+                return_value = task(*args, **kwargs)
+
+                end_time = time.time()
+                run_time = (end_time - start_time) * 1000.0
+
+                worker = self._workers_in_use.get_and_dec()
```
peterschwarz: Should probably move this to outside the try/catch so that it always gets decremented.

validator/sawtooth_validator/concurrent/threadpool.py
```diff
@@ -0,0 +1,114 @@
...
+                        time_in_queue)
+                    LOGGER.debug(
+                        '(%s) Workers already in use %s/%s',
+                         self._name,
+                         workers_already_in_use,
+                         self._max_workers)
+                    LOGGER.debug(
+                        '(%s) Executing task %s', self._name, task_details)
+
+                return_value = task(*args, **kwargs)
```
peterschwarz: Maybe limit the try/catch just around the task.

aludvik:
Fixes for @peterschwarz

PR #1047, opened 2017-11-01
Add dowser to cli.py DON'T MERGE

agunde406:


vaporos:
To be clear, I'm proposing moving this entire block into main after arg parsing/config loading.  Will probably require a pylint exception.

agunde406:
I was going off the assumption this will not go into master. And Ryan will do his magic to make it work but he needed a passed build to do it, hence the simple import error catch.

PR #1054, opened 2017-11-02
State delete in JS and Go

nick-drozd:
This PR adds state delete to the Go and JS SDKs. It also adds a game delete transaction to xo in order to test state deletes. (This transaction is not in the xo transaction family spec, so it should be considered a nonstandard extension for now.)

### DIFF
--- a/sdk/examples/xo_go/src/sawtooth_xo/handler/handler.go
+++ b/sdk/examples/xo_go/src/sawtooth_xo/handler/handler.go
@@ -86,6 +86,8 @@ func (self *XoHandler) Apply(request *processor_pb2.TpProcessRequest, context *p
 	switch payload.Action {
 	case "create":
 		return applyCreate(payload.Name, context)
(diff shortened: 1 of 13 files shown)


sdk/examples/xo_go/src/sawtooth_xo/handler/handler.go
```diff
@@ -86,6 +86,8 @@ func (self *XoHandler) Apply(request *processor_pb2.TpProcessRequest, context *p
 	switch payload.Action {
 	case "create":
 		return applyCreate(payload.Name, context)
+    case "delete":
```
peterschwarz: Formatting is off - probably should run `gofmt` on the file.

nick-drozd: I had run `run_go_fmt fix`, but xo wasn't being checked. I added it.

PR #1059, opened 2017-11-02
Make TpProcessRequest header unserialized

nick-drozd:
Previously the header was bytes, requiring TP handlers to go through
the extra step of deserializing it before being able to access it.

Addresses https://jira.hyperledger.org/browse/STL-817


nick-drozd:
@peterschwarz smallbank had one unused function that had to do with transaction headers. I removed it.

PR #1063, opened 2017-11-02
Improve Block List Performance

peterschwarz:
**Introduce indexes and cursors to the Database Implementation**

Introduce indexes and cursors to the Database to improve interaction
with the Database in several ways:  First, indexing is managed by the
database, which is provided a series of index name, index key function
pairs.  Indexes are stored atomically with each update to the database.
The data is segregated using LMDB's child databases.  Currently, indexes
are only sorted by natural key order (this is string-based).

The block store has been updated to support this indexing by adding
indexes for block_num, batch ids, and transaction ids within the block.

Cursors give an efficient way to navigate the underlying database.

**Use iterators, based on the cursors for block list client requests**

Use a block iterator from the block store to list the blocks
efficiently.  The block list has been simplified to only support sorting
based on block number in forward and reverse order. Paging returns just
a `next_id` value.

All tests that no longer fit into this model have been removed.

Additionally,

- Database has been made and Abstract Class
- Database implementations handle serialization/deserialization of the
  primary record
- LMDBNoLockDatabase has been updated to match the new Database
  Interface, but doesn't implement cursor
- DictDatabase has been updated to support indexes and cursors
- chain controller has been modified to check the value of the chain
  head against block store as the source of truth.
- tests have been updated to take indexes into account


peterschwarz:
One open item here is that this will require a migration tool for upgrading 0.8.9 and below block stores.  Do we want to support that sort of upgrade?

peterschwarz:
[STL-815](https://jira.hyperledger.org/browse/STL-815) has also been addressed by this change, for the block list command.

dcmiddle:
I don't think we need to back port an optimization. This will all be ready in 1.0 soon enough.

peterschwarz:
It's not a back port.  This includes a significant structural change to how data is laid out in the LMDB for the blocks (and in future PR's the same may be true for the State DB).

dcmiddle:
I see. Let's not support that kind of upgrade.

validator/tests/test_block_store/tests.py
```diff
@@ -94,22 +89,22 @@ def test_has(self):
...
-        self.assertTrue(block_store.has_batch('batch'))
+        self.assertTrue(block_store.has_batch(
+            _get_first_batch_id(block)))
         self.assertFalse(block_store.has_transaction('batch_missing'))
 
-        self.assertTrue('block' in block_store)
-        self.assertTrue('batch' in block_store)
-        self.assertTrue('txn' in block_store)
+        self.assertTrue(block.header_signature in block_store)
+        # self.assertTrue(_get_first_batch_id(block) in block_store)
```
jsmitchell: Commented outcode

peterschwarz:
Rebased before merging

PR #1064, opened 2017-11-02
Fix memory leak in schedulers when .cancel is called

boydjohnson:
Cancel is called when, during block publishing, the validator receives
a new chain head and stops the construction of that candidate block
because the predecessor is a different block, with a different state
root hash. No new merkle root will be constructed out of contexts
in the context manager, but they need to be cleaned up. This commit
adds a call to squash with persist=False, and cleanup=True in .cancel.

### DIFF
--- a/validator/sawtooth_validator/execution/scheduler_parallel.py
+++ b/validator/sawtooth_validator/execution/scheduler_parallel.py
@@ -857,8 +857,15 @@ class ParallelScheduler(Scheduler):
 
     def cancel(self):
         with self._condition:
-            self._cancelled = True
-            self._condition.notify_all()
+            if not self._cancelled and not self._final:
+                contexts = [tr.context_id for tr in self._txn_results.values()
+                            if tr.context_id]
+                self._squash(self._first_state_hash,
+                             contexts,
+                             persist=False,
+                             clean_up=True)
+                self._cancelled = True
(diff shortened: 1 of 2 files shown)


validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -857,8 +857,15 @@ def get_transaction(self, index):
 
     def cancel(self):
         with self._condition:
-            self._cancelled = True
-            self._condition.notify_all()
+            if not self._cancelled and not self._final:
+                contexts = [tr.context_id for tr in self._txn_results.values()
+                            if tr.context_id]
+                self._squash(self._first_state_hash,
```
peterschwarz: Why do the contexts need to be squashed?  Can't the just be removed?  Seems like a lot of unnecessary computation to just clean up the contexts.

peterschwarz: Granted, it fixes the leak, and I can address the performance implications while working on [STL-778](https://jira.hyperledger.org/browse/STL-778)

boydjohnson: There would have to be a second method from the context manager that the schedulers would take that would delete contexts.

PR #1066, opened 2017-11-03
Handle network.send when a connection has gone away

agunde406:


validator/sawtooth_validator/gossip/gossip.py
```diff
@@ -737,9 +744,12 @@ def _remove_temporary_connection(self, connection_id):
...
             msg = DisconnectMessage()
-            self._network.send(validator_pb2.Message.NETWORK_DISCONNECT,
-                               msg.SerializeToString(),
-                               connection_id)
+            try:
+                self._network.send(validator_pb2.Message.NETWORK_DISCONNECT,
+                                   msg.SerializeToString(),
+                                   connection_id)
+            except ValueError:
+                pass
```
peterschwarz: If we pass here, doesn't this imply that the connection has been removed and the next line will throw a key error?

validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -913,12 +923,14 @@ def _inbound_challenge_authorization_callback(self, request, result,
...
-            connection_id
-            )
+        try:
+            self.send(
+                validator_pb2.Message.AUTHORIZATION_CHALLENGE_SUBMIT,
+                auth_challenge_submit.SerializeToString(),
+                connection_id
+                )
+        except ValueError:
+            LOGGER.debug("Connection disconnected: %s", connection_id)
```
peterschwarz: This gets repeated a lot - can we put this in a `_safe_send` function?

validator/sawtooth_validator/gossip/gossip.py
agunde406: No, It only gets into this block if the connection_id is in self._connection_statuses[connection_id]  and needs to be removed. Connection_statues are gossips own copy of known connections. Also, network.remove_connection does not throw an error if the connection_id is already gone.

validator/sawtooth_validator/networking/interconnect.py
agunde406: I will add this

PR #1082, opened 2017-11-07
Renamed state_context Entry to TpStateEntry

peterschwarz:
The change makes the Entry message consistently named with the `TpState-` prefix that the rest of the messages in `state_context`


peterschwarz:
Rebased from master.

PR #1085, opened 2017-11-07
Make BlockValidator less stateful

aludvik:
Previously a new BlockValidator had to be created for every block. This PR changes the BlockValidator so it only needs to be created once. It also makes some minor opinionated tweaks here and there .


aludvik:
I forgot to lint first

aludvik:
I took some additional notes on how the BlockValidator could be further changed here: https://github.com/aludvik/sawtooth-notes/blob/master/block_validator_notes

PR #1086, opened 2017-11-07
Move config.py and log.py into processor directory

agunde406:


utility/ias_proxy/sawtooth_ias_proxy/ias_proxy.py
```diff
@@ -31,7 +31,7 @@
 
 from sawtooth_ias_client import ias_client
 from sawtooth_ias_proxy.utils import LruCache
-from sawtooth_sdk.client import config
+from sawtooth_poet_cli import config
```
nick-drozd: Should this be `sawtooth_sdk.processor`?

agunde406: Hmm. for the sake of this Pr, yes, but in the future the ias proxy should use its own way to get the configs instead of using the sdk

PR #1090, opened 2017-11-08
SDK docstrings / TP versions

nick-drozd:
This PR fills in missing TP SDK docstrings (with an eye towards doc autogeneration) and changes the Go and JS transaction handlers to accept multiple versions.

### DIFF
--- a/sdk/javascript/processor/index.js
+++ b/sdk/javascript/processor/index.js
@@ -155,7 +155,8 @@ class TransactionProcessor {
         .then(content => TpRegisterResponse.decode(content))
         .then(ack => {
           let {transactionFamilyName: familyName, version} = handler
-          let status = ack.status === 0 ? 'succeeded' : 'failed'
+          let status = ack.status ===
+              TpRegisterResponse.Status.OK ? 'succeeded' : 'failed'
           console.log(`Registration of [${familyName} ${version}] ${status}`)
         })
         .catch(e => {
--- a/sdk/javascript/processor/index.js
+++ b/sdk/javascript/processor/index.js
@@ -181,54 +181,7 @@ class TransactionProcessor {
   }
 }
 
-const _readOnlyProperty = (instance, propertyName, value) =>
-    Object.defineProperty(instance, propertyName, {
-      writeable: false,
-      enumerable: true,
-      congigurable: true,
-      value})
-
-/**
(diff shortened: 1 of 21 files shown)


sdk/python/sawtooth_sdk/processor/core.py
```diff
@@ -239,6 +248,10 @@ def _unregister(self):
                         vce)
 
     def start(self):
+        """Connects the TransactionProcessor to a Validator and starts
```
chenette: Same question as above ("transaction processor" vs "TransactionProcessor).  Also, should it be "validator" instead of "Validator"?

nick-drozd: I was hoping the doc generator would convert those to links, but it didn't :(

nick-drozd:
@agunde406 I added a commit to get rid of the deprecated context methods.

sdk/python/sawtooth_sdk/processor/core.py
chenette: In a perfect world, this ought to be possible.  I wonder if Shawn knows how to make it work?

sdk/javascript/processor/index.js
```diff
@@ -155,7 +155,7 @@ class TransactionProcessor {
         .then(content => TpRegisterResponse.decode(content))
         .then(ack => {
           let {transactionFamilyName: familyName, version} = handler
-          let status = ack.status === 0 ? 'succeeded' : 'failed'
+          let status = ack.status === 1 ? 'succeeded' : 'failed'
```
peterschwarz: This should use the enum here, which would avoid the issue of having to correct this kind of magic number.

sdk/javascript/processor/index.js
```diff
@@ -144,25 +145,27 @@ class TransactionProcessor {
...
+            Message.MessageType.TP_REGISTER_REQUEST,
+            TpRegisterRequest.encode({
+              family: handler.transactionFamilyName,
+              version: version,
+              namespaces: handler.namespaces
+            }).finish())
+            .then(content => TpRegisterResponse.decode(content))
+            .then(ack => {
+              let {transactionFamilyName: familyName} = handler
+              let status = ack.status === 1 ? 'succeeded' : 'failed'
```
peterschwarz: This should use the enum, not a magic number.

PR #1092, opened 2017-11-08
Separate BlockValidator from ChainController

aludvik:
This supersedes PR #1085 (it contains all the same commits and then some).


validator/sawtooth_validator/journal/block_validator.py
```diff
@@ -125,7 +123,8 @@ def __init__(self,
              process transactions.
              squash_handler: A parameter passed when creating transaction
              schedulers.
-             identity_signing_key: Private key for signing blocks.
+             identity_public_key: Public key use for this validator's
```
nick-drozd: use -> used

validator/sawtooth_validator/journal/block_validator.py
```diff
@@ -329,48 +329,48 @@ def validate_block(self, blkw):
...
+        common height between the two chains. Note that the chains may not
+        have the same block id at the greatest common height.
+
+        Args:
+            head_long (BlockWrapper)
+            head_short (BlockWrapper)
+
+        Returns:
+            (list of BlockWrapper) All blocks in the longer chain since the
+            last block in the shorter chain. Ordered newest to oldest."""
```
peterschwarz: Could add
```
Raises:
    BlockValidationAborted: <whatever reason>
```
If you're being really diligent about your docstrings

validator/sawtooth_validator/journal/block_validator.py
```diff
@@ -440,25 +437,24 @@ def get_batch_commit_changes(new_chain, cur_chain):
 
         return (committed_batches, uncommitted_batches)
 
-    def run(self, consensus, callback):
+    def run(self, new_block, consensus, callback):
```
peterschwarz: Suggest calling this `candidate_block`

validator/sawtooth_validator/journal/block_validator.py
```diff
@@ -140,8 +143,14 @@ def __init__(self,
...
-        self._validation_rule_enforcer = \
-            ValidationRuleEnforcer(SettingsViewFactory(state_view_factory))
+        self._validation_rule_enforcer = ValidationRuleEnforcer(
+            SettingsViewFactory(state_view_factory))
+
+        self._thread_pool = InstrumentedThreadPoolExecutor(1) \
+            if thread_pool is None else thread_pool
+
+    def stop(self):
+        self._thread_pool.shutdown(wait=True)
```
peterschwarz: This thread pool should only be shutdown by this class if it was created by this class.  Otherwise, the creator of this class may end up having a shutdown threadpool that is used in other situations.

aludvik: The threadpool is owned by the class. I just needed a way to pass in the synchronous executor for the unit tests. Unless you know another way to swap out the thread pool without making it an argument to init.

aludvik:
Addressed @peterschwarz and @nick-drozd comments

PR #1093, opened 2017-11-08
Fix top comment in config files

chenette:
- Change "Sawtooth Lake" to "Sawtooth".
- Add a "title" comment to poet_enclave_sgx.toml.example
  to match the others.


chenette:
It appears that there are other occurrences of "Sawtooth Lake" in the comments.  I could add another commit to change the rest, or could fix this one.

chenette:
FIXED: I removed the second "Lake" in the path.toml.example comment.

PR #1098, opened 2017-11-09
Add a handler to gossip message that checks if we already have the block

agunde406:

Also changes the check in chain controller for if we have a block to check the block cache instead of block store

### DIFF
--- a/validator/sawtooth_validator/journal/chain.py
+++ b/validator/sawtooth_validator/journal/chain.py
@@ -849,7 +849,7 @@ class ChainController(object):
     def on_block_received(self, block):
         try:
             with self._lock:
-                if block.header_signature in self._block_store:
+                if self.has_block(block.header_signature):
                     # do we already have this block
                     return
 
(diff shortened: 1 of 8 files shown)


validator/sawtooth_validator/journal/chain.py
```diff
@@ -849,7 +849,7 @@ def on_block_validated(self, commit_new_block, result):
     def on_block_received(self, block):
         try:
             with self._lock:
-                if block.header_signature in self._block_store:
+                if block.header_signature in self._block_cache:
```
aludvik: Should this also use `has_block()`?

validator/sawtooth_validator/journal/chain.py
```diff
@@ -889,6 +889,18 @@ def on_block_received(self, block):
             LOGGER.exception(
                 "Unhandled exception in ChainController.on_block_received()")
 
+    def has_block(self, block_id):
```
aludvik: Does this need to be inside a lock?

validator/sawtooth_validator/server/network_handlers.py
```diff
@@ -214,7 +222,7 @@ def add(
             completer=completer),
         thread_pool)
 
-    # GOSSIP_MESSAGE 6) Send message to completer
+    # GOSSIP_MESSAGE 8) Send message to completer
```
aludvik: Maybe just delete the numbers?

validator/sawtooth_validator/journal/chain.py
agunde406: Yes, I will update

validator/sawtooth_validator/journal/chain.py
agunde406: Yes i will update

validator/sawtooth_validator/server/network_handlers.py
agunde406: Sure

PR #1101, opened 2017-11-10
CLI renaming

nick-drozd:
This PR renames some CLI commands. Are there any packaging issues that need to be taken care of?

PR #1102, opened 2017-11-11
Add batch option to poet genesis CLI

dcmiddle:
Add optional batch argument so you can create a poet registry enrollment
transaction from the CLI beyond the genesis block.

### DIFF
--- a/consensus/poet/cli/sawtooth_poet_cli/genesis.py
+++ b/consensus/poet/cli/sawtooth_poet_cli/genesis.py
@@ -68,6 +68,12 @@ def add_genesis_parser(subparsers, parent_parser):
         type=str,
         help='the name of the file to output the resulting batches')
 
+    parser.add_argument(
+        '-b', '--block',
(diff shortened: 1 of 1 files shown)


nick-drozd:
https://github.com/hyperledger/sawtooth-core/pull/1101 renames `poet genesis` to `poet registration create`.

dcmiddle:
Thanks @nick-drozd!  I knew if I ignored that part of the problem it would fix itself. ;)

chenette:
Dan, is this task (creating a poet registry enrollment transaction from the CLI beyond the genesis block) already described in the Sawtooth doc?  If not, should it be?

dcmiddle:
@chenette I think the CLI docs will automatically update with the new option. As for a larger discussion of a how-to that might be good. I've been describing this general ability of sawtooth recently as "Unpluggable" consensus. I have a script I can put up as a gist or something but may as well get #1101 through first so it uses the updated syntax.

PR #1105, opened 2017-11-13
STL-816 Remove head_id

aludvik:

PR #1111, opened 2017-11-14
Validate client block get by id input

aludvik:
Based on #1105

### DIFF
--- a/validator/sawtooth_validator/database/indexed_database.py
+++ b/validator/sawtooth_validator/database/indexed_database.py
@@ -139,7 +139,11 @@ class IndexedDatabase(database.Database):
                     if not read_key:
                         continue
 
-                packed = cursor.get(read_key)
+                try:
+                    packed = cursor.get(read_key)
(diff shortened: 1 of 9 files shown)


validator/sawtooth_validator/database/indexed_database.py
```diff
@@ -139,7 +139,11 @@ def get_multi(self, keys, index=None):
                     if not read_key:
                         continue
 
-                packed = cursor.get(read_key)
+                try:
+                    packed = cursor.get(read_key)
+                except lmdb.BadValsizeError:
+                    raise KeyError("Invalid key: %s" % read_key)
```
delventhalz: This try/except also needs to be added around line 120. It will throw a `BadValsizeError` if an empty string is passed into `contains_key`. This can be trigged with `/batch_statuses?id=,`

Also, I'm not sure if it can be triggered in regular validator operation, but I imagine `update` (line 167) and `seek` (line 282) might trigger the same behavior if passed empty strings.

PR #1113, opened 2017-11-14
Update Python Signing API

peterschwarz:
Update the signing API to follow a more flexible, pluggable model.
Signing algorithms can be encapsulated with a `Context` and keys with
`PrivateKey` and `PublicKey` objects.

A `Signer` can be used to wrap a private key and a context, which can
then be used repeatedly during operation.  For example, the identity key
for a given validator is loaded at startup time, parsed into a Secp256k1
key, encapsulated in a Signer with the appropriate context, and passed
to all users of the Signer - e.g. the BlockPublisher.

All uses of the old API have been updated to reflect this change.


signing/sawtooth_signing/core.py
```diff
@@ -0,0 +1,141 @@
...
+
+class ParseError(Exception):
+    """Thrown when an error occurs during deserialization of a Private or
+    Public key from various formats.
+    """
+    pass
+
+
+class PrivateKey(metaclass=ABCMeta):
+    """A private key instance.
```
dcmiddle: Any reason not to add new_random() to this ABC?

vaporos: new_random() doesn't make sense on an instance; however, maybe it would make sense on context - context.new_random_private_key().

dcmiddle: what about just creating a new PrivateKey(). You don't need a separate member method. A create without parameters should call a private new random method.

dcmiddle:
I thought about the key classes and we might not be able to make those abstract. They have to be associated with a specific curve.
So then I wonder if we even need a "context" class. Once you instantiate a private key you can create a signature and/or a public key.
Once you instantiate a public key you can use it to verify a signature.

PublicKey's would have constructor logic to handle PrivateKey's as objects, hex, or bytes.

my_private = PrivateKey()
signature = my_private.sign(message)

my_public = PublicKey(my_private)
my_public_key_file.write(my_public.as_hex())
some_socket.write(my_public.as_bytes())
my_public = PublicKey(my_public_key_file.read())
my_public.verify(signature, message)

vaporos:
Context is important for wrapping any state necessary.  That is more clear in the other languages where we avoid global variables.  For secp256k1, this corresponds to the secp256k1 context.

signing/sawtooth_signing/core.py
peterschwarz: I'll add `new_random_private_key` to context.  It should remove many imports of `Secpk2561PrivateKey`

peterschwarz:
Added a commit to add `new_random_private_key` to the Context

PR #1114, opened 2017-11-15
Remove CORS handlers from REST API

delventhalz:
Blanket CORS approval represents a security vulnerability and
should be removed previous to 1.0. The REST API is intended to
be lightweight, and features like authentication or CORS should
be handled by a proxy server when needed.


homarne:
agreed - functions is extraneous to normal sawtooth operation

PR #1116, opened 2017-11-15
Drop Block/Batches the sig verifier has already validated

agunde406:


validator/sawtooth_validator/gossip/signature_verifier.py
```diff
@@ -34,6 +34,7 @@
 from sawtooth_validator.networking.dispatch import HandlerStatus
 from sawtooth_validator.networking.dispatch import Handler
 from sawtooth_validator.protobuf.validator_pb2 import Message
+from sawtooth_validator.journal.timed_cache import TimedCache
```
jsmitchell: Since we're using this outside journal, it would be nice to put it somewhere else at some point

PR #1117, opened 2017-11-15
transaction family namespace permission checked by the validator

benoitrazet:
We propose to modify the validator to enforce namespace permission declared on-chain. The change has its own test added to the `run_tests` script.

### DIFF
--- a/bin/run_tests
+++ b/bin/run_tests
@@ -273,6 +273,8 @@ test_validator() {
     copy_coverage .coverage.test_shutdown_smoke
     run_docker_test test_events_and_receipts
     copy_coverage .coverage.test_events_and_receipts
+    run_docker_test test_namespace_permission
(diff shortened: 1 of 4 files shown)


benoitrazet:
We propose to modify the file `executor.py` of the validator, to enforce namespace permission declared on-chain. The change has its own test added to `run_tests`.

peterschwarz:
This looks pretty good.  My only comment would be to improve the commit message to have some more detail about the change, and to follow the [message format](https://chris.beams.io/posts/git-commit/#seven-rules) that is referenced in the [contribution guide](https://sawtooth.hyperledger.org/docs/core/releases/latest/community/contributing.html#commit-process) in the sawtooth docs

benoitrazet:
Thanks. I improved the commit message by adding some description of the change proposed. Let me know if anything needs more clarifications.

PR #1123, opened 2017-11-16
Delete Seth from Core (and move to new repo)

aludvik:


aludvik:
https://github.com/hyperledger/sawtooth-seth

PR #1124, opened 2017-11-16
Remove the use of CBOR in Merkle Trie

peterschwarz:
Remove the reliance on CBOR and encode the merkle nodes using protocol
buffers. Introduces the messages `MerkleNode` and `MerklePath`, which
provide the serializable containers for what is stored in the dict
version of the nodes.

This only replaces the nodes at the serialization seam.  The nodes
during the operation of the trie are materialized into standard python
dicts, as per previous operation.


peterschwarz:
This results in far slower behavior, so closing

PR #1130, opened 2017-11-17
Fix fatal error in sawtooth batch submit --wait

delventhalz:


cli/sawtooth_cli/batch.py
```diff
@@ -272,11 +272,11 @@ def do_batch_submit(args):
                 args.wait - int(wait_time))
             wait_time = time.time() - start_time
 
-            if all(s.status == 'COMMITTED' for s in statuses):
+            if all(s['status'] == 'COMMITTED' for s in statuses):
```
peterschwarz: do a `get('status')`, otherwise it will throw a key error if it is missing for whatever reason.

delventhalz: Is a KeyError if it is missing not the correct behavior here? It _shouldn't_ be missing, and if it was for some reason, then we would just swallow the problem, and the command would loop until it timed out.

I could see maybe doing `s.get('status', 'COMMITTED')`, and then the loop will just terminate if the statuses are missing. Though it would erroneously report back that everything had been committed.

Perhaps the most correct answer would be to add a new condition before this one with
```
if any(s.get('status', None) == None):
```
and throw some RunTime error. Not sure that gives you much more than just letting the KeyError get raised though.

PR #1133, opened 2017-11-17
Switch default REST API port from 8080 to 8008

delventhalz:
8080 is a common port for HTTP services, and is likely to cause collisions. 8008 is similarly derived from the default HTTP port (80), is much less likely to collide, and happens to be twice the default
validator port.


agunde406:
Looks like you missed one int test_network_permisssioning

PR #1138, opened 2017-11-18
Update the paging and sort controls in the Rest API

agunde406:
Update paging to only have use start which should be an start id and
limit which is the number to be returned. The response will only
contain next.

Sort endpoint is changed to /reverse, compare_lengths have been
removed. When reversing, it is done by the order they are returned
from the blockstore and the key used is "default"

### DIFF
--- a/protos/client_list_control.proto
+++ b/protos/client_list_control.proto
@@ -21,40 +21,31 @@ option go_package = "client_list_control_pb2";
 
 // Paging controls to be sent with List requests.
 // Attributes:
-//     start_id: The id of a resource to start the page with
-//     end_id: The id of a resource to end the page with
-//     min_index: A resource index relative to the query, the newest being 0
-//     count: The number of results per page, defaults to and maxes out at 1000
+//     start: The id of a resource to start the page with
+//     limit: The number of results per page, defaults to 100 and maxes out at 1000
 message ClientPagingControls {
-    oneof location_marker {
-        string start_id = 1;
-        string end_id = 2;
-        int32 start_index = 3;
-    }
-    int32 count = 4;
+    string start = 1;
+    int32 limit = 2;
 }
 
 // Information about the pagination used, sent back with List responses.
 // Attributes:
-//     next_id: The id of the first resource in the next page
-//     previous_id: The id of the last resource in the previous page
-//     start_index: The index of the first resource in this page
-//     total_resources: The total resources available from the requested query
+//     next: The id of the first resource in the next page
+//     start: The id of the first resource in the returned page
+//     limit: The number of results per page, defaults to 100 and maxes out at 1000
 message ClientPagingResponse {
-    string next_id = 1;
-    string previous_id = 2;
-    int32 start_index = 3;
-    int32 total_resources = 4;
+    string next = 1;
+    string start = 2;
+    int32 limit = 3;
+
 }
 
 // Sorting controls to be sent with List requests. More than one can be sent.
 // If so, the first is used, and additional controls are tie-breakers.
 // Attributes:
-//     keys: Nested set of keys to sort by (i.e. ['header', 'signer_public_key'])
+//     keys: Nested set of keys to sort by (i.e. ['default, block_num'])
 //     reverse: Whether or not to reverse the sort (i.e. descending order)
-//     compare_length: Sorts by value length, rather than the property itself
 message ClientSortControls {
     repeated string keys = 1;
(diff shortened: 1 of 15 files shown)


rest_api/sawtooth_rest_api/route_handlers.py
```diff
@@ -719,43 +719,24 @@ def _wrap_paginated_response(
         link = cls._build_url(request, head=head)
 
         paging_response = response['paging']
```
delventhalz: This behavior is missing from `master` (I think because it would've been a pain to implement), but the response `"link"` should include explicit `start` and `limit` queries that reflect any validator imposed defaults or maximums. If `start` and `limit` are added to the `ClientPagingResponse` this becomes trivial.

Simply move line 721 before line 719, then change line 719 to:
```
link = cls._build_url(request, head=head, start=paging_response['start'], limit=paging_response['limit'])
```

protos/client_list_control.proto
```diff
@@ -21,40 +21,26 @@ option go_package = "client_list_control_pb2";
...
-//     previous_id: The id of the last resource in the previous page
-//     start_index: The index of the first resource in this page
-//     total_resources: The total resources available from the requested query
+//     next: The id of the first resource in the next page
 message ClientPagingResponse {
-    string next_id = 1;
-    string previous_id = 2;
-    int32 start_index = 3;
-    int32 total_resources = 4;
+    string next = 1;
```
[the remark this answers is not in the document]
agunde406: done

rest_api/openapi.yaml
```diff
@@ -504,33 +501,22 @@ parameters:
...
     description: Number of items to return
-  min:
-    name: min
-    in: query
-    type: string
-    description: Id or index to start paging (inclusive)
-  max:
-    name: max
+  start:
+    name: start
```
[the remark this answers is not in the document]
agunde406: done

rest_api/openapi.yaml
```diff
@@ -190,10 +189,10 @@ paths:
           in: query
           type: string
           description: A partial address to filter leaves by
-        - $ref: "#/parameters/count"
-        - $ref: "#/parameters/min"
-        - $ref: "#/parameters/max"
-        - $ref: "#/parameters/sort"
+        - $ref: "#/parameters/head"
```
delventhalz: You already have head on line 187 above.

rest_api/openapi.yaml
```diff
@@ -540,18 +526,18 @@ definitions:
     example: https://api.sawtooth.com/state?head=65cd3a3ce088b265b626f704b7f3db97b6f12e848dccb35d7806f3d0324c71b709ed360d602b8b658b94695374717e3bdb4b76f77886953777d5d008558247dd
   Paging:
     properties:
-      start_index:
-        type: integer
-        example: 1000
-      total_count:
+      start:
+        type: strings
```
delventhalz: `strings` -> `string`

rest_api/openapi.yaml
```diff
@@ -540,18 +526,18 @@ definitions:
...
+        example: "65cd3a3ce088b265b626f704b7f3db97b6f12e848dccb35d7806f3d0324c71b709ed360d602b8b658b94695374717e3bdb4b76f77886953777d5d008558247dd"
+      limit:
         type: integer
         example: 54321
-      previous:
-        type: string
-        example: https://api.sawtooth.com/state?head=65cd3a3ce088b265b626f704b7f3db97b6f12e848dccb35d7806f3d0324c71b709ed360d602b8b658b94695374717e3bdb4b76f77886953777d5d008558247dd&min=0&count=1000
+      next_position:
+        type: integer
```
delventhalz: `integer` -> `string`

rest_api/sawtooth_rest_api/route_handlers.py
```diff
@@ -702,53 +709,39 @@ def _wrap_response(request, data=None, metadata=None, status=200):
...
                 data=data,
                 metadata={'head': head, 'link': link, 'paging': paging})
 
-        count = controls.get('count', len(data))
-        start = paging_response['start_index']
-        paging['start_index'] = start
+        limit = controls.get('limit')
+        start = controls.get("start")
+        paging["limit"] = limit
+        paging["start"] = start
```
delventhalz: This need to go before the conditional on line 726, so that start and limit get returned on the last page as well.

PR #1139, opened 2017-11-18
Update JavaScript Signing API

peterschwarz:
Update the signing API to follow a more flexible, pluggable model. Signing algorithms can be encapsulated with a `Context` and keys with `PrivateKey` and `PublicKey` objects.

A `Signer` can be used to wrap a private key and a context, which can then be used repeatedly during operation.  For example, the private key for a given client can be loaded at startup time, parsed into a Secp256k1 key, encapsulated in a Signer with the appropriate context, and passed
to all users of the Signer - e.g. the submitting transactions and batches to a server.


sdk/javascript/signing/index.js
```diff
@@ -0,0 +1,121 @@
...
+   * @param {PrivateKey} privateKey - a private key
+   * @return {Signer} - a signer instance
+   */
+  newSigner (privateKey) {
+    return new Signer(this._context, privateKey)
+  }
+}
+
+/**
+ * Returns an Context instance by algoritm name.
```
nick-drozd: algoritm -> algorithm (next several lines)

sdk/javascript/spec/signing/secp256k1_test.js
```diff
@@ -0,0 +1,141 @@
...
+ * distributed under the License is distributed on an 'AS IS' BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ------------------------------------------------------------------------------
+ */
+
+'use strict'
+
+const assert = require('assert')
```
nick-drozd: This should be added to `run_tests`.

peterschwarz: Will do in a future PR

PR #1140, opened 2017-11-19
Update Go signing API

nick-drozd:


sdk/go/src/sawtooth_sdk/signing/core.go
```diff
@@ -0,0 +1,115 @@
...
+		return NewSecp256k1Context()
+	}
+
+	panic("No such algorithm")
+}
+
+// -- Signer --
+
+// A convenient wrapper of Context and PrivateKey.
+type signer struct {
```
aludvik: Is this supposed to be private?

sdk/go/src/sawtooth_sdk/signing/core.go
```diff
@@ -0,0 +1,115 @@
...
+
+// Returns the public key for this Signer instance.
+func (self *signer) GetPublicKey() PublicKey {
+	return self.context.GetPublicKey(self.private_key)
+}
+
+// -- CryptoFactory --
+
+// A factory for generating Signers.
+type cryptoFactory struct {
```
aludvik: Is this supposed to be private? Making it private prevents the type from being used in any user-defined function.

nick-drozd: I'll change those to public.

sdk/go/src/sawtooth_sdk/signing/core.go
peterschwarz: It should be public

PR #1142, opened 2017-11-19
Fix REST API hanging bug, validate ids

delventhalz:
This PR fixes the bug, where the validator could be caused to raise an error if sent an empty string as a batch id.

It also adds a validation to all input ids, checking to make sure they are 128 character hex strings.

### DIFF
--- a/validator/sawtooth_validator/database/indexed_database.py
+++ b/validator/sawtooth_validator/database/indexed_database.py
@@ -135,7 +135,10 @@ class IndexedDatabase(database.Database):
                 read_key = key.encode()
                 # If we're looking at an index, check the index first
                 if index_cursor:
-                    read_key = index_cursor.get(read_key)
+                    try:
+                        read_key = index_cursor.get(read_key)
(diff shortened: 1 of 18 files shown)


validator/sawtooth_validator/database/indexed_database.py
```diff
@@ -117,7 +117,10 @@ def contains_key(self, key, index=None):
             search_db = self._main_db
 
         with self._lmdb.begin(db=search_db) as txn:
-            return txn.cursor().set_key(key.encode())
+            try:
+                return txn.cursor().set_key(key.encode())
+            except lmdb.BadValsizeError:
```
vaporos: This should probably log something.

delventhalz: What are you thinking? Something like _"Checked for existence of empty key"_?

delventhalz:
The first commit has been heavily refactored, and now does a RegEx validation on many  BlockStore methods.

PR #1149, opened 2017-11-20
Add newRandomPrivateKey to JS Signing Context

peterschwarz:
Update exports from the root index.js.  Includes:

- signing
- protobuf
    - all protobufs are exported, as well

### DIFF
--- a/sdk/javascript/index.js
+++ b/sdk/javascript/index.js
@@ -20,5 +20,6 @@
 module.exports = {
   processor: require('./processor'),
   exceptions: require('./processor/exceptions'),
-  protobuf: require('./protobuf')
+  protobuf: require('./protobuf'),
+  signing: require('./signing')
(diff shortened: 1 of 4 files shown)


sdk/javascript/index.js
```diff
@@ -20,5 +20,6 @@
 module.exports = {
   processor: require('./processor'),
   exceptions: require('./processor/exceptions'),
```
delventhalz: Should the signing errors be included here?

If not maybe this should be removed and there should be a `signing.errors` and a `processor.errors`

peterschwarz: The are exported again into  `./signing/index.js`, since usage, in browser, would probably be doing:

```
let signing = require('sawtooth-sdk/signing')
...
```
to avoid the issues with ZMQ

PR #1150, opened 2017-11-20
Add links to API docs

nick-drozd:
This PR adds a file that links to the published nightly docs. This will work for now, but a better way should be figured out.

### DIFF
--- a/docs/source/sdks.rst
+++ b/docs/source/sdks.rst
@@ -1,0 +1,24 @@
+=================
+Sawtooth SDK APIs
+=================
+
+- Python
+
+  - `Transaction Processor
+    <python_sdk/processor/sawtooth_sdk.processor.html#http://>`__
+  - `Signing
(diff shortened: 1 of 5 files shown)


docs/source/sdks.rst
```diff
@@ -0,0 +1,22 @@
+=================
+Sawtooth SDK APIs
+=================
+
+- Python
+
+  - `Transaction Processor
+    <https://sawtooth.hyperledger.org/docs/core/nightly/master/python_sdk/processor/sawtooth_sdk.processor.html>`__
```
peterschwarz: Is there any way to make these relative links?

chenette: I agree that relative links would be good.  Here's a possible way to do this (kinda hacky): https://stackoverflow.com/questions/27979803/external-relative-link-in-sphinx-toctree-directive

PR #1161, opened 2017-11-21
(consensus) Move trackers into validator

aludvik:
This increments on STL-848. It handles 1 and 3 in the description.

### DIFF
--- a/validator/sawtooth_validator/concurrent/atomic.py
+++ b/validator/sawtooth_validator/concurrent/atomic.py
@@ -14,6 +14,7 @@
 # ------------------------------------------------------------------------------
 
 from threading import Lock
+from threading import RLock
 
 
 class Counter:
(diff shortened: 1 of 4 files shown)


validator/sawtooth_validator/concurrent/atomic.py
```diff
@@ -46,3 +47,80 @@ def inc(self, step=1):
     def dec(self, step=1):
         with self._lock:
             self._value -= step
+
+
+class Set:
```
peterschwarz: This should probably be called `ConcurrentSet`

validator/sawtooth_validator/concurrent/atomic.py
```diff
@@ -46,3 +47,80 @@ def inc(self, step=1):
...
+    def remove(self, element):
+        with self.lock:
+            self.set.remove(element)
+
+    def __contains__(self, element):
+        with self.lock:
+            return element in self.set
+
+
+class MapOfList:
```
peterschwarz: This is often called a `MultiMap`.  Given that it has a lock, maybe even `ConcurrentMultiMap`

aludvik:
@peterschwarz fixed

PR #1164, opened 2017-11-22
 Use Sphinx for Python API docs

nick-drozd:
The pydoc API docs were ugly and broken, but it turns out that Sphinx can be used instead.

Unrelated to that, I added a commit to fix a version compatibility hack in `sawtooth-dev-go`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,7 @@
 /docs/build/
 /docs/source/_autogen/
 /docs/source/cli/output/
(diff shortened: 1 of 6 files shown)


nick-drozd:
Example page: https://build.sawtooth.me/job/Sawtooth-Hyperledger/job/sawtooth-core/job/PR-1164/1/artifact/docs/build/html/sdks/python_sdk/processor.html#module-processor.context

PR #1172, opened 2017-11-22
Change default scheduler from serial to parallel

jsmitchell:
For discussion


dcmiddle:
Does parallel have the same test coverage as serial? Have we been testing it in e.g. LR networks?

agunde406:
@dcmiddle Both schedulers run the same unit tests (where it makes sense) and we switched to using the parallel scheduler for LR networks a couple weeks ago.

jsmitchell:
Any concerns before this is merged @vaporos @boydjohnson @aludvik

PR #1177, opened 2017-11-23
except ParseError: --> except ParseError as e:

cclauss:
Make the same fix in four different files where we are catching a ParseError and then re-raising it.  In the except statement, we need to capture the exception into the variable in __e__ so that we can re-raise it in the lines immediately following.  Without this change, we would raise a NameError instead of the intended exception.

Discovered via [flake8](http://flake8.pycqa.org) testing of https://github.com/hyperledger/sawtooth-core on Python 3.6.3

$ __flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics__
```
./cli/sawtooth_cli/identity.py:567:76: F821 undefined name 'e'
            raise CliException('Unable to read key in file: {}'.format(str(e)))
                                                                           ^

./cli/sawtooth_cli/sawset.py:290:76: F821 undefined name 'e'
            raise CliException('Unable to read key in file: {}'.format(str(e)))
                                                                           ^

./consensus/poet/cli/sawtooth_poet_cli/registration.py:258:76: F821 undefined name 'e'
            raise CliException('Unable to read key in file: {}'.format(str(e)))
                                                                           ^

./validator/sawtooth_validator/server/keys.py:62:67: F821 undefined name 'e'
                "Invalid key in file {}: {}".format(key_path, str(e)))
                                                                  ^

4     F821 undefined name 'e'
```

Signed-off-by: cclauss <cclauss@bluewin.ch>

### DIFF
--- a/cli/sawtooth_cli/identity.py
+++ b/cli/sawtooth_cli/identity.py
@@ -560,7 +560,7 @@ def _read_signer(key_filename):
 
     try:
         private_key = Secp256k1PrivateKey.from_hex(signing_key)
-    except ParseError:
+    except ParseError as e:
         try:
             private_key = Secp256k1PrivateKey.from_wif(signing_key)
         except ParseError:
--- a/cli/sawtooth_cli/sawset.py
+++ b/cli/sawtooth_cli/sawset.py
@@ -286,7 +286,7 @@ def _read_signer(key_filename):
     except ParseError:
         try:
             private_key = Secp256k1PrivateKey.from_wif(signing_key)
-        except ParseError:
+        except ParseError as e:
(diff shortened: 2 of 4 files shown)


nick-drozd:
This is a known pylint bug: https://github.com/PyCQA/pylint/issues/626

cclauss:
This is not a bug in the linting tools...

1) I did not use pylint.
2) https://github.com/PyCQA/pylint/issues/626 is correct.  Things changed in Python 3.4 and the code will raise an UnboundLocalError in Python 3.4+

nick-drozd:
@cclauss That was an incomplete thought on my part. What I meant was: These errors should have been caught by pylint (which runs as part of the build process), but, because of that bug, were not.

PR #1180, opened 2017-11-27
Consensus engine protobuf messages

nick-drozd:


protos/consensus.proto
```diff
@@ -0,0 +1,325 @@
...
+  }
+
+  Status status = 1;
+  bytes data = 2;
+}
+
+// This is needed to read values from arbitrary addresses used by
+// consensus (eg., validator registry).
+message ConsensusStateGetRequest {
+  string address = 1;
```
peterschwarz: Who handles what state root hash this address is relative to?

protos/consensus.proto
```diff
@@ -0,0 +1,325 @@
...
+  }
+
+  Status status = 1;
+}
+
+// Queries
+
+// Read the current value of the setting.
+message ConsensusSettingGetRequest {
+  string setting = 1;
```
peterschwarz: Who handles what state root hash this is relative to?

protos/consensus.proto
```diff
@@ -0,0 +1,325 @@
...
+}
+
+// When used in a blockchain context, this block definition is
+// equivalent to the those parts of the block that are relevant to
+// consensus. When used outside the blockchain context, the
+// interpretation of this data structure is context dependent.
+message ConsensusBlock {
+  string id = 1;
+  string previous_id = 2;
+  uint32 index = 3;
```
peterschwarz: What is "index"?  Is it the same as block num?  If so, this should be uint64

aludvik: This is the same as block num. I think I chose index because we wanted this to work in a context where there weren't blocks, but I am fine with calling it `block_num` for now. Yeah, it should be `uint64`.

protos/consensus.proto
```diff
@@ -0,0 +1,325 @@
...
+message ConsensusBroadcastRequest {
+  ConsensusMessage message = 1;
+}
+
+message ConsensusBroadcastResponse {
+  enum Status {
+    STATUS_UNSET = 0;
+    OK = 1;
+    INTERNAL_ERROR = 2;
+    NOT_READY = 3;
```
aludvik: What would `NOT_READY` mean?

protos/consensus.proto
```diff
@@ -0,0 +1,325 @@
...
+
+  Status status = 1;
+}
+
+// Block Creation
+
+// Initialize a new block based on the current chain head and start
+// adding batches to it.
+message ConsensusInitializeBlockRequest {
+
```
aludvik: We may want to add head_id to this later on, to allow the consensus engine to build blocks on top of multiple forks at once or as a check to make sure we are building on top of the block we think we are building on top of.

protos/consensus.proto
```diff
@@ -0,0 +1,325 @@
...
+}
+
+// When used in a blockchain context, this block definition is
+// equivalent to the those parts of the block that are relevant to
+// consensus. When used outside the blockchain context, the
+// interpretation of this data structure is context dependent.
+message ConsensusBlock {
+  string id = 1;
+  string previous_id = 2;
+  uint64 index = 3;
```
aludvik: @peterschwarz I think I chose index over block num here because there was a secondary goal of making this interface compatible with non-blockchain consensus.

protos/consensus.proto
```diff
@@ -0,0 +1,325 @@
...
+    OK = 1;
+    INTERNAL_ERROR = 2;
+  }
+
+  Status status = 1;
+}
+
+// Called when a new block is received and validated
+message ConsensusOnNewBlockReceivedRequest {
+  ConsensusBlock block = 1;
```
aludvik: I think we also need to know who the block came from, so we can keep track of who is publishing blocks and whether it was our block or not. This could go in either this message or `ConsensusBlock`, I'm not sure which is correct yet.

protos/consensus.proto
aludvik: Probably need a head_id for that.

protos/consensus.proto
```diff
@@ -0,0 +1,325 @@
...
+  string id = 1;
+}
+
+message ConsensusBlockGetResponse {
+  enum Status {
+    STATUS_UNSET = 0;
+    OK = 1;
+    INTERNAL_ERROR = 2;
+    NOT_READY = 3;
+    INVALID_ADDRESS = 6;
```
aludvik: Should be`NO_RESOURCE`, not invalid address.

protos/consensus.proto
```diff
@@ -0,0 +1,325 @@
...
+}
+
+message ConsensusStateGetResponse {
+  enum Status {
+    STATUS_UNSET = 0;
+    OK = 1;
+    INTERNAL_ERROR = 2;
+    NOT_READY = 3;
+    NO_RESOURCE = 5;
+    INVALID_ADDRESS = 6;
```
aludvik: I don't think `INVALID_ADDRESS` is needed here.

protos/consensus.proto
nick-drozd: According to the status numbering scheme in the other proto files: `* NOT_READY - the validator does not yet have a genesis block`

It seems like that would be relevant here, but I don't know.

aludvik: hmm, we will need to solve the genesis block issue too. this is fine for now.

PR #1181, opened 2017-11-27
Add Reference Counting  to BlockCache

agunde406:
If a block has a count above 0, it will not be purged from the catch.

Also, catch KeyErrors in _validate_permissions and _validate_on_chain_rules and update the completer to check if the journal has the predecessor for the block it is verifying.

### DIFF
--- a/docs/source/architecture/journal.rst
+++ b/docs/source/architecture/journal.rst
@@ -89,8 +89,8 @@ valid.
 
 The BlockCache keeps blocks that are currently relevant, tracked by the last
 time the block was accessed. Periodically, the blocks that have not been
-accessed recently are purged from the block cache. The time horizon for purging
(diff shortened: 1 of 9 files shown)


docs/source/architecture/journal.rst
```diff
@@ -89,8 +89,8 @@ valid.
 
 The BlockCache keeps blocks that are currently relevant, tracked by the last
 time the block was accessed. Periodically, the blocks that have not been
-accessed recently are purged from the block cache. The time horizon for purging
-and the frequency of purging are set in the Completer.
+accessed recently are purged from the block cache, but only if none of the
+other blocks in the BlockCache reference those blocks as a predecessors.
```
jsmitchell: s/as a predecessors/as predecessors/

PR #1182, opened 2017-11-28
Remove python3-ecdsa dependency from PoET

dcmiddle:
Wait certificates are intended to be verifiable outside of an enclave.
The python portion of PoET took this on to verify the NIST 256p ECDSA
signature created in the enclave using the SGX SDK. It used an ECC library
available for python. Sawtooth security reviewers did not have time
to review that library ahead of Sawtooth's planned 1.0 release and it
was not yet listed in available whitelists. It was deemed more secure
in the near term not to use the python ecdsa library until reviewers
can gain more familiarity with it.

As such this commit reverts wait certificate signature verification to
use the SGX SDK's ECC library call.

This commit does not preclude an observer node verifying wait
certificates using its own ECC lib without SGX.


dcmiddle:
Well 5th build got through. Other failures were all unrelated to the PR. Test timeout sensitivity in docker or something.

PR #1191, opened 2017-11-30
Relax PoET signup freshness default

dcmiddle:
PoET Validator registrations must be checked to prevent reusing old signups
and/or creating a valid signup and then compromising the machine in such a
way as to arrive on the EPID revocation list.

Our first default value was to require signups within one block of the
head of the chain. This proves to be far to likely to miss as a validator
may signup as the chain is updating.

In general any congestion on the network decreases the chance of getting the
signup transaction to commit within a block or two.

All we need is to know that the signup was created relatively recently.
Changing to within 10 blocks should stop a lot of spurious errors without
introducing any risk.


dcmiddle:
Yeah it would be good to get all the state-manageable settings documented (for other components too). Where do you think that would go? 
Maybe the sysadmin guide? Make section for poet?

jsmitchell:
Would there be benefit in having this an on-chain setting?

dcmiddle:
@jsmitchell I believe it is an on-chain setting. 
sawtooth.poet.signup_commit_maximum_delay
consensus/poet/core/sawtooth_poet/poet_consensus/poet_settings_view.py:251

PR #1193, opened 2017-11-30
Fix variable name typos

aludvik:
Fix a couple typos


aludvik:
Added another typo fix I remembered

PR #1199, opened 2017-12-01
Prevent users from writing to /etc/sawtooth

rberg2:
This prevents sawtooth and other users from being
able to read private keys while allowing the group
to read them.

create ~/.sawtooth in a way that it is not group
writable and further restrict the keys there.

### DIFF
--- a/cli/sawtooth_cli/admin_command/keygen.py
+++ b/cli/sawtooth_cli/admin_command/keygen.py
@@ -106,6 +106,14 @@ def do_keygen(args):
                     print('writing file: {}'.format(priv_filename))
             priv_fd.write(private_key.as_hex())
             priv_fd.write('\n')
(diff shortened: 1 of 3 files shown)


cli/sawtooth_cli/admin_command/keygen.py
```diff
@@ -106,6 +106,11 @@ def do_keygen(args):
                     print('writing file: {}'.format(priv_filename))
             priv_fd.write(private_key.as_hex())
             priv_fd.write('\n')
+            keydir_info = os.stat(key_dir)
```
aludvik: Can you add a comment explaining what these lines are doing for future readers?

rberg2:
@aludvik Comments added and rebased. Care to take another look?

rberg2:
addresses ticket STL-900

PR #1200, opened 2017-12-01
Add block in correct order to block_cache

agunde406:

Block was being added in descending order (5, 4, 3, ..) but to correctly add reference counts they need to be added in ascending order.

### DIFF
--- a/validator/sawtooth_validator/journal/chain.py
+++ b/validator/sawtooth_validator/journal/chain.py
@@ -777,10 +777,7 @@ class ChainController(object):
                                                        result["cur_chain"])
 
                         # make sure old chain is in the block_caches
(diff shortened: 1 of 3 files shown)


validator/sawtooth_validator/journal/chain.py
```diff
@@ -777,7 +777,7 @@ def on_block_validated(self, commit_new_block, result):
                                                        result["cur_chain"])
 
                         # make sure old chain is in the block_caches
-                        for block in result["cur_chain"]:
+                        for block in reversed(result["cur_chain"]):
```
peterschwarz: Seems like this chain should be handed to the block cache in total, and let the block cache worry about how to order it.

PR #1205, opened 2017-12-04
Update NULL_BLOCK_IDENTIFIER to be 128 characters long.

cianx:


peterschwarz:
I should have had a chance to weigh in on this.  It should be rolled back:  It's a breaking change.

 NULL_BLOCK_ID is in the genesis block, so, changing that, means any checks for "is this the genesis block" will be wrong for existing chains that have upgraded from the previous version.

PR #1215, opened 2017-12-08
Remove Vagrant tools

peterschwarz:
Remove the use of Vagrant, as it has be supplanted by the docker workflow.


rberg2:
There are a few references to vagrant in the documentation as well.
docs/source/introduction.rst
docs/source/community/issue_tracking.rst

PR #1216, opened 2017-12-08
Implement ability to send a message without keeping the future around.  

agunde406:
This will fix several areas where we have a small memory leak around futures that never get results. Also, this will greatly reduce the number of Network Acks the network will have to deal with.

### DIFF
--- a/validator/sawtooth_validator/journal/responder.py
+++ b/validator/sawtooth_validator/journal/responder.py
@@ -162,13 +162,8 @@ class ResponderBlockResponseHandler(Handler):
                              connection)
 
         self._responder.remove_request(block.header_signature)
-        ack = network_pb2.NetworkAcknowledgement()
-        ack.status = ack.OK
 
-        return HandlerResult(
-            HandlerStatus.RETURN,
-            message_out=ack,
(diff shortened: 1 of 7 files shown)


validator/sawtooth_validator/journal/responder.py
```diff
@@ -93,13 +93,8 @@ def handle(self, connection_id, message_content):
...
             LOGGER.debug("Received repeat GossipBlockRequest from %s",
                          connection_id)
-            ack = network_pb2.NetworkAcknowledgement()
-            ack.status = ack.OK
 
-            return HandlerResult(
-                HandlerStatus.RETURN,
-                message_out=ack,
-                message_type=validator_pb2.Message.NETWORK_ACK)
+            return HandlerResult(HandlerStatus.DROP)
```
peterschwarz: Should we use `PASS` for these?  Drop should only be used when we absolutely know no other handler in the handler chain should happen (e.g. the request is malformed).

agunde406: This is dropping a duplicate message we have already handled (no need to pass it along), so I thougt drop was fitting.

PR #1219, opened 2017-12-11
Bug fixes on cxx code

askmish:

- Fixed wrong size zeroing in enclave.cpp file
- Removed extra code in c11_support.cpp/c files


dcmiddle:
Per vaporos, please update your commit message. Otherwise looks good to me and tests clean locally.

vaporos:
In the commit message, you could re-use what you provided in the PR description:

"
    Fixed wrong size zeroing in enclave.cpp file
    Removed extra code in c11_support.cpp/c files
"

If this were two commits, the above would be fairly good commit messages.  Keeping them in the same commit, you can add another paragraph with this level of detail.

PR #1221, opened 2017-12-11
Add Configurable Back Pressure to Batch queue

peterschwarz:
Add a configurable high water mark for the batch queue which will supply back pressure on batch submission.  This adds a new CLI parameter/toml parameter `--maximum-batch-queue-size`/`maximum_batch_queue_size` to configure this value. 

Batches are rejected during a batch submit, and a QUEUE_FULL status is returned via the ZMQ component interface.  Inbound (i.e. not requested by the completer) Gossip batch messages are dropped.

The REST API will return a `429: Too Many Requests` error.

This is a local configuration option, as node performance will vary based on local limitations, such as hardware configuration.

Additionally:

- Fixes an issue with the Dispatcher when a previous handler fails with an exception
- Fixes an issue with the intkey-workload, such that any non-202 status code is used to check the batch status later.

### DIFF
--- a/docs/source/rest_api/error_codes.rst
+++ b/docs/source/rest_api/error_codes.rst
@@ -114,6 +114,10 @@ Error Codes and Descriptions
      - Submitted Batches Invalid
      - The submitted BatchList failed initial validation by the validator. It
        may have a bad signature, or be poorly formed.
+   * - 31
+     - Unable to Accept Batches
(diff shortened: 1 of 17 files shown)


peterschwarz:
Updated description based on new changes

peterschwarz:
Rebased to pick up network changes for testing

peterschwarz:
Closing in favor of an adaptive back-pressure solution.

PR #1224, opened 2017-12-11
Create command for comparing across multiple nodes

aludvik:
Do `sawnet compare --help` for usage.

### DIFF
--- a/cli/sawtooth_cli/network_command/compare.py
+++ b/cli/sawtooth_cli/network_command/compare.py
@@ -1,0 +1,329 @@
+# Copyright 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


aludvik:
There are a few gross things in here that I need to fix, but Andi should be able to use it as is

chenette:
I believe that the new "sawnet" CLI should be documented in the CLI Reference.  You can follow the model of other parent-commands-with-subcommands, such as "sawtooth block" or "sawset". I'm available to help (aka edit the doc information).

aludvik:
Fixed gross things

agunde406:
Do test you actually have to do `sawnet compare-chains --help`

cli/sawtooth_cli/network_command/compare.py
```diff
@@ -0,0 +1,330 @@
...
+
+        for block_id, node_ids in siblings.items():
+            for node_id in node_ids:
+                node_list[node_id] = block_id[:8]
+
+    # Print the last one
+    print(format_str.format(prev_block_num, *node_list))
+
+
+def print_tree(graph, tails):
```
agunde406: Remove reference to --tree if it is not implemented

peterschwarz: Also, delete the function

aludvik:
Added an okay implementation of `--tree`. I have a wip commit that is approaching something better, but don't want to spend anymore time on this for now.

PR #1227, opened 2017-12-12
Drop duplicate responses

nick-drozd:


validator/sawtooth_validator/server/network_handlers.py
```diff
@@ -325,7 +325,7 @@ def add(
     # GOSSIP_BATCH_RESPONSE 1) Sends ack to the sender
```
peterschwarz: Not sure this comment is still accurate

validator/sawtooth_validator/gossip/gossip_handlers.py
```diff
@@ -156,30 +156,72 @@ def handle(self, connection_id, message_content):
...
         block_response_message = GossipBlockResponse()
         block_response_message.ParseFromString(message_content)
+        block = Block()
+        block.ParseFromString(block_response_message.content)
+
+        block_id = block.header_signature
+
+        if not self._has_open_requests(block_id) and self._has_block(block_id):
+            LOGGER.debug('Drop duplicate block: %s', block_id)
+            return HandlerResult(HandlerStatus.DROP)
```
agunde406: This should return a Network Ack to close the future from the sender. HandlerStatus.RETURN will return the Ack but stop processing of the message. Same for below.

PR #1230, opened 2017-12-13
Cleanup poet base64 includes and licensing

dcmiddle:
Remove unneeded includes from poet.cpp and enclave.cpp
Add modification notice to base64 source file required for license
attribution.

See also STL-398


dcmiddle:
@TomBarnes fixed

PR #1231, opened 2017-12-13
Move fork resolution logic from block validator to chain controller

aludvik:
Because of all the architectural changes, the block validator and chain controller unit tests need some work. I am waiting to make those changes until I implement the block validation pipeline so that I don't have to make the changes twice and because I think it will be easier to write unit tests after all the components are in that pipeline.

### DIFF
--- a/validator/sawtooth_validator/journal/block_validator.py
+++ b/validator/sawtooth_validator/journal/block_validator.py
@@ -42,41 +42,11 @@ LOGGER = logging.getLogger(__name__)
 
 class BlockValidationAborted(Exception):
     """
-    Indication that the validation of this fork has terminated for an
-    expected(handled) case and that the processing should exit.
+    Indication that the validation of this block has terminated for an expected
+    case and that the processing should exit.
     """
     pass
 
-
-class ChainHeadUpdated(Exception):
-    """ Raised when a chain head changed event is detected and we need to abort
-    processing and restart processing with the new chain head.
-    """
-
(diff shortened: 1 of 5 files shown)


validator/sawtooth_validator/journal/chain.py
```diff
@@ -280,16 +280,16 @@ def resolve_fork(self, chain_head, block):
         self.extend_fork_diff_to_common_ancestor(
             new_block, current_block, result.new_chain, result.current_chain)
 
-        # First check if any predecessors are invalid or unknown
-        # TODO: If we never pass blocks onto the chain controller unless they
-        # are valid, and the block validator never marks a block as valid
-        # unless all its predecessors are valid, then this isn't needed
+        # First, double check if any predecessors are invalid or unknown.
+        # Because the BlockValidator only passes the ChainController blocks
+        # if they are valid, this should always be the case..
```
nick-drozd: extra `.`

validator/sawtooth_validator/journal/block_validator.py
```diff
@@ -169,11 +104,6 @@ def stop(self):
 
 
 class BlockValidator(object):
-    """
```
peterschwarz: Is doc going to be restored in a future PR?

aludvik: Yeah, I am planning to break this class up further.

PR #1233, opened 2017-12-13
Add Time To Live to Gossip Message and Requests

agunde406:
also remove unused function from IdentityCache

### DIFF
--- a/validator/sawtooth_validator/journal/responder.py
+++ b/validator/sawtooth_validator/journal/responder.py
@@ -268,7 +268,7 @@ class BatchByTransactionIdResponderHandler(Handler):
         if batches == [] and len(not_requested) == \
                 len(batch_request_message.ids):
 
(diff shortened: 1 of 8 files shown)


validator/sawtooth_validator/journal/responder.py
```diff
@@ -180,26 +185,33 @@ def handle(self, connection_id, message_content):
...
+
+                    self._responder.add_request(batch_request_message.id,
+                                                connection_id)
             else:
                 LOGGER.debug("Batch %s has already been requested",
                              batch_request_message.id)
 
-            self._responder.add_request(batch_request_message.id,
-                                        connection_id)
+                self._responder.add_request(batch_request_message.id,
```
peterschwarz: This may explain duplicates batch requests...

agunde406:
updated int32 -> uint32

PR #1236, opened 2017-12-14
Move TP list to own section; add missing TPs

chenette:
This PR does two things to the list of transaction processor commands (aka executables) in the Application Developer's Guide:

Adds the missing TP commands:
- identity-tp
- block-info-tp
- smallbank-tp

Moves the list of transaction processor commands into a separate section so that the info appears only once (and is easier to find in the Contents). The list was in both the Ubuntu and AWS procedures, but the info isn't necessary for the steps in those procedures.

### DIFF
--- a/docs/source/app_developers_guide/example_transaction_processors.rst
+++ b/docs/source/app_developers_guide/example_transaction_processors.rst
@@ -1,0 +1,40 @@
+******************************
+Example Transaction Processors
+******************************
+
+Sawtooth includes several transaction families as examples for developing a
+transaction processor. The following executables are available:
+
+* ``block-info-tp`` - BlockInfo transaction processor, written in Python
+
+* ``identity-tp`` - Identity transaction processor, written in Python
+
+* ``intkey-tp-go`` - IntegerKey transaction processor, written in Go
+
+* ``intkey-tp-java`` - IntegerKey transaction processor, written in Java
+
+* ``intkey-tp-javascript`` - IntegerKey transaction processor, written in
+  JavaScript (Node.js)
+
+* ``intkey-tp-python`` - IntegerKey transaction processor, written in Python
+
+* ``poet-validator-registry-tp`` - Validator Registry transaction processor,
+  which is used by the PoET consensus algorithm implementation to keep track of
+  other validators
+
+* ``settings-tp`` - Settings family transaction processor, written in Python
+
+  .. note::
+
+    In a production environment, you should always run a transaction processor
+    that supports the Settings transaction family.
+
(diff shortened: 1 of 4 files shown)


docs/source/app_developers_guide/example_transaction_processors.rst
```diff
@@ -0,0 +1,40 @@
...
+* ``block-info-tp`` - BlockInfo transaction processor (written in Python)
+
+* ``identity-tp`` - Identity transaction processor (written in Python)
+
+* ``intkey-tp-go`` - IntegerKey transaction processor written in Go
+
+* ``intkey-tp-java`` - IntegerKey transaction processor written in Java
+
+* ``intkey-tp-javascript`` - IntegerKey transaction processor written in
+  JavaScript (requires node.js)
```
delventhalz: Node.js is typically capitalized.

The "required" wording also feels a little odd to me, since Node is a runtime environment (as opposed to a browser), rather than real a dependency. I think just `(Node.js)`, would be preferable, and is more typical of what you'd see when people are trying to emphasize JS is not meant for the browser.

docs/source/app_developers_guide/example_transaction_processors.rst
```diff
@@ -0,0 +1,40 @@
...
+
+  .. note::
+
+    In a production environment, you should always run a transaction processor
+    that supports the Settings transaction family.
+
+* ``smallbank-tp`` - Smallbank transaction processor (written in Go)
+
+* ``xo-tp-javascript`` - XO transaction processor written in JavaScript
+  (requires node.js)
```
delventhalz: Same Node.js note above.

docs/source/app_developers_guide/example_transaction_processors.rst
```diff
@@ -0,0 +1,40 @@
...
+******************************
+
+Sawtooth includes several transaction families as examples for developing a
+transaction processor. The following executables are available:
+
+* ``block-info-tp`` - BlockInfo transaction processor (written in Python)
+
+* ``identity-tp`` - Identity transaction processor (written in Python)
+
+* ``intkey-tp-go`` - IntegerKey transaction processor written in Go
```
peterschwarz: Inconsistent use of parens around the "written in X"

docs/source/app_developers_guide/example_transaction_processors.rst
chenette: Fixed.

docs/source/app_developers_guide/example_transaction_processors.rst
chenette: Fixed.

PR #1241, opened 2017-12-14
Add PoET sgx build to build_all script.

cianx:

### DIFF
--- a/bin/build_all
+++ b/bin/build_all
@@ -28,6 +28,7 @@ ALL_TARGETS="
         go
         java
         javascript
+        poet_sgx
         python
         rust
 "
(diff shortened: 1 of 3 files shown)


bin/build_all
```diff
@@ -281,7 +284,10 @@ then
         docker_build docker/sawtooth-dev-javascript docker/ sawtooth-xo-tp-javascript
         docker_run sawtooth-dev-javascript
     }
-
+    build_poet_sgx() {
+        docker_build docker/sawtooth-dev-poet-sgx docker/ sawtooth-dev-poet-sgx
+        docker_run sawtooth-dev-poet-sgx
+    }
```
aludvik: Since you're not supporting installed mode, I would follow the pattern used by `build_external()` instead.

PR #1243, opened 2017-12-14
DO NOT MERGE Experimental change to clamp poet wait times.

dcmiddle:
This change should not be merged into master.

Clamp one of the wait time variables to cut off long tail of
exponentially distributed wait times.


dcmiddle:
this was a red herring. the actual likelihood of this event is something like (1/10)^n where n is your validator population size.

PR #1250, opened 2017-12-17
Improve logging around poet block initialization

dcmiddle:


dcmiddle:
#1262 is a better solution

PR #1253, opened 2017-12-18
More lint stuff, add web socket subscription test

nick-drozd:
1. In a shocking coincidence, a new version of `pylint` was released the day after I merged [the last lint PR](https://github.com/hyperledger/sawtooth-core/pull/1235). This PR fixes the new issues raised (and some old ones that were missed previously).

2. I was having so much fun running Python formatters that I figured I would try running one on the Javascript code to see what happened. This PR updates the JS SDK code to comply with [Standard JS](https://standardjs.com/). Running the `standard` formatter also led me to find that

3. `test_web_socket_subscription` didn't get added to `run_tests`. I added it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,13 +17,6 @@
 /.idea/
 *.sublime*
 
-# Exclude vagrant and config files in the tools/ directory
-/tools/.vagrant/
-/tools/conf-local.sh
-/tools/vagrant.local.yaml
-/tools/cache/
-/tools/ubuntu-xenial-16.04-cloudimg-console.log
-
 # Exclude poet build artifacts
 /consensus/**/__pycache__/
 /consensus/**/build/
@@ -47,8 +40,6 @@
 /cli/build/
 /cli/sawtooth_cli/protobuf/
 /cli/**/__pycache__/
-
-/core/sawtooth_core.egg-info/
 
 /families/**/__pycache__/
 /families/settings/sawtooth_settings/protobuf/
--- a/sdk/examples/intkey_javascript/index.js
+++ b/sdk/examples/intkey_javascript/index.js
@@ -17,7 +17,7 @@
 
 'use strict'
 
-const {TransactionProcessor} = require('sawtooth-sdk/processor')
+const { TransactionProcessor } = require('sawtooth-sdk/processor')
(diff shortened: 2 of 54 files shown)


sdk/examples/intkey_javascript/index.js
```diff
@@ -17,7 +17,7 @@
 
 'use strict'
 
-const {TransactionProcessor} = require('sawtooth-sdk/processor')
```
peterschwarz: Were these changes based on the output from Standard.js?  If not, this commit should be dropped

nick-drozd: `standard` is not very rigorous. In particular, it doesn't care about line length (see https://github.com/standard/standard/issues/28). But the long lines certainly need to be taken care of, if for no other reason than that they make the docs look bad:

screen shot 2017-12-18 at 2 04 32 pm

I don't care about stuff like 
```
-const {Message} = require('../protobuf')
+const { Message } = require('../protobuf')
```
but I do think that most of these changes are worth making.

peterschwarz: I just want to ensure that the changes being made are consistent with Standard JS.  If so, I'll approve it.

peterschwarz: You could also pull the Javascript changes into a separate PR, since they're unrelated to fixes  to get the build to pass

nick-drozd: The changes here are a strict superset of what is required by `standard`.

.gitignore
```diff
@@ -13,17 +13,6 @@
 # limitations under the License.
 # ------------------------------------------------------------------------------
 
-# Exclude IDE and Editor files
-/.idea/
-*.sublime*
```
dcmiddle: We should still ignore these

validator/sawtooth_validator/state/merkle.py
```diff
@@ -43,7 +44,7 @@ def _yield_iter(self, path, hash_key):
         try:
             node = self._get_by_addr(path)
         except KeyError:
-            raise StopIteration()
```
peterschwarz: This change actual changes behaviour.  It should remain

PR #1255, opened 2017-12-18
Introduce Adaptive Back Pressure to Pending Batches

peterschwarz:
Introduces adaptive back pressure on the pending batch queue.  

The amount of batches allowed in the pending queue at any point in time is predicted using a rolling average of batches committed per block.  This is initialized to 30 (the equivalent to 1 transaction per second, with a 30 second wait time), and updated on each block commit.

New batches are rejected before any processing can take place if the pending batch queue exceeds 2 * the average, giving it a geometric increase in batches if all are committed.

Supersedes #1221

### DIFF
--- a/integration/sawtooth_integration/tests/intkey_client.py
+++ b/integration/sawtooth_integration/tests/intkey_client.py
@@ -17,6 +17,7 @@ import time
 from urllib.parse import urlparse, parse_qs
 
 from sawtooth_cli.rest_client import RestClient
+from sawtooth_cli.exceptions import CliException
 from sawtooth_cli.exceptions import RestClientException
 from sawtooth_intkey.intkey_message_factory import IntkeyMessageFactory
 
@@ -31,9 +32,20 @@ class IntkeyClient(RestClient):
     def send_txns(self, txns):
         batch = self.factory.create_batch(txns)
 
-        response = self.send_batches(batch)
-        id_query = urlparse(response['link']).query
-        return parse_qs(id_query)['id'][0]
+        attempts = 0
+        response = None
+        while True:
+            try:
+                response = self.send_batches(batch)
+                id_query = urlparse(response['link']).query
+                return parse_qs(id_query)['id'][0]
+            except CliException:
+                if attempts < 8:
+                    print('responding to back-pressure, retrying...')
+                    attempts += 1
+                    time.sleep(0.2 * (2 ** attempts))
+                else:
+                    raise
 
     def recent_block_signatures(self, tolerance):
         return self.list_block_signatures()[:tolerance]
--- a/protos/client_batch_submit.proto
+++ b/protos/client_batch_submit.proto
@@ -62,12 +62,15 @@ message ClientBatchSubmitRequest {
 //   * OK - everything with the request worked as expected
 //   * INTERNAL_ERROR - general error, such as protobuf failing to deserialize
 //   * INVALID_BATCH - the batch failed validation, likely due to a bad signature
+//   * QUEUE_FULL - the batch is unable to be queued for processing, due to
+//        a full processing queue.  The batch may be submitted again.
 message ClientBatchSubmitResponse {
     enum Status {
         STATUS_UNSET = 0;
         OK = 1;
         INTERNAL_ERROR = 2;
         INVALID_BATCH = 3;
+        QUEUE_FULL = 4;
     }
     Status status = 1;
 }
--- a/rest_api/openapi.yaml
+++ b/rest_api/openapi.yaml
@@ -53,6 +53,8 @@ paths:
                 $ref: "#/definitions/Link"
         400:
           $ref: "#/responses/400BadRequest"
+        429:
+          $ref: "#/responses/429TooManyRequests"
         500:
           $ref: "#/responses/500ServerError"
         503:
@@ -472,6 +474,10 @@ responses:
       $ref: "#/definitions/Error"
   404NotFound:
     description: Address or id did not match any resource
+    schema:
+      $ref: "#/definitions/Error"
+  429TooManyRequests:
+    description: Too many requests have been made to process batches
     schema:
       $ref: "#/definitions/Error"
   500ServerError:
--- a/sdk/examples/intkey_python/sawtooth_intkey/client_cli/intkey_workload.py
+++ b/sdk/examples/intkey_python/sawtooth_intkey/client_cli/intkey_workload.py
@@ -133,15 +133,17 @@ class IntKeyWorkload(Workload):
 
                 batch_list = batch_pb2.BatchList(batches=[batch])
 
-                post_batches(key.url, batch_list, auth_info=self._auth_info)
+                (code, _) = post_batches(key.url, batch_list,
+                                         auth_info=self._auth_info)
 
-                with self._lock:
-                    self._pending_batches[batch.header_signature] = \
-                        IntKeyState(
-                        name=key.name,
-                        url=key.url,
-                        value=key.value + 1)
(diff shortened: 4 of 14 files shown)


integration/sawtooth_integration/tests/intkey_client.py
```diff
@@ -31,9 +32,20 @@ def __init__(self, url, wait=30):
     def send_txns(self, txns):
         batch = self.factory.create_batch(txns)
 
-        response = self.send_batches(batch)
-        id_query = urlparse(response['link']).query
-        return parse_qs(id_query)['id'][0]
+        attempts = 25
```
vaporos: This should perhaps use exponential back-off up to some limit.  This would allow retrying sub-second initially but then back off to something slow like 8s or 16s.

validator/sawtooth_validator/networking/dispatch.py
```diff
@@ -202,6 +202,10 @@ def _determine_next(self, message_id, future):
             LOGGER.exception("Dispatcher timeout waiting on handler result.")
             raise
 
+        if res is None:
+            LOGGER.debug('Ignoring result previous error')
```
vaporos: This logging statement should be clearer and include more information as to the context.

sdk/examples/intkey_python/sawtooth_intkey/client_cli/intkey_workload.py
```diff
@@ -133,15 +133,17 @@ def on_batch_committed(self, batch_id):
...
+                                         auth_info=self._auth_info)
+
+                if code == 202:
+                    with self._lock:
+                        self._pending_batches[batch.header_signature] = \
+                            IntKeyState(
+                            name=key.name,
+                            url=key.url,
+                            value=key.value + 1)
+                    self.delegate.on_new_batch(batch_id, key.url)
```
aludvik: Do we want any visibility on when a batch is rejected? In this case, is the batch just dropped?

validator/sawtooth_validator/journal/publisher.py
```diff
@@ -714,3 +726,25 @@ def has_batch(self, batch_id):
...
+    @property
+    def value(self):
+        return self._current_average
+
+    def update(self, sample):
+        """Add the sample and return the updated average.
+        """
+        self._samples.append(sample)
+
+        self._current_average = sum(list(self._samples)) / len(self._samples)
```
aludvik: You do not need to convert deque to a list: https://docs.python.org/3/library/collections.html#deque-recipes

validator/sawtooth_validator/journal/publisher.py
```diff
@@ -498,6 +498,16 @@ def queue_batch(self, batch):
         for observer in self._batch_observers:
             observer.notify_batch_pending(batch)
 
+    def can_accept_batch(self):
+        can_accept = len(self._pending_batches) < \
+            (2 * self._publish_count_average.value)
```
aludvik: Can you add a comment explaining the logic behind using this calculation to determine the high-water mark?

validator/sawtooth_validator/journal/back_pressure_handlers.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+
+    def __init__(self, can_accept_fn):
+        self._can_accept = can_accept_fn
+
+    def handle(self, connection_id, message_content):
+        gossip_message = GossipMessage()
+        gossip_message.ParseFromString(message_content)
+        if gossip_message.content_type == GossipMessage.BATCH and \
+                not self._can_accept():
```
aludvik: Why do we need to deserialize before checking can_accept()?

peterschwarz: We don't want to reject blocks, which also come on this message, so we need to know what the content type is.

sdk/examples/intkey_python/sawtooth_intkey/client_cli/intkey_workload.py
peterschwarz: We do log a message on the failure with the error code and status message, in the call to post batches.  We could add one that logs something more specific to the 429 error that results here.  

This fix also corrects a case where the rest api is returning any non-202 status - it previously was assuming that all submits resulted in a successful submission (regardless if the batch is committed).

peterschwarz:
Also fixes: [STL-958](https://jira.hyperledger.org/browse/STL-958) with commit 
5e99768

peterschwarz:
Taking out the logging of back pressure.  Will re-consider what and how to log this state in a future PR, as the implementation in this PR is noisy.

peterschwarz:
Added a commit for logging entering and exiting back pressure conditions

protos/client_batch_submit.proto
```diff
@@ -62,12 +62,15 @@ message ClientBatchSubmitRequest {
...
 //   * INVALID_BATCH - the batch failed validation, likely due to a bad signature
+//   * QUEUE_FULL - the batch is unable to be queued for processing, due to
+//        a full processing queue.  The batch may be submitted again.
 message ClientBatchSubmitResponse {
     enum Status {
         STATUS_UNSET = 0;
         OK = 1;
         INTERNAL_ERROR = 2;
         INVALID_BATCH = 3;
+        QUEUE_FULL = 4;
```
vaporos: This compatibility change should be noted in the release notes for the next RC version.

rest_api/openapi.yaml
```diff
@@ -53,6 +53,8 @@ paths:
                 $ref: "#/definitions/Link"
         400:
           $ref: "#/responses/400BadRequest"
+        429:
```
vaporos: This should also be mentioned explicitly in the release notes for the next RC.

validator/sawtooth_validator/journal/back_pressure_handlers.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def __init__(self, can_accept_fn):
+        self._can_accept = can_accept_fn
+        self._applying_backpressure = False
+
+    def handle(self, connection_id, message_content):
+        if not self._can_accept():
+            if not self._applying_backpressure:
+                self._applying_backpressure = True
+                LOGGER.info(
```
vaporos: This logs too frequently and can be determined client-side.  Suggest either not logging or replacing with a metric.

validator/sawtooth_validator/journal/publisher.py
```diff
@@ -499,6 +506,12 @@ def queue_batch(self, batch):
         for observer in self._batch_observers:
             observer.notify_batch_pending(batch)
 
+    def can_accept_batch(self):
+        # Limit the number of batches to 2 times the publishing average.  This
+        # allows the queue to grow geometrically, if the queue is drained.
+        return len(self._pending_batches) < \
+            (2 * self._publish_count_average.value)
```
vaporos: It may be desirable for "2" to be a configurable settings (a float) so that we can do some testing for the right value to use.  Short of that, perhaps defining as a constant at the top of the file so it's easy to change for that purpose.

validator/sawtooth_validator/journal/back_pressure_handlers.py
peterschwarz: This only logs when entering the back pressure state, and another when exiting the state.  So at most, it logs twice per block commit

jsmitchell:
To what degree do we need to update the workload tools to reattempt the same transaction on QUEUE_FULL rather than moving onto the next transaction?

peterschwarz:
We would need to add a backoff resubmit, like I included in the integration intkey client.  It should be relatively simple.

peterschwarz:
Rebased against master

PR #1258, opened 2017-12-19
CXX SDK Rework

EugeneYYY:
Rework to CXX SDK build and API. Protobuf example is added to intkey CXX TP and intkey CLI.

### DIFF
--- a/sdk/cxx/include/sawtooth/transaction_handler.h
+++ b/sdk/cxx/include/sawtooth/transaction_handler.h
@@ -19,98 +19,106 @@
 #include <list>
 #include <string>
 
+#include "sawtooth_sdk.h"
 #include "proto/transaction.pb.h"
 
 #include "sawtooth/global_state.h"
 
 namespace sawtooth {
 
-// Utility types to allow for managed instances of std::string on the heap.
-typedef std::unique_ptr<std::string> StringUPtr;
-typedef std::shared_ptr<std::string> StringPtr;
+class TransactionHeaderImpl: public TransactionHeader {
+public:
+    TransactionHeaderImpl(::TransactionHeader* header): header_(header) {};
+    virtual ~TransactionHeaderImpl() {
+        if (header_ != NULL) {
+            delete header_;
+        }
+    };
 
-// The transaction data for a Transaction Processing request.
-class Transaction final {
- public:
-    Transaction(TransactionHeader* header, StringPtr payload, StringPtr signature):
-            payload_(payload), signature_(signature) {
-    }
+    virtual int GetCount(TransactionHeaderField field) {
+        int count = 0;
+        switch (field) {
+            case TransactionHeaderStringDependencies:
+                count = header_->dependencies_size();
+                break;
 
-    Transaction (const Transaction&) = delete;
-    Transaction (const Transaction&&) = delete;
-    Transaction& operator= (const Transaction&) = delete;
+            case TransactionHeaderInputs:
+                count = header_->inputs_size();
+                break;
 
-    const TransactionHeader& header() const {
-        return this->header_;
-    }
+            case TransactionHeaderOutputs:
+                count = header_->outputs_size();
+                break;
 
-    const std::string& payload() const {
-        return *(this->payload_);
-    }
+            case TransactionHeaderNonce:
+            case TransactionHeaderFamilyName:
+            case TransactionHeaderFamilyVersion:
+            case TransactionHeaderPayloadSha512:
+            case TransactionHeaderBatcherPublicKey:
+            case TransactionHeaderSignerPublicKey:
+                count = 1;
+                break;
 
-    const std::string& signature() const {
-        return *(this->signature_);
-    }
+            default:
+                count = 0;
+                break;
+        }
 
- private:
-    TransactionHeader header_;
-    StringPtr payload_;
-    StringPtr signature_;
+        return count;
+    };
+
+    virtual const ::std::string& GetValue(TransactionHeaderField field, int index) {
+        int count = 0;
+
+        switch (field) {
+            case TransactionHeaderStringDependencies:
+                return header_->dependencies(index);
+                break;
+
+            case TransactionHeaderInputs:
+                return header_->inputs(index);
+                break;
+
+            case TransactionHeaderOutputs:
+                return header_->outputs(index);
+                break;
+
+            case TransactionHeaderNonce:
+                return header_->nonce();
+                break;
+
+            case TransactionHeaderFamilyName:
+                return header_->family_name();
+                break;
+
+            case TransactionHeaderFamilyVersion:
+                return header_->family_version();
+                break;
+
+            case TransactionHeaderPayloadSha512:
+                return header_->payload_sha512();
+                break;
+
+            case TransactionHeaderBatcherPublicKey:
+                return header_->batcher_public_key();
+                break;
+
+            case TransactionHeaderSignerPublicKey:
+                return header_->signer_public_key();
+                break;
+
+            default:
+                return dummy;
+                break;
+        }
+        return dummy;
+     };
+
+private:
+    ::TransactionHeader* header_; // pointer to protobuf generated object
+    ::std::string dummy;
 };
-typedef std::unique_ptr<Transaction> TransactionUPtr;
 
 
-// TransactionApplicator is an interface that defined to represent the
-// processing of a single transaction. This object should make no assumptions
-// about the thread and timing of when it will execute. The
-// TransactionApplicator owns the reference to the transaction and the global
-// state it is passed.
-class TransactionApplicator {
- public:
-    TransactionApplicator(sawtooth::TransactionUPtr txn,
-        sawtooth::GlobalStateUPtr state) : txn(std::move(txn)),
-        state(std::move(state)) {}
-    virtual ~TransactionApplicator() {}
-
-
-    // Process the transaction, this function should do all of the processing
-    // of the given transaction. The general framework should be:
-    // 1) Validate the transaction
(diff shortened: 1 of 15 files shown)


bin/build_cxx
```diff
@@ -15,16 +15,28 @@
...
-
-echo -e "\033[0;32m--- Building cxx sdk ---\n\033[0m"
-cd $top_dir/sdk/cxx
-mkdir -p build
-cd build
-# To build with debug symbols use:
-# cmake -DCMAKE_BUILD_TYPE=Debug ..
-cmake ..
-make
+#set -e
```
vaporos: The changes here have multiple flaws: a) does not properly use top_dir, which then requires the script to be run in a specific directory; b) commented out code.

sdk/cxx/include/sawtooth/transaction_handler.h
```diff
@@ -19,98 +19,113 @@
...
-
-    const std::string& signature() const {
-        return *(this->signature_);
-    }
-
- private:
-    TransactionHeader header_;
-    StringPtr payload_;
-    StringPtr signature_;
+class TxHeaderWrapper: public TxHeaderIF
```
vaporos: We do not use 'Tx' to mean Transaction anywhere else in the code, and generally frown on abbreviations.  The C++ SDK should generally be made to look closer to the other SDKs.

sdk/cxx/src/transaction_processor.cpp
```diff
@@ -88,11 +88,12 @@ void TransactionProcessor::HandleProcessingRequest(const void* msg,
         request.ParseFromArray(msg, msg_size);
 
         TransactionHeader* txn_header(request.release_header());
+        TxHeaderPtr txHeader(new TxHeaderWrapper(txn_header)); // YYY TBD
```
vaporos: What does YYY TBD mean?

EugeneYYY: My initials

EugeneYYY:
PROTOBUF example is added per request from the POC team. It doesn't affect existing functionality. By default it is not used. PROTOBUF is used instead of CBOR only if -p or --protobuf option is specified. 
Overall, the changes make CXX SDK and example to behave in line with examples in other languages at least in terms of command line behavior.

sdk/cxx/include/sawtooth/transaction_handler.h
EugeneYYY: I can change Tx to Transaction, but curios why Tx is not welcomed. It makes variables shorter

sdk/cxx/src/transaction_processor.cpp
peterschwarz: We don't allow TODO comments in the code base - they should be reflected in tasks on the Jira board.  Likewise, a comment like that gives no context to its meaning to anyone other than the original commenter. E.g. what is TBD on that line?

sdk/cxx/include/sawtooth/transaction_handler.h
vaporos: The reason for avoiding abbreviations in general is for code clarity, both for English and non-native-English speakers. Abbreviations tend to cause the reader to require more information.  This is obviously subjective but generally the rule we use.  There are exceptions too, of course, especially when we are declaring local variables, etc.  But specifically here, I think we want consistency with the rest of the SDKs to the extent that makes sense.

vaporos:
We have an intkey implementation for every language supported and multiple clients.  The overhead in adding protobuf support, however simplistic it may seem on the surface, is undesirable both from adding the support everywhere, but in addition the added build and maintenance overhead.  Many other transaction families can be used as reference for how to nicely use protobuf.

EugeneYYY:
The pull request was amended. All comments should be addressed including removing protobuf changes that were temporarily moved to a separate branch "protobuf_example" oin my fork

sdk/cxx/include/sawtooth_sdk.h
```diff
@@ -0,0 +1,258 @@
...
+ private:
+    TransactionHeaderPtr header_;
+    StringPtr payload_;
+    StringPtr signature_;
+};
+typedef std::unique_ptr<Transaction> TransactionUPtr;
+
+
+// Provides access to the global state
+class GlobalStateIF {
```
vaporos: There should be only one implementation of state, which suggests it should not be an interface.

sdk/cxx/include/sawtooth_sdk.h
```diff
@@ -0,0 +1,258 @@
...
+    virtual void DeleteState(const std::string& address) const = 0;
+
+    // Delete multiple entries from global state.
+    virtual void DeleteState(const std::vector<std::string>& address) const = 0;
+};
+typedef std::shared_ptr<GlobalStateIF> GlobalStatePtr;
+typedef std::unique_ptr<GlobalStateIF> GlobalStateUPtr;
+
+
+// TransactionApplicator is an interface that defined to represent the
```
vaporos: This class is inconsistent with the other SDKs.  It seems to largely duplicate the purpose of TransactionHandler.

sdk/cxx/include/sawtooth_sdk.h
```diff
@@ -0,0 +1,258 @@
...
+    // as possible should be done in this function.
+    virtual TransactionApplicatorUPtr GetApplicator(
+        TransactionUPtr txn,
+        GlobalStateUPtr state) = 0;
+};
+typedef std::unique_ptr<TransactionHandler> TransactionHandlerUPtr;
+typedef std::shared_ptr<TransactionHandler> TransactionHandlerPtr;
+
+
+// Helper class to provide mappings from domain identifiers
```
vaporos: AddressMapper and AddressMapperIF are inconsistent with the other SDKs.

For the other SDKs, we are currently designing the specific approach to suggest to users. It will likely not have a generic key/value interface as presented here (which is only interesting to very simplistic mappings) and instead present an approach for mapping the state object with a TF-specific API.  Suggest renaming AddressMapper to IntkeyAddressMapper and moving it to the intkey example.

sdk/cxx/include/sawtooth_sdk.h
```diff
@@ -0,0 +1,258 @@
...
+
+    static AddressMapperIF* Create(const std::string& namespace_);
+};
+
+typedef std::shared_ptr<AddressMapperIF> AddressMapperPtr;
+typedef std::unique_ptr<AddressMapperIF> AddressMapperUPtr;
+
+
+
+class TransactionProcessorIF
```
vaporos: There is only one implementation fo the TransactionProcessor, so the interface does not seem necessary.

sdk/cxx/include/sawtooth_sdk.h
```diff
@@ -0,0 +1,258 @@
...
+    TransactionHeaderFamilyName,
+    TransactionHeaderFamilyVersion,
+    TransactionHeaderInputs,
+    TransactionHeaderNonce,
+    TransactionHeaderOutputs,
+    TransactionHeaderPayloadSha512,
+    TransactionHeaderSignerPublicKey
+};
+
+class TransactionHeaderIF
```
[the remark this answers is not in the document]
EugeneYYY: The goal is to isolate TP developers from SDK details. Most of protobuf is implicitly isolated by other interfaces.  The header object is an exception and, yes, this interface' goal is to isolate header protobuf from the app developers

sdk/cxx/include/sawtooth_sdk.h
EugeneYYY: The primary goal of this interface is to isolate app developers from underlying details of GlobalState implementation, no SDK code is required for building a TP.

sdk/cxx/include/sawtooth_sdk.h
EugeneYYY: It is a good point, but it was taken as is from the existing code as is. Its redesign was not in the scope of this change

sdk/cxx/include/sawtooth_sdk.h
EugeneYYY: The primary goal of this interface is to isolate app developers from underlying details of TransactionProcessor implementation so no SDK code is required for building a TP.

sdk/examples/intkey_cxx/int_key_transaction_processor.cpp
```diff
@@ -226,53 +251,225 @@ class IntKeyApplicator:  public sawtooth::TransactionApplicator {
...
         return sawtooth::TransactionApplicatorUPtr(
             new IntKeyApplicator(std::move(txn), std::move(state)));
     }
+private:
+    std::string namespacePrefix;
 };
 
 
+void Usage(bool bExit = false, int exitCode = 1)
+{
```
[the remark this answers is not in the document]
EugeneYYY: Fixed

sdk/cxx/include/sawtooth_sdk.h
cianx: Unclear which interfaces in the other sdks you are refering to. It rather hard to guess what design work is in progress. This has been present in the CXX  SDK since the first revision. This feed back would have been better suited to the original review, rather then on a small incremental improvement.

sdk/cxx/include/sawtooth_sdk.h
cianx: The transaction handler is a singleton and the CXX sdk allows for multithread processing of transactions. It is not good practice to have multiple threads calling into a function on the same object. The transaction applicator is a transaction specific instance of a processor that can be dispatched to a thread for processing with an explicit lifespan of the processing of that transaction. Again this feedback should have been givin on the initial commit.

PR #1260, opened 2017-12-19
Fix postinst rules for validator package

rberg2:
Add a check that "$2" is not set. On an upgrade "$2"
is the previous package version, on a new install it's
undefined.

### DIFF
--- a/validator/packaging/ubuntu/postinst
+++ b/validator/packaging/ubuntu/postinst
@@ -25,7 +25,8 @@ directories="
 user="sawtooth"
 group="sawtooth"
 
-if [ "$1" = install ]; then
+# Run on install and not upgrade
+if [ "$1" = configure ] && [ -z "$2" ]; then
     if ! getent group $group > /dev/null; then
         addgroup --quiet --system $group
     fi


validator/packaging/ubuntu/postinst
```diff
@@ -25,7 +25,8 @@ directories="
 user="sawtooth"
 group="sawtooth"
 
-if [ "$1" = install ]; then
+# Run on install and not upgrade
+if [ "$1" = configure ] && [ -z "$2" ]; then
```
peterschwarz: What is the expected value of `$2`?  It's not clear from looking at this code.

rberg2: -z returns True if the string length is zero. When dpkg does an new install it calls 'postinst configure' but on an upgrade it calls 'postinst configure previousversion'

rberg2:
RCA: 'install' in an invalid option in a postinst script causing that code to be skipped, it is valid in preinst scripts. I am changing this to 'configure' and checking that we are not upgrading the package.

https://www.debian.org/doc/debian-policy/#maintainer-script-flowcharts

PR #1262, opened 2017-12-19
Remove noisy debug line in poet block publisher

dcmiddle:
If a validator is not yet registered block initialization will
repeatedly return false prompting a debug message that it's not
ready yet. That message hides the useful debug/info/warn messages
that indicate what the validator is waiting on for registration.


dcmiddle:
@vaporos here's a simpler alternative to https://github.com/hyperledger/sawtooth-core/pull/1250. Just get rid of the noisy line then it should be easier to see the root causes which already have meaningful info like block numbers.

dcmiddle:
It was felt this will just mask the problem. #1288 is intended to correct the underlying issue, whereas #1250 was a different approach the logging we may revisit after 1288 is evaluated.

PR #1263, opened 2017-12-19
Disable the LMDB writemap and map_async features when creating a db environment

jsmitchell:


peterschwarz:
Does this potentially address [STL-957](https://jira.hyperledger.org/browse/STL-957)?

PR #1267, opened 2017-12-20
Add block-info package installation to validator image

grkvlt:
Required to add the block-info transaction injector that is configured in `sawtooth-seth` images and compose files

### DIFF
--- a/docker/sawtooth-int-validator
+++ b/docker/sawtooth-int-validator
@@ -45,6 +45,7 @@ RUN cd /debs \
 
 RUN apt-get install -y -q --allow-unauthenticated \
     python3-sawtooth-validator \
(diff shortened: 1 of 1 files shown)


grkvlt:
Hi @aludvik, this is *not* for *running* the `block-info-tp`, it  is to add the python block-info library to the validator so that the validator process can be configured with the `block-info` injector, as is done in many of the example Docker Compose files. This currently only works if the `/project/sawtooth-core` directory is mounted as a volume. It is similar to how `sawtooth-poet` has been installed, too, to enable PoET consensus config.

aludvik:
@grkvlt Oh, yeah you are right and I was wrong.

PR #1270, opened 2017-12-20
Improve logging

nick-drozd:
This PR makes small changes to make logging cheaper and less noisy. The biggest change in terms of output is removing the size of messages from the `%s receiving %s` message in `interconnect`.

### DIFF
--- a/validator/sawtooth_validator/gossip/gossip.py
+++ b/validator/sawtooth_validator/gossip/gossip.py
@@ -448,7 +448,8 @@ class ConnectionManager(InstrumentedThread):
                 if self._peering_mode == 'dynamic':
                     self._refresh_peer_list(self._gossip.get_peers())
                     peers = self._gossip.get_peers()
-                    if len(peers) < self._min_peers:
+                    peer_count = len(peers)
+                    if peer_count < self._min_peers:
                         LOGGER.debug("Below minimum peer threshold. "
                                      "Doing topology search.")
 
@@ -476,8 +477,8 @@ class ConnectionManager(InstrumentedThread):
                                 set(peered_endpoints) -
                                 set([self._endpoint]))
 
-                        LOGGER.debug("Number of peers: %s", len(peers))
-                        LOGGER.debug("Peers are: %s", list(peers.values()))
+                        LOGGER.debug("Number of peers: %s", peer_count)
+                        LOGGER.debug("Peers are: %s", peered_endpoints)
                         LOGGER.debug("Unpeered candidates are: %s",
                                      unpeered_candidates)
 
--- a/validator/sawtooth_validator/gossip/gossip.py
+++ b/validator/sawtooth_validator/gossip/gossip.py
@@ -450,8 +450,12 @@ class ConnectionManager(InstrumentedThread):
                     peers = self._gossip.get_peers()
                     peer_count = len(peers)
                     if peer_count < self._min_peers:
-                        LOGGER.debug("Below minimum peer threshold. "
-                                     "Doing topology search.")
+                        LOGGER.debug(
+                            "Number of peers (%s) below "
+                            "minimum peer threshold (%s). "
+                            "Doing topology search.",
+                            peer_count,
+                            self._min_peers)
 
                         self._reset_candidate_peer_endpoints()
                         self._refresh_peer_list(peers)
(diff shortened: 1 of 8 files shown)


validator/sawtooth_validator/gossip/gossip_handlers.py
```diff
@@ -40,8 +40,8 @@ def __init__(self, gossip):
     def handle(self, connection_id, message_content):
```
nick-drozd: `accuracte` -> `accurate` in commit message

validator/sawtooth_validator/gossip/gossip.py
```diff
@@ -476,8 +477,8 @@ def run(self):
                                 set(peered_endpoints) -
                                 set([self._endpoint]))
 
-                        LOGGER.debug("Number of peers: %s", len(peers))
-                        LOGGER.debug("Peers are: %s", list(peers.values()))
+                        LOGGER.debug("Number of peers: %s", peer_count)
```
vaporos: These should be on 1 line instead of three. We need less lines, more info. (generally speaking)

validator/sawtooth_validator/gossip/gossip_handlers.py
```diff
@@ -61,8 +61,8 @@ def __init__(self, gossip):
...
         response = GetPeersResponse()
         response.ParseFromString(message_content)
-        LOGGER.debug("got peers response message "
-                     "from %s. sending ack", connection_id)
+
+        LOGGER.debug("Got peers response message from %s", connection_id)
```
vaporos: This should be combined with the line below.  Few lines, more information.

validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -278,28 +277,30 @@ def _dispatch_message(self):
...
                 self._get_received_message_counter(tag).inc()
 
-                if zmq_identity is not None:
-                    connection_id = \
-                        self._identity_to_connection_id(zmq_identity)
-                else:
-                    connection_id = \
-                        self._identity_to_connection_id(
-                            self._connection.encode())
+                connection_id = \
```
vaporos: I find the previous format easier to read than this inline if across multiple lines.

validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -309,10 +310,10 @@ def _dispatch_message(self):
                 else:
                     my_future = self._futures.get(message.correlation_id)
 
-                    LOGGER.debug("message round "
-                                 "trip: %s %s",
-                                 get_enum_name(message.message_type),
+                    LOGGER.debug("message round trip: %s %s",
```
vaporos: Seems way too granular, maybe better for metrics than logging?

nick-drozd: I can take that statement out, but anything more complicated is outside the scope of this PR.

validator/sawtooth_validator/gossip/gossip_handlers.py
nick-drozd: fixed

validator/sawtooth_validator/networking/interconnect.py
jsmitchell: yeah, i vote for removing this and any other loggers that happen on every message

nick-drozd: That's a more extreme change. I'll do a separate PR for that in order to keep the resulting flame war from holding up this one.

validator/sawtooth_validator/gossip/gossip.py
aludvik: second

validator/sawtooth_validator/gossip/gossip_handlers.py
aludvik: second

vaporos:
This needs a rebase.

nick-drozd:
Rebased, although since https://github.com/hyperledger/sawtooth-core/pull/1279 went in this PR doesn't do much.

validator/sawtooth_validator/gossip/gossip.py
```diff
@@ -448,7 +448,8 @@ def run(self):
                 if self._peering_mode == 'dynamic':
                     self._refresh_peer_list(self._gossip.get_peers())
                     peers = self._gossip.get_peers()
-                    if len(peers) < self._min_peers:
+                    peer_count = len(peers)
+                    if peer_count < self._min_peers:
                         LOGGER.debug("Below minimum peer threshold. "
```
peterschwarz: You should add the `peer_count` and the `_min_peers`

PR #1271, opened 2017-12-20
Add SettingsCache and SettingsObserver

agunde406:

### DIFF
--- a/families/settings/tests/test_tp_settings.py
+++ b/families/settings/tests/test_tp_settings.py
@@ -56,6 +56,14 @@ class TestSettings(TransactionProcessorTestCase):
         self.validator.respond(
             self.factory.create_set_response(key), received)
 
+    def _expect_add_event(self, key):
+        received = self.validator.expect(
+            self.factory.create_add_event_request(key))
+
+        self.validator.respond(
+            self.factory.create_add_event_response(),
+            received)
+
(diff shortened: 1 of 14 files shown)


validator/sawtooth_validator/execution/executor.py
```diff
@@ -39,6 +40,10 @@
 LOGGER = logging.getLogger(__name__)
 
 
+TpSetting = namedtuple('TpSetting',
```
peterschwarz: Is this just used for storing one setting?  If so, it should probably be named for that particular setting

agunde406: Okay, I was following the format of `self._tp_settings_key = "sawtooth.validator.transaction_families`. Should it be called TransactionFamilies?

peterschwarz: That sounds like a good name

families/settings/tests/test_tp_settings.py
```diff
@@ -56,6 +56,14 @@ def _expect_set(self, key, expected_value):
         self.validator.respond(
             self.factory.create_set_response(key), received)
 
+    def _expect_add_event(self, key):
+        recieved = self.validator.expect(
```
jsmitchell: s/recieved/received/

PR #1273, opened 2017-12-21
Set type of validator args for cli command

grkvlt:
The default arg type is `str` causing the validator to crash if the min or max peer `--minimum-peer-connectivity` and `--maximum-peer-connectivity` arguments are set.


peterschwarz:
Please rebase to get whitelist updates

PR #1274, opened 2017-12-21
Fix try/or use in workload commands

grkvlt:
Use `or_else` to choose between executing `from_hex` or `from_wif` when parsing key files, instead of `or` which executes both.

### DIFF
--- a/perf/sawtooth_workload/src/main.rs
+++ b/perf/sawtooth_workload/src/main.rs
@@ -120,9 +120,7 @@ fn run_batch_command(args: &ArgMatches) -> Result<(), Box<Error>> {
     try!(key_file.read_to_string(&mut buf));
     buf.pop(); // remove the new line
 
-    let private_key = try!(
-        Secp256k1PrivateKey::from_hex(&buf).or(
-            Secp256k1PrivateKey::from_wif(&buf)));
+    let private_key = try!(Secp256k1PrivateKey::from_hex(&buf).or_else(|_| Secp256k1PrivateKey::from_wif(&buf)));
     let context = try!(signing::create_context("secp256k1"));
 
     if let Err(err) = generate_signed_batches(&mut in_file, &mut out_file,
--- a/perf/smallbank_workload/src/main.rs
+++ b/perf/smallbank_workload/src/main.rs
@@ -127,9 +127,7 @@ fn run_batch_command(args: &ArgMatches) -> Result<(), Box<Error>> {
     try!(key_file.read_to_string(&mut buf));
     buf.pop(); // remove the new line
 
-    let private_key = try!(
-        Secp256k1PrivateKey::from_hex(&buf).or(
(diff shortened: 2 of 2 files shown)


perf/sawtooth_workload/src/main.rs
```diff
@@ -120,9 +120,7 @@ fn run_batch_command(args: &ArgMatches) -> Result<(), Box<Error>> {
     try!(key_file.read_to_string(&mut buf));
     buf.pop(); // remove the new line
 
-    let private_key = try!(
-        Secp256k1PrivateKey::from_hex(&buf).or(
-            Secp256k1PrivateKey::from_wif(&buf)));
+    let private_key = try!(Secp256k1PrivateKey::from_wif(&buf)).or(Secp256k1PrivateKey::from_hex(&buf))
```
peterschwarz: Good catch, but wrong fix - it should be:
```
let private_key = try!(Secp256k1PrivateKey::from_hex(&buf)).or_else(|_| Secp256k1PrivateKey::from_wif(&buf))
```

peterschwarz:
Also, recommend that you rebase on master, so that you can take advantage of being on the whitelist now.

perf/sawtooth_workload/src/main.rs
grkvlt: Not sure what the difference between the `or` and `or_else` syntax (I'm kindof new to Rust) but I trust your judgement here and changed to use `or_else` and changed the order back to `from_hex` followed by `from_wif`. I guess it is theoretically possible to have a wif format key that has a base58 encoding that is actually only using the 0-9a-f characters, but that's probably so rare that we don't need to worry about it here!

peterschwarz: Oh, right.  Hadn't thought of that - so probably both reordering _and_ using `or_else` is the solution.

As for the why `or_else` over `or`?  In the `or_else` case, the code is only executed on a failure.  In the `or` case, both are executed, and the successful one is chosen.  I.e. less work in the former over the later.

grkvlt: OK, I understand the `or_else` rationale now, thanks. Also, you were right with the ordering, so I changed it back to `from_hex` then `from_wif`. Once I thought about it a bit more, you _need_ to try hex first, because any valid hex is _also_ valid base58 (so most wif keys will fail to parse as hex, but hex keys will parse as wif) and I believe the wif keys have a different length anyway, so they should _still_ fail to parse even if they have pathological formatting of only `0-9a-f` characters!

peterschwarz:
Just checking in on the re-ordering comment under the line comment (that is now hidden by github, so maybe it was missed): Any more changes to this PR coming or is this good to go?

peterschwarz:
This is superseded by #1318, as we're going to remove WIF all together due an underlying library being removed.

peterschwarz:
Closing due to previous comment.

PR #1275, opened 2017-12-21
DO NOT MERGE Experimental change to limit population est.

dcmiddle:
Cap population estimate using validator registry length.

Needs new unit tests and LR testing before merge.

### DIFF
--- a/consensus/poet/common/sawtooth_poet_common/validator_registry_view/validator_registry_view.py
+++ b/consensus/poet/common/sawtooth_poet_common/validator_registry_view/validator_registry_view.py
@@ -15,7 +15,7 @@
 import hashlib
 
 from sawtooth_poet_common.protobuf.validator_registry_pb2 import ValidatorInfo
-
+from sawtooth_poet_common.protobuf.validator_registry_pb2 import ValidatorMap
 
 _NAMESPACE = hashlib.sha256('validator_registry'.encode()).hexdigest()[0:6]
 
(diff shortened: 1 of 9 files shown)


dcmiddle:
Capping the population estimate seems to have narrowed the standard deviation and reduced long wait times as desired. However, because it cuts off the top of the sample distribution it has the effect of lowering overall wait times. To prevent that from happening we'd need a slightly messier solution that would continue to pass the original estimates (without limit capping).  

// Control output from a multi-day run on LR7
There were 8846 distinct local means
Ignoring 0 one-offs that might be from forks
Average local mean: 539.9448462581959
StdDev local mean: 115.58741689268248
Average local mean including one-offs: 539.9448462581959

Avg min wait time: 23.147107167081167
Max min wait time: 414.3
Min min wait time: 1.26
StdDev min wait time: 37.12503388731272

// Output using this PR on 23 hours of an LR network
There were 508 distinct local means
Ignoring 0 one-offs that might be from forks
Average local mean: 123.2353346456693
StdDev local mean: 82.57227688710711
Average local mean including one-offs: 123.2353346456693

Avg min wait time: 9.12527559055118
Max min wait time: 296.68
Min min wait time: 1.24
StdDev min wait time: 25.09715253719495

dcmiddle:
Ok, not hearing anyone saying they see outlier times so going to close this for now. we can reopen and make it better preserve the intended distribution if this is recognized as a problem in the future.

PR #1281, opened 2017-12-21
Reduce locking acquisition on receiving batches

aludvik:
Need to run a long-running network experiment on this to make sure it doesn't affect stability

### DIFF
--- a/validator/sawtooth_validator/journal/publisher.py
+++ b/validator/sawtooth_validator/journal/publisher.py
@@ -66,12 +66,12 @@ class PendingBatchObserver(metaclass=abc.ABCMeta):
 
 class _PublisherThread(InstrumentedThread):
     def __init__(self, block_publisher, batch_queue,
-                 check_publish_block_frequency):
+                 check_publish_block_interval):
         super().__init__(name='_PublisherThread')
         self._block_publisher = block_publisher
         self._batch_queue = batch_queue
-        self._check_publish_block_frequency = \
-            check_publish_block_frequency
+        self._check_publish_block_interval = \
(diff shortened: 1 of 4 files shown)


validator/sawtooth_validator/journal/publisher.py
```diff
@@ -591,26 +595,30 @@ def _build_candidate_block(self, chain_head):
...
-                # execution.
-                if self._candidate_block and \
-                        self._candidate_block.can_add_batch:
-                    self._candidate_block.add_batch(batch)
-            else:
-                LOGGER.debug("Batch has an unauthorized signer. Batch: %s",
-                             batch.header_signature)
+        self._queued_batch_ids = self._queued_batch_ids[:1]
+        if self._permission_verifier.is_batch_signer_authorized(batch):
+            self._pending_batches.append(batch)
```
dcmiddle: _pending_batches is fiddled with in a few places. I'm a little wary of removing locks from it. 
Have you already assessed risks on these sections:
validator/sawtooth_validator/journal/publisher.py:735

```                    last_batch_index = self._pending_batches.index(last_batch)
                    unsent_batches = \
                        self._pending_batches[last_batch_index + 1:]
                    self._pending_batches = pending_batches + unsent_batches

                    self._pending_batch_gauge.set_value(
                        len(self._pending_batches))
```

And this section
validator/sawtooth_validator/journal/publisher.py:597

```        for batch in self._pending_batches:
            if self._candidate_block.can_add_batch:
                self._candidate_block.add_batch(batch)
            else:
                break
```

PR #1282, opened 2017-12-22
Improve sawnet command usability

aludvik:
@jsmitchell I am pretty sure I got the tree printing to work. Let me know if you can break it. It is based on git's tree printing.

### DIFF
--- a/cli/sawtooth_cli/network_command/compare.py
+++ b/cli/sawtooth_cli/network_command/compare.py
@@ -187,48 +187,214 @@ def print_table(graph, tails):
 
 def print_tree(graph, tails):
     """Print out a tree of blocks starting from the common ancestor."""
-    num_col_width = max(
-        floor(log(max(get_heights(tails)), 10)) + 1,
-        len("NUM"))
-    col_n = 8
+    # Example:
+    # |
+    # | 5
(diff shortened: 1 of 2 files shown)


cli/sawtooth_cli/network_command/compare.py
```diff
@@ -77,31 +79,57 @@ def do_compare_chains(args):
     urls = split_comma_append_args(args.urls)
     users = split_comma_append_args(args.users)
     clients = make_rest_apis(urls, users)
-    chains = get_chain_generators(clients)
+    chains, unreporting = get_chain_generators(clients)
```
peterschwarz: Would be useful to print the unreporting nodes.

aludvik: They get printed here: https://github.com/hyperledger/sawtooth-core/pull/1282/commits/a2748f2de4a7ae191d720dd0ff8e12bc30df40fb#diff-43dc287f9866e819f9773544270cc762R112

peterschwarz: Seems like a side-effect, rather than printing all the unreporting that you return at once.  Makes it hard to track down.

aludvik:
Print unresponsive nodes without depending on a side-effect

PR #1285, opened 2017-12-22
Add limit to block list

agunde406:
This makes the number of blocks returned from block list
configurable. Before, if a block list is run all, blocks would
be listed which pulls a lot of blocks from the database
which causes a Ram increase.

### DIFF
--- a/cli/sawtooth_cli/block.py
+++ b/cli/sawtooth_cli/block.py
@@ -46,12 +46,18 @@ def add_block_parser(subparsers, parent_parser):
         'blockchain, including the block id and number, public keys all '
         'of allsigners, and number of transactions and batches.')
 
-    grand_parsers.add_parser(
+    list_parser = grand_parsers.add_parser(
         'list',
         help='Displays information for all blocks on the current blockchain',
         description=description,
         parents=[base_http_parser(), base_list_parser()],
         formatter_class=argparse.RawDescriptionHelpFormatter)
+
+    list_parser.add_argument(
+        '--limit',
+        type=int,
+        help='the number of blocks to list',
+    )
 
     description = (
         'Displays information about the specified block on '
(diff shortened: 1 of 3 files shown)


cli/sawtooth_cli/block.py
```diff
@@ -46,13 +46,20 @@ def add_block_parser(subparsers, parent_parser):
...
+    list_parser = grand_parsers.add_parser(
         'list',
         help='Displays information for all blocks on the current blockchain',
         description=description,
         parents=[base_http_parser(), base_list_parser()],
         formatter_class=argparse.RawDescriptionHelpFormatter)
 
+    list_parser.add_argument(
+        '--limit',
+        type=str,
```
nick-drozd: Should be `type=int`

cli/sawtooth_cli/rest_client.py
```diff
@@ -114,10 +114,16 @@ def _get(self, path, **queries):
 
     def _get_data(self, path, **queries):
         url = self._base_url + path
+        params = self._format_queries(queries)
+
+        limit = None
+        if "limit" in params:
+            limit = int(params["limit"])
```
nick-drozd: If the argument is parsed as an `int`, this type conversion is unnecessary.

cli/sawtooth_cli/block.py
```diff
@@ -46,13 +46,20 @@ def add_block_parser(subparsers, parent_parser):
...
+        help='the number of blocks to list',
+        default="100"
```
peterschwarz: How do I handle the unlimited case?  I.e. what if I want to fetch all and just stop when I find what I'm looking for?

agunde406: I will remove the default limit of 100 and it if limit is not provided, it will return like it does currently.

PR #1286, opened 2017-12-22
WIP sawnet peers {list,graph}

nick-drozd:
This is a WIP PR to solicit comments for the commands `sawnet peers {list,graph}`.

Currently `sawnet peers list` takes a CSV list of URLs and prints a JSON dict with each URL matched to its peers. The peers are the ones returned by `sawtooth peer list`, ie a list of the network endpoints of the validators. A `--pretty` flag can be passed to pretty-print the JSON. It has been suggested that the default should be to print columns and that a `--json` (or `--yaml` or whatever) flag should be passed if that format is desired.

`sawnet peers graph` takes a CSV list of URLs and writes a Graphviz graph to a file called `peers.dot` (overwriting an existing file if the `--force` flag is passed). If a `--png` flag is passed,  the Graphviz command `dot -Tpng peers.dot -o peers.png` will be executed, converting the graph file to a pretty picture. I assume more file formats can be supported; this particular command was copied from a shell script of @vaporos 's.

The problem with the graph command right now is that there's no general way to correlate a REST API's URL with the network endpoint of its associated validator, so the image produced actually shows two isomorphic but distinct graphs with different names for the nodes. It would be extremely helpful to be able to query for the validator endpoint. There are two options for this:

1) Add a `/status` endpoint that would return the validator's network endpoint, its public key, and maybe other stuff.

2) Change the `/peers` endpoint to include information about the validator being queried.

### DIFF
--- a/bin/generate_cli_output
+++ b/bin/generate_cli_output
@@ -81,6 +81,8 @@ save_usage sawadm genesis
 save_usage sawadm keygen
 
 save_usage sawnet
+save_usage sawnet peers
+save_usage sawnet peers list
 save_usage sawnet compare-chains
 
 save_usage sawtooth-validator
--- a/bin/generate_cli_output
+++ b/bin/generate_cli_output
@@ -83,6 +83,7 @@ save_usage sawadm keygen
 save_usage sawnet
 save_usage sawnet peers
 save_usage sawnet peers list
+save_usage sawnet peers graph
 save_usage sawnet compare-chains
 
 save_usage sawtooth-validator
--- a/cli/sawtooth_cli/network_command/peers.py
+++ b/cli/sawtooth_cli/network_command/peers.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 # ------------------------------------------------------------------------------
 
+import os
 import json
 import shlex
 import subprocess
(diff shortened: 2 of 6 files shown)


grkvlt:
Nice, I'll give this a try shortly and see how it copes with some of my more complex multi-region Sawtooth deployments.

However, I'm a bit wary of adding a `sawnet` dependency on the presence of the `dot` tool on the user's `PATH` and, in general, about installing it... If we can't generate a PNG natively then just output the `.dot` file (which is just text generation, no libraries required, one of the reasons Graphviz is so useful) and the user can deal with it themselves on some other machine/their laptop/etc. Remember the `sawnet` command will be being run in a Docker container a lot of the time, which should have the absolute _minimum_ of packages installed. I know this is not really the case right now, but I'm working on reducing image size, and would rather not do things that increase it further.

I prefer the idea of a new `/status` endpoint, it give better separation of concerns and I can see it being usefully extended for other things, too. Clients can easily make the extra HTTP call to get the information, it's not a huge overhead. Plus, it would be really useful as a health check endpoint for tools like Apache Brooklyn and Kubernetes that are managing a running Sawtooth deployment.

peterschwarz:
This requires a new feature for the REST API to provide proper value, so I suggest we close this PR and re-open when that feature is implemented.

PR #1288, opened 2017-12-24
DO NOT MERGE: Add time-out check for poet signups

dcmiddle:
Fix issue where the validator never figures out its PoET signup
transaction failed.
Add a poet setting for a retry after some number of blocks.
Add logic in block initializer to timeout waiting on stalled
registrations.


consensus/poet/core/sawtooth_poet/poet_consensus/poet_block_publisher.py
```diff
@@ -321,6 +346,28 @@ def initialize_block(self, block_header):
...
+                                                        poet_settings_view,
+                                                        self._block_cache):
+                LOGGER.error('Assuming registration txn failed using PPK '
+                             '%s Retry registration',
+                             poet_key_state.poet_public_key)
+                del self._poet_key_state_store[poet_key_state.poet_public_key]
+                self._register_signup_information(
+                    block_header=block_header,
+                    poet_enclave_module=poet_enclave_module)
+
```
aludvik: Can you refactor this into a function since it looks mostly identical to the above on line 275? Also, the code to construct the consensus_state object is duplicated three times in this function. It should probably be moved to a separate function and we should make sure that consensus_state is only constructed once per call to this function at most, since it is probably expensive.

PR #1290, opened 2017-12-24
Move determine_next/dispatch to own thread pool

peterschwarz:
Moving the determine_next/initial dispatch to its own thread pool should reduce message slowdowns.

Additionally, 

- externalized the priority settings for various message types in the dispatcher
- changed the priorities to a proper enum


peterschwarz:
Additional pool provided no discernible difference.  Will open a PR with the refactor alone.

PR #1291, opened 2017-12-26
Use transaction receipts for tracking  batch updates

peterschwarz:
Instead of incurring the cost of going to disk when checking to see if
a batch has been committed, use the transaction receipts provided  to
determine if a batch has been committed.

### DIFF
--- a/validator/sawtooth_validator/state/batch_tracker.py
+++ b/validator/sawtooth_validator/state/batch_tracker.py
@@ -52,7 +52,8 @@ class BatchTracker(ChainObserver,
         self._block_store = block_store
         self._batch_info = TimedCache(cache_keep_time, cache_purge_frequency)
         self._invalid = TimedCache(cache_keep_time, cache_purge_frequency)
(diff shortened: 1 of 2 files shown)


validator/sawtooth_validator/state/batch_tracker.py
```diff
@@ -61,12 +62,24 @@ def chain_update(self, block, receipts):
...
         with self._lock:
-            for batch_id in self._pending.copy():
-                if self._block_store.has_batch(batch_id):
-                    self._pending.remove(batch_id)
-                    self._update_observers(batch_id,
-                                           ClientBatchStatus.COMMITTED)
+            committed_txn_ids = [receipt.transaction_id
+                                 for receipt in receipts]
+
+            for txn_id in committed_txn_ids:
```
aludvik: Can you add a comment clarifying that this works because batches are committed atomically?

peterschwarz:
Rebased

PR #1292, opened 2017-12-27
Revert "Increase poet sample length"

vaporos:
This reverts commit 495201ca459e6e3da851c3801d99faacd7c6bcc5.

The reverted change is not backwards compatible with networks running
1.0.0rc5, and thus should be mitigated in another manner. The
incompatibility occurs when running a mixed network of validators
(before and after this change), with the following error occurring:

  Block X rejected: Wait certificate check failed - Local mean does not
  match: A != B

Ideally, all PoET settings should be set in the genesis block (or when
switching to PoET from another consensus algorithm), so that these
settings are always on-chain and the default settings are never used on
new networks.


vaporos:
How about we add a note to the release notes for 1.0.0rc6 suggesting this setting (and others) be set along with other PoET settings at genesis time, and then update the documentation to reflect this process prior to 1.0.0rc7?

PR #1294, opened 2018-01-03
Handle communication errors gracefully

aludvik:
Previously the sawnet compare-chains command would blow up if there was
a communication error. This commit adds error handling around this
problem and displays information about all nodes that didn't have a
communication error.

### DIFF
--- a/cli/sawtooth_cli/network_command/compare.py
+++ b/cli/sawtooth_cli/network_command/compare.py
@@ -82,17 +82,47 @@ def do_compare_chains(args):
     urls = split_comma_append_args(args.urls)
     users = split_comma_append_args(args.users)
     clients = make_rest_apis(urls, users)
-    chains, unreporting = get_chain_generators(clients)
-    for node in unreporting:
+
+    broken = []
+
+    chains, errors = get_chain_generators(clients)
+    broken.extend(errors)
+    for node in errors:
         print("Error connecting to node %d: %s" % (node, urls[node]))
     if not chains:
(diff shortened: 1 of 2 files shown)


cli/sawtooth_cli/network_command/compare.py
```diff
@@ -532,28 +586,41 @@ def build_fork_graph(chains, tails):
...
+    Returns:
+        A ForkGraph
+        A list of indexes of the chains that had communication problems.
     """
     graph = ForkGraph()
+    bad_chains = []
 
     # Add tails to the graph first
-    for i, tail in enumerate(tails):
+    for i, tail in tails.items():
```
peterschwarz: Is `i` still an integer value here  or is it now a key?

cli/sawtooth_cli/network_command/compare.py
```diff
@@ -532,28 +586,41 @@ def build_fork_graph(chains, tails):
...
-        return graph
+        return graph, bad_chains
 
     # Chains should now all be at the same height, so we can walk back
     # to common ancestor
     while True:
-        heads = [next(chain) for chain in chains]
-        for i, block in enumerate(heads):
+        heads = []
+        for i, chain in chains.items():
```
peterschwarz: Same question here

cli/sawtooth_cli/network_command/compare.py
aludvik: yes to both

PR #1295, opened 2018-01-03
Add Pyformance metrics around forking behavior

dplumb94:


validator/sawtooth_validator/journal/publisher.py
```diff
@@ -477,6 +479,11 @@ def __init__(self,
         self._pending_batch_gauge = \
             metrics_registry.gauge('pending_batch_gauge') \
             if metrics_registry else None
+        if metrics_registry:
+            self._blocks_published_count = CounterWrapper(
+                metrics_registry.counter('blocks_published_count'))
```
peterschwarz: Is this really 'block_claimed_count'?

validator/sawtooth_validator/journal/back_pressure_handlers.py
```diff
@@ -31,11 +34,21 @@ class ClientBatchSubmitBackpressureHandler(Handler):
...
 
-    def __init__(self, can_accept_fn, queue_info_fn):
+    def __init__(self, can_accept_fn, queue_info_fn, metrics_registry=None):
         self._can_accept = can_accept_fn
         self._queue_info = queue_info_fn
         self._applying_backpressure = False
 
+        if metrics_registry:
+            self._bp_batches_rejected_count = CounterWrapper(
+                metrics_registry.counter('bp_batches_rejected_count'))
```
vaporos: The abbreviation is problematic here, as no one knows what bp means.  Does the longer string fail to work?

validator/sawtooth_validator/journal/chain.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 # ------------------------------------------------------------------------------
 
+# pylint: disable=too-many-lines
```
vaporos: This should really be disabled in pylintrc, not per-file.  This must have been re-enabled accidentally or due to changes in pylint.

validator/sawtooth_validator/journal/chain.py
```diff
@@ -508,6 +516,10 @@ def run(self):
                  self._result["uncommitted_batches"]) = \
                     self._compute_batch_change(new_chain, cur_chain)
 
+                if new_chain[0].previous_block_id != \
+                        self._chain_head.identifier:
+                    self._fork_changes_count.inc()
```
vaporos: Consider "chain_head_moved_to_fork_count".

validator/sawtooth_validator/journal/chain.py
```diff
@@ -747,6 +764,7 @@ def on_block_validated(self, commit_new_block, result):
         """
         try:
             with self._lock:
+                self._blocks_considered_count.inc()
```
vaporos: If this is the number of blocks that passed validation, consider "block_validated_count".  But really it seems like we should have metrics on the resulting status instead.

validator/sawtooth_validator/journal/back_pressure_handlers.py
dplumb94: It just required an ugly line break. I'll change it as the variable names are more important in this case.

dplumb94: Also fails to pass lint as `backpressure_batches_rejected_count` is over 30 characters. Is `batches_rejected_count` too vague in this context?

validator/sawtooth_validator/journal/chain.py
dplumb94: This is the number of blocks that have gone through validation but are not necessarily valid.

PR #1296, opened 2018-01-04
Network optimizations

nick-drozd:
This PR makes some low-level optimizations to interconnect (and one tiny change in the threadpool). It would be nice to be able to quantify exactly how much of a difference (if any) these optimizations make.

### DIFF
--- a/validator/sawtooth_validator/networking/interconnect.py
+++ b/validator/sawtooth_validator/networking/interconnect.py
@@ -172,11 +172,12 @@ class _SendReceive(object):
                 self._connection_timeout)
 
     def _identity_to_connection_id(self, zmq_identity):
-        if zmq_identity not in self._identities_to_connection_ids:
+        try:
+            return self._identities_to_connection_ids[zmq_identity]
+        except KeyError:
             self._identities_to_connection_ids[zmq_identity] = \
                 hashlib.sha512(zmq_identity).hexdigest()
-
-        return self._identities_to_connection_ids[zmq_identity]
+            return self._identities_to_connection_ids[zmq_identity]
 
     def _get_received_message_counter(self, tag):
         if tag not in self._received_message_counters:
@@ -351,15 +352,19 @@ class _SendReceive(object):
         :param msg: protobuf validator_pb2.Message
         """
         zmq_identity = None
-        if connection_id is not None and self._connections is not None:
-            if connection_id in self._connections:
+
+        if self._connections is not None:
+            try:
                 connection_info = self._connections.get(connection_id)
-                if connection_info.connection_type == \
-                        ConnectionType.ZMQ_IDENTITY:
-                    zmq_identity = connection_info.connection
-            else:
+            except KeyError:
                 LOGGER.debug("Can't send to %s, not in self._connections",
                              connection_id)
+
+            try:
+                if connection_info.connection_type == ConnectionType.ZMQ_IDENTITY:
+                    zmq_identity = connection_info.connection
+            except AttributeError:
+                pass
 
         self._ready.wait()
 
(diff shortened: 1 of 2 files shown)


validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -351,15 +348,18 @@ def send_message(self, msg, connection_id=None):
...
-                             connection_id)
+
+        try:
+            connection_info = self._connections.get(connection_id)
+        except KeyError:
+            LOGGER.debug("Can't send to %s, not in self._connections",
+                         connection_id)
+
+        try:
+            if connection_info.connection_type == ConnectionType.ZMQ_IDENTITY:
```
peterschwarz: This could probably just be part of the previous `try/exception` block

validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -406,18 +406,21 @@ def send_last_message(self, msg, connection_id=None):
...
-                return
+        try:
+            connection_info = self._connections.get(connection_id)
+        except KeyError:
+            LOGGER.debug("Can't send to %s, not in self._connections",
+                         connection_id)
+            return
+
+        try:
+            if connection_info.connection_type == ConnectionType.ZMQ_IDENTITY:
```
peterschwarz: This could probably just be part of the previous `try/exception` block

validator/sawtooth_validator/networking/interconnect.py
nick-drozd: `connection_info = self._connections.get(connection_id)` returns `None` if the id isn't there, so the extra `except` has to be there. 

Changing `self._connections.get(connection_id)` to `self._connections[connection_id]` would allow putting everything in one block. Is there a reason it isn't like that already?

nick-drozd:
Some reasoning behind these changes:

`if` checks are not free, so if the `if` clause is executed a lot more often than the `else` clause, it's probably cheaper to do away with the `if` check and handle the exceptional case if it arises. `interconnect` has a lot of `if` checks, and its functions are called frequently, so the question is: are the `else` clauses exceptional cases or not? If they are, they should be handled as exceptions.

Here's an experiment you can run. Apply the following changes to master:

```
diff --git a/validator/sawtooth_validator/networking/interconnect.py b/validator/sawtooth_validator/networking/interconnect.py
index 8702a6c..ecca4ed 100644
--- a/validator/sawtooth_validator/networking/interconnect.py
+++ b/validator/sawtooth_validator/networking/interconnect.py
@@ -353,11 +353,13 @@ class _SendReceive(object):
         zmq_identity = None
         if connection_id is not None and self._connections is not None:
             if connection_id in self._connections:
+                LOGGER.error('send_message IF')
                 connection_info = self._connections.get(connection_id)
                 if connection_info.connection_type == \
                         ConnectionType.ZMQ_IDENTITY:
                     zmq_identity = connection_info.connection
             else:
+                LOGGER.error('send_message ELSE')
                 LOGGER.debug("Can't send to %s, not in self._connections",
                              connection_id)
```

Then run `test_poet_liveness` and count the instances of each of the messages. By my count, `send_message IF` occurs 26173 times, which `send_message ELSE` occurs 4 times (all close to the beginning of the test). The `else` case is exceptional.

Here's another example. Apply the following changes to master:

```
diff --git a/validator/sawtooth_validator/networking/interconnect.py b/validator/sawtooth_validator/networking/interconnect.py
index 8702a6c..6f5fea4 100644
--- a/validator/sawtooth_validator/networking/interconnect.py
+++ b/validator/sawtooth_validator/networking/interconnect.py
@@ -173,8 +173,11 @@ class _SendReceive(object):
 
     def _identity_to_connection_id(self, zmq_identity):
         if zmq_identity not in self._identities_to_connection_ids:
+            LOGGER.error('_identity_to_connection_id IF')
             self._identities_to_connection_ids[zmq_identity] = \
                 hashlib.sha512(zmq_identity).hexdigest()
+        else:
+            LOGGER.error('_identity_to_connection_id ELSE')
 
         return self._identities_to_connection_ids[zmq_identity]
```

Again, run `test_poet_liveness`. By my count, `_identity_to_connection_id IF` occurs 54 times (again, all close to the beginning of the test), while `_identity_to_connection_id ELSE` occurs 51712 times. So it looks like `zmq_identity not in self._identities_to_connection_ids` is an exceptional case that does not need to be checked every time.

The other cases are similar.

nick-drozd:
I've profiled this PR with `yappi` a few times, and it doesn't seem to make much of a difference in the diagrams. I still think the changes here are a good idea, but I'm closing this until I have better results to show.

PR #1299, opened 2018-01-04
Database optimizations

nick-drozd:
This PR makes some small optimizations to hotspots in the database code. At least I think they're optimizations -- it's hard to tell without reliable profiling.

### DIFF
--- a/validator/sawtooth_validator/state/merkle.py
+++ b/validator/sawtooth_validator/state/merkle.py
@@ -31,6 +31,28 @@ NODE_PROTO = {
 TOKEN_SIZE = 2
 
 
+def _decode(encoded):
+    return cbor.loads(encoded)
+
+
+def _encode(value):
+    return cbor.dumps(value, sort_keys=True)
+
+
+def _hash(stuff):
+    return hashlib.sha512(stuff).hexdigest()[:64]
+
+
+def _encode_and_hash(value):
+    packed = _encode(value)
+    return _hash(packed), packed
+
+
+NODE_PROTO_PACKED = _encode(NODE_PROTO)
+
+NODE_PROTO_HASHED = _hash(NODE_PROTO_PACKED)
+
+
 class MerkleDatabase(object):
     def __init__(self, database, merkle_root=INIT_ROOT_KEY):
         self._database = database
@@ -51,7 +73,7 @@ class MerkleDatabase(object):
             node = self._get_by_hash(hash_key)
 
         if node["v"] is not None:
-            yield (path, self._decode(node["v"]))
+            yield (path, _decode(node["v"]))
 
         for child in node["c"]:
             for value in self._yield_iter(path + child, node["c"][child]):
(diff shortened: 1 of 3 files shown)


validator/sawtooth_validator/state/merkle.py
```diff
@@ -31,6 +31,23 @@
 TOKEN_SIZE = 2
 
 
+def _decode(encoded):
```
peterschwarz: These should replace the `self._decode` and `self._encode` calls

nick-drozd: Done. I also went ahead and moved out `_encode_and_hash` for consistency.

validator/sawtooth_validator/state/merkle.py
```diff
@@ -31,6 +31,28 @@
...
+def _hash(stuff):
+    return hashlib.sha512(stuff).hexdigest()[:64]
+
+
+def _encode_and_hash(value):
+    packed = _encode(value)
+    return _hash(packed), packed
+
+
+NODE_PROTO_PACKED = _encode(NODE_PROTO)
```
nick-drozd: Note that because these are constants, there's no need to calculate these in the code at all -- the calculations could be done in the REPL and then the literal results could be inserted into the code. That would be slightly more efficient, but at a severe cost to maintainability.

Instead, calculate them once at the start. A separate `_encode` function is used here because the calculations are done prior to the initialization of any particular `MerkleDatabase`, so `MerkleDatabase._encode` can't be used. I don't think `MerkleDatabase.hash` can be used either because of the way the statements are ordered in the file.

nick-drozd:
Okay, here are some numbers. Actually, here are some pictures with numbers on them.

What I did was run a modified version of the poet liveness test with three validators, parallel schedulers, smallbank TPs. 1000 smallbank transactions were submitted (using the workload generator tool) to validator-0 (the genesis node), and the network ran until everything was processed. Validators were running with the `yappi` profiler. For some reason I couldn't get validator-0 to dump its stats, so these numbers are from the other validators, which got all of their transactions from the genesis node.

I ran the test four times, twice against master and twice with this PR's changes applied.

- validator-1, master
liveness-3-n1-p
_liveness-3-n1-p

- validator-1, PR
liveness-3-n1-p-db
_liveness-3-n1-p-db

- validator-2, master
liveness-3-n2-p
_liveness-3-n2-p

- validator-2, PR
liveness-3-n2-p-db
_liveness-3-n2-p-db

The nodes in the graph contain three numbers: total time spent in the function, total time spent in the function but not its subcalls, and total number of calls. The nodes are colored according to the second metric: brighter-colored nodes represent bigger chunks of time. Dark blue is cheap (either fast or called infrequently), red is expensive (slow or called frequently).

To compare the data, look for the cluster of nodes dealing with database / LMDB functions. As far as I can tell, the two big gains from this PR are seen in two functions: `LMDBNoLockDatabase.contains_key` and `MerkleDatabase._set_kv`. The former doesn't show up on the PR graphs, because it isn't being called nearly as often. I think this is due to this change:

```
     def _get_by_hash(self, key_hash):
-        if key_hash in self._database:
+        try:
             return _decode(self._database.get(key_hash))
-        else:
+        except ValueError:   # value returned from database was None
             raise KeyError("hash {} not found in database".format(key_hash))
```

`_set_kv` doesn't show up on any of the PR graphs but one, and that one with reduced numbers. This is because it was repeatedly hashing and encoding a constant value. Now the constant is hashed and encoded just once.

***

I wouldn't be my life on these numbers or on my interpretation thereof, but they are consistent with what I was expecting.

PR #1301, opened 2018-01-04
Send blocks in increasing order during catchup

aludvik:
Rather than requesting and sending blocks in decreasing order when a new
validator joins a network, send blocks in increasing order so the
validator can start processing them immediately.


aludvik:
This is my first pass at improving the memory usage on catchup.

PR #1302, opened 2018-01-05
PredecessorTree optimizations

nick-drozd:
The `PredecessorTree` takes up more program execution time than you might think. In particular, `PredecessorTreeNode.__init__` takes up a shocking amount of time. This PR makes a few small optimizations there.

### DIFF
--- a/validator/sawtooth_validator/execution/scheduler_parallel.py
+++ b/validator/sawtooth_validator/execution/scheduler_parallel.py
@@ -70,7 +70,7 @@ class PredecessorTree:
             for i in range(0, len(address), self._token_size)
         ]
 
-    def _get(self, address, create=False):
+    def _get_creating(self, address):
         tokens = self._tokenize_address(address)
 
         node = self._root
@@ -78,8 +78,6 @@ class PredecessorTree:
             if token in node.children:
                 node = node.children[token]
             else:
-                if not create:
-                    return None
                 child = PredecessorTreeNode(children={}, readers=[])
                 node.children[token] = child
                 node = child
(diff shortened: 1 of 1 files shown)


nick-drozd:
Some context for the reasoning behind these changes:

I ran the test against PRs 1296 and 1299 combined. The graph showed `PredecessorTreeNode.__init__` with the following numbers: `0.75% (0.75%) 186014x`. That seems pretty high for such an insignificant function, right? So I look and see what's in it ("in it", get it?):

```
class PredecessorTreeNode:
    def __init__(self, children=None, readers=None, writer=None):
        self.children = children if children is not None else {}
        self.readers = readers if readers is not None else []
        self.writer = writer
```

It turns out that `PredecessorTreeNode`s are initialized in exactly two places: `PredecessorTree.__init__` and `PredecessorTree._get`. In both cases, they are initialized without any arguments. So every time a `PredecessorTreeNode` is intitialized, two useless default argument checks are executed. Okay, so get rid of the checks and require the two callers to specify their arguments. (Another option would be to do away with specified arguments altogether and initialize all nodes as empty.)

While we're at it, why not get rid of the `get` wrapper and the `create` argument? `_get` is called exactly twice (in `add_reader` and `set_writer`), and both calls have `create=True`. `_get` has an `if not create` clause, which in every single case is useless, because its callers always want `create`. So why keep the argument at all? Well, the predecessor tree tests use `create=False` to inspect trees, but never use `create=True`. So to make everyone happy, split the function into two slightly different versions, one which always creates and one which never does. This allows getting rid of the extra function call and the useless argument check. This function gets called a lot, so the savings might be noticeable, though at a slight cost to maintainability.

nick-drozd:
The numbers aren't coming out the way I expected, so I'm closing this for now.

PR #1303, opened 2018-01-05
Remove non injected batches from block before broacasting

dplumb94:
This PR removes all batches (besides injected batches) from a block before it is broadcasted to the validator network. This will be an improvement for two main reasons:

1. Most nodes are already likely to have these batches in their pending batch queues from previous gossip messages, so sending these batches is likely to repeat work.
2. This will greatly reduce the message size of broadcasted blocks.

Injected batches will still be broadcasted. At the time of broadcast, only the injecting node will have these batches, and if they were omitted from the block all other nodes would request the batches at the same time.

### DIFF
--- a/validator/sawtooth_validator/journal/block_sender.py
+++ b/validator/sawtooth_validator/journal/block_sender.py
@@ -15,6 +15,8 @@
 
 import abc
 
+from sawtooth_validator.protobuf.block_pb2 import Block
+
 
 class BlockSender(object, metaclass=abc.ABCMeta):
     """Implementations should take classes like completer,
@@ -23,7 +25,7 @@ class BlockSender(object, metaclass=abc.ABCMeta):
     """
 
     @abc.abstractmethod
-    def send(self, block):
+    def send(self, block, keep_batches=None):
         """Sends the block to the completer and also to the
            gossip network.
         :param block:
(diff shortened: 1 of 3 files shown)


aludvik:
Also, in response to "Injected batches will remain in the block because there is no way that another node would have them." that isn't exactly true. The reason for leaving them in is to prevent all the nodes on the network from requesting the injected batches at the same time when only the injecting node will have them.

aludvik:
Per offline conversation, the batches should be stripped in the block sender so they aren't removed prior to sending to the completer. But I would still suggest passing the batches to keep through the send method instead of attaching it to the BlockWrapper

validator/sawtooth_validator/journal/block_sender.py
```diff
@@ -37,6 +40,24 @@ def __init__(self, completer, gossip):
...
+            self._remove_batches(block, keep_batches=injected_batch_ids))
         self._completer.add_block(block)
+
+    def _remove_batches(self, block, keep_batches):
+        """Returns a copy of the block without the non-injected
+        batches, which other validators are likely to already
+        have. This reduces the size of the broadcasted block.
+        """
+        block_header = BlockHeader()
+        block_header.ParseFromString(block.header)
```
aludvik: This introduces a redundant serialization/deserialization step, which will affect performance. I would probably do this clone manually rather than use the block builder since it isn't a lot of work.

```
clone = Block()
clone.header = block.header
clone.header_signature = block.header_signature
clone.batches = [batch for batch in block.batches if batch.header_signature in keep_batches]
return clone
```

validator/sawtooth_validator/journal/block_sender.py
```diff
@@ -37,6 +40,24 @@ def __init__(self, completer, gossip):
...
+        block_builder = BlockBuilder(block_header)
+        block_builder.set_signature(block.header_signature)
+
+        if keep_batches is not None:
```
aludvik: I would move this check up to the send function and only call `_remove_batches` when there are batches to remove.

PR #1310, opened 2018-01-11
Request Chain Head from Peers, if None

peterschwarz:
In the edge case where a node that joins a new network, where the only
block produced is the genesis block, and the network is larger than the
mimimum peering threashold, the a node can be left in a state where it
has no blocks, until new blocks are published.  This leaves the node in
a precarious state, whereby it cannot publish any blocks for the batches
it receives until its network is ready.

This change modifieds the peer search to include a check for an existing
chain.  If no chain exists, it will query its connected peers until it
receives a block.

Additionally:

- Update doc string with missing params
- separate dynamic peering retry to own function

### DIFF
--- a/validator/sawtooth_validator/gossip/gossip.py
+++ b/validator/sawtooth_validator/gossip/gossip.py
@@ -444,52 +444,8 @@ class ConnectionManager(InstrumentedThread):
         while not self._stopped:
             try:
                 if self._peering_mode == 'dynamic':
-                    self._refresh_peer_list(self._gossip.get_peers())
-                    peers = self._gossip.get_peers()
-                    peer_count = len(peers)
-                    if peer_count < self._min_peers:
-                        LOGGER.debug(
-                            "Number of peers (%s) below "
-                            "minimum peer threshold (%s). "
-                            "Doing topology search.",
-                            peer_count,
-                            self._min_peers)
-
-                        self._reset_candidate_peer_endpoints()
-                        self._refresh_peer_list(peers)
-                        # Cleans out any old connections that have disconnected
-                        self._refresh_connection_list()
-                        self._check_temp_endpoints()
-
-                        peers = self._gossip.get_peers()
-
-                        self._get_peers_of_peers(peers)
-                        self._get_peers_of_endpoints(
-                            peers,
-                            self._initial_seed_endpoints)
-
-                        # Wait for GOSSIP_GET_PEER_RESPONSE messages to arrive
-                        time.sleep(self._response_duration)
-
-                        peered_endpoints = list(peers.values())
-
-                        with self._lock:
-                            unpeered_candidates = list(
-                                set(self._candidate_peer_endpoints) -
-                                set(peered_endpoints) -
-                                set([self._endpoint]))
-
-                        LOGGER.debug(
-                            "Peers are: %s. "
-                            "Unpeered candidates are: %s",
-                            peered_endpoints,
-                            unpeered_candidates)
-
-                        if unpeered_candidates:
-                            self._attempt_to_peer_with_endpoint(
-                                random.choice(unpeered_candidates))
-
-                if self._peering_mode == 'static':
+                    self.retry_dynamic_peering()
+                elif self._peering_mode == 'static':
                     self.retry_static_peering()
 
                 time.sleep(self._check_frequency)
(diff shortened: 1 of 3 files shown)


nick-drozd:
Typos in third commit message

validator/sawtooth_validator/gossip/gossip.py
```diff
@@ -445,13 +453,19 @@ def start(self):
         super().start()
 
     def run(self):
+        has_chain_head = self._current_chain_head_func is not None
```
nick-drozd: Should the function be called here?

peterschwarz:
Fixed typos

validator/sawtooth_validator/gossip/gossip.py
```diff
@@ -445,13 +453,19 @@ def start(self):
...
         while not self._stopped:
             try:
                 if self._peering_mode == 'dynamic':
                     self.retry_dynamic_peering()
                 elif self._peering_mode == 'static':
                     self.retry_static_peering()
 
+                has_chain_head = has_chain_head or \
```
nick-drozd: I think this would be clearer as
```
if not has_chain_head:
    has_chain_head = self._current_chain_head_func() is not None
```

peterschwarz: In that case, you need a second if block to react to the chain head still being not none, since it being set is an asynchronous operation - I.e. it may have been reset.  Using the or is equivalent, and makes the operation extremely cheap for all subsequent checks (no trips to the block store).

peterschwarz:
Fixed up test_peer_list to remove a sleep and replace it with a `wait_for_rest_apis` call.  This also has the added benefit of testing this change.

validator/sawtooth_validator/gossip/gossip.py
```diff
@@ -474,6 +488,23 @@ def stop(self):
...
+        peered_connecions = [conn_id for conn_id in peers
+                             if self._connection_statuses[conn_id] ==
+                             PeerStatus.PEER]
+
+        if not peered_connecions:
+            return
+
+        LOGGER.debug('Have not received a chain head from peers. '
+                     'Requesting from %s',
+                     peered_connecions)
```
aludvik: Per @nick-drozd logging guidelines, "Logging should not be done by low-level functions" this log message should be removed or moved to a higher level function.

validator/sawtooth_validator/gossip/gossip.py
```diff
@@ -474,6 +488,23 @@ def stop(self):
...
+
+        for conn_id in peered_connecions:
+            self._gossip.send_block_request("HEAD", conn_id)
```
aludvik: This request already happens whenever peering with a potential peer is successful: https://github.com/hyperledger/sawtooth-core/blob/master/validator/sawtooth_validator/gossip/gossip.py#L774

With that context, my interpretation of this change is that a node will now repeatedly send HEAD requests to the peers it has already successfully connected to until it has a chain head, while it is in dynamic peering. Is that correct?

aludvik: If so, can the other request for HEAD be removed now, since it will happen through this channel now?

peterschwarz: No - the added channel is only for the degenerate case where a node has no chain, and neither do its immediate peers.  

This initial request is still valid, as it still would need to query peers when reconnecting to a network (e.g. after it disconnected or crashed) - it has a chain, but it may be out-of-date.

validator/sawtooth_validator/gossip/gossip.py
peterschwarz: This is potentially the only place to detect this change, so the log message lives at its appropriate level in the code.

peterschwarz:
@aludvik Added comments to the last commit

validator/sawtooth_validator/gossip/gossip.py
aludvik: I was suggesting moving it to the `run()` method which is high-level instead of in this lower level helper function. Line 471 in this file

validator/sawtooth_validator/gossip/gossip.py
```diff
@@ -474,6 +499,25 @@ def stop(self):
...
+                if self._connection_statuses[conn_id] == PeerStatus.PEER]
+
+    def _request_chain_head(self, peered_connections):
+        """Request chain head from the given peer ids.
+
+        Args:
+            peered_connecions (:list:str): a list of peer connection ids where
+                the requests will be sent.
+        """
+        if not peered_connections:
```
nick-drozd: This check is redundant, both because it's already checked in `run` and because iterating over an empty is fine.

PR #1317, opened 2018-01-12
Clarify which commands must run as sawtooth

chenette:
Add a [sawtooth@system] prompt to mark the commands that must be run
as the sawtooth user.

Reorganize content to make this part of the procedure less confusing:
- Move the "skip for multiple validators" note to the right place.
- Make the "sawtooth proposal" details into a note and move after
  the command output.

These changes address STL-989.


peterschwarz:
The opening part of the commit messages should not have the jira issue in it (it's in the contributor guide under commit messages).

PR #1319, opened 2018-01-12
Make Long Running Smallbank workload command

boydjohnson:
@nick-drozd had mentioned that it was beneficial to have a way to produce a set number of batches for testing, so I left in that functionality.

After this PR there needs to be work to allow this tool to run for days. That work is: better error handling in the futures::Stream object that does the timing of http request as well as the actual requests. i think that every closure used in a combinator off of the Stream object should take and return Result<Option<_>, WorkloadError>, to allow for handling errors without the stream ending.

The second issue is that the Iterator for Smallbank payloads was originally built to output a certain number of transactions and that number was passed to the iterator. The iterator should output endless payloads and the caller should use .take(num_txns) to take that number.


perf/sawtooth_perf/src/batch_gen.rs
```diff
@@ -130,36 +130,70 @@ impl<'a> SignedBatchProducer<'a> {
...
+    batch.set_transactions(protobuf::RepeatedField::from_vec(txns));
 
-        // set signer_public_key
-        let txn_ids = txns.iter().cloned().map(|mut txn| txn.take_header_signature()).collect();
-        batch_header.set_transaction_ids(protobuf::RepeatedField::from_vec(txn_ids));
-        batch_header.set_signer_public_key(self.signer.get_public_key()?.as_hex());
+    Ok(Some(batch))
+}
+
+pub trait BatchIter {
```
peterschwarz: This should just be the `std::iter::Iterator`

perf/sawtooth_perf/src/batch_submit.rs
```diff
@@ -222,3 +224,37 @@ impl<'a> BatchListFeeder<'a> {
         Ok(Some(batch_list))
     }
 }
+
+pub trait BatchListIter {
```
peterschwarz: This should just be the `std::iter::Iterator`

perf/sawtooth_perf/src/batch_submit.rs
```diff
@@ -153,19 +166,343 @@ pub fn http_submitter(target: String, rate: u64, receiver: Arc<Mutex<mpsc::Recei
...
+        .map(|batch_list| {
+            let bytes = batch_list?.write_to_bytes()?;
+            let mut req = Request::new(Method::Post, Uri::from_str(batch_url.as_ref())?);
+            let content_len = bytes.len() as u64;
+            req.set_body(bytes);
+            req.headers_mut().set(ContentType::octet_stream());
+            req.headers_mut().set(ContentLength(content_len));
+            Ok(req)
+        }).and_then(|req: Result<Request, WorkloadError>| {
+            let counter_cc = counter_c.clone();
```
peterschwarz: This seems like a lot of extra cloning

boydjohnson: There are also a lot of scopes, with each scope passing it off to another closure. These examples in the docs seem to show calling clone() on an Arc being how to share it. https://doc.rust-lang.org/std/sync/struct.Arc.html#examples

perf/sawtooth_perf/src/batch_submit.rs
boydjohnson: Sounds good. I will update.

perf/sawtooth_perf/src/batch_gen.rs
boydjohnson: Sounds good.

perf/sawtooth_perf/src/batch_gen.rs
```diff
@@ -169,6 +215,7 @@ mod tests {
     use super::LengthDelimitedMessageSource;
     use super::TransactionSource;
     use super::SignedBatchProducer;
+    use super::BatchIter;
```
peterschwarz: This looks left-over from replacing this with Iterator

perf/sawtooth_perf/src/batch_gen.rs
```diff
@@ -130,36 +130,82 @@ impl<'a> SignedBatchProducer<'a> {
...
-        if txns.len() == 0 {
-            return Ok(None);
-        }
+        batch_transactions(txns, self.signer)
+    }
+}
 
-        let mut batch_header = BatchHeader::new();
+fn batch_transactions(txns: Vec<Transaction>, signer: &signing::Signer) -> Option<BatchResult> {
+    if txns.len() == 0 {
```
aludvik: Why should this be treated as invalid? Within the context of this function, it seems like passing a zero length transaction vector should be valid and checking the length should be done outside the function? Adding this check makes the return type more complex.

perf/sawtooth_perf/src/batch_gen.rs
```diff
@@ -130,36 +130,82 @@ impl<'a> SignedBatchProducer<'a> {
...
+    let public_key = pk.as_hex();
 
-        // set signer_public_key
-        let txn_ids = txns.iter().cloned().map(|mut txn| txn.take_header_signature()).collect();
-        batch_header.set_transaction_ids(protobuf::RepeatedField::from_vec(txn_ids));
-        batch_header.set_signer_public_key(self.signer.get_public_key()?.as_hex());
 
-        let header_bytes = batch_header.write_to_bytes().unwrap();
-        let signature = try!(self.signer.sign(&header_bytes).map_err(BatchingError::SigningError));
+    let txn_ids = txns.iter().cloned().map(|mut txn| txn.take_header_signature()).collect();
```
aludvik: Since you are only taking the header signatures, cloning the entire transaction is probably wasted effort. I would favor something that only clones the header signature like:
`let txn_ids = txns.iter().map(|txn| String::from(txn.get_header_signature())).collect();`

perf/sawtooth_perf/src/batch_gen.rs
```diff
@@ -130,36 +130,82 @@ impl<'a> SignedBatchProducer<'a> {
...
+    batch_header.set_transaction_ids(protobuf::RepeatedField::from_vec(txn_ids));
+    batch_header.set_signer_public_key(public_key);
 
-        let mut batch = Batch::new();
-        batch.set_header_signature(signature);
-        batch.set_header(header_bytes);
-        batch.set_transactions(protobuf::RepeatedField::from_vec(txns));
+    let header_bytes = batch_header.write_to_bytes().unwrap();
```
aludvik: Is it safe to unwrap this?

perf/sawtooth_perf/src/batch_gen.rs
```diff
@@ -130,36 +130,82 @@ impl<'a> SignedBatchProducer<'a> {
...
+    let signature = signer.sign(&header_bytes).map_err(BatchingError::SigningError);
```
aludvik: Favor the `?` operator to matching on an error. If you make this function return BatchResult instead of Option<BatchResult> this is possible.

perf/sawtooth_perf/src/batch_submit.rs
```diff
@@ -154,19 +167,388 @@ pub fn http_submitter(target: String, rate: u64, receiver: Arc<Mutex<mpsc::Recei
...
+                    }).and_then(|_| future::ok(()) ).map_err(|_| () );
+                Ok(handle_clone_3.spawn(work))
+            },
+            
+            Err(err) => Err(err),
+            }
+            }).for_each(|_| future::ok(())).map_err(|err| WorkloadError::from(err));
+
+    core.run(stream_work)
+}
```
aludvik: Some of the above closures should be moved to first class functions if possible so it is easier to understand what is happening above. The above is very hard to make sense of. It may even be a good idea to refactor the above into a struct/impl and move it out of this file.

perf/smallbank_workload/src/main.rs
```diff
@@ -127,9 +143,70 @@ fn run_load_command(args: &ArgMatches) -> Result<(), Box<Error>> {
...
+    let mut transaction_iterator = SmallbankGeneratingIter::new(accounts, usize::max_value(), seed.as_slice())
+        .map(|payload| {transformer.payload_to_transaction(payload)}).map(|item| item.unwrap());
+
+    let mut batch_iter = SignedBatchIterator::new(&mut transaction_iterator, max_txns, &signer);
+    let mut batchlist_iter = InfiniteBatchListIterator::new(&mut batch_iter);
 
+    let time_to_wait: u32 = 1_000_000_000 / rate as u32;
+
+    let result = run_workload(&mut batchlist_iter, time_to_wait, update, target);
+    match result {
```
aludvik: Prefer `match run_workload(...) {`

perf/sawtooth_perf/src/batch_submit.rs
peterschwarz: This could be done in a separate PR - it doesn't necessarily gate the functionality.

boydjohnson:
@aludvik @peterschwarz I have fixed each comment except for the refactor of the combinators with closures. I have worked on that but it will be part of a later PR.

perf/smallbank_workload/src/main.rs
```diff
@@ -344,6 +479,58 @@ fn run_playlist_process_command(args: &ArgMatches) -> Result<(), Box<Error>> {
...
+        let hash: &mut [u8] = & mut [0; 64];
+        sha.result(hash);
+
+        txn_header.set_payload_sha512(bytes_to_hex_str(hash));
+        txn_header.set_signer_public_key(self.signer.get_public_key()?.as_hex());
+        txn_header.set_batcher_public_key(self.signer.get_public_key()?.as_hex());
+
+        let header_bytes = try!(txn_header.write_to_bytes());
+
+        let signature = try!(self.signer.sign(&header_bytes.to_vec()));
```
aludvik: Use `?;` instead of `try!()`: https://doc.rust-lang.org/std/macro.try.html

perf/sawtooth_perf/src/batch_submit.rs
boydjohnson: https://jira.hyperledger.org/browse/STL-1014 I opened a Jira story to do the refactor.

perf/smallbank_workload/src/main.rs
boydjohnson: Thanks, @aludvik . I will fix those try!s and https://jira.hyperledger.org/browse/STL-1014 here is the Jira story.

boydjohnson:
@aludvik Here is the story to refactor the closures: https://jira.hyperledger.org/browse/STL-1014. I have removed the try!s for ?.

PR #1320, opened 2018-01-12
Make Settings View LRU Cache per Instance

peterschwarz:
Use of lru_cache on SettingsView.get_setting created a global setting
cache, which cuased the networks to fork for a significant amount of
time, early in their runs.  Changing this to a per-instance cache, which
was the original intention, fixes the issue.

### DIFF
--- a/validator/sawtooth_validator/state/settings_view.py
+++ b/validator/sawtooth_validator/state/settings_view.py
@@ -47,8 +47,13 @@ class SettingsView(object):
         """
         self._state_view = state_view
 
-    @lru_cache(maxsize=128)
-    def get_setting(self, key, default_value=None, value_type=str):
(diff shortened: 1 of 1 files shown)


validator/sawtooth_validator/state/settings_view.py
```diff
@@ -47,8 +47,13 @@ def __init__(self, state_view):
         """
         self._state_view = state_view
 
-    @lru_cache(maxsize=128)
-    def get_setting(self, key, default_value=None, value_type=str):
+        # The public method for get_settings should have its results  memoized
```
jsmitchell: extra space between results and memoized

dcmiddle: so you're going to call out an extra space but not a double the?

PR #1325, opened 2018-01-15
Enhance rust sdk build

vaporos:
Transitions the Rust SDK from using protogen to build.rs.

### DIFF
--- a/docker/sawtooth-dev-rust
+++ b/docker/sawtooth-dev-rust
@@ -41,8 +41,13 @@ RUN echo "deb http://repo.sawtooth.me/ubuntu/ci xenial universe" >> /etc/apt/sou
     openssl \
     pkg-config \
     python3-grpcio-tools=1.1.3-1 \
+    unzip \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/*
+
+RUN curl -OLsS https://github.com/google/protobuf/releases/download/v3.5.1/protoc-3.5.1-linux-x86_64.zip \
+ && unzip protoc-3.5.1-linux-x86_64.zip -d protoc3 \
+ && rm protoc-3.5.1-linux-x86_64.zip
 
 RUN curl https://sh.rustup.rs -sSf > /usr/bin/rustup-init \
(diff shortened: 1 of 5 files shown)


docker/sawtooth-dev-rust
```diff
@@ -41,24 +41,28 @@ RUN echo "deb http://repo.sawtooth.me/ubuntu/ci xenial universe" >> /etc/apt/sou
     openssl \
     pkg-config \
     python3-grpcio-tools=1.1.3-1 \
+    unzip \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/*
 
+RUN curl -OL https://github.com/google/protobuf/releases/download/v3.5.1/protoc-3.5.1-linux-x86_64.zip \
+ && unzip protoc-3.5.1-linux-x86_64.zip -d protoc3
```
peterschwarz: Could probably clean up the zip file once you're done with it, for image size reasons.

docker/sawtooth-dev-rust
```diff
@@ -41,24 +41,28 @@ RUN echo "deb http://repo.sawtooth.me/ubuntu/ci xenial universe" >> /etc/apt/sou
...
  && rustup-init -y
 
-EXPOSE 4004/tcp
+#EXPOSE 4004/tcp
 
-ENV PATH=$PATH:/project/sawtooth-core/bin:/root/.cargo/bin
+ENV PATH=$PATH:/protoc3/bin:/project/sawtooth-core/bin:/root/.cargo/bin
 
-RUN cargo install protobuf
+#RUN cargo install protobuf
```
agunde406: There are several lines that are commented out but still in the file. These should be removed.

docker/sawtooth-dev-rust
vaporos: Changed this so it removes the zip file. Also added -sS to curl.

docker/sawtooth-dev-rust
vaporos: Well, that's embarrassing.  Updated to actually remove those lines.

PR #1326, opened 2018-01-15
Create new sawadm blockstore command (in rust)

aludvik:
This creates a new subcommand for the `sawadm` command called `blockstore` which allows users to directly inspect and modify the blockstore database. The `blockstore` subcommand names and arguments should be treated as unstable as additional work is being done on this command.

The new command was written in rust and so the old command has been removed.

### DIFF
--- a/adm/src/blockstore.rs
+++ b/adm/src/blockstore.rs
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2018 Intel Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ------------------------------------------------------------------------------
+ */
+
+use protobuf;
+use protobuf::Message;
+use sawtooth_sdk::messages::block::{Block, BlockHeader};
+
+use database::database::DatabaseError;
+use database::lmdb::LmdbDatabase;
+
+pub struct Blockstore<'a> {
+    db: LmdbDatabase<'a>,
+}
+
+impl<'a> Blockstore<'a> {
+    pub fn new(db: LmdbDatabase<'a>) -> Self {
+        Blockstore{
+            db: db,
+        }
+    }
+
+    pub fn get(&self, block_id: &str) -> Result<Block, DatabaseError> {
+        let reader = self.db.reader()?;
+        let packed = reader.get(&block_id.as_bytes())
+            .ok_or(DatabaseError::NotFoundError(format!("Block not found: {}", block_id)))?;
+        let block: Block = protobuf::parse_from_bytes(&packed)
+            .map_err(|err|
+                DatabaseError::CorruptionError(
+                    format!("Could not interpret stored data as a block: {}", err)))?;
+        Ok(block)
+    }
+
+    pub fn put(&self, block: Block) -> Result<(), DatabaseError> {
+        let block_header: BlockHeader = protobuf::parse_from_bytes(&block.header).map_err(|err|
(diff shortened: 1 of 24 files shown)


adm/src/blockstore.rs
```diff
@@ -0,0 +1,122 @@
...
+                DatabaseError::CorruptionError(
+                    format!("Could not interpret stored data as a block: {}", err)))?;
+        Ok(block)
+    }
+
+    pub fn put(&self, block: Block) -> Result<(), DatabaseError> {
+        let block_header: BlockHeader = protobuf::parse_from_bytes(&block.header).map_err(|err|
+            DatabaseError::CorruptionError(format!("Invalid block header: {}", err)))?;
+        { // Add block to main db
+            let mut writer = self.db.writer(None)?;
```
peterschwarz: All of these updates need to happen in the context of a transaction.

adm/src/database/lmdb.rs
```diff
@@ -0,0 +1,161 @@
...
+                DatabaseError::ReaderError(format!("Not an index: {}", index))),
+        }?;
+        Ok(LmdbDatabaseReader{
+            db: db,
+            txn: txn,
+        })
+    }
+
+    pub fn writer(&self, index: Option<&str>) -> Result<LmdbDatabaseWriter, DatabaseError> {
+        let txn = lmdb::WriteTransaction::new(&self.ctx.env).map_err(|err|
```
peterschwarz: Need a way to have a single transaction for multiply writers (for the case of updating multiple indexes per record in a single transaction).

adm/src/blockstore.rs
```diff
@@ -0,0 +1,122 @@
+use protobuf;
```
vaporos: Need copyright statement at beginning of all files.

adm/src/blockstore.rs
```diff
@@ -0,0 +1,122 @@
...
+use protobuf::Message;
+use sawtooth_sdk::messages::block::{Block, BlockHeader};
+
+use database::database::DatabaseError;
+use database::lmdb::LmdbDatabase;
+
+// TODO: Wrap block protobuf in BlockWrapper
```
vaporos: Remove the todo.

adm/src/config.rs
```diff
@@ -0,0 +1,31 @@
+use std::path::Path;
```
vaporos: Should add support for SAWTOOTH_HOME.

adm/src/database/lmdb.rs
```diff
@@ -0,0 +1,161 @@
+use std::collections::HashMap;
```
vaporos: Suggest adding api docs to the database code.

sdk/rust/src/signing/secp256k1.rs
```diff
@@ -151,6 +151,13 @@ impl Context for Secp256k1Context {
             Ok(pk) => Ok(Box::new(pk))
         }
     }
+
+    fn new_random_private_key(&self) -> Result<Box<PrivateKey>, Error> {
```
vaporos: Any reason this shouldn't be Secp256k1PrivateKey::new_random()?

adm/src/blockstore.rs
aludvik: Needed to implement this in order for the show command to work.

adm/src/config.rs
aludvik: added

sdk/rust/src/signing/secp256k1.rs
aludvik: It looks like the other SDKs provide both

adm/src/blockstore.rs
aludvik: woops

adm/src/database/lmdb.rs
aludvik: I want to finalize this API before adding docs, which won't happen until I finish STL-1007, STL-1008, and STL-1009. I also want to create a Database trait that is implemented and used, which may influence this API.

sdk/rust/src/signing/secp256k1.rs
aludvik: Decided to add `Secp256k1PrivateKey::new_random()` later.

aludvik:
Updated the copyright headers to 2018

adm/src/database/lmdb.rs
vaporos: ok

PR #1330, opened 2018-01-16
API Change remove WIF file decoding

dcmiddle:
Remove WIF as an option in CLI's and libraries.
Remove supporting libraries incl. pybitcointools and base58
WIF was originally used in Sawtooth because of some ease of using
bitcoin signing libraries. It's since been deprecated in Sawtooth.
Finally, there were concerns raised about some of the requisite libraries
including end of life of pybitcointools.


dcmiddle:
FYI, this is a backport of https://github.com/hyperledger/sawtooth-core/pull/1318 to address 1.0 release criteria.

PR #1332, opened 2018-01-16
Improve block validator logging

nick-drozd:
Currently when a block fails validation, the reason it failed is logged separately from the message saying that it failed. This makes it hard to track down what happened. This PR rearranges things in the block validator so that the messages are logged together.

Unrelated to that, there is also a commit to remove some noisy and expensive route handler logging statements.

### DIFF
--- a/validator/sawtooth_validator/journal/chain.py
+++ b/validator/sawtooth_validator/journal/chain.py
@@ -339,9 +339,6 @@ class BlockValidator(object):
                     block_store.chain_head.identifier):
                 raise ChainHeadUpdated()
 
-            blkw.status = \
-                BlockStatus.Valid if valid else BlockStatus.Invalid
-
             return valid
         except Exception:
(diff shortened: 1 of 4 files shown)


rest_api/sawtooth_rest_api/route_handlers.py
```diff
@@ -580,19 +580,10 @@ def __init__(
                                payload, error_traps=None):
         """Sends a request to the validator and parses the response.
         """
-        LOGGER.debug(
```
peterschwarz: I don't think these should be removed, since they are valid log messages.  I would wrap them in an `if log_level == DEBUG` (or how ever that needs to be done), such that the expensive calls are not done on info or higher.

peterschwarz: Also, I'd pull this commit out all together and do it in a separate PR

PR #1333, opened 2018-01-16
Improve block validator logging

nick-drozd:
Currently when a block fails validation, the reason it failed is logged separately from the message saying that it failed. This makes it hard to track down what happened. This PR rearranges things in the block validator so that the messages are logged together.

### DIFF
--- a/validator/sawtooth_validator/journal/chain.py
+++ b/validator/sawtooth_validator/journal/chain.py
@@ -339,9 +339,6 @@ class BlockValidator(object):
                     block_store.chain_head.identifier):
                 raise ChainHeadUpdated()
 
-            blkw.status = \
-                BlockStatus.Valid if valid else BlockStatus.Invalid
-
             return valid
         except Exception:
             LOGGER.exception(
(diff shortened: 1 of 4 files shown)


validator/sawtooth_validator/journal/chain.py
```diff
@@ -359,19 +357,19 @@ def _find_common_height(self, new_chain, cur_chain):
             # new chain is longer
             # walk the current chain back until we find the block that is the
             # same height as the current chain.
-            while new_blkw.block_num > cur_blkw.block_num and \
-                    new_blkw.previous_block_id != NULL_BLOCK_IDENTIFIER:
+            while (new_blkw.block_num > cur_blkw.block_num
```
peterschwarz: Unless the linter is complaining, let's not make these sorts of stylistic changes

validator/sawtooth_validator/journal/chain.py
```diff
@@ -380,7 +378,8 @@ def _find_common_height(self, new_chain, cur_chain):
                    and new_blkw.previous_block_id != NULL_BLOCK_IDENTIFIER):
                 cur_chain.append(cur_blkw)
                 cur_blkw = self._block_cache[cur_blkw.previous_block_id]
-        return (new_blkw, cur_blkw)
+
+        return new_blkw, cur_blkw
```
peterschwarz: Likewise here.

PR #1335, opened 2018-01-16
Backport of Add time-out check for poet signups

dcmiddle:
backport of #1288 for LR testing purposes


dcmiddle:
Deferring this patch to a point release. No need to test on 1-0.

PR #1340, opened 2018-01-17
Refactor Chain Controller (Part 1 of 2)

aludvik:
This PR refactors the chain controller to more cleanly separate out block validation. Some of the commits are intended to improve readability.

The functional changes are:
1. A rewrite of the function used to find the common block height between two forks
2. Replacing the result dictionary with a python object to better control the result's fields
3. Promoting the BlockValidator to a long-lived, top-level object.

The last of these functional changes is the most likely to affect performance, as it eliminates the need to instantiate a new block validator object for every block that is validated.

Finally, many BlockValidator methods were modified to be more functional in design, which was needed to support 3 but also improves readability.

These commits are more granular than usual, and I can squash them down if that is preferred. They are granular to reflect the incremental approach I took to make the changes.

I have some additional improvements to make after this PR which further separate the ChainController and BlockValidator, rewrites the ChainCommitState class, and a couple other things.


aludvik:
Fixed missing signed off commit

nick-drozd:
The PR looks great overall, just a few comments:

1. Several methods were changed from "private" to "public", e.g.
   `_verify_batch_transactions` to `verify_batch_transactions`. Are
   those intended to be part of the block validator API now? If so,
   that should be explained in the commit message. If not, they should
   be moved out into separate functions rather than be made static
   methods (maybe I'm wrong, but having a private static method in
   Python seems pointless to me, and very "unpythonic").
   
2. Several of the commit messages lack meaningful context. For
   instance, "Clarify what the remaining steps do". What steps? Where?
   It should be something like "Update BlockValidator comments".
   Similarly, "Remove unused instance variables" should be "Remove
   unused ChainController instance variables", and so on for the
   others.
   
3. The commit message "Move thread_pool to BlockValidator" needs an
   explanatory body.
   
4. I don't want to hold up the PR with this gripe, but I would like to
   reserve the right in the future to add a space before all linebreak
   backslashes.

vaporos:
When merging this, to make it clear that these go together, please use a merge commit and not a github rebase.

validator/sawtooth_validator/journal/block_validator.py
```diff
@@ -323,8 +321,8 @@ def validate_block(self, blkw, consensus, chain=None):
                 # since changes to the chain-head can change the state of the
                 # blocks in BlockStore we have to revalidate this block.
                 block_store = self._block_cache.block_store
-                if self._chain_head is not None and\
-                        self._chain_head.identifier !=\
+                if chain_head is not None and\
```
vaporos: chain_head is set above, thus will never be None here.

validator/sawtooth_validator/journal/block_validator.py
```diff
@@ -284,63 +286,71 @@ def validate_on_chain_rules(self, blkw, prev_state_root):
         return True
 
     def validate_block(self, blkw, consensus, chain_head=None, chain=None):
+        if blkw.status == BlockStatus.Valid:
```
vaporos: The changes in this commit do not reflect the simplicity of the commit message.  It is thus difficult to determine the intent in some of these changes.

validator/sawtooth_validator/journal/block_validator.py
```diff
@@ -348,9 +347,6 @@ def validate_block(self, blkw, consensus, chain_head=None, chain=None):
             blkw.status = BlockStatus.Valid
             return True
 
-        except ChainHeadUpdated:
```
vaporos: Is this never raised? It's not clear why this is a linting-only change.

validator/sawtooth_validator/journal/block_validator.py
```diff
@@ -82,6 +80,7 @@ def __str__(self):
         return out[:-1] + "}"
 
 
+# pylint: disable=stop-iteration-return
```
vaporos: Is this related to - https://github.com/PyCQA/pylint/issues/1830 ?  If so, might be worth a mention in the commit message or next to this comment.

validator/sawtooth_validator/journal/block_validator.py
aludvik: This is not actually true. When the chain head is retrieved from the block store, it can still be None if there isn't a genesis block yet.

validator/sawtooth_validator/journal/block_validator.py
aludvik: I split the commit up into two and added more information to this part of the change.

validator/sawtooth_validator/journal/block_validator.py
aludvik: Catching the exception only to immediately raise it again has not affect and causes the linter to break.

validator/sawtooth_validator/journal/block_validator.py
aludvik: yeah

validator/sawtooth_validator/journal/block_validator.py
vaporos: Isn't the intent of this line to propogate it up (skip the broad except in this case)?  After this, won't ChainHeadUpdated be caught by the broad except?

nick-drozd: I think it would work to change it to
```
except ChainHeadUpdated as e:
    raise e
```
Leaving it as-is crashes pylint due to a bug in the release version.

vaporos:
I think these changes should also go into 1-0 branch.

PR #1342, opened 2018-01-17
Add transaction processor support to the Rust SDK

vaporos:


sdk/rust/src/processor/handler.rs
```diff
@@ -0,0 +1,243 @@
+/*
+ * Copyright 2017 Bitwise IO, Inc.
```
peterschwarz: 2018?

sdk/rust/src/processor/mod.rs
```diff
@@ -0,0 +1,144 @@
+/*
+ * Copyright 2017 Bitwise IO, Inc.
```
peterschwarz: 2018?

sdk/rust/src/processor/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+
+                            let serialized = response.write_to_bytes().unwrap();
+                            let x : &[u8] = &serialized;
+                            sender.reply(
+                                Message_MessageType::TP_PROCESS_RESPONSE,
+                                message.get_correlation_id(),
+                                x).unwrap();
+                        },
+                        _ => {
+                            let mut response = TpProcessResponse::new();
```
peterschwarz: I'm not sure what sort of behavior this will result in: sending back a response is good (i.e. the future on the other end will resolve), but it will potentially be the wrong type of message.

peterschwarz: Missing handling of `PING_REQUEST`, thought this could probably be done in a future PR

agunde406: I can include the Ping functionality with the reconnect work.

vaporos: We could also include the default handling rework in with reconnect work (probably some affinity to testing it all together).

sdk/rust/src/processor/handler.rs
vaporos: 2017 is accurate, it started late last year.

sdk/rust/src/processor/handler.rs
```diff
@@ -0,0 +1,243 @@
...
+        }
+    }
+}
+
+impl std::fmt::Display for ContextError {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        match *self {
+            ContextError::AuthorizationError(ref s) =>
+                write!(f, "AuthorizationError: {}", s),
+            ContextError::UnknownError(ref s) =>
```
aludvik: UnknownError seems like it could be abused rather easily. It looks like this is just being used when validating a response from the validator, so maybe something like `ConsistencyError` or `IntegrityError`?

sdk/rust/src/processor/handler.rs
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    pub fn get_state(&mut self, address: &str) -> Result<Option<Vec<u8>>, ContextError> {
+        let mut request = TpStateGetRequest::new();
+        request.set_context_id(self.context_id.clone());
+        request.set_addresses(RepeatedField::from_vec(vec!(String::from(address))));
+        let serialized = request.write_to_bytes()?;
+        let x : &[u8] = &serialized;
+
+        let correlation_id: String = rand::thread_rng().gen_ascii_chars().take(16).collect();
```
aludvik: In other implementations, correlation id generation is behind another function, which is what I would expect here. If not, at a minimum 16 should be moved to a constant.

sdk/rust/src/processor/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+                let x : &[u8] = &serialized;
+
+                let correlation_id: String = rand::thread_rng().gen_ascii_chars().take(16).collect();
+                let mut future = sender.send(
+                    Message_MessageType::TP_REGISTER_REQUEST,
+                    &correlation_id,
+                    x).unwrap();
+
+                // Absorb the TpRegisterResponse message
+                let _ = future.get().unwrap();
```
aludvik: Why are we unwrapping here?

sdk/rust/src/processor/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+            }
+        }
+
+        loop {
+            match receiver.recv() {
+                Ok(r) => {
+                    let message = r.unwrap();
```
aludvik: Why are we unrwapping?

sdk/rust/src/processor/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+                                    info!("TP_PROCESS_REQUEST sending TpProcessResponse: {}", msg);
+                                },
+                                Err(err) => {
+                                    response.set_status(TpProcessResponse_Status::INTERNAL_ERROR);
+                                    response.set_message(String::from(err.description()));
+                                    info!("TP_PROCESS_REQUEST sending TpProcessResponse: {}", err.description());
+                                }
+                            };
+
+                            let serialized = response.write_to_bytes().unwrap();
```
aludvik: Why wrapping unwrapping?

sdk/rust/src/processor/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+                        _ => {
+                            let mut response = TpProcessResponse::new();
+                            response.set_status(TpProcessResponse_Status::INTERNAL_ERROR);
+                            response.set_message(String::from("not implemented..."));
+                            let serialized = response.write_to_bytes().unwrap();
+                            let x : &[u8] = &serialized;
+                            sender.reply(
+                                Message_MessageType::TP_PROCESS_RESPONSE,
+                                message.get_correlation_id(),
+                                x).unwrap();
```
aludvik: Why are we unrwapping?

vaporos:
Rebased and updated "Add transaction processor support to Rust SDK" to not add rand to Cargo.toml, as that was added in master recently as part of the sawadm PR.

sdk/rust/src/processor/handler.rs
vaporos: This is now a function generate_correlation_id().

vaporos:
With respect to the error handling deficiencies in start() (the unwrap()s), I think we need to reconne-ction logic to properly handle many of these.  Or at least it seems related from a flow perspective.

For example, if we receive an error prior to the current loop(), such as during registration, we may want to disconnect and try again.  Because this could happen on re-connection, we shouldn't return an error from start() as we don't want transaction processors to exit due to transient errors.

PR #1344, opened 2018-01-17
Avoid db write via StateViewFactory.create_view

peterschwarz:
This fixes an issue where the call to create_view separates the
construciton and a call to set_merkle_root, in the case where no
root is provided.  However, the constructor also calls set_merkle_root,
which does a db put when the INIT_ROOT_KEY is the root. Therefore,
every instance of StateView is incurring the cost of a db write, after
block 0.


nick-drozd:
Nice performance gains for such a small change.

Running against 1-0:

Node 1 before:
before-1

Node 1 after:
after-1

Node 2 before:
before-2

Node 2 after:
after-2

PR #1345, opened 2018-01-18
add cxx debian to set of published debs from build.

cianx:


vaporos:
This LGTM, will approve when conflicts are resolved.

PR #1347, opened 2018-01-18
Backport: Add a from_state option to identity cache

agunde406:
This will remove the case where a value in the cache could
be replaced by an outdated value while validating an old
block. Instead the value will be checked from state and it
will not be added to the cache.


aludvik:
Before you merge this can you:
1. Confirm a successful LR regression run?
2. Get approval from @dcmiddle?
3. Confirm that this fixes a known stability issue on 1-0?

agunde406:
Closing for now.

PR #1348, opened 2018-01-18
1.0: Avoid db write via StateViewFactory.create_view

peterschwarz:
This fixes an issue where the call to create_view separates the
construciton and a call to set_merkle_root, in the case where no
root is provided.  However, the constructor also calls set_merkle_root,
which does a db put when the INIT_ROOT_KEY is the root. Therefore,
every instance of StateView is incurring the cost of a db write, after
block 0.


aludvik:
Before you merge this can you:
1. Confirm a successful LR regression run?
2. Get approval from @dcmiddle?
3. Confirm that this fixes a known stability issue on 1-0?

PR #1350, opened 2018-01-18
Adds NOTICES file to 1-0 branch

TomBarnes:
Signed-off-by: Tom Barnes <thomas.j.barnes@intel.com>

### DIFF
--- a/NOTICES
+++ b/NOTICES
@@ -1,0 +1,2022 @@
+Sawtooth relies on the following open source components for run-time execution.
+These components are released under separate copyright notices and license terms.
+
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+
+=> Alex Crichton futures-rs
+=> Apache Commons Lang
+=> CBOR for Go and Python
+=> C_based_gRPC
+=> Maven CBOR for Java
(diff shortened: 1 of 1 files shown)


NOTICES
```diff
@@ -0,0 +1,2022 @@
...
+=> Node.js
+Copyright Node.js contributors. All rights reserved
+Copyright Joyent, Inc. and other Node contributors. All rights reserved.
+=> PyYAML - a YAML parser and emitter for Python
+Copyright (c) 2006 Kirill Simonov
+=> aiodns
+Copyright (C) 2014 by Saúl Ibarra Corretgé
+=> cffi
+Copyright (C) 2005-2007, James Bielman  <jamesjb@jamesjb.com
+=> cppzmq
```
boydjohnson: This one seems missing the who: https://github.com/zeromq/cppzmq/blob/master/LICENSE.

TomBarnes: Although it is an MIT license, the author has not chosen to include a copyright notice in the license file, and so I have not include a copyright notice in this notices file.  I  am trying to make the artifacts in this file match the project licence artifacts as closely as possible while minimizing duplication.

PR #1351, opened 2018-01-18
Fix typo in _find_common_height

aludvik:
This was checking the wrong block to see if it reached genesis. Possibly
a copy-paste error.

### DIFF
--- a/validator/sawtooth_validator/journal/chain.py
+++ b/validator/sawtooth_validator/journal/chain.py
@@ -359,11 +359,9 @@ class BlockValidator(object):
         # 1) find the common ancestor of this block in the current chain
         # Walk back until we have both chains at the same length
 
-        # Walk back the new chain to find the block that is the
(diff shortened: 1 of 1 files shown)


validator/sawtooth_validator/journal/chain.py
```diff
@@ -383,7 +383,7 @@ def _find_common_height(self, new_chain, cur_chain):
             # walk the current chain back until we find the block that is the
             # same height as the new chain.
             while (cur_blkw.block_num > new_blkw.block_num
-                   and new_blkw.previous_block_id != NULL_BLOCK_IDENTIFIER):
+                   and cur_blkw.previous_block_id != NULL_BLOCK_IDENTIFIER):
```
dcmiddle: Could you please also correct the pre-existing cut and paste comments at lines? remove ~362 and correct ~366.

PR #1352, opened 2018-01-18
updated to use long form apache license and add 3rd party licenses

TomBarnes:
Signed-off-by: Tom Barnes <thomas.j.barnes@intel.com>

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,13 +1,292 @@
-Copyright 2016 Intel Corporation
+                                Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
 
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
 
-    http://www.apache.org/licenses/LICENSE-2.0
+   1. Definitions.
 
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -1,13 +1,292 @@
...
+      License. However, in accepting such obligations, You may act only
+      on Your own behalf and on Your sole responsibility, not on behalf
+      of any other Contributor, and only if You agree to indemnify,
+      defend, and hold each Contributor harmless for any liability
+      incurred by, or claims asserted against, such Contributor by reason
+      of your accepting any such warranty or additional liability.
+
+   END OF TERMS AND CONDITIONS
+
+   APPENDIX: How to apply the Apache License to your work.
```
dcmiddle: Should this appendix be cut? It's instructions on how to attache the short form.

LICENSE
```diff
@@ -1,13 +1,292 @@
-Copyright 2016 Intel Corporation
```
dcmiddle: I think this copyright notice should remain 
See around line 189 "... To apply the Apache License to your work, attach the following boilerplate notice, with the fields enclosed by brackets "[]" replaced with your own identifying information. (Don't include the brackets!) ... Copyright [yyyy] [name of copyright owner]"

TomBarnes: That is in the section that provides directions on how to apply the license to "your work"; that is, all the other files you've created, not the license itself.  See for example how its handled in this major Apache-sponsored project: https://github.com/apache/spark/blob/master/LICENSE

TomBarnes: ... on which was used as an example for this revision

LICENSE
TomBarnes: see above

TomBarnes: ...and also see: http://www.apache.org/dev/apply-license.html#new

LICENSE
dcmiddle: ok, makes sense

PR #1358, opened 2018-01-19
Add Intkey TP Written in Rust

agunde406:
Copyright Years differ depending on when the file was created.

### DIFF
--- a/bin/intkey-tp-rust
+++ b/bin/intkey-tp-rust
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright 2017 Intel Corporation
+# Copyright 2018 Bitwise IO, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


bin/intkey-tp-rust
```diff
@@ -0,0 +1,21 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# -----------------------------------------------------------------------------
+
+top_dir=$(cd $(dirname $(dirname $0)) && pwd)
+cd $top_dir/sdk/examples/intkey_rust
+
+exec cargo run --release -- $*
```
peterschwarz: This might not be the best thing to do here, since this currently triggers a full build, due to source generation from the proto files.  Something to look into soon after this is merged.

agunde406: I will add it to my todo list.

PR #1359, opened 2018-01-19
Add multiple rest api targets and basic auth to smallbank workload

boydjohnson:


boydjohnson:
I almost have the refactor done so it seems better to have the refactor as a PR and then these changes.

PR #1361, opened 2018-01-20
Add new peer reporting CLI commands

nick-drozd:
This PR adds `sawtooth status show`, which currently reports a node's public network endpoint and peers, and `sawnet peers {list,graph}`, which display the peering arrangement of a network of nodes.

### DIFF
--- a/cli/sawtooth_cli/network_command/peers.py
+++ b/cli/sawtooth_cli/network_command/peers.py
@@ -1,0 +1,142 @@
+# Copyright 2018 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+
+import os
+import json
+
+from sawtooth_cli.network_command.parent_parsers import base_multinode_parser
+from sawtooth_cli.network_command.parent_parsers import split_comma_append_args
+from sawtooth_cli.network_command.parent_parsers import make_rest_apis
+
+from sawtooth_cli.exceptions import CliException
+
+
+DOT_FILE = 'peers.dot'
+
+
+def add_peers_parser(subparsers, parent_parser):
+    help_text = 'Shows the peering arrangment of a network'
+
+    parser = subparsers.add_parser(
+        'peers',
+        help=help_text,
+        description='{}.'.format(help_text))
+
+    peers_parsers = parser.add_subparsers(
+        title='subcommands',
+        dest='peers_command')
+    peers_parsers.required = True
+
+    _add_list_parser(peers_parsers, parent_parser)
+    _add_graph_parser(peers_parsers, parent_parser)
+
+
+def _add_list_parser(parser, parent_parser):
+    help_text = 'Lists peers for validators with given URLs'
+
+    list_parser = parser.add_parser(
+        'list',
+        help=help_text,
+        description='{}.'.format(help_text),
+        parents=[parent_parser, base_multinode_parser()])
+
+    list_parser.add_argument(
+        '--pretty', '-p',
+        action='store_true',
+        help='Pretty-print the results')
+
+
+def do_peers(args):
+    if args.peers_command == 'list':
+        _do_peers_list(args)
+    elif args.peers_command == 'graph':
+        _do_peers_graph(args)
+    else:
+        raise CliException('Invalid command: {}'.format(args.subcommand))
+
+
+def _do_peers_list(args):
+    urls = split_comma_append_args(args.urls)
+    users = split_comma_append_args(args.users)
+    clients = make_rest_apis(urls, users)
+
+    print(
+        json.dumps(
+            _get_peer_endpoints(clients),
+            sort_keys=True,
+            indent=(4 if args.pretty else 0)
+        )
+    )
+
+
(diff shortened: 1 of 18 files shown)


nick-drozd:
Here is the graph produced by the network laid out in `test_peer_list`:

peers

nick-drozd:
Per a suggestion from @peterschwarz, the `/status` endpoint now returns
```
{
	"endpoint": "tcp://endpoint",
	"peers": [
		{
			"endpoint": "tcp://peer-1"
		},
		{
			"endpoint": "tcp://peer-2"
		}
	]
}
```
This allows the endpoint to be extended to include not just more information about the validator, but also more information about the validator's peers.

protos/client_status.proto
```diff
@@ -0,0 +1,39 @@
...
+syntax = "proto3";
+
+option java_multiple_files = true;
+option java_package = "sawtooth.sdk.protobuf";
+option go_package = "client_status";
+
+message ClientStatusGetRequest{
+}
+
+message ClientStatusGetResponse {
```
peterschwarz: Add some doc

cli/sawtooth_cli/network_command/peers.py
```diff
@@ -0,0 +1,152 @@
...
+
+DOT_FILE = 'peers.dot'
+PNG_FILE = 'peers.png'
+
+
+def add_peers_parser(subparsers, parent_parser):
+    parser = subparsers.add_parser(
+        'peers',
+        help='TODO',
+        description='TODO',
```
aludvik: Create a story to resolve these TODOs and reference in this PR

cli/sawtooth_cli/network_command/peers.py
```diff
@@ -0,0 +1,152 @@
...
+            indent=(4 if args.pretty else 0)
+        )
+    )
+
+
+def _get_status(url):
+    status = json.loads(
+        subprocess.check_output(
+            shlex.split(
+                'sawtooth status show --url {}'.format(url))
```
aludvik: This should not drop into a shell to execute this. You should do the request manually using the requests library.

cli/sawtooth_cli/network_command/peers.py
```diff
@@ -0,0 +1,152 @@
...
+        'list',
+        help='List peers for validators with given URLs',
+        description=(
+            'List peers for validators with given URLs separated by commas.'))
+
+    list_parser.add_argument(
+        'urls',
+        nargs=1,
+        help='the comma-separated URLs of the validators to be queried',
+        type=str)
```
aludvik: You should use `cli/sawtooth_cli/network_command/parent_parsers.py` for this as it supports authenticated access with the rest apis. It also has a function to create the rest clients for all the peers.

cli/sawtooth_cli/network_command/peers.py
```diff
@@ -0,0 +1,152 @@
...
+    graph_parser = parser.add_parser(
+        'graph',
+        help='TODO',
+        description='TODO')
+
+    graph_parser.add_argument(
+        'urls',
+        nargs=1,
+        type=str,
+        help='TODO')
```
aludvik: This should also use parent_parsers here. Checkout how it is used in compare-chains.

cli/sawtooth_cli/network_command/peers.py
```diff
@@ -0,0 +1,152 @@
...
+
+    graph_parser.add_argument(
+        '--force',
+        action='store_true',
+        help='TODO')
+
+    graph_parser.add_argument(
+        '--png',
+        action='store_true',
+        help='TODO')
```
aludvik: This should also take an `--output` flag to tell the command where to write the file, instead of always creating it in the cwd.

cli/sawtooth_cli/network_command/peers.py
aludvik: It also supports passing the urls as a space separated list, which is how compare-chains works, and we should provide a consistent way to pass urls.

nick-drozd:
Addressed comments.

cli/sawtooth_cli/network_command/peers.py
```diff
@@ -0,0 +1,150 @@
...
+        'list',
+        help=help_text,
+        description='{}.'.format(help_text),
+        parents=[parent_parser, base_multinode_parser()])
+
+    list_parser.add_argument(
+        'urls',
+        nargs=1,
+        help='the URLs of the validators to be queried',
+        type=str)
```
aludvik: Since you are using `base_multinode_parser()` this creates a duplicate arg:
```
lradmin@frost:~/core-clones/sawtooth-drozd$ ./bin/sawnet peers list --help
usage: sawnet peers list [-h] [-v] [-V] [--users USERNAME[:PASSWORD]]
                         [--pretty]
                         urls [urls ...] urls

Lists peers for validators with given URLs.

positional arguments:
  urls                  The URLs of the validator's REST APIs of interest,
                        separated by commas or spaces. (no default)
  urls                  the URLs of the validators to be queried
```

It looks like you got it right for `peers graph`

cli/sawtooth_cli/network_command/peers.py
```diff
@@ -0,0 +1,150 @@
...
+        description='{}.'.format(help_text),
+        parents=[parent_parser, base_multinode_parser()])
+
+    graph_parser.add_argument(
+        '--output-file',
+        help='The name of the dot file to be produced (defaults to peers.dot)')
+
+    graph_parser.add_argument(
+        '--output-dir',
+        help='The directory in which the dot file is created')
```
aludvik: Typically output is a single flag, see `cli/sawtooth_cli/admin_command/genesis.py` L43

nick-drozd: `sawtooth keygen` separates the file name and the directory (https://sawtooth.hyperledger.org/docs/core/releases/latest/cli/sawtooth.html#sawtooth-keygen). We can do it either way, but separating them seems more flexible to me.

nick-drozd: Actually `keygen` does something slightly different, so I'll just go with `--output`.

PR #1362, opened 2018-01-20
 The path C:\dir is mapped as /mnt/c/dir on Windows Subsystem for Lin…

arthurgreef:
The path C:\dir is mapped as /mnt/c/dir on Windows Subsystem for Linux (WSL) Bash shell and as /c/dir/ on Docker For Windows (D4W). This fix adds a -w flag to the build_all script to remove the /mnt string when mapping WSL paths for D4W volumes. The documentation is updated for build environments that use the Docker for Linux client in WSL with the D4W engine.

### DIFF
--- a/bin/build_all
+++ b/bin/build_all
@@ -36,16 +36,18 @@ ALL_TARGETS="
 MOUNTED="mounted"
 INSTALLED="installed"
 DEBS="debs"
+WSL_D4W=false
 
 usage() {
     echo "Usage:"
-    echo "  $0 [-l Language] [-x Language] {$MOUNTED,$INSTALLED,$DEBS}"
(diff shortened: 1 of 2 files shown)


arthurgreef:
Closing this pull request. Add the following to .bashrc instead.

 .bashrc:
# mount /mnt/c to /c if not already done
if [ ! -d "/c" ] || [ ! "$(ls -A /c)" ]; then
  echo "Requiring root password to $(tput setaf 6)mount --bind /mnt/c /c$(tput sgr 0)"
  sudo mkdir -p /c
  sudo mount --bind /mnt/c /c
fi

# Change from /mnt/c/... to /c/...
if [ "$(pwd | cut -c -7)" == "/mnt/c/" ]; then
  cd "$(pwd | cut -c 5-)"
fi

# Provide docker for bash and docker-compose
docker() {
  docker.exe "$@"
}

PR #1363, opened 2018-01-22
Support looking up blocks by height, batch, txn

aludvik:
Adds support to sawadm blockstore for looking up blocks based on block
height, batch header signature, and transaction header signature.


adm/src/main.rs
```diff
@@ -74,7 +74,13 @@ fn parse_args<'a>() -> ArgMatches<'a> {
                 (@arg start: --start +takes_value "the first block to list"))
             (@subcommand show =>
                 (about: "inspect a block in the blockstore")
-                (@arg block: +required "the block to show"))
+                (@arg identifier: +required "the identifier used to determine which bloc to show")
```
agunde406: bloc -> block

adm/src/main.rs
```diff
@@ -74,7 +74,13 @@ fn parse_args<'a>() -> ArgMatches<'a> {
...
+                (@arg batch: --batch conflicts_with[transaction height]
+                    "show a block based on batch id")
+                (@arg transaction: --transaction conflicts_with[batch height]
+                    "show a block based on transaction id")
+                (@arg height: --height conflicts_with[batch transaction]
```
vaporos: --block-num may make more sense, as it's a field within the block.

adm/src/main.rs
```diff
@@ -74,7 +74,13 @@ fn parse_args<'a>() -> ArgMatches<'a> {
                 (@arg start: --start +takes_value "the first block to list"))
             (@subcommand show =>
                 (about: "inspect a block in the blockstore")
-                (@arg block: +required "the block to show"))
+                (@arg identifier: +required "the identifier used to determine which bloc to show")
+                (@arg batch: --batch conflicts_with[transaction height]
```
vaporos: These arguments are a bit odd, in that you can't do "--batch=xxx" because these are flags.

Suggest -b,--block=xxx, -N,--block-num=xxx, -T,--transaction=xxx, -B,--batch=xxx, and identifier all be allowed with identifier being a search (you can differentate block-num from signatures based on length and digits) and you could search for block, batch, and then transaction id in that order.

aludvik:
Addressed @vaporos feedback

PR #1365, opened 2018-01-22
Add heartbeat logging to threadpools and _ChainThread

aludvik:
This is to confirm that the chain controller and the thread pools are not silently dying and causing the flat line bug.

### DIFF
--- a/validator/sawtooth_validator/concurrent/heartbeat.py
+++ b/validator/sawtooth_validator/concurrent/heartbeat.py
@@ -1,0 +1,18 @@
+import time
+import logging
+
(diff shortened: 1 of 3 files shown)


validator/sawtooth_validator/concurrent/heartbeat.py
```diff
@@ -0,0 +1,18 @@
+import time
+import logging
+
+LOGGER = logging.getLogger(__name__)
+
+
+class Heartbeat:
+    """Logs a message every `period` seconds."""
+    def __init__(self, message="heartbeat", period=3):
```
peterschwarz: Why 3?

PR #1367, opened 2018-01-23
Update minimum wait timer in poet

askmish:

### DIFF
--- a/consensus/poet/core/sawtooth_poet/poet_consensus/consensus_state.py
+++ b/consensus/poet/core/sawtooth_poet/poet_consensus/consensus_state.py
@@ -63,6 +63,10 @@ class ConsensusState(object):
             claimed by all validators
     """
 
+    # MINIMUM_WAIT_TIME should be same as the value
+    # at ecall_CreateWaitTimer
+    MINIMUM_WAIT_TIME = 1.0
+
     _BlockInfo = \
         collections.namedtuple(
             '_BlockInfo',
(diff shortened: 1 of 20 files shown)


dcmiddle:
@vaporos I discussed this approach with Ashish also with @aludvik and @peterschwarz . This is the best option given the issue. This value should not have been originally created as a parameter/setting.

consensus/poet/core/tests/test_consensus/test_consensus_state.py
```diff
@@ -27,8 +27,10 @@
 from sawtooth_poet_common.protobuf.validator_registry_pb2 \
     import SignUpInfo
 
-
+print(consensus_state.MINIMUM_WAIT_TIME)
```
dcmiddle: remove the print() call

consensus/poet/core/sawtooth_poet/poet_consensus/consensus_state.py
```diff
@@ -63,6 +63,10 @@ class ConsensusState(object):
             claimed by all validators
     """
 
+    # MINIMUM_WAIT_TIME should be same as the value
+    # at ecall_CreateWaitTimer
```
dcmiddle: Maybe a better comment is that it should match the constant in the enclaves. That should include both the simulated enclave and the sgx enclave.

consensus/poet/simulator/sawtooth_poet_simulator/poet_enclave_simulator/poet_enclave_simulator.py
```diff
@@ -352,7 +352,8 @@ def create_wait_timer(cls,
             tagd = float(struct.unpack('Q', tag[-8:])[0]) / (2**64 - 1)
 
             # Now compute the duration with a minimum wait time guaranteed
-            duration = minimum_wait_time - local_mean * math.log(tagd)
+            duration = _PoetEnclaveSimulator.MINIMUM_WAIT_TIME
+                           - local_mean * math.log(tagd)
```
dcmiddle: line continuation.

vaporos:
I'm okay with this PR and backporting it to 1.0 (I don't believe this setting is being used currently).

However, docs/source/architecture/journal.rst mentions this setting, so we need to remove that example.

dcmiddle:
I'll take care of the doc change separately for sake of getting the code change into the testing pipeline sooner than later.

PR #1368, opened 2018-01-23
Refactor sawtooth_perf long running batch submission

boydjohnson:
This PR also implements submitting to multiple rest apis and basic auth support.

After this PR, STL-978, 981, 1014 are completed. The smallbank-workload tool has been manually tested at low rates on a mac (it bumps up against the 256 file handle limit), and at higher rates withing a Ubuntu docker container (ulimit -Sn 100000, --rate 30).


boydjohnson:
@vaporos should they be buildable at each commit, or is it good enough that they make the refactor more clear?

perf/smallbank_workload/src/main.rs
```diff
@@ -16,6 +16,7 @@
  */
 extern crate clap;
 extern crate crypto;
+extern crate env_logger;
```
peterschwarz: We used `simple_logging` in the Seth RPC code: See https://github.com/hyperledger/sawtooth-seth/blob/master/families/seth/rpc/src/main.rs

perf/sawtooth_perf/src/batch_submit.rs
```diff
@@ -181,17 +185,19 @@ pub fn run_workload<'a>(batch_list_iter: &'a mut Iterator<Item = BatchListResult
         .map(move |_| -> Result<BatchList, workload::WorkloadError> {
             workload::get_next_batchlist(batch_list_iter)
         })
-        .map(move |batch_list: Result<BatchList, workload::WorkloadError>| {
+        .map(|batch_list: Result<BatchList, workload::WorkloadError>| {
+            let basic_auth_c = basic_auth.clone();
```
peterschwarz: This is a strange naming convention.  We should talk as a team about what we want for these.

boydjohnson:
Since there are two approvals and @vaporos has not responded since his initial comment, I am merging this.

PR #1372, opened 2018-01-23
Complete the Transaction Processor for the rust sdk

agunde406:
This pr includes adding a monitor socket and handling a
disconnection from the validator, and ability to respond to
ping requests. Also the handling of wrong message types of
has been updated to match the other SDK and unwraps have
been removed in the processor.

Also includes some other random fixups.

### DIFF
--- a/sdk/rust/Cargo.toml
+++ b/sdk/rust/Cargo.toml
@@ -24,7 +24,7 @@ protobuf="1.4.1"
 secp256k1 = "0.7.1"
 rand = "0.4.2"
 rust-crypto = "0.2.36"
-zmq = "0.8"
+zmq = { git = "https://github.com/erickt/rust-zmq", branch = "release/v0.8" }
 uuid = { version = "0.5", features = ["v4"] }
 log = "0.3"
 libc = "0.2"
--- a/sdk/rust/src/messaging/zmq_stream.rs
+++ b/sdk/rust/src/messaging/zmq_stream.rs
@@ -63,13 +63,18 @@ impl MessageConnection<ZmqMessageSender> for ZmqMessageConnection {
     }
 }
 
+#[derive(Debug)]
+enum SocketCommand {
+    Send(Message),
+    Shutdown
+}
 
 #[derive(Clone)]
 pub struct ZmqMessageSender {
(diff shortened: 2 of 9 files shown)


sdk/rust/src/processor/mod.rs
```diff
@@ -225,6 +226,34 @@ impl<'a> TransactionProcessor<'a> {
...
+                                let response = PingResponse::new();
+                                let serialized = match response.write_to_bytes() {
+                                    Ok(serialized) => serialized,
+                                    Err(err) => {
+                                        error!("Serialization failed: {}", err.description());
+                                        continue
+                                    }
+                                };
+                                let x : &[u8] = &serialized;
+                                match sender.reply(
```
peterschwarz: Have we determined if we can use `rustfmt` yet?  Not wanting to ding this line based on formatting without a tool, but I'm registering my opinion that I'm not a fan of the way this line is wrapped.

agunde406: Opinion registered. I do not believe that we can run rustfmt yet but that is something I can look into shortly.

sdk/rust/Cargo.toml
```diff
@@ -24,7 +24,8 @@ protobuf="1.4.1"
 secp256k1 = "0.7.1"
 rand = "0.4.2"
 rust-crypto = "0.2.36"
-zmq = "0.8"
+rust-base58 = "0.0.4"
```
aludvik: We can't use this library due to licensing

sdk/rust/src/messaging/zmq_stream.rs
```diff
@@ -171,14 +175,25 @@ impl InboundRouter {
...
+    fn route(&mut self, message_result: MessageResult) {
+        match message_result  {
+            Ok(message) => {
+                let mut expected_replies = self.expected_replies.lock().unwrap();
+                match expected_replies.remove(
+                message.get_correlation_id())
+                {
+                Some(sender) => sender.send(Ok(message)).unwrap(),
+                None => self.inbound_tx.send(Ok(message)).unwrap()
+                }
```
aludvik: Above indentation is weird

PR #1375, opened 2018-01-24
Add delete_state, add_receipt_data, add_event  to rust TransactionContext

agunde406:


sdk/rust/src/processor/handler.rs
```diff
@@ -91,7 +91,7 @@ pub enum ContextError {
     SerializationError(Box<StdError>),
     /// Raised when an error is returned when sending a message
     SendError(Box<StdError>),
-    /// Raise when an error is return when sending a message
+    /// Raised when an error is returned when sending a message
```
peterschwarz: Technically, these aren't raised, but returned...

PR #1376, opened 2018-01-24
Add protocol_version to  PeerRegisterRequest

peterschwarz:
Add `protocol_version` to `PeerRegisterRequest`, to create a future avenue for dealing with network protocol changes.

### DIFF
--- a/protos/network.proto
+++ b/protos/network.proto
@@ -25,6 +25,9 @@ message DisconnectMessage {
 // The registration request from a peer to the validator
 message PeerRegisterRequest {
     string endpoint = 1;
+    // The current version of the network protocol that is being used by the
+    // sender.  This version is an increasing value.
+    uint32 protocol_version = 2;
 }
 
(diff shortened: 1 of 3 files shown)


validator/sawtooth_validator/gossip/gossip.py
```diff
@@ -68,6 +68,8 @@ class EndpointStatus(Enum):
 
 TIME_TO_LIVE = 3
 
+NETWORK_PROTOCOL_VERSION = 1
```
vaporos: Should add a comment above this on when it should be incremented.

protos/network.proto
```diff
@@ -25,6 +25,9 @@ message DisconnectMessage {
 // The registration request from a peer to the validator
 message PeerRegisterRequest {
     string endpoint = 1;
+    // THe current version of the network protocol that is being used by the
```
agunde406: THe -> the

PR #1377, opened 2018-01-25
Add some completer metrics

nick-drozd:
These are some completer metrics that will be useful in measuring proposed fixes for the flatline bug. There will be more.

### DIFF
--- a/validator/sawtooth_validator/concurrent/threadpool.py
+++ b/validator/sawtooth_validator/concurrent/threadpool.py
@@ -22,6 +22,7 @@ from concurrent.futures import ThreadPoolExecutor
 
 from sawtooth_validator.concurrent import atomic
 from sawtooth_validator.metrics.wrappers import GaugeWrapper
+from sawtooth_validator.metrics.wrappers import TimerWrapper
 
 
 LOGGER = logging.getLogger(__name__)
@@ -55,8 +56,13 @@ class InstrumentedThreadPoolExecutor(ThreadPoolExecutor):
             self._workers_already_in_use_gauge = GaugeWrapper(
                 metrics_registry.gauge(
                     '{}-threadpool.workers_already_in_use'.format(self._name)))
+            # Tracks how long tasks take to run
+            self._task_run_timer = TimerWrapper(
+                metrics_registry.timer(
+                    '{}-threadpool.task_run_time'.format(self._name)))
         else:
             self._workers_already_in_use_gauge = GaugeWrapper()
+            self._task_run_timer = TimerWrapper()
 
     def submit(self, fn, *args, **kwargs):
         submitted_time = time.time()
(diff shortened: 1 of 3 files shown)


validator/sawtooth_validator/journal/completer.py
```diff
@@ -87,8 +88,13 @@ def __init__(self,
             self._unsatisfied_dependency_count = CounterWrapper(
                 metrics_registry.counter(
                     'completer_unsatisfied_dependency_count'))
+            self._seen_txns_length = GaugeWrapper(
+                metrics_registry.gauge(
+                    'completer_seen_txns_length'))
         else:
             self._unsatisfied_dependency_count = CounterWrapper()
+            self._seen_txns_length = GaugeWrapper()
+            self._incomplete_blocks_length = GaugeWrapper()
```
aludvik: I think this line belongs in another commit

validator/sawtooth_validator/concurrent/threadpool.py
```diff
@@ -41,13 +42,26 @@ def __init__(self, max_workers=None, name='', trace=None):
...
-            # This is the same default as ThreadPoolExecutor, but we want to
-            # know how many workers there are for logging
-            self._max_workers = multiprocessing.cpu_count() * 5
         super().__init__(max_workers)
 
+        if metrics_registry:
+            # Tracks how long tasks are queued
+            self._time_in_queue_gauge = GaugeWrapper(
+                metrics_registry.gauge(
+                    'threadpool.time_in_queue'))
```
aludvik: This needs the threadpool name prefixed to it, or else all of the threadpools will be aggregated into one metric.

validator/sawtooth_validator/concurrent/threadpool.py
```diff
@@ -41,13 +42,26 @@ def __init__(self, max_workers=None, name='', trace=None):
...
+            # Tracks how many workers are already in use
+            self._workers_already_in_use_gauge = GaugeWrapper(
+                metrics_registry.gauge(
+                    'threadpool.workers_already_in_use'))
+            # Tracks how long tasks take to run
+            self._task_run_time_gauge = GaugeWrapper(
+                metrics_registry.gauge(
+                    'threadpool.task_run_time'))
+        else:
+            self._time_in_queue_gauge = GaugeWrapper()
```
aludvik: What was the reason for not using the pyformance Timer type for this?

validator/sawtooth_validator/concurrent/threadpool.py
```diff
@@ -88,31 +94,23 @@ def wrapper():
...
-            # pylint: disable=broad-except
-            except Exception:
-                LOGGER.exception(
-                    '(%s) Unhandled exception during execution of task %s',
-                    self._name,
-                    task_details)
-
-            end_time = time.time()
-            run_time = (end_time - start_time) * 1000.0
+            with self._task_run_timer.time():
```
peterschwarz: This needs to be a timer per thread in the pool.  This measurement won't make sense otherwise.

PR #1379, opened 2018-01-25
Swap completer cache keep times

aludvik:
Setting the requested keep time higher than the cache keep time can
cause the validator to get into a state where it stops processing blocks
all together because it thinks it has already requested the blocks that
it needs to request.

Swapping these keep times should eliminate this bug an allow a validator
to catchup successfully, provided it does not run out of memory (a
different bug).


nick-drozd:
Typo in commit message: `eliminate this bug an allow`

`an` -> `and`

PR #1380, opened 2018-01-25
Add shutdown on CTL-C to rust transaction processor

agunde406:
Looking for suggestions on ways to improve this. Currently, I needed to add timeouts to 3 places we were blocking on  (using arbitrary timeouts for now). I originally tried adding unregister directly as the handler for ctl-c but the handler can only be set once but we need to be able to update the sender used in case of reconnection.

### DIFF
--- a/sdk/rust/src/processor/mod.rs
+++ b/sdk/rust/src/processor/mod.rs
@@ -18,18 +18,24 @@
 extern crate zmq;
 extern crate protobuf;
 extern crate rand;
+extern crate ctrlc;
+
+use std::sync::mpsc::RecvTimeoutError;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+use std::time::Duration;
+use std::error::Error;
 
 use self::rand::Rng;
 
 pub mod handler;
-
-use std::error::Error;
 
 use protobuf::Message as M;
 use protobuf::repeated::RepeatedField;
 use messages::validator::Message_MessageType;
 use messages::network::PingResponse;
 use messages::processor::TpRegisterRequest;
+use messages::processor::TpUnregisterRequest;
 use messages::processor::TpProcessRequest;
 use messages::processor::TpProcessResponse;
 use messages::processor::TpProcessResponse_Status;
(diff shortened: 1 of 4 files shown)


sdk/rust/src/processor/mod.rs
```diff
@@ -110,7 +116,7 @@ impl<'a> TransactionProcessor<'a> {
                     };
 
                 // Absorb the TpRegisterResponse message
-                let _ = match future.get(){
+                let _ = match future.get_timeout(Duration::from_millis(40000)) {
```
vaporos: Is the duration arbitrary?

agunde406: Yes, you looked at the PR before I added a message asking for better ways to handle this.

peterschwarz: In the other SDK's we don't time out on registration.

sdk/rust/src/processor/mod.rs
```diff
@@ -123,12 +129,51 @@ impl<'a> TransactionProcessor<'a> {
...
+            &generate_correlation_id(),
+            x) {
+                Ok(fut) => fut,
+                Err(err) => {
+                    error!("Unregistration failed: {}", err.description());
+                    return
+                }
+            };
+        // Absorb the TpUnregisterResponse message
+        let _ = match future.get_timeout(Duration::from_millis(40000)){
```
peterschwarz: In the Python SDK, we only wait 1 second for this response.

sdk/rust/src/processor/mod.rs
agunde406: It updated it to timeout out at 10 seconds (instead of 40). The timeout is required because the control handler will not stop the blocking on future.get. Instead, if registration is not completely in 10 seconds, registration will be retried if a control c had not been received.

sdk/rust/src/processor/mod.rs
agunde406: updated to 1 second

agunde406:
Added commit to update gitignore

agunde406:
Updated behavior around register to loop on getting the future instead of resubmitting registration.

PR #1383, opened 2018-01-25
1-0 backport: Fix for two errors dealing with unscheduling

boydjohnson:
I am putting this PR here so I can build debs to run further tests on it. The explanation of the errors are in PR #1382 .


dcmiddle:
I'm not familiar enough with these internals to find nuanced issues with how e.g. _outstanding is managed. That said, I don't see any reason to stop this backporting to 1-0.

PR #1391, opened 2018-01-28
Reorganize doc to improve clarity and ease of use

chenette:
Executive summary of the four commits in this PR:

Top-level docs: Move "Transaction Family Specifications" up (right after the app dev guide),
change titles for three chapters, and add a new "API Reference" chapter (now contains
the SDK API and REST API references).

Transaction Family Specifications: Streamline section titles by dropping "Specification" 
(for example, "IntegerKey Transaction Family").

CLI Command Reference: Reorder the sections - sawtooth first, then the sawtooth-* commands, 
then sawadm/sawnet/sawset, then poet, and then the transaction processor commands in 
alphabetical order.

CLI Command Reference: Change titles to be the command name only (sawtooth, identity-tp, 
xo, etc.) and edit the intro text to correct terminology and to clarify the connection between a 
TP command and the associated transaction family.


docs/source/_templates/indexcontent.html
```diff
@@ -14,29 +14,29 @@ <h1>Hyperledger Sawtooth documentation</h1>
...
+    <span class="linkdescr">{% trans %}perform installation, configuration, and maintenance{% endtrans %}</span></p>
 
-  <p class="biglink"><a class="biglink" href="{{ pathto("cli") }}">{% trans %}CLI Reference{% endtrans %}</a><br/>
-    <span class="linkdescr">{% trans %}command-line tool reference{% endtrans %}</span></p>
+  <p class="biglink"><a class="biglink" href="{{ pathto("api_references") }}">{% trans %}API References{% endtrans %}</a><br/>
+    <span class="linkdescr">{% trans %}explore the Sawtooth API reference guides{% endtrans %}</span></p>
 
-  <p class="biglink"><a class="biglink" href="{{ pathto("transaction_family_specifications") }}">{% trans %}Transaction Family Specifications{% endtrans %}</a><br/>
-    <span class="linkdescr">{% trans %}dive into design of existing transaction families{% endtrans %}</span></p>
+  <p class="biglink"><a class="biglink" href="{{ pathto("cli") }}">{% trans %}CLI Command Reference{% endtrans %}</a><br/>
```
delventhalz: I suppose this has already been hashed out, and decisions have been made, but it seems odd to me that we have the CLI reference at the top level, but the REST API and SDK references tucked away in an API References section. Are all three not APIs? If not, it seems like the difference is more technicality than material. I would think all three should be top level, or all three in one section (whatever you want to call it).

Just my two cents.

chenette: I don't think of a CLI as an application programming interface (or, as Wikipedia calls it, an "application protocol interface").  Yes, an app could run Sawtooth commands, but I believe that the main intended use is by humans and shell scripts.  Also, this organization is per @vaporos' recommendation - he wanted to see the CLI reference at the top level. (That was his two cents.)

PR #1400, opened 2018-01-29
More parallel scheduler optimizations

nick-drozd:
This PR makes two changes to the parallel scheduler.

The first change affects the function `_index_of_txn_in_schedule`. This function was taking around .7% of total CPU time, with most of that being taking up by a list comprehension. It's in the upper left corner of the graph below.

index-before-2

It turns out that the function, which is called frequently, was taking a list of transactions and then building out a new list of transaction ids every time just to get the index of one id. Switching from list comprehension / index to an enumeration causes the function to drop off the graph entirely, indicating that it no longer takes a significant amount of time.

index-after-1

The second change has to do with a function called `_txn_is_in_valid_batch`. This function takes around 5% of total CPU time. About 2/5 of that time (2% of total CPU time) is spent in a little function called `_txn_result_is_invalid`, which checks that a signature is in a dictionary called `_txn_results` and, if it is, whether its result is valid. My thought was that this "look before you leap" approach in such a tight loop was not efficient, and that it would be faster to construct the set of ids with results beforehand and then check whether they are all valid.

The results in the `yappi` graph are a little strange:

valid-after-1

If you look where those functions should be (under `next_transaction`), you'll see that they aren't there at all, and that `next_transaction` has gone from 8-9% to less than 4%. Pretty great, right? Well, if you look in the upper left corner, you'll see two new nodes labeled only `<genexpr>`, representing the generators used in the rewritten function. Together they add up to around 4%, which is a savings of 1% of total CPU time.

I don't quite understand why the nodes are floating off in the corner rather than being attached to their calling chain. Is that computation being executed in another thread? If so, is that good?

### DIFF
--- a/validator/sawtooth_validator/execution/scheduler_parallel.py
+++ b/validator/sawtooth_validator/execution/scheduler_parallel.py
@@ -694,10 +694,6 @@ class ParallelScheduler(Scheduler):
     def _is_outstanding(self, txn_id):
         return txn_id in self._outstanding
 
-    def _txn_result_is_invalid(self, sig):
-        return sig in self._txn_results and \
-            not self._txn_results[sig].is_valid
-
     def _txn_is_in_valid_batch(self, txn_id):
         """Returns whether the transaction is in a valid batch.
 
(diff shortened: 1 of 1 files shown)


nick-drozd:
@boydjohnson I will make those changes after other comments / approvals.

PR #1401, opened 2018-01-30
Two small smallbank workload fixes

boydjohnson:
Use full length addresses, and generate random account names.


perf/smallbank_workload/src/playlist.rs
```diff
@@ -177,7 +177,7 @@ fn customer_id_address(customer_id: u32) -> String {
 
     let hex = bytes_to_hex_str(hash);
     // Using the precomputed Sha512 hash of "supplychain"
```
peterschwarz: supplychain -> smallbank?

PR #1402, opened 2018-01-30
Add key-file option to intkey workload

boydjohnson:

### DIFF
--- a/sdk/examples/intkey_python/sawtooth_intkey/client_cli/intkey_workload.py
+++ b/sdk/examples/intkey_python/sawtooth_intkey/client_cli/intkey_workload.py
@@ -26,11 +26,14 @@ from base64 import b64encode
 import requests
 from sawtooth_signing import create_context
 from sawtooth_signing import CryptoFactory
+from sawtooth_signing import ParseError
+from sawtooth_signing.secp256k1 import Secp256k1PrivateKey
 from sawtooth_intkey.client_cli.workload.workload_generator import \
     WorkloadGenerator
 from sawtooth_intkey.client_cli.workload.sawtooth_workload import Workload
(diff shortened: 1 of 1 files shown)


boydjohnson:
@peterschwarz I updated to not catch IOError.

sdk/examples/intkey_python/sawtooth_intkey/client_cli/intkey_workload.py
```diff
@@ -89,8 +92,22 @@ def __init__(self, delegate, args):
...
+        if args.key_file is not None:
+            try:
+                with open(args.key_file, 'r') as infile:
+                    signing_key = infile.read().strip()
+                private_key = Secp256k1PrivateKey.from_hex(signing_key)
+
+                self._signer = crypto_factory.new_signer(
+                    private_key=private_key)
+            except ParseError:
+                self._signer = crypto_factory.new_signer(
```
peterschwarz: Same situation as previously mentioned: the user provided a bad key and they should be informed of that error.  Otherwise, it will silently go on to create a new key.

boydjohnson: Makes sense. I will log an error and not make the key.

boydjohnson: Since we don't have an IntkeyCliError, I think it makes sense to just not catch the error.

boydjohnson:
@peterschwarz Since we don't have an IntkeyCliError, I think it makes sense to just not catch the error.

peterschwarz:
You could use `from sawtooth_intkey.client_cli.exceptions import IntKeyCliException`

PR #1404, opened 2018-01-30
Ignore unscheduled transaction results

peterschwarz:
When an transaction has been removed from the schedule, but was being
processed by a transaction processor, the scheduler will ignore any
results provided.

Includes a pair of unit tests for both schedulers to verify this
behavior.

### DIFF
--- a/validator/sawtooth_validator/execution/scheduler_parallel.py
+++ b/validator/sawtooth_validator/execution/scheduler_parallel.py
@@ -651,8 +651,6 @@ class ParallelScheduler(Scheduler):
                     "transaction not scheduled: {}".format(txn_signature))
 
             if txn_signature not in self._batches_by_txn_id:
-                LOGGER.debug('Received result for %s, but was unscheduled',
-                             txn_signature)
                 return
 
             self._set_least_batch_id(txn_signature=txn_signature)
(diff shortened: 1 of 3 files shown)


validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -649,6 +649,12 @@ def set_transaction_execution_result(
             if txn_signature not in self._scheduled:
                 raise SchedulerError(
                     "transaction not scheduled: {}".format(txn_signature))
+
+            if txn_signature not in self._batches_by_txn_id:
+                LOGGER.debug('Received result for %s, but was unscheduled',
```
nick-drozd: This feels like it will clutter the logs.

peterschwarz: I was using the same logging strategy from the serial scheduler.  I could remove it from here and the serial scheduler.

validator/tests/test_scheduler/tests.py
```diff
@@ -428,6 +428,101 @@ def _test_wildcarded_unschedule(self, scheduler, context_manager):
...
+        self._test_unschedule(
+            scheduler=scheduler,
+            context_manager=context_manager)
+
+    def _test_unschedule(self, scheduler, context_manager):
+        """Tests that each of the top level methods on scheduler can be
+        called after unscheduling.
+
+        Notes:
+
```
agunde406: Missing Notes

peterschwarz: Notes added.

validator/sawtooth_validator/execution/scheduler_parallel.py
nick-drozd: They could also be replaced with metrics.

PR #1405, opened 2018-01-30
Log consensus readiness at boundaries

peterschwarz:
Adds a dict to maintina logging states that may be entered or exited
asynchronously within the publishers processing cycle.  Adds the
"consensus_not_ready" state, which tracks that condition related to
block publishing readiness, according to the publisher.

### DIFF
--- a/validator/sawtooth_validator/journal/publisher.py
+++ b/validator/sawtooth_validator/journal/publisher.py
@@ -409,6 +409,14 @@ class _CandidateBlock(object):
         return builder.build_block()
 
 
+class _PublisherLoggingStates:
+    """Collects and tracks the changes in various states of the Publisher.  For
+    example it tracks 'consensus_not_ready', which denotes entering or exiting
+    of that state."""
+    def __init__(self, consensus_not_ready=False):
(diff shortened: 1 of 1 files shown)


validator/sawtooth_validator/journal/publisher.py
```diff
@@ -496,6 +496,10 @@ def __init__(self,
         self._check_publish_block_frequency = check_publish_block_frequency
         self._publisher_thread = None
 
+        # A series of states that allow us to check for condition changes.
+        # These can be used to log only at the boundary of condition changes.
+        self._logging_states = {}
```
aludvik: I would prefer a python object with explicit attributes to a dictionary, since it is easy to abuse the dictionary

aludvik: Since this is a pattern we are likely to repeat elsewhere, it may also be worth thinking about how to do this consistently in other objects.

peterschwarz: Not all states will potentially be boolean, hence the choice of dict here.  How about we split the difference and go with a named tuple?

PR #1406, opened 2018-01-31
Rewrite duplicate/missing transaction/batch checks

aludvik:
This depends on #1403 and includes a number of other chain controller refactor commits from the consensus-prototype branch

### DIFF
--- a/validator/sawtooth_validator/concurrent/atomic.py
+++ b/validator/sawtooth_validator/concurrent/atomic.py
@@ -65,3 +65,62 @@ class ConcurrentSet:
     def __contains__(self, element):
         with self._lock:
             return element in self._set
+
+
+class ConcurrentMultiMap:
+    """A dictionary that maps keys to lists of items. All methods are
+    modifications to the referenced list."""
+    def __init__(self):
+        self._dict = dict()
+        self._lock = RLock()
+
+    def __contains__(self, key):
+        with self._lock:
+            return key in self._dict
+
+    def append(self, key, item):
+        """Append item to the list at key. Creates the list at key if it
+        doesn't exist.
+        """
+        with self._lock:
+            if key in self._dict:
+                self._dict[key].append(item)
+            else:
+                self._dict[key] = [item]
+
+    def set(self, key, items):
+        """Set key to a copy of items"""
+        if not isinstance(items, list):
+            raise ValueError("items must be a list")
+        with self._lock:
+            self._dict[key] = items.copy()
+
+    def swap(self, key, items):
+        """Set key to a copy of items and return the list that was previously
+        stored if the key was set. If not key was set, returns an empty list.
+        """
+        if not isinstance(items, list):
+            raise ValueError("items must be a list")
+        return_value = []
+        with self._lock:
+            if key in self._dict:
+                return_value = self._dict[key]
+            # Make a copy since we don't want users keeping a reference that is
+            # outside the lock
+            self._dict[key] = items.copy()
+        return return_value
+
+    def pop(self, key, default):
+        """If the key is set, remove and return the list stored at key.
+        Otherwise return default."""
+        with self._lock:
+            return self._dict.pop(key, default)
+
+    def get(self, key, default):
+        """If the key is set, return a copy of the list stored at key.
+        Otherwise return default."""
+        with self._lock:
+            try:
+                return self._dict[key].copy()
+            except KeyError:
+                return default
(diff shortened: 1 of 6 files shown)


validator/sawtooth_validator/journal/block_validator.py
```diff
@@ -477,7 +477,7 @@ def validate_block(self, blkw, chain_head=None, chain=None):
 
         except Exception as e:
             LOGGER.exception(
-                "Unhandled exception BlockPublisher.validate_block()")
```
nick-drozd: `validate_block` is called in just two places, and both of those calls are wrapped in broad `except`s, so this catch isn't necessary at all.

aludvik:
In general, this PR tries to do a better job of abstracting some difficult book keeping tasks.

aludvik:
Also, the ordering of commits got weird because I dropped some other commits that were causing problems.

validator/sawtooth_validator/concurrent/atomic.py
```diff
@@ -46,3 +47,80 @@ def inc(self, step=1):
     def dec(self, step=1):
         with self._lock:
             self._value -= step
+
+
+class ConcurrentSet:
+    def __init__(self):
+        self.set = set()
+        self.lock = RLock()
```
boydjohnson: Though RLock can be used in place of Lock, it doesn't seem necessary here. https://stackoverflow.com/a/22885810/1766739 This stackoverflow answer describes the use of RLock; when calling functions that are synchronized with the same lock. I would also use `self._lock` instead of `self.lock`. The protection we get from the linter that an internal lock doesn't get called externally to the object by accident is important.

validator/sawtooth_validator/journal/chain_commit_state.py
```diff
@@ -13,6 +13,174 @@
...
+        self.uncommitted_batch_ids = uncommitted_batch_ids
+        self.uncommitted_txn_ids = uncommitted_txn_ids
+
+    def _block_in_chain(self, block):
+        if self.common_ancestor is not None:
+            return block.block_num <= self.common_ancestor.block_num
+        return False
+
+    @staticmethod
+    def _check_for_duplicates_within(key_fn, items):
```
nick-drozd: This one feels like an exam question. Could you add a description?

validator/sawtooth_validator/concurrent/atomic.py
aludvik: I'll fix `self._lock`. I am not going to change RLock to Lock in this PR because I was not able to convince myself that the lock didn't need to be re-entrant when I wrote this in the first place and RLock has no performance cost over Lock: https://docs.python.org/3/whatsnew/3.2.html#optimizations "Recursive locks (created with the threading.RLock() API) now benefit from a C implementation which makes them as fast as regular locks, and between 10x and 15x faster than their previous pure Python implementation."

validator/sawtooth_validator/journal/block_validator.py
```diff
@@ -185,73 +180,65 @@ def _get_previous_block_state_root(self, blkw):
...
+            BlockValidationError:
+                If validation fails, raises this error with the reason.
+            MissingDependency:
+                Validation failed because of a missing dependency.
+            DuplicateTransaction:
+                Validation failed because of a duplicate transaction.
+            DuplicateBatch:
+                Validation failed because of a duplicate batch.
+        """
         if blkw.block.batches:
```
nick-drozd: In order to save a level of indentation, it would be nice if this whole section were rewritten as

```
if not blkw.block.batches:
    return

try:
    blah blah blah
```

validator/sawtooth_validator/concurrent/atomic.py
boydjohnson: Thanks, That was news to me.

validator/sawtooth_validator/journal/block_validator.py
```diff
@@ -488,10 +493,37 @@ def _get_batch_commit_changes(new_chain, cur_chain):
     def submit_blocks_for_verification(
         self, blocks, consensus, callback
     ):
+
         for block in blocks:
+            LOGGER.debug(
+                "Adding block %s for processing", block.identifier[:6])
```
nick-drozd: Identifiers should be logged in full (unless this message is going to be logged frequently, in which case it should be dropped or replaced with a metric).

aludvik:
@boydjohnson changed concurrent class instance variables to protected.
@nick-drozd Added docstring, reduced indentation, and logged full identifier

PR #1410, opened 2018-02-01
Relocate metrics dockerfiles and improve dashboard

dplumb94:
- The influxdb and grafana dockerfiles do not follow the same format as the rest of our Docker stuff. This moves the dockerfiles into the docker directory, and moves the scripts and json files to their own directory.

- This PR adds some new metrics to the grafana dashboard. It also utilizes a different json structure for the dashboard that allows us to use a template variable for the metrics datasource. This means we will not have to create a new dashboard for each LR network that show the same graphs. It will also allow us to create new dashboards for specialized metrics without having to edit the main dashboard.

- The sawtooth-metrics.yaml docker-compose file starts up a network that is connected to influxdb and grafana. This allows us to easily test new metrics or conduct other experiments using Docker. The containers are built from mounted files. The grafana and influxdb images will automatically be built if they are not already.

[Creating and Editing Dashboards](https://docs.google.com/document/d/1yAOzW_Ez7zXPSAzakY1o4gYUFbnKmr95Xwmc9BWUgE4/edit#)

### DIFF
--- a/bin/build_all
+++ b/bin/build_all
@@ -237,10 +237,10 @@ docker_build() {
 build_external() {
     docker_build docker/apache-basic_auth_proxy docker/ apache-basic_auth_proxy
     docker_build docker/grafana/sawtooth-stats-grafana \
-        docker/grafana \
+        docker/ \
         sawtooth-stats-grafana
     docker_build docker/influxdb/sawtooth-stats-influxdb \
-        docker/influxdb \
+        docker/ \
         sawtooth-stats-influxdb
 }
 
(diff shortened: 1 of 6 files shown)


dplumb94:
Updated based on @vaporos first two comments.


> The multiple config files seem unnecessary - we could template them and use sed to replace values to generate the specific files.

I am not sure how this would be implemented. Where would the sed commands be called?

bin/build_all
```diff
@@ -237,10 +237,10 @@ docker_build() {
 build_external() {
     docker_build docker/apache-basic_auth_proxy docker/ apache-basic_auth_proxy
     docker_build docker/grafana/sawtooth-stats-grafana \
-        docker/grafana \
+        . \
```
vaporos: This change seems unnecessary.  Is it?

dplumb94: The build context has to be changed so that the grafana and influxdb startup scripts can be accessed from docker-compose. But it can be improved to just be `docker/`, which is the same format as the other build scripts. I will make this change

PR #1411, opened 2018-02-02
Fix STL-1050 incorrect poet key state store use

dcmiddle:
Fix root cause of STL-1050 and add protections for backwards
compatibility with Sawtooth v1.0.1

Issue introduced in #1288 

Will want to backport #1288 and this PR together. Presumably into v1.0.2.

### DIFF
--- a/consensus/poet/core/sawtooth_poet/poet_consensus/poet_block_publisher.py
+++ b/consensus/poet/core/sawtooth_poet/poet_consensus/poet_block_publisher.py
@@ -233,6 +233,7 @@ class PoetBlockPublisher(BlockPublisherInterface):
                          'committed by block %s. Create new registration',
                          poet_public_key,
                          block_header.previous_block_id)
+
             del self._poet_key_state_store[poet_public_key]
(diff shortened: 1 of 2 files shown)


consensus/poet/core/sawtooth_poet/poet_consensus/poet_block_publisher.py
```diff
@@ -233,7 +233,10 @@ def _handle_registration_timeout(self, block_header, poet_enclave_module,
                          'committed by block %s. Create new registration',
                          poet_public_key,
                          block_header.previous_block_id)
-            del self._poet_key_state_store[poet_public_key]
+            try:
+                del self._poet_key_state_store[poet_public_key]
+            except IndexError:
```
peterschwarz: Doesn't this throw a `KeyError`?

dcmiddle: good point. the delete method actually eats that exception anyway. I'll pull it out.

PR #1412, opened 2018-02-05
Split off predecessor tree's tree

nick-drozd:
Currently the predecessor tree's data structure is not particularly efficient. Separating the data structure for the predecessor tree itself will make it easier to improve.

### DIFF
--- a/validator/sawtooth_validator/execution/scheduler_parallel.py
+++ b/validator/sawtooth_validator/execution/scheduler_parallel.py
@@ -65,10 +65,10 @@ class PredecessorTree:
         return repr(self._root)
 
     def _tokenize_address(self, address):
-        return [
+        return (
             address[i:i + self._token_size]
             for i in range(0, len(address), self._token_size)
-        ]
+        )
 
     def insert(self, address):
         tokens = self._tokenize_address(address)
@@ -95,6 +95,21 @@ class PredecessorTree:
                 return None
 
         return node
+
+    def get_ancestors(self, address):
+        '''
+        Returns a stream of the ancestors of the node at address,
+        starting at the root and walking down.
+
+        Raises KeyError if there is no node at address.
+        '''
+        node = self._root
+
+        yield node
+
+        for token in self._tokenize_address(address):
+            node = node.children[token]
+            yield node
 
     def add_reader(self, address, reader):
         node = self.insert(address)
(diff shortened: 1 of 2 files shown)


validator/sawtooth_validator/execution/predecessor_tree.py
```diff
@@ -37,7 +37,7 @@ def __repr__(self):
         return repr(retval)
 
 
-class PredecessorTree:
+class Tree:
```
vaporos: What we want to end up with here is a generic RadixTrie implementation, so 'RadixTree' or 'RadixTrie' would be sensible names.

validator/sawtooth_validator/execution/predecessor_tree.py
```diff
@@ -86,12 +86,17 @@ def get_nodes_above(self, address):
...
             yield node
 
+
+class PredecessorTree:
+    def __init__(self, token_size=2):
+        self._tree = Tree(token_size)
+
     def add_reader(self, address, reader):
-        node = self.insert(address)
+        node = self._tree.insert(address)
```
vaporos: The radix tree should not allocate the stored data structure; instead, this should be "self._tree.insert(address, node_data)" where node could be a PredecessorTreeNode or any other python data structure.

nick-drozd: fixed

validator/sawtooth_validator/execution/predecessor_tree.py
nick-drozd: The name can be changed when the data structure changes. Alternatively, the name can be left vague, and the specifics of the data structure can be changed without the possibility of rendering the name inaccurate.

validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -65,10 +65,10 @@ def __repr__(self):
         return repr(self._root)
 
     def _tokenize_address(self, address):
-        return [
+        return (
             address[i:i + self._token_size]
             for i in range(0, len(address), self._token_size)
-        ]
+        )
```
aludvik: This change should probably be in a separate commit.

validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -96,6 +96,15 @@ def get(self, address):
 
         return node
 
+    def get_nodes_above(self, address):
```
aludvik: A docstring would be helpful here, to clarify what `above` means

nick-drozd: Renamed to `get_ancestors` and docstring added

validator/sawtooth_validator/execution/scheduler_parallel.py
nick-drozd: I don't think it's worth a full commit, but I added a comment about it in the `get_ancestors` commit.

validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -111,14 +105,44 @@ def get_ancestors(self, address):
...
+        self._tree = Tree(token_size=token_size)
+
+    def __repr__(self):
+        return repr(self._tree)
+
+    def _insert(self, address):
+        node = self._tree.insert(address)
+
+        if node.data is None:
+            node.data = Predecessors()
```
aludvik: Can you just pass the new instance of Predecessor to `self._tree.insert`?

nick-drozd: No. The logic of `insert` (maybe it needs a different name) is to get the node at address if it exists, and create it if it doesn't. In the former case, we don't necessarily want to overwrite the existing data.

peterschwarz:
Is there anyway to add a unit test for the `Tree` class by itself?  I'm having trouble understanding what it is going to look like in use outside of the PredecessorTree.

validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -36,67 +35,114 @@
...
-        if self.children:
-            retval['children'] = \
-                {k: literal_eval(repr(v)) for k, v in self.children.items()}
 
-        return repr(retval)
-
-
-class PredecessorTree:
-    def __init__(self, token_size=2):
+class Tree:
```
jsmitchell: Calling this a Tree is confusing, since its interface doesn't follow expected Tree semantics. Call this a RadixTree or AddressedTree or something better.

PR #1414, opened 2018-02-05
Refactor XO python

boydjohnson:


vaporos:
@chenette Is breaking it up into multiple files friendly from a documentation perspective? (Hoping yes.)

chenette:
@vaporos - I'm not sure what "it" refers to in "...breaking it up into multiple files...".  This PR is code only, right?

sdk/examples/xo_python/sawtooth_xo/processor/xo_state.py
```diff
@@ -0,0 +1,203 @@
...
+        Returns (str): The state that the game is in.
+        """
+
+        return self._game[self._game_name].state
+
+    @property
+    def board(self):
+        """The game board before the current transaction.
+
+
```
nick-drozd: Missing `Returns` comment

sdk/examples/xo_python/sawtooth_xo/processor/handler.py
```diff
@@ -38,88 +39,66 @@ def family_versions(self):
 
     @property
     def namespaces(self):
-        return [self._namespace_prefix]
+        return [xo_state.XO_NAMESPACE]
 
     def apply(self, transaction, context):
 
-        # 1. Deserialize the transaction and verify it is valid
```
nick-drozd: I realize there is going to be accompanying documentation, but it still seems to me like it would be a good idea to keep some narrative comments in the code (they don't have to be the existing ones).

sdk/examples/xo_python/sawtooth_xo/processor/xo_state.py
boydjohnson: Thanks. I will update this after I get more reviews.

sdk/examples/xo_python/sawtooth_xo/processor/handler.py
boydjohnson: I would like the code to be as clear as possible so that comments are almost not needed. The accompanying documentation can call out particular steps. I would wonder if particular method names could be extended to be clear about what they do. XoPayload.from_bytes() could be something else more descriptive or there could be a comment that it also validates the payload.

sdk/examples/xo_python/sawtooth_xo/processor/handler.py
```diff
@@ -50,47 +48,57 @@ def apply(self, transaction, context):
 
         xo_payload = XoPayload.from_bytes(transaction.payload)
 
-        # 2. Retrieve the game data from state storage
-        board, state, player1, player2, game_list = \
-            _get_state_data(context, self._namespace_prefix, xo_payload.name)
+        xo_context = xo_state.XoState(context, xo_payload.name)
```
peterschwarz: Let's call this `xo_state` to match the class

sdk/examples/xo_python/sawtooth_xo/processor/xo_state.py
```diff
@@ -0,0 +1,198 @@
...
+        self._game[self._game_name] = (board, state, player1, player2)
+
+    def create_player2(self, player2):
+
+        board, state, player1, _ = self._game[self._game_name]
+
+        self._game[self._game_name] = (board, state, player1, player2)
+
+    @property
+    def state(self):
```
peterschwarz: So as not confuse users, can we call this game state or something?  Calling a field state in a class called XoState is moderately confusing.

boydjohnson: Yeah, agreed.

sdk/examples/xo_python/sawtooth_xo/processor/handler.py
boydjohnson: Sounds good.

sdk/examples/xo_python/sawtooth_xo/processor/handler.py
```diff
@@ -38,211 +41,95 @@ def family_versions(self):
...
-        raise InternalError("State Error")
-
-
-def _delete_game(context, name, namespace_prefix):
-    LOGGER.warning('Deleting game %s', name)
-
-    address = _make_xo_address(namespace_prefix, name)
-
-    addresses = context.delete_state([address])
+            xo_state.set_game(game=game)
```
peterschwarz: This all looks good.  The only suggestion I would make is to either:

a) Make this follow the analogous key/value model with
```
xo_state.set_game(xo_payload.name, game)
```
b) Rename this to `save_game`

sdk/examples/xo_python/sawtooth_xo/processor/xo_state.py
```diff
@@ -0,0 +1,150 @@
...
+        self.state = state
+        self.player1 = player1
+        self.player2 = player2
+
+
+class XoState(object):
+
+    TIMEOUT = 3
+
+    def __init__(self, context, game_name):
```
vaporos: This constructor shouldn't take a game_name, as XoState should appear to be stateless, and you should be able to access/store any number of games using this class. (Because it's a better example, not because it is used this way by the Xo game logic.)

sdk/examples/xo_python/sawtooth_xo/processor/xo_state.py
```diff
@@ -0,0 +1,150 @@
...
+                validator state from within the transaction processor.
+
+            game_name (str): The name of the game currently being acted on.
+        """
+
+        self._context = context
+        self._game_name = game_name
+        self._game_address = _make_xo_address(game_name)
+
+        state_entries = context.get_state(
```
vaporos: The constructor should do no operations on context. Those should be lazily done in the methods requiring them.

sdk/examples/xo_python/sawtooth_xo/processor/xo_state.py
```diff
@@ -0,0 +1,150 @@
...
+
+        Returns:
+            (Game): All the information specifying a game.
+        """
+        try:
+            return self._games[game_name]
+        except KeyError:
+            return None
+
+    def _store_game(self):
```
vaporos: This is too confusing for example code.  Maybe we can break seralization/deserializion of the game state out to separate functions.

sdk/examples/xo_python/sawtooth_xo/processor/xo_state.py
```diff
@@ -0,0 +1,150 @@
...
+        state_data = '|'.join(sorted([
+            ','.join([name, g.board, g.state, g.player1, g.player2])
+            for name, g in self._games.items()
+        ])).encode()
+
+        addresses = self._context.set_state(
+            {self._game_address: state_data},
+            timeout=self.TIMEOUT)
+
+        if len(addresses) < 1:
```
vaporos: Is this check actually necessary?  Will we ever get back addresses that were not saved but have no error message stating why, without context.set_state() throwing an error?

sdk/examples/xo_python/sawtooth_xo/processor/xo_state.py
```diff
@@ -0,0 +1,150 @@
...
+            raise InternalError("State set error: {}".format(
+                self._game_address))
+
+    def _delete_game(self):
+        addresses = self._context.delete_state(
+            [self._game_address],
+            timeout=self.TIMEOUT)
+
+        if len(addresses) < 1:
```
vaporos: Same question here - is this necessary?

sdk/examples/xo_python/sawtooth_xo/processor/xo_payload.py
```diff
@@ -0,0 +1,72 @@
...
+    @property
+    def action(self):
+        return self._action
+
+    @property
+    def space(self):
+        return self._space
+
+
+def _validate_transaction(name, action, space):
```
vaporos: This doesn't take a transaction, so validate_transaction is confusing. Would this be cleaner as a _validate() method on XoPayload or even inline in the XoPayload constructor?

sdk/examples/xo_python/sawtooth_xo/processor/handler.py
vaporos: Let's keep it set_ so it matches set_state() which is in the implementation of this method. set_game(name, game) seems good to me.

sdk/examples/xo_python/sawtooth_xo/processor/handler.py
```diff
@@ -38,211 +41,95 @@ def family_versions(self):
...
+            _display("Player {} created a game.".format(signer[:6]))
 
-    if not addresses:
-        raise InternalError('State delete error')
+        elif xo_payload.action == 'take':
+            game = xo_state.get_game(xo_payload.name)
 
+            if game is None:
+                raise InvalidTransaction(
+                    'Invalid action: Take requires an existing game.')
```
vaporos: Let's lose the ending periods since it's not friendly to chaining exception error messages together. (Ideally as another commit since this is a hold-over from the previous code.)

sdk/examples/xo_python/sawtooth_xo/processor/xo_state.py
boydjohnson: So in the case that the context Id is not known (which happens often during unscheduling) the context.set_state would get back an empty list. It shouldn't be an InternalError.

boydjohnson:
@vaporos I have fixed according to your comments. In the fix it was hard to keep two of the commits separate, so the "Refactor to use XoState" and "Refactor validation rules" commits are now one commit.

sdk/examples/xo_python/sawtooth_xo/processor/xo_state.py
```diff
@@ -0,0 +1,160 @@
...
+            [address],
+            timeout=self.TIMEOUT)
+        if state_entries:
+
+            games = self._deserialize(data=state_entries[0].data)
+
+        else:
+            games = {}
+
+        self._games = games
```
aludvik: Per our discussion, I think the `XoState` class should be a thin wrapper around the context object that handles (de)serialization of games, hash collisions, and address creation. In this model, it makes sense to maintain an up-to-date cache between the handler and the validator to reduce the number of messages that need to be exchanges and keep track of any games at the address of interest that are not the game of interest because of a hash collision.

I think another valid API, which it sounds like you started with, would be to have one XoState object per game you are interacting with. However in that case, you would probably want the class to be called XoGame and to take the game name as a constructor argument. In that case, the class is more of a network storage backed RPC object.

While both are valid, it looks like we are already using the first one in other places such as intkey rust: https://github.com/hyperledger/sawtooth-core/blob/master/sdk/examples/intkey_rust/src/handler.rs#L139

sdk/examples/xo_python/sawtooth_xo/processor/xo_state.py
```diff
@@ -0,0 +1,175 @@
...
+
+        if address in self._address_cache:
+            serialized_games = self._address_cache[address]
+            games = self._deserialize(serialized_games)
+        else:
+            state_entries = self._context.get_state(
+                [address],
+                timeout=self.TIMEOUT)
+            if state_entries:
+
```
vaporos: Extra whitespace here

sdk/examples/xo_python/sawtooth_xo/processor/xo_state.py
```diff
@@ -0,0 +1,175 @@
...
+        Args:
+            game_name (str): The name.
+
+        Raises:
+            KeyError: The Game with game_name does not exist.
+        """
+
+        address = _make_xo_address(game_name)
+        self._context.delete_state(
```
vaporos: This does not make sense here, as it will delete space even if there are other games present.

sdk/examples/xo_python/sawtooth_xo/processor/xo_state.py
```diff
@@ -0,0 +1,175 @@
...
+            [address],
+            timeout=self.TIMEOUT)
+
+    def _load_games(self, game_name):
+        address = _make_xo_address(game_name)
+
+        if address in self._address_cache:
+            serialized_games = self._address_cache[address]
+            games = self._deserialize(serialized_games)
+        else:
```
aludvik: When you are getting/setting/deleting state the address cache needs to be updated to accurately reflect whatever the validator has, otherwise the address cache will cause problems or not be useful.

sdk/examples/xo_python/sawtooth_xo/processor/xo_state.py
boydjohnson: Yeah, you are right. I was moving code around and somehow that happened.

sdk/examples/xo_python/sawtooth_xo/processor/xo_state.py
boydjohnson: There is an intentional line to aid in readability.

sdk/examples/xo_python/sawtooth_xo/processor/xo_state.py
```diff
@@ -0,0 +1,180 @@
...
+        """
+
+        games = {}
+        try:
+            for game in data.decode().split("|"):
+                name, board, state, player1, player2 = game.split(",")
+
+                games[name] = Game(name, board, state, player1, player2)
+        except ValueError:
+            raise InternalError("Failed to deserialize game data.")
```
nick-drozd: period

PR #1415, opened 2018-02-05
Add generating proto files to smallbank_workload

boydjohnson:
This PR has three main fixes:
  - #1325 took out generating rust files from protogen script and handled it by using a build.rs file. This PR adds it back in for smallbank_workload.
 - Removing the wait from the sawtooth_perf long running batch submitter. This is only a short-term fix and will be handled better by work done in STL-1052. 
- ~Adding the Cargo.lock to version control since smallbank_workload is a bin. https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html This cargo documentation explains.~
- Changing the default transactions per batch to 1.

### DIFF
--- a/perf/smallbank_workload/src/main.rs
+++ b/perf/smallbank_workload/src/main.rs
@@ -109,7 +109,7 @@ fn create_load_subcommand_args<'a, 'b>() -> App<'a, 'b> {
               .short("n")
               .long("max-batch-size")
               .value_name("NUMBER")
-              .help("The number of transaction in a batch. Defaults to 100."))
+              .help("The number of transaction in a batch. Defaults to 1."))
         .arg(Arg::with_name("num-accounts")
               .short("a")
               .long("accounts")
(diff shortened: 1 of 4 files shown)


aludvik:
@vaporos do we like Cargo.lock files?

perf/smallbank_workload/src/main.rs
```diff
@@ -147,7 +147,7 @@ fn create_load_subcommand_args<'a, 'b>() -> App<'a, 'b> {
 
 fn run_load_command(args: &ArgMatches) -> Result<(), Box<Error>> {
     let max_txns: usize = match args.value_of("max-batch-size")
-        .unwrap_or("100")
+        .unwrap_or("1")
         .parse() {
             Ok(n) => n,
             Err(_) => 0,
```
peterschwarz: This error should be returned, not silently set to `0`

boydjohnson: Yeah, I will update that.

boydjohnson: Thinking about it, the next check is whether max_txns is not equal to 0. All failures to parse as unsigned int state that "max-batch-size must be a number greater than 0" which seems like a better statement to the user than panicking on a ParseIntError kind: INvalidDigit.

PR #1416, opened 2018-02-05
Reject Batches That Will Violate Block Validation Rules

peterschwarz:
Validate that a batch added to a candidate block does not violate the block validation rules specified in the settings.  Before this change, batches that violated these rules were only verified before the block was committed, but this left open the possibility of rule-violating transactions being processed, which could affect downstream state in a given schedule.

Additionally:

* Adds block-info-tp to PoET Liveness test

### DIFF
--- a/validator/sawtooth_validator/journal/validation_rule_enforcer.py
+++ b/validator/sawtooth_validator/journal/validation_rule_enforcer.py
@@ -18,162 +18,164 @@ from sawtooth_validator.protobuf.transaction_pb2 import TransactionHeader
 LOGGER = logging.getLogger(__name__)
 
 
-class ValidationRuleEnforcer(object):
-    def __init__(self, setting_view_factory):
-        self._settings_view_factory = setting_view_factory
+def enforce_validation_rules(settings_view, expected_signer, batches):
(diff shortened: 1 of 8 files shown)


peterschwarz:
This should also be tested to see if it fixes the conditions discussed in [STL-1034](https://jira.hyperledger.org/browse/STL-1034)

validator/sawtooth_validator/journal/validation_rule_enforcer.py
```diff
@@ -62,118 +69,121 @@ def validate(self, block, state_root):
...
-            header = TransactionHeader()
-            header.ParseFromString(txn.header)
-            if header.family_name == family:
-                count += 1
-
-            if count > limit:
-                LOGGER.debug("Too many transactions of type %s", family)
-                return False
 
+def _do_nofx(transactions, arguments):
```
nick-drozd: These low-level functions should pass error messages upwards rather than doing the logging themselves.

peterschwarz:
Added a commit to fix [STL-1048](https://jira.hyperledger.org/browse/STL-1048)

PR #1417, opened 2018-02-06
Implement blockstore stats command

aludvik:
This adds a subcommand to compute some basic summary stats for the
blockstore database. More detailed stats can be added later.

### DIFF
--- a/adm/src/commands/blockstore.rs
+++ b/adm/src/commands/blockstore.rs
@@ -15,6 +15,7 @@
  * ------------------------------------------------------------------------------
  */
 
+use std::collections::HashMap;
 use std::io::{self, Read, Write};
 use std::fs::File;
 
@@ -24,6 +25,7 @@ use protobuf::Message;
 use serde_yaml;
 
 use sawtooth_sdk::messages::block::{Block, BlockHeader};
+use sawtooth_sdk::messages::transaction::{TransactionHeader};
 
 use blockstore::Blockstore;
 use database::lmdb;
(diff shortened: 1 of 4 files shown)


adm/src/commands/blockstore.rs
```diff
@@ -265,6 +266,42 @@ fn run_import_command<'a>(args: &ArgMatches<'a>) -> Result<(), CliError> {
...
+    // Get the chain head
+    let chain_head = blockstore.get_chain_head()
+        .map_err(|err| CliError::EnvironmentError(format!("{}", err)))?;
+
+    let mut block = blockstore.get(&chain_head)
+        .map_err(|err| CliError::EnvironmentError(format!("{}", err)))?;
+
+    loop {
+        block_count += 1;
+        batch_count += block.batches.len();
```
peterschwarz: You can just get the number of entries in `batch` index

adm/src/commands/blockstore.rs
```diff
@@ -265,6 +266,42 @@ fn run_import_command<'a>(args: &ArgMatches<'a>) -> Result<(), CliError> {
...
+        for batch in &block.batches {
+            txn_count += batch.transactions.len();
```
peterschwarz: Likewise, you can just get the number of entries in the `transaction` index

adm/src/commands/blockstore.rs
```diff
@@ -265,6 +266,42 @@ fn run_import_command<'a>(args: &ArgMatches<'a>) -> Result<(), CliError> {
...
+
+    // Get the chain head
+    let chain_head = blockstore.get_chain_head()
+        .map_err(|err| CliError::EnvironmentError(format!("{}", err)))?;
+
+    let mut block = blockstore.get(&chain_head)
+        .map_err(|err| CliError::EnvironmentError(format!("{}", err)))?;
+
+    loop {
+        block_count += 1;
```
peterschwarz: block count is just the chain heads block number

peterschwarz:
We manage those indexes - they are just databases where the keys are the indexer (e.g. batch id) and the value is the block num.  They are all updated in a single LMDB transaction, so if we trust those to operate correctly, we can trust that the indexes are correct.

PR #1419, opened 2018-02-06
Log TP internal error on validator

dplumb94:
This commit logs any transaction processor internal errors at the error
level on the validator. It also includes information about the
transaction that was being executed when the error ocurred.

### DIFF
--- a/validator/sawtooth_validator/execution/executor.py
+++ b/validator/sawtooth_validator/execution/executor.py
@@ -132,6 +132,13 @@ class TransactionExecutorThread(object):
                 data=data)
 
         elif response.status == processor_pb2.TpProcessResponse.INTERNAL_ERROR:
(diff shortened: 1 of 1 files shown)


validator/sawtooth_validator/execution/executor.py
```diff
@@ -133,6 +133,15 @@ def _future_done_callback(self, request, result):
 
         elif response.status == processor_pb2.TpProcessResponse.INTERNAL_ERROR:
 
+            LOGGER.exception(
+                "Transaction Processor Internal Error:\n %s",
```
nick-drozd: Log messages shouldn't have newlines (it makes it more difficult to grep).

aludvik: The log level should also be error, not exception

PR #1423, opened 2018-02-07
Remove explicit priority from within dispatcher

peterschwarz:
Remove the explicit priority setting from within the dispatcher,
allowing priority be configured as either part of the handler addition
or explicitly for handlers grouped by message type.


peterschwarz:
Rebased with master

PR #1424, opened 2018-02-07
Add logging implementation to Java sdk tests

boydjohnson:
This cleans up some warnings that are printed during `mvn package`.

### DIFF
--- a/sdk/java/pom.xml
+++ b/sdk/java/pom.xml
@@ -23,6 +23,10 @@
     <artifactId>sdk</artifactId>
     <version>1.0-SNAPSHOT</version>
 
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    </properties>
+
     <build>
         <extensions>
             <extension>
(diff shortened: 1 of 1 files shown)


sdk/java/src/test/java/sawtooth/sdk/client/SigningTest.java
```diff
@@ -19,7 +19,6 @@
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.lang.String;
-import java.math.BigInteger;
```
peterschwarz: Is this change relevant of for this PR, or should it at least be in a separate commit?

boydjohnson: It could be in a separate commit. I removed 2 unneeded imports.

boydjohnson: I think I am going to drop the removal of the imports from the PR.

PR #1425, opened 2018-02-08
Add transactions in process metric

dplumb94:
This PR adds a new metric to track the transactions that are currently in process for each node. In order to do this, the metrics counter wrapper had to be extended to allow the `dec()` method. This PR also adds the metric to the dashboard.

In order to test this metric, run the `sawtooth-local-poet.yaml` docker-compose file, and set a workload. Since the validators in this compose file use the serial scheduler, we see on the graph that each node has between 0 and 2 transactions in process at a time. Next, edit the validators in `sawtooth-local-poet.yaml` to use the parallel scheduler. We see that many more transactions can be processed in parallel with this scheduler.

### DIFF
--- a/docker/grafana/dashboards/dashboard.json
+++ b/docker/grafana/dashboards/dashboard.json
@@ -971,7 +971,7 @@
           "renderer": "flot",
           "seriesOverrides": [],
           "spaceLength": 10,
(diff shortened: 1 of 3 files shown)


aludvik:
Tested locally

PR #1426, opened 2018-02-08
Cleanup standard metrics dashboard

aludvik:
- Removes the legend from most graphs
- Fixes any derivatives to be over 1s instead of 10s
- Updates Packets Per Second and Bytes Per Second to use a derivative
  correctly and show sent as negative
- Fixes a variety of issues with the new threadpool metrics


dplumb94:
For “Transactions in Process” I think we want `GROUP BY ... fill(none)`

aludvik:
Updated to use fill(previous) per @dplumb94 comment

PR #1427, opened 2018-02-08
Apply Back-pressure to Gossip Batch Messages

peterschwarz:
The GossipMessages of type "Batch" have back pressure applied via the [token bucket](https://en.wikipedia.org/wiki/Token_bucket) method of network rate limiting.  In this implementation, a token pool maintains a set of token bearers, which can acquire and release tokens.  As tokens come in, a token is acquired for that batch.  When the batch is either marked as invalid, or committed, the token is released back into the pool.  The size of the pool is altered at the same time as the pending batch queue.

This includes a number of additional changes:

- Schedulers can return the batch id for a given transaction id
- Containing batch id is included when an invalid transaction notification occurs
- Validator constructor is wrapped in the CLI's try/except (this helps catch syntax errors that can easily occur in that large block of code)
- `GossipMessageDuplicateHandler` is renamed to `GossipMessageDropHandler`: it drops items based on multiple criteria


validator/sawtooth_validator/server/cli.py
```diff
@@ -349,27 +349,27 @@ def main(args=None):
...
-        validator_config.minimum_peer_connectivity,
-        validator_config.maximum_peer_connectivity,
-        validator_config.network_public_key,
-        validator_config.network_private_key,
-        roles=validator_config.roles,
-        metrics_registry=wrapped_registry)
-
     # pylint: disable=broad-except
     try:
+        validator = Validator(
```
vaporos: To move this within the block, you should set validator = None prior to the block, then make sure the finally clause checks "if validator is not None" before using it.  As it is, a well-timed KeyboardInterrupt may cause an exception in the finally block.

validator/sawtooth_validator/gossip/token_pool.py
```diff
@@ -0,0 +1,94 @@
...
+
+from threading import Lock
+
+
+DEFAULT_POOL_SIZE = 60
+"""The default number of tokens in a TokenPool.
+"""
+
+
+class NoTokenAvailabe(Exception):
```
vaporos: Should be "NoTokenAvailable".

validator/sawtooth_validator/gossip/gossip_handlers.py
```diff
@@ -151,6 +158,12 @@ def handle(self, connection_id, message_content):
             if has_batch:
                 return HandlerResult(HandlerStatus.DROP)
 
+            try:
+                self._batch_token_pool.acquire_token(batch.header_signature)
+            except NoTokenAvailabe:
+                # The available tokesn may have changed since our last check
```
vaporos: tokesn should be tokens

validator/sawtooth_validator/server/core.py
```diff
@@ -203,7 +207,7 @@ def __init__(self,
             context_manager=context_manager,
             settings_view_factory=SettingsViewFactory(state_view_factory),
             scheduler_type=scheduler_type,
-            invalid_observers=[batch_tracker],
+            invalid_observers=[batch_tracker, batch_token_pool_updater],
```
vaporos: Can we move this observer to the caller of the scheduler? For the purposes of the batch_token_pool_updater, it seems like this could actually be just the block publisher since it is related to accepting batches to insert into the pending queue?

validator/sawtooth_validator/execution/scheduler.py
```diff
@@ -187,6 +187,17 @@ def get_transaction(self, index):
         """
         raise NotImplementedError()
 
+    def get_batch_id_for_transaction(self, transaction_id):
```
vaporos: I think that we might want to make the caller keep a record of the batches/transactions it added to the scheduler. Eventually this API may be across a 0MQ socket.  I'll approve either way, but something to consider I think.

peterschwarz:
Closing this - requires some major rework

PR #1429, opened 2018-02-09
Export blocks as regular protobuf messages

aludvik:
Previously, this was including a length delimiter byte, which makes
reading exported blocks more difficult. The length delimiter is only
needed when exporting multiple blocks to one file, such as with the
backup command.

### DIFF
--- a/adm/src/commands/blockstore.rs
+++ b/adm/src/commands/blockstore.rs
@@ -223,15 +223,20 @@ fn run_export_command<'a>(args: &ArgMatches<'a>) -> Result<(), CliError> {
     let block = blockstore.get(block_id).map_err(|_|
         CliError::ArgumentError(format!("Block not found: {}", block_id)))?;
 
-    let packed = block.write_to_bytes().map_err(|err|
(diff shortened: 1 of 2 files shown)


adm/src/commands/blockstore.rs
```diff
@@ -223,10 +223,15 @@ fn run_export_command<'a>(args: &ArgMatches<'a>) -> Result<(), CliError> {
...
         CliError::ArgumentError(format!("Block not found: {}", block_id)))?;
 
+    let packed = block.write_to_bytes().map_err(|err|
+        CliError::EnvironmentError(format!("{}", err)))?;
+
     let stdout = io::stdout();
     let mut handle = stdout.lock();
 
-    backup_block(&block, &mut handle)
+    handle.write(&packed)
```
peterschwarz: You could probably use `block.write_to_writer(handle)` and avoid separate `write_to_bytes`

aludvik:
Added a commit to address @peterschwarz comment and implement exporting to a file

PR #1433, opened 2018-02-12
Fix STL-166 cxx intkey out of spec

dcmiddle:
Add range checks to set, inc, and dec


dcmiddle:
updated types from int to uint32_t

PR #1434, opened 2018-02-12
Fix off by one error in parallel scheduler

aludvik:

### DIFF
--- a/validator/tests/test_scheduler/test_schedulers_with_yaml.py
+++ b/validator/tests/test_scheduler/test_schedulers_with_yaml.py
@@ -285,6 +285,30 @@ class TestSchedulersWithYaml(unittest.TestCase):
             context_manager=context_manager,
             name='complex_dependency_failures.yaml')
 
+    def test_long_chain_of_inputs_outputs(self):
+        """Tests the schedulers against the
+        test_scheduler/data/long_chain_of_inputs_outputs.yaml file.
+
+        Tests that transactions are sorted correctly in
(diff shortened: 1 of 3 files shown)


aludvik:
This closes STL-1055

validator/tests/test_scheduler/test_schedulers_with_yaml.py
```diff
@@ -285,6 +285,23 @@ def test_complex_dependency_failures(self):
             context_manager=context_manager,
             name='complex_dependency_failures.yaml')
 
+    def test_single_txn_per_batch(self):
```
peterschwarz: How does this test prevent a regression for this issue?  It's not at all clear from the test case itself. 

Can you speak a bit more about these items?

validator/tests/test_scheduler/test_schedulers_with_yaml.py
```diff
@@ -285,6 +285,23 @@ def test_complex_dependency_failures(self):
...
+        """Tests the schedulers against the
+        test_scheduler/data/long_chain_of_inputs_outputs.yaml file.
```
peterschwarz: Missing this file.

validator/tests/test_scheduler/test_schedulers_with_yaml.py
aludvik: Yeah, sorry the explanation was in the file I forgot to commit...

aludvik:
This fix is currently running on PSIM-LR1

PR #1435, opened 2018-02-13
Split off predecessor tree's tree

nick-drozd:
Separate the predecessor tree's predecessor logic from its data structure and add a `Tree` class with the following methods: `update`, `prune`, and `walk`.

Note that the data structure implemented by `Tree` is not yet a radix tree.

Two of the predecessor test assertions have been dropped, as they test low-level implementation details rather than an interface.

### DIFF
--- a/validator/sawtooth_validator/execution/scheduler_parallel.py
+++ b/validator/sawtooth_validator/execution/scheduler_parallel.py
@@ -79,6 +79,19 @@ class Tree:
 
         return child, address[len(addr):]
 
+    def update(self, address, updater):
+        '''
+        Walk to ADDRESS, creating nodes if necessary, and set the data
+        there to UPDATER(data).
+
+        Arguments:
+            address (str): the address to be updated
+        '''
+
+        node = self._get(address, create=True)
+
+        node.data = updater(node.data)
+
     def prune(self, address):
         '''
         Remove all children (and descendants) below ADDRESS.
@@ -144,9 +157,6 @@ class Tree:
                 node.children[token] = child
                 node = child
 
-        if node.data is None:
-            node.data = Predecessors(readers=[], writer=None)
-
         return node
 
     def get(self, address):
@@ -173,13 +183,27 @@ class PredecessorTree:
         return repr(self._tree)
 
     def add_reader(self, address, reader):
-        node = self._tree._get(address, create=True)
-        node.data.readers.append(reader)
+        def updater(data):
+            if data is None:
+                return Predecessors(readers=[reader], writer=None)
+
+            data.readers.append(reader)
+
+            return data
+
+        self._tree.update(address, updater)
 
     def set_writer(self, address, writer):
-        node = self._tree._get(address, create=True)
-        node.data.readers = []
-        node.data.writer = writer
+        def updater(data):
+            if data is None:
+                return Predecessors(readers=[], writer=writer)
+
+            data.writer = writer
+            data.readers.clear()
+
+            return data
+
+        self._tree.update(address, updater)
 
         self._tree.prune(address)
 
--- a/validator/sawtooth_validator/execution/scheduler_parallel.py
+++ b/validator/sawtooth_validator/execution/scheduler_parallel.py
@@ -77,6 +77,18 @@ class PredecessorTree:
             for i in range(0, len(address), self._token_size)
         ]
 
+    def _get_child_and_remainder(self, node, address):
+        try:
+            child, addr = next(
+                (child, addr)
+                for addr, child in node.children.items()
+                if address.startswith(addr[:self._token_size])
+            )
+        except StopIteration:
+            raise AddressNotInTree()
+
+        return child, address[len(addr):]
+
     def prune(self, address):
         '''
         Remove all children (and descendants) below ADDRESS.
@@ -87,6 +99,34 @@ class PredecessorTree:
 
         node = self.get(address)
         node.children.clear()
+
+    def walk(self, address):
+        node = self._root
+
+        yield node, node.address
+
+        current_address = address
+
+        while current_address:
+            node, current_address = \
+                self._get_child_and_remainder(
+                    node, current_address)
+
+            yield node, node.address
+
+        to_process = deque()
+
+        to_process.extendleft(
+            node.children.values())
+
+        while to_process:
+            node = to_process.pop()
+
+            yield node, node.address
+
+            if node.children:
+                to_process.extendleft(
+                    node.children.values())
 
     def _get(self, address, create=False):
         tokens = self._tokenize_address(address)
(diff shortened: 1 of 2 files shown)


validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -70,24 +70,37 @@ def _tokenize_address(self, address):
...
         ]
 
-    def _get(self, address, create=False):
-        tokens = self._tokenize_address(address)
+    def update(self, address, updater):
+        '''
+        Sets the data at ADDRESS to UPDATER(address).
+
+        If ADDRESS is not in the tree, it is added (along with any
+        intermediate nodes).
```
peterschwarz: Add some doc string for the arguments (this does not follow the usual format)

validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -89,6 +98,44 @@ def _get(self, address, create=False):
...
+
+        node = self._root
+
+        yield node
+
+        for token in self._tokenize_address(address):
+            try:
+                node = node.children[token]
+            except KeyError:
+                raise AddressNotInTree()
```
aludvik: I think it is probably okay to just raise KeyError here if you call out in the docstring that it will raise that on missing address.

validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -89,6 +98,44 @@ def _get(self, address, create=False):
...
+
+            yield node
+
+        # indicates that all ancestors have been yielded
+        yield self.DUMMY
```
aludvik: This seems like more of a hack than a good design. We should discuss some alternatives.

validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -89,6 +98,44 @@ def _get(self, address, create=False):
...
+    def walk(self, address):
+        '''
+        Returns a stream of node data, raising AddressNotInTree if
+        ADDRESS is not in the tree.
+
+        First, the ancestors of ADDRESS (including itself) are
+        yielded, earliest to latest. Next, a dummy node is yielded to
+        indicate that all the ancestors have been returned. Finally,
+        the descendants of ADDRESS are yielded in an unspecified
+        order.
```
aludvik: Instead of unspecified, can we commit to depth- or breadth-first search? Is there a reason not to?

validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -137,51 +184,51 @@ def find_write_predecessors(self, address):
...
+            return predecessors
 
-            if node.writer is not None:
-                enclosing_writer = node.writer
-
-        if enclosing_writer is not None:
-            predecessors.add(enclosing_writer)
+        # Next, descend down the tree starting at the address node and
+        # find all children writers and readers. Uses breadth first
+        # search.
```
aludvik: Please update the comments to be consistent. You mention earlier that the order is unspecified, but this says it is breadth first.

validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -70,24 +70,37 @@ def _tokenize_address(self, address):
             for i in range(0, len(address), self._token_size)
         ]
 
-    def _get(self, address, create=False):
-        tokens = self._tokenize_address(address)
+    def update(self, address, updater):
+        '''
+        Sets the data at ADDRESS to UPDATER(address).
```
aludvik: Should this say something like `UPDATER(data_at_address)`?

validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -140,6 +140,14 @@ def walk(self, address):
             if node.children:
                 to_process.extendleft(node.children.values())
 
+    def prune(self, address):
+        '''
+        Removes all children (and descendants) at ADDRESS.
```
aludvik: Should "at" be "below"?

validator/sawtooth_validator/execution/scheduler_parallel.py
nick-drozd: yes

validator/sawtooth_validator/execution/scheduler_parallel.py
nick-drozd: Is there a specific use case for one or the other? If not, why constrain the implementation?

validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -77,6 +77,18 @@ def _tokenize_address(self, address):
             for i in range(0, len(address), self._token_size)
         ]
 
+    def _get_child_and_remainder(self, node, address):
+        try:
+            child, addr = next(
+                (child, addr)
+                for addr, child in node.children.items()
+                if address.startswith(addr[:self._token_size])
+            )
```
aludvik: Why is this implementation preferred to the following?
```
for addr, child in node.children.items() :
    if address.startswith(addr[:self._token_size])
        return child, addr
raise AddressNotInTree()
```

This implementation seems harder to follow and less standard.

validator/tests/test_scheduler/test_predecessor_tree.py
```diff
@@ -109,20 +109,6 @@ def test_predecessor_tree(self):
         #           i:
         #             c: Readers: [5]
 
-        self.assert_rwc_at_addresses({
```
aludvik: These tests should be rewritten to work with the new API so that we do not lose coverage instead of being deleted.

validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -70,34 +67,147 @@ def _tokenize_address(self, address):
...
+            raise AddressNotInTree()
+
+        return child
+
+    def _walk_to_address(self, address):
+        node = self._root
+
+        yield node
+
+        while node.address < address:
```
boydjohnson: Could you talk about the importance of the lexicographical ordering here? For example 'aaaa' < 'aaab' but 'aaac > 'aaab'.

validator/tests/test_scheduler/test_predecessor_tree.py
nick-drozd: Those assertions check specific implementation details of the tree data structure, and so would have to be rewritten for any change to the tree, including this PR.

validator/sawtooth_validator/execution/scheduler_parallel.py
nick-drozd: A node's address is always a strict prefix of all the addresses of its children. An equivalent test would be:
```
        while address != node.address and address.startswith(node.address):
            node = self._get_child(node, address)
```

validator/tests/test_scheduler/test_predecessor_tree.py
peterschwarz: Removing these tests are fine: We shouldn't be using unit tests to verify internal state.

nick-drozd: I added commit message to reflect that change.

validator/sawtooth_validator/execution/scheduler_parallel.py
aludvik: I think he meant put it in a comment

nick-drozd: comment added

PR #1437, opened 2018-02-14
Recompute state on startup if missing

aludvik:
This PR adds a feature to the validator where it checks if any state is missing on startup and, if it is, recomputes it before starting normal operation. This can be used in conjuction with the `sawadm blockstore restore` command to restore a validator from a backup.

### DIFF
--- a/validator/sawtooth_validator/server/state_verifier.py
+++ b/validator/sawtooth_validator/server/state_verifier.py
@@ -1,0 +1,294 @@
+# Copyright 2018 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+
+import logging
+import os
+
+from sawtooth_validator.protobuf import validator_pb2
+from sawtooth_validator.state.merkle import INIT_ROOT_KEY
+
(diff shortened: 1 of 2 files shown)


validator/sawtooth_validator/server/state_verifier.py
```diff
@@ -0,0 +1,287 @@
...
+
+    # If there is a head, do a binary search to find the last block with state
+    max_num = block.block_num
+    min_num = 0
+    next_num = (max_num - min_num) // 2
+
+    while max_num != min_num:
+        block = blockstore.get_block_by_number(next_num)
+
+        # If this block's state root is in state, assume all the lower blocks
```
boydjohnson: Is this a safe assumption? Have you timed sequential lookup of each block's state root hash from the state database?

validator/sawtooth_validator/server/state_verifier.py
```diff
@@ -0,0 +1,287 @@
...
+
+    component_dispatcher.start()
+    component_service.start()
+
+    process_blocks(
+        blocks=blockstore.get_block_iter(start_block=start, reverse=False),
+        transaction_executor=transaction_executor,
+        context_manager=context_manager,
+        state_view_factory=state_view_factory)
+
```
boydjohnson: the ContextManager and the Executor are both active components that need to be shut down. I think they both have a stop method.

aludvik: fixed

validator/sawtooth_validator/server/state_verifier.py
```diff
@@ -0,0 +1,293 @@
...
+    prev_state_root = INIT_ROOT_KEY
+    for block in blockstore.get_block_iter(reverse=False):
+        if not state_db_has_root(state_view_factory, block.state_root_hash):
+            return block, prev_state_root
+        prev_state_root = block.state_root_hash
+
+    # This should never happen, since we already checked that the chain head
+    # didn't have a state root
+    return None, None
```
peterschwarz: "This should never happen" should probably result in an error of some kind.

aludvik:
Raise exception per @peterschwarz request

PR #1440, opened 2018-02-15
Create block-info docker

rberg2:


docker/sawtooth-int-block-info-tp
```diff
@@ -21,7 +21,7 @@
 #   This image should be built using `build_all installed`.
 #
 # Run:
-#   $ docker run sawtooth-block-info-tp
+#   $ docker run sawtooth-int-block-info-tp
```
aludvik: This is not a typo, the Dockerfile is `sawtooth-int-block-info-tp` but the image created by build_all is called `sawtooth-block-info-tp`.

rberg2:
@aludvik I have dropped the commit to docker/sawtooth-int-block-info-tp

ci/sawtooth-validator
```diff
@@ -33,6 +33,7 @@ RUN echo "deb http://repo.sawtooth.me/ubuntu/1.0/stable xenial universe" >> /etc
  && apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 8AA7AF1F1091A5FD \
  && apt-get update \
  && apt-get install -y -q \
+    python3-sawtooth-block-info \
```
aludvik: This should be in a separate commit since it isn't related to the original commit message.

rberg2: I moved this to a separate commit as requested.

PR #1441, opened 2018-02-15
Create block-info docker - Backport

rberg2:
This is merged to master in PR #1440 . This was not tested in LR because we don't run docker containers there. This was tested locally.


vaporos:
Retro-approved for backport.

PR #1442, opened 2018-02-15
Add separate threadpool for client handlers

peterschwarz:
Add a separate threadpool for handling client messages.  This removes
contention between components like transaction processors being starved
for thread attention by high-volume client requests.


benoitrazet:
I tested the branch against the scenario described in STL-1064 and it looks like this PR is fixing the bug described in the ticket. I even tried to increase the load of testing to see where would be the next breaking point and it appears that the validator crashed with an error code 137, which can very well be docker killing the containers because of heavy load.

peterschwarz:
Added code to shutdown the threadpool.  Please re-review.

PR #1444, opened 2018-02-16
Reorganize validator.toml.example so it is valid toml

agunde406:
The subcategory for roles and permissions should come last
in the files. I also commented those categories out as they
are less likely to be used when starting the validator for
the first time.

@dcmiddle This should fix the issues people have been seeing when trying to start the validator up with the full example file.


dcmiddle:
@peterschwarz can you add this to 1.0.2 backports?

PR #1445, opened 2018-02-16
Fix issue Unscheduled InternalError Results

peterschwarz:
Transaction that return INTERNAL_ERROR on their result will immediately be resent to a transaction processor.  This can trigger a death spiral of transaction executions, over and over again.  We can mitigate this problem by not resending the transaction once its containing batch has been unscheduled.

This requires the addition of a `is_transaction_in_schedule` method to the Scheduler interface.


benoitrazet:
I tested the branch against the scenario described in https://jira.hyperledger.org/browse/STL-1063 and the bug do not show up again. It's all good.

PR #1449, opened 2018-02-19
Javascript XO Refactor

boydjohnson:


sdk/examples/xo_javascript/xo_handler.js
```diff
@@ -167,182 +106,142 @@ const _isWin = (board, letter) => {
...
-          handlerFn = _handleCreate
-        } else if (update.action === 'take') {
-          handlerFn = _handleTake
-        } else if (update.action === 'delete') {
-          handlerFn = _handleDelete
+        if (payload.action === 'create') {
+          return xo_state.getGame(payload.name)
+            .then((game) => {
+
+              let name = payload.name
```
peterschwarz: This is unused.

sdk/examples/xo_javascript/xo_handler.js
```diff
@@ -167,182 +106,142 @@ const _isWin = (board, letter) => {
...
+
+              _display(`Player ${player.toString().substring(0, 6)} created game ${payload.name}`)
+
+              return xo_state.setGame(payload.name, createdGame)
+            })
+
+        } else if (payload.action === 'take') {
+          return xo_state.getGame(payload.name)
+            .then((game) => {
+              let name = payload.name
```
peterschwarz: This is also not used.

sdk/examples/xo_javascript/xo_payload.js
```diff
@@ -0,0 +1,45 @@
...
+ * ------------------------------------------------------------------------------
+ */
+
+'use strict'
+
+const { InvalidTransaction } = require('sawtooth-sdk/processor/exceptions')
+
+class XoPayload {
+
+    fromBytes(payload) {
```
peterschwarz: This looks like it should be a static function that returns the promise, and the promise returns an XoPayload with those fields.

nick-drozd:
I suggest running all this through Standard JS.

From the `xo_javascript` directory:

```
$ standard --fix
standard: Use JavaScript Standard Style (https://standardjs.com)
  /Users/drozd/project/sawtooth-core/sdk/examples/xo_javascript/xo_handler.js:20:7: 'crypto' is assigned a value but never used.
  /Users/drozd/project/sawtooth-core/sdk/examples/xo_javascript/xo_handler.js:117:9: Identifier 'xo_state' is not in camel case.
  /Users/drozd/project/sawtooth-core/sdk/examples/xo_javascript/xo_handler.js:135:20: Expected '!==' and instead saw '!='.
```

Those three errors have to be fixed manually, but the `--fix` flag fixes all boring whitespace errors in-place.

Is Standard JS actually a widely followed standard? I don't know, but the rest of the JS SDK conforms to it, and Intkey JS too I think. Maybe it should be added as another pass during the linting stage of the build. It's not configurable, but [that's touted as a feature](https://standardjs.com/#i-disagree-with-rule-x-can-you-change-it):

> I disagree with rule X, can you change it?
>
> No. The whole point of standard is to save you time by avoiding bikeshedding about code style. There are lots of debates online about tabs vs. spaces, etc. that will never be resolved. These debates just distract from getting stuff done. At the end of the day you have to 'just pick something', and that's the whole philosophy of standard -- its a bunch of sensible 'just pick something' opinions. Hopefully, users see the value in that over defending their own opinions.

boydjohnson:
Due to a rebase the commits show up on Github as out of order.

sdk/javascript/processor/index.js
```diff
@@ -103,55 +103,60 @@ class TransactionProcessor {
...
-                    TpProcessResponse.encode(response).finish()
-                  )
-                }
+            let applyPromise
+            try {
+              applyPromise = Promise.resolve(handler.apply(request, context))
+            } catch(err) {
+              applyPromise = Promise.reject(err)
+            }
+            applyPromise
```
peterschwarz: If this is undefined, it will thrown an error.

peterschwarz: Nevermind, I think I misinterpreted the try block's content

PR #1450, opened 2018-02-19
Add dispatcher queue size metric

dplumb94:
This PR replaces the dispatcher queue size logging message with a gauge metric, and adds a dashboard panel to track it.


validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -162,6 +163,12 @@ def __init__(self, connection, address, futures, connections,
         self._received_message_counters = {}
         self._dispatcher_queue = None
 
+        if metrics_registry:
+            self._queue_size_gauge = GaugeWrapper(
+                metrics_registry.gauge('interconnect.dispatcher_queue_size'))
```
aludvik: Since there are multiple instances of interconnect, we need some way to differentiate those queue sizes, otherwise this graph will be misleading.

dplumb94: fix'd

PR #1453, opened 2018-02-19
Compute pending batch queue limit by block bounds

peterschwarz:
Compute the pending batch queue limit by adding the number of batches in
a chain head (i.e. a block) to the rolling average, instead of using the
number of committed batches, which may be across multiple blocks (in the
case of switching forks of length > 1).  For a network that is able to
process transactions at a high rate, this will max out as a function of
max batches per block.


aludvik:
@peterschwarz create a 1-0 backport PR after merging

PR #1455, opened 2018-02-19
Update Metrics API

aludvik:
Updates the metrics API to be similar to logging and then updates the usage in the validator. Also adds configurable reporting levels, with the intention of extending configuration to include per-module and per-class reporting levels.


nick-drozd:
The first commit could use a more informative commit message -- "Improve usability" is vague and subjective. Maybe "Change metrics API to match logging" or something like that.

validator/sawtooth_validator/metrics/USAGE.md
```diff
@@ -0,0 +1,63 @@
+# Metric types
```
vaporos: This should really be in the API docs (you should be able to do module-level).

aludvik:
Moved usage to docstring and fixed dashboard

validator/sawtooth_validator/metrics/metrics.py
```diff
@@ -0,0 +1,196 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+
+import os
+import platform
+
+
+class MetricsConfig:
+    DEBUG = 10
```
peterschwarz: Are these reporting levels?  Should they be "constant" instances of `MetricsConfig`?  e.g. `DEBUG = MetricsConfig(10)`

validator/sawtooth_validator/metrics/metrics.py
```diff
@@ -0,0 +1,196 @@
...
+    INFO = 20
+    PERF = 30
+
+    def __init__(self, reporting_level):
```
peterschwarz: Alternatively, I don't see this being used as a class, so it could just be an `IntEnum`

validator/sawtooth_validator/metrics/metrics.py
```diff
@@ -0,0 +1,196 @@
...
+            instance=instance,
+            tags=tags)
+
+    def _create_identifier(self, metric_name, instance=None):
+        if instance is None:
+            return (self._module_name, metric_name)
+        return (self._module_name, instance.__class__.__name__, metric_name)
+
+
+class MockMetricsRegistry:
```
peterschwarz: I'm not sure I'd call this "Mock", as that has test connotations.  I would consider naming these as "NoOp"

validator/sawtooth_validator/metrics/metrics.py
aludvik: This config will be extended in a future PR to include per-module configuration, so it doesn't make sense to have constant instances. You also need to be able to specify a reporting level when you create a metric.

validator/sawtooth_validator/metrics/metrics.py
aludvik: As I mentioned above, this will be extended to include more information about which modules and classes are enabled/disabed, not just the reporting level.

validator/sawtooth_validator/metrics/metrics.py
aludvik: NoOp seems fine. Originally I had Fake which was even worse.

validator/sawtooth_validator/metrics/metrics.py
peterschwarz: That makes this class premature then. It should be part of that future PR.

aludvik:
@peterschwarz updated based on feedback

PR #1456, opened 2018-02-19
Implement parallelism hint and handling

dplumb94:
Adds a maxOccupancy field to the TpRegisterRequest proto. This will act as a limit to the number of in-flight transactions a validator can have with a given TP

Tracks the number of current in-flight transactions a validator has with a TP

Removes the _execute_or_wait_for_processor_type() method and the waiting threadpool from the executor, and shifts all wait functionality to the ProcessorManager (previously ProcessorIteratorCollection). Execution will now happen as follows:

- The executor will attempt to grab a processor from the processor iterator collection.
- The processor iterator collection will check to make sure that aprocessor of the requested type is registered. If not, the thread will wait until a suitable processor registers.
- The processor iterator collection will attempt to get the next processor out of the processor iterator of the requested type.
- The processor iterator will cycle through its processors to find a processor that has occupancy to handle another transaction. If all processors are at max occupancy it will throw an exception.
- If the processor iterator encounters an exception, it will wait for a processor to become available and return the processor. If not, it  will just return the processor.

### DIFF
--- a/docker/grafana/dashboards/dashboard.json
+++ b/docker/grafana/dashboards/dashboard.json
@@ -1583,6 +1583,390 @@
           "dashes": false,
           "datasource": "$DATASOURCE",
           "fill": 0,
+          "id": 43,
+          "legend": {
+            "avg": false,
+            "current": false,
+            "max": false,
+            "min": false,
+            "show": false,
+            "total": false,
+            "values": false
+          },
+          "lines": true,
+          "linewidth": 1,
+          "links": [],
+          "nullPointMode": "null",
+          "percentage": false,
+          "pointradius": 5,
+          "points": false,
+          "renderer": "flot",
+          "seriesOverrides": [],
+          "spaceLength": 10,
+          "span": 4,
+          "stack": false,
+          "steppedLine": false,
+          "targets": [
+            {
+              "alias": "[[tag_host]]",
+              "dsType": "influxdb",
+              "groupBy": [
+                {
+                  "params": [
+                    "$__interval"
+                  ],
+                  "type": "time"
+                },
+                {
+                  "params": [
+                    "host"
+                  ],
+                  "type": "tag"
+                },
+                {
+                  "params": [
+                    "null"
+                  ],
+                  "type": "fill"
+                }
+              ],
+              "measurement": "sawtooth_validator.executor.TransactionExecutorThread.tp_process_response_count",
+              "orderByTime": "ASC",
+              "policy": "default",
+              "refId": "A",
+              "resultFormat": "time_series",
+              "select": [
+                [
+                  {
+                    "params": [
+                      "count"
+                    ],
+                    "type": "field"
+                  },
+                  {
+                    "params": [],
+                    "type": "mean"
+                  },
+                  {
+                    "params": [
+                      "1s"
+                    ],
+                    "type": "non_negative_derivative"
+                  }
+                ]
+              ],
+              "tags": [
+                {
+                  "key": "response_type",
+                  "operator": "=",
+                  "value": "OK"
+                }
+              ]
+            }
+          ],
+          "thresholds": [],
+          "timeFrom": null,
+          "timeShift": null,
+          "title": "Valid Transaction Response Rate",
+          "tooltip": {
+            "shared": true,
+            "sort": 0,
+            "value_type": "individual"
+          },
+          "type": "graph",
+          "xaxis": {
+            "buckets": null,
+            "mode": "time",
+            "name": null,
+            "show": true,
+            "values": []
+          },
+          "yaxes": [
+            {
+              "format": "short",
+              "label": null,
+              "logBase": 1,
+              "max": null,
+              "min": null,
+              "show": true
+            },
+            {
+              "format": "short",
+              "label": null,
+              "logBase": 1,
+              "max": null,
+              "min": null,
+              "show": true
+            }
+          ]
+        },
+        {
+          "aliasColors": {},
+          "bars": false,
+          "dashLength": 10,
+          "dashes": false,
+          "datasource": "$DATASOURCE",
+          "fill": 0,
+          "id": 44,
+          "legend": {
+            "avg": false,
+            "current": false,
+            "max": false,
+            "min": false,
+            "show": false,
+            "total": false,
+            "values": false
+          },
+          "lines": true,
+          "linewidth": 1,
+          "links": [],
+          "nullPointMode": "null",
+          "percentage": false,
+          "pointradius": 5,
+          "points": false,
+          "renderer": "flot",
+          "seriesOverrides": [],
+          "spaceLength": 10,
+          "span": 4,
+          "stack": false,
+          "steppedLine": false,
+          "targets": [
+            {
+              "alias": "[[tag_host]]",
+              "dsType": "influxdb",
+              "groupBy": [
+                {
+                  "params": [
+                    "$__interval"
+                  ],
+                  "type": "time"
+                },
+                {
+                  "params": [
+                    "host"
+                  ],
+                  "type": "tag"
+                },
+                {
+                  "params": [
+                    "null"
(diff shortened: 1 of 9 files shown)


validator/sawtooth_validator/exceptions.py
```diff
@@ -72,3 +72,7 @@ class PossibleForkDetectedError(Exception):
     through the block store.
     """
     pass
+
+
+class ProcessorQueuesFullError(Exception):
```
aludvik: Maybe rename this `NoProcessorVacancyError` to keep with the language used in the code.

validator/sawtooth_validator/execution/processor_iterator.py
```diff
@@ -63,7 +67,13 @@ def get_next_of_type(self, processor_type):
...
         with self._condition:
             if processor_type in self:
-                return self[processor_type].next_processor()
+                for _ in range(len(self[processor_type])):
+                    processor = self[processor_type].next_processor()
+                    occupancy = self.occupancy_map.get(
+                        processor.connection_id)
+                    if occupancy['maximum'] > occupancy['current']:
+                        return processor
+                raise ProcessorQueuesFullError()
```
aludvik: This function has kind of a funky interface. It returns None if there are no processors of the given type registered but raises an exception if they are all full. Then when it is used, you aren't checking for None and are checking whether the processor is registered manually (which theoretically could change in between the line you check and the line you call this function). Instead, I would create a new error type for no processor registered and catch that where you call this function.

aludvik: This is where I am talking about adding another catch https://github.com/hyperledger/sawtooth-core/pull/1456/commits/5dbe78a9c39e69639685a4bf00b315f0a5f96b57#diff-6e2f0b2b49825b88875126268a6cff5fR340

validator/sawtooth_validator/execution/executor.py
```diff
@@ -304,37 +318,71 @@ def _execute_schedule(self):
...
+        except ProcessorQueuesFullError:
+            self._wait_for_processor_vacancy(
+                processor_type=processor_type,
+                content=content,
+                signature=signature)
+
+    def _wait_for_processor_vacancy(self, processor_type, content, signature):
+        LOGGER.debug("all transaction processors of type "
+                     "%s are at max occupancy", processor_type)
+        if processor_type not in self._waiters_by_type_vacancy:
```
aludvik: I think this is a data race, since if there are multiple threads executing it multiple threads could see there is no waiter and create them and then overwrite the others when they are set in the `_waiters_by_type_vacancy`. It looks like this was an existing data race.

aludvik: Since this was an existing data race, I will accept this if you open a bug report on JIRA.

validator/sawtooth_validator/execution/processor_iterator.py
```diff
@@ -259,3 +259,40 @@ def remove_processor(self, processor_identity):
...
+        self._dict = dict()
+        self._lock = RLock()
+
+    def get(self, key, default=None):
+        with self._lock:
+            return self._dict.get(key)
+
+    def add_if_unique(self, key, maximum):
+        with self._lock:
+            if not self._dict.get(key):
```
boydjohnson: I would think that `if key not in self._dict` would be more meaningful and possibly more performant.

validator/sawtooth_validator/execution/processor_iterator.py
```diff
@@ -259,3 +259,40 @@ def remove_processor(self, processor_identity):
...
+                self._dict[key] = {
+                    'maximum': maximum,
+                    'current': 0
+                }
+
+    def inc(self, key):
+        """Increments the current counter for a given key if it is below the
+        maximum count"""
+        with self._lock:
+            if self._dict[key]['maximum'] > self._dict[key]['current']:
```
boydjohnson: This is a small preference: This statement reads as, generally, "if the maximum is greater than current, do", when the more meaningful way to check would be "If current is less than maximum, do". This is just a small stylistic preference.

validator/sawtooth_validator/execution/processor_iterator.py
```diff
@@ -144,6 +156,34 @@ def cancellable_wait(self, processor_type, cancelled_event):
...
+            None
+        """
+
+        with self._condition:
+            self._condition.wait_for(lambda: (
+                self.processor_available(processor_type)
+                or cancelled_event.is_set()))
+
+    def processor_available(self, processor_type):
+        with self._condition:
```
aludvik: I do not think you need to reacquire this condition variable here. I would make this a protected function (start it with an underscore) to show that it should only be used internally.

validator/sawtooth_validator/execution/processor_iterator.py
```diff
@@ -259,3 +259,40 @@ def remove_processor(self, processor_identity):
...
+        with self._lock:
+            if key not in self._dict:
+                self._dict[key] = {
+                    'maximum': maximum,
+                    'current': 0
+                }
+
+    def inc(self, key):
+        """Increments the current counter for a given key if it is below the
+        maximum count"""
```
aludvik: Add a note about how this returns True/False

validator/sawtooth_validator/execution/processor_handlers.py
```diff
@@ -33,22 +35,25 @@ def __init__(self, processor_collection):
     def handle(self, connection_id, message_content):
         request = processor_pb2.TpRegisterRequest()
         request.ParseFromString(message_content)
+        max_occupancy = request.max_occupancy or DEFAULT_MAX_OCCUPANCY
```
peterschwarz: Suggest using an explicit check for `== 0` here.  It will be more clear about types

peterschwarz:
We should consider refactoring some of this. There's a great deal of intertwined responsibility between the Executor and the ProcessorIteratorCollection with respect to processors and their availability.  There should be an appropriate separation of concerns here.

validator/sawtooth_validator/execution/executor.py
```diff
@@ -299,37 +294,24 @@ def _execute_schedule(self):
...
-            connection_id = processor.connection_id
-            self._send_and_process_result(content, connection_id, signature)
+    def _execute(self, processor_type, content, signature):
+        try:
+            processor = self._processors.get_next_of_type(
+                processor_type=processor_type)
+        except WaitCancelledException:
+            LOGGER.exception("Transaction %s cancelled while "
+                             "waiting for available processor",
+                             content.signature)
```
aludvik: This should probably return so that we don't try to execute `_send_and_process_result` without a processor

validator/sawtooth_validator/execution/executor.py
```diff
@@ -314,12 +318,12 @@ def _execute(self, processor_type, content, signature):
             content, processor.connection_id, signature)
 
     def _send_and_process_result(self, content, connection_id, signature):
+        self._in_process_transactions_count.inc()
```
aludvik: Is this change functional?

docker/grafana/dashboards/dashboard.json
```diff
@@ -3686,10 +4444,11 @@
                   "type": "fill"
                 }
               ],
+              "hide": false,
               "measurement": "sawtooth_validator.threadpool.InstrumentedThreadPoolExecutor.task_run_time",
               "orderByTime": "ASC",
               "policy": "default",
-              "query": "",
+              "query": "SELECT last(\"99_percentile\") FROM \"sawtooth_validator.Component-threadpool.task_run_time\" WHERE $timeFilter GROUP BY time($__interval), \"host\" fill(none)",
```
aludvik: This is reverting a previous change to the dashboard, where the names were updated and the query field removed.

aludvik: There are also a few more below

PR #1457, opened 2018-02-20
Convert predecessor tree's trie to a prefix tree

nick-drozd:
This PR changes the data structure of the parallel scheduler's predecessor tree from a trie to a prefix tree (there's a lot of confusing and overlapping terminology for this family of data structures; these are the names I've settled on). For a nice comparison of these data structures, see https://medium.com/basecs/compressing-radix-trees-without-too-many-tears-a2e658adb9a0. The tl;dr is that the new one should use a lot less memory than the old one. How much less, you ask? I ran the dynamic network test before and after this change with the following logging statement added:

```
class Node:
    def __init__(self, address, data=None):
        LOGGER.error('NEW NODE')  # <--
        self.address = address
        self.children = set()
        self.data = data
```

Then I counted all the occurrences of that statement. Before this change, the message logged 24140 times. With this change, the message was logged 959 times.

### DIFF
--- a/validator/sawtooth_validator/execution/scheduler_parallel.py
+++ b/validator/sawtooth_validator/execution/scheduler_parallel.py
@@ -36,43 +36,39 @@ _AnnotatedBatch = namedtuple('ScheduledBatch',
 
 
 class AddressNotInTree(Exception):
-    pass
+    def __init__(self, match=None):
+        super().__init__()
+        self.match = match
 
 
 class Node:
     def __init__(self, address, data=None):
         self.address = address
-        self.children = {}
+        self.children = set()
         self.data = data
-
-    def __repr__(self):
-        return repr({
-            'address': self.address,
-            'children': self.children,
-            'data': self.data,
-        })
 
 
 class Tree:
-    def __init__(self, token_size):
+    '''
+    This tree is a prefix tree: a node's address is always a strict
+    prefix of the addresses of its children, and every node either has
+    data or has multiple children.
+    '''
+    def __init__(self):
         self._root = Node('')
-        self._token_size = token_size
-
-    def __repr__(self):
-        return repr(self._root)
-
-    def _tokenize_address(self, address):
-        return [
-            address[i:i + self._token_size]
-            for i in range(0, len(address), self._token_size)
-        ]
 
     def _get_child(self, node, address):
-        for child in node.children.values():
+        for child in node.children:
             if address.startswith(child.address):
                 return child
 
-        raise AddressNotInTree()
+        match = None
+
+        for child in node.children:
+            if child.address.startswith(address):
+                match = child.address
+
+        raise AddressNotInTree(match=match)
 
     def _walk_to_address(self, address):
         node = self._root
@@ -100,7 +96,7 @@ class Tree:
             address (str): the address to be updated
         '''
 
-        node = self._get(address)
+        node = self._get_or_create(address)
 
         node.data = updater(node.data)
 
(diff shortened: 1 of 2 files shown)


validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -36,43 +36,44 @@
...
+        self.children = set()
         self.data = data
 
     def __repr__(self):
-        return repr({
-            'address': self.address,
-            'children': self.children,
-            'data': self.data,
-        })
+        return repr((
```
peterschwarz: Since you're modifying this content, perhaps this should be changed to actually be the correct usage of `__repr__` vs what this looks like, which more of a debugging string.

See: https://docs.python.org/3.5/library/functions.html#repr

validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -149,21 +150,83 @@ def walk(self, address):
 
             if node.children:
                 to_process.extendleft(
-                    node.children.values())
+                    node.children)
 
     def _get(self, address):
```
peterschwarz: This function should probably be named `_get_or_create`, since it does quite a bit more than just getting a node.

validator/sawtooth_validator/execution/scheduler_parallel.py
aludvik: Why do we even both implementing this? I have always considered __repr__ and __str__ to just be for debugging purposes. If you don't implement __repr__ or __str__, a default implementation is provided which "is a string enclosed in angle brackets that contains the name of the type of the object ..." Do you mean this should be reimplemented so that it is "an attempt to return a string that would yield an object with the same value when passed to eval()"? That seems kind of absurd since you should never be passing this to eval() unless you are copy-pasting some logging message directly into an interactive python interpreter.

validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -36,43 +36,44 @@
...
+        return repr((
+            self.address,
+            self.data,
+            self.children,
+        ))
 
 
 class Tree:
-    def __init__(self, token_size):
+    def __init__(self):
```
aludvik: Great job writing a very descriptive commit message. It would probably be a good idea to include some of that description about how addresses are represented in this Tree in a docstring under the `class Tree:` line for future reference.

validator/sawtooth_validator/execution/scheduler_parallel.py
nick-drozd: I don't think `__repr__` here was ever anything other than a debug string.

validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -149,21 +155,83 @@ def walk(self, address):
...
+
+            node.children.add(new_node)
 
-        return node
+            return new_node
+
+        # If node address is 'rustic' and the address being added is
+        # 'rust', then 'rust' will be the intermediate node taking
+        # 'rustic' as a child.
+        if match.address.startswith(address):
```
vaporos: The double spacing here should be single spaced (below too where it occurs).

validator/sawtooth_validator/execution/scheduler_parallel.py
peterschwarz: Change it to `__str__`, then

PR #1461, opened 2018-02-21
Adds missing handlers for GetBlock client messages

grkvlt:
The messages `CLIENT_BLOCK_GET_BY_TRANSACTION_ID_REQUEST` and `CLIENT_BLOCK_GET_BY_BATCH_ID` are now properly handled by the validator code.

### DIFF
--- a/validator/sawtooth_validator/server/component_handlers.py
+++ b/validator/sawtooth_validator/server/component_handlers.py
@@ -178,6 +178,16 @@ def add(
         client_handlers.BlockGetByNumRequest(block_store),
         client_thread_pool)
 
+    dispatcher.add_handler(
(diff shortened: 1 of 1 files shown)


validator/sawtooth_validator/server/component_handlers.py
```diff
@@ -178,6 +178,11 @@ def add(
         client_handlers.BlockGetByNumRequest(block_store),
         client_thread_pool)
 
+    dispatcher.add_handler(
+        validator_pb2.Message.CLIENT_BLOCK_GET_BY_TRANSACTION_REQUEST,
+        client_handlers.BlockGetByTransactionRequest(block_store),
+        thread_pool)
```
peterschwarz: This should use the `client_thread_pool`

grkvlt: Changed, thx

PR #1452, opened 2018-02-19
Add yaml test generator (fuzzer)

boydjohnson:
This PR adds a tool, `bin/gen-scheduler-test`, that can generate a scheduler yaml test according to specifications on the number of addresses used, inputs, outputs, portion of transactions that are invalid, whether inputs and outputs are wildcarded and by how much, etc.

PR #1468, opened 2018-02-22
Parallel scheduler fixes

boydjohnson:
This PR is based on #1452. The last 4 commits are of this PR.

I can (and probably should) remove some of the yaml tests. I just wanted to demonstrate the extent of the correctness provided by these fixes.

There are 2 bug fixes and then an improvement. The 2 bug fixes are around checking if a transaction is a predecessor of a predecessor (and so on), where the previous check had just been if the transaction was a predecessor of a predecessor. The improvement sorts the predecessors and stores them in a list when they are first calculated instead of when they are accessed.


boydjohnson:
I rebased on master after #1452 was merged. I am still working on organizing the test files with the commits and trying to come up with a means of reducing the overall lines committed.

nick-drozd:
Two suggestions:

1. The sorting commit is an improvement rather than a bug fix, so it
   should probably go in a separate PR so it can be discussed,
   profiled, etc. without holding up the bug fixes.
   
2. The data structure commit looks like a refactor, but it also
   includes a change in behavior. It would be easier to see what's
   going on if the refactor were done before the behavior change.

nick-drozd:
Closing in favor of https://github.com/hyperledger/sawtooth-core/pull/1499

PR #1472, opened 2018-02-23
Update to enable rust build behind proxy

askmish:
build_rust fails if https_proxy contains https.
Downgrade https to use http

Signed-off-by: ashish kumar mishra <ashish.k.mishra@intel.com>


askmish:
More appropriate as a documentation than script change. Closing this PR.

PR #1478, opened 2018-02-27
Updated committed transactions metric

aludvik:
This adds a gauge metric for committed transactions which shows the actual number of transactions currently in the block store, which can go up and down over time as forks are resolved.

### DIFF
--- a/validator/sawtooth_validator/database/indexed_database.py
+++ b/validator/sawtooth_validator/database/indexed_database.py
@@ -109,6 +109,16 @@ class IndexedDatabase(database.Database):
         with self._lmdb.begin(db=self._main_db) as txn:
             return txn.stat()['entries']
 
+    def count(self, index=None):
+        if index is None:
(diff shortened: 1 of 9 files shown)


validator/sawtooth_validator/database/indexed_database.py
```diff
@@ -109,6 +109,16 @@ def __len__(self):
         with self._lmdb.begin(db=self._main_db) as txn:
             return txn.stat()['entries']
 
+    def len(self, index=None):
```
peterschwarz: I don't think you should shadow the built-in function here.  I would suggest naming this either `size` or `length` and then calling it with the default args in the `__len__` implementation.

aludvik:
Re-requested reviews after some substantial additions:
- renamed len() to count()
- added count() to database abc and other implementations
- added unit tests for new methods
- updated dashboard

PR #1483, opened 2018-03-01
Update package versioning to include commit hash

rbuysse:


aludvik:
We should test that upgrading still works with this

PR #1484, opened 2018-03-01
Ensure that possible successor in schedule

peterschwarz:
When determining if a transaction should be replayed, ensure that any possible successors have not been unscheduled.  If they have, a `KeyError` could occur.


peterschwarz:
Also fixes [STL-1071](https://jira.hyperledger.org/browse/STL-1071)

PR #1485, opened 2018-03-01
Fix empty lines in literalinclude blocks (remove :linenos:)

chenette:
There's a line-height bug when line numbers are displayed with :linenos:

### DIFF
--- a/docs/source/architecture/transactions_and_batches.rst
+++ b/docs/source/architecture/transactions_and_batches.rst
@@ -27,7 +27,6 @@ message types:
 .. literalinclude:: ../../../protos/transaction.proto
    :language: protobuf
    :caption: File: protos/transaction.proto
-   :linenos:
 
 Header, Signature, and Public Keys
 ----------------------------------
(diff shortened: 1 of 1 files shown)


chenette:
@vaporos - Are you saying that we should ADD line numbers to all the code examples?  

Currently, only these two items in "Transactions and Batches" have line numbers.  Dozens of other code examples don't have line numbers (at a rough estimate, 113 code blocks).

chenette:
Basically, my question is "Why do two protobuf code chunks use line numbers and the others don't"?  But that's a different issue than fixing the line-height mismatch when :linenos: is used. I agree that we should know about the correct fix in case we should want to include line numbers on a code block.

Likewise, line numbers on --help usage chunks is also a separate issue.  Should I remove the CLI and SysAdmin changes from this PR and put them in a separate PR?

chenette:
I moved the --help text  changes to their own PR, because they're a separate issue.  See PR #1491.

chenette:
Closing this PR. I don't know enough about sphinx themes and Python to apply the real fix (see my info in STL-103).

PR #1486, opened 2018-03-02
Sawtooth SDK for .NET Core

tmarkovski:
This is a full implementation of the SDK for .NET Core 2.0.
The SDK provides complete Signer, Processor and State Delta support.
An implementation for IntKey example is also provided as Processor as well a Client tool that can be run to make sample transaction calls.
I've added docker images and build scripts accordingly following the standard syntax, taking example from the existing scripts and languages.
Docker image uses microsoft/dotnet image.
I will be submitting README.md's for the SDK as well as the IntKey processor and cli tool to complement the SDK.


tmarkovski:
Closing to resubmit with DCO

PR #1487, opened 2018-03-02
Sawtooth SDK for .NET Core

tmarkovski:
This is a full implementation of the SDK for .NET Core 2.0.
The SDK provides complete Signer, Processor and State Delta support.
An implementation for IntKey example is also provided as Processor as well a Client tool that can be run to make sample transaction calls.
I've added docker images and build scripts accordingly following the standard syntax, taking example from the existing scripts and languages.
Docker image uses microsoft/dotnet image.
I will be submitting README.md's for the SDK as well as the IntKey processor and cli tool to complement the SDK.

Signed-off-by: Tomislav Markovski <tmarkovski@gmail.com>


vaporos:
This is fantastic.

Recently, we have been discussing breaking off some of the SDKs from the core repo (Java and C++ for example) since the rest of core does not depend upon them. Our only barrier is really the time to do it (as it requires incremental build changes, etc.).

Thoughts on starting the dotnet SDK in a sawtooth-dotnet repo)?  I could coordinate repo creation and we could do the PR review there, if that is how we want to proceed.

tmarkovski:
I’m good either way, if the question was directed at me. I can resubmit to another repo.

tmarkovski:
Accidental close/reopen :)

tmarkovski:
Moving PR to sawtooth-sdk-dotnet

PR #1489, opened 2018-03-05
Message handler preprocessors

nick-drozd:
This PR implements @aludvik 's idea for cutting down on repeated
message deserialization, namely, adding "preprocessors" on a
message-by-message basis. These preprocessors act on messages before
they reach the handlers, deserializing the message content and also
extracting whatever other message information is required by the
handlers. With gossip messages, for instance, there are handlers that
require the message type tag and the message time-to-live in addition
to the batch or block object contained in the message, so those three
things are extracted and stored in a tuple that is passed to the
handlers.

With these changes, the poet liveness test seems to be able to run
with a higher workload, so I've included a commit to bump the workload
rates. This consistently passes when running locally, but maybe it
won't work on Jenkins.

### DIFF
--- a/validator/sawtooth_validator/gossip/gossip_handlers.py
+++ b/validator/sawtooth_validator/gossip/gossip_handlers.py
@@ -194,10 +194,7 @@ class GossipBatchResponseHandler(Handler):
         self._block_publisher_has_batch = block_publisher_has_batch
 
     def handle(self, connection_id, message_content):
-        batch_response_message = GossipBatchResponse()
-        batch_response_message.ParseFromString(message_content)
-        batch = Batch()
-        batch.ParseFromString(batch_response_message.content)
+        batch, _ = message_content
 
         batch_id = batch.header_signature
 
@@ -278,5 +275,16 @@ def gossip_block_response_preprocessor(message_content_bytes):
     block.ParseFromString(block_response.content)
 
     content = block, message_content_bytes
+
+    return PreprocessorResult(content=content)
+
+
+def gossip_batch_response_preprocessor(message_content_bytes):
+    batch_response = GossipBatchResponse()
+    batch_response.ParseFromString(message_content_bytes)
+    batch = Batch()
+    batch.ParseFromString(batch_response.content)
+
+    content = batch, message_content_bytes
 
     return PreprocessorResult(content=content)
--- a/validator/sawtooth_validator/gossip/gossip_handlers.py
+++ b/validator/sawtooth_validator/gossip/gossip_handlers.py
@@ -160,10 +160,7 @@ class GossipBlockResponseHandler(Handler):
         self._chain_controller_has_block = chain_controller_has_block
 
     def handle(self, connection_id, message_content):
-        block_response_message = GossipBlockResponse()
-        block_response_message.ParseFromString(message_content)
-        block = Block()
-        block.ParseFromString(block_response_message.content)
+        block, _ = message_content
 
         block_id = block.header_signature
 
@@ -270,5 +267,16 @@ def gossip_message_preprocessor(message_content_bytes):
         obj.ParseFromString(gossip_message.content)
 
     content = obj, tag, gossip_message.time_to_live
+
+    return PreprocessorResult(content=content)
+
+
+def gossip_block_response_preprocessor(message_content_bytes):
+    block_response = GossipBlockResponse()
+    block_response.ParseFromString(message_content_bytes)
+    block = Block()
+    block.ParseFromString(block_response.content)
+
+    content = block, message_content_bytes
 
     return PreprocessorResult(content=content)
--- a/validator/sawtooth_validator/gossip/gossip_handlers.py
+++ b/validator/sawtooth_validator/gossip/gossip_handlers.py
@@ -17,6 +17,7 @@ import logging
 from sawtooth_validator.networking.dispatch import Handler
 from sawtooth_validator.networking.dispatch import HandlerResult
 from sawtooth_validator.networking.dispatch import HandlerStatus
+from sawtooth_validator.networking.dispatch import PreprocessorResult
 from sawtooth_validator.protobuf import validator_pb2
 from sawtooth_validator.protobuf.batch_pb2 import Batch
 from sawtooth_validator.protobuf.block_pb2 import Block
@@ -123,29 +124,27 @@ class GossipMessageDuplicateHandler(Handler):
         self._has_batch = has_batch
 
     def handle(self, connection_id, message_content):
-        gossip_message = GossipMessage()
-        gossip_message.ParseFromString(message_content)
-        if gossip_message.content_type == gossip_message.BLOCK:
-            block = Block()
-            block.ParseFromString(gossip_message.content)
+        obj, tag, _ = message_content
+
+        header_signature = obj.header_signature
+
+        if tag == GossipMessage.BLOCK:
             has_block = False
-            if self._completer.get_block(block.header_signature) is not None:
+            if self._completer.get_block(header_signature) is not None:
                 has_block = True
 
-            if not has_block and self._has_block(block.header_signature):
+            if not has_block and self._has_block(header_signature):
                 has_block = True
 
             if has_block:
                 return HandlerResult(HandlerStatus.DROP)
 
-        if gossip_message.content_type == gossip_message.BATCH:
-            batch = Batch()
-            batch.ParseFromString(gossip_message.content)
+        if tag == GossipMessage.BATCH:
             has_batch = False
-            if self._completer.get_batch(batch.header_signature) is not None:
+            if self._completer.get_batch(header_signature) is not None:
                 has_batch = True
 
-            if not has_batch and self._has_batch(batch.header_signature):
+            if not has_batch and self._has_batch(header_signature):
                 has_batch = True
 
             if has_batch:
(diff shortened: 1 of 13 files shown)


validator/sawtooth_validator/gossip/gossip_handlers.py
```diff
@@ -122,30 +120,29 @@ def __init__(self, completer, has_block, has_batch):
...
         self._has_batch = has_batch
 
-    def handle(self, connection_id, message_content):
-        gossip_message = GossipMessage()
-        gossip_message.ParseFromString(message_content)
-        if gossip_message.content_type == gossip_message.BLOCK:
-            block = Block()
-            block.ParseFromString(gossip_message.content)
+    # pylint:disable=arguments-differ
+    def handle(self, connection_id, obj_tag_ttl):
```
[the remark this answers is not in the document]
nick-drozd: In a sense it doesn't implement the abstract base class correctly. The docstring for `Handler.handle` has 
```
        :param message_content: The bytes to be deserialized
                                into a protobuf python class
```
but this handler now expects something other than message content bytes. Leaving the argument as `message_content` would be misleading.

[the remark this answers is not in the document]
peterschwarz: I would agree.  Update the docstring to indicate that message content is either the raw bytes or a deserialized object. 

I would point out  that the name `obj_tag_ttl` provides very little indication as to what its value is, especially since you haven't documented its new meaning.

validator/sawtooth_validator/gossip/gossip_handlers.py
```diff
@@ -233,32 +232,40 @@ def __init__(self, gossip, completer):
...
+    tag = gossip_message.content_type
+
+    if tag == GossipMessage.BLOCK:
+        obj = Block()
+        obj.ParseFromString(gossip_message.content)
+    elif tag == GossipMessage.BATCH:
+        obj = Batch()
+        obj.ParseFromString(gossip_message.content)
+
+    return obj, tag, gossip_message.time_to_live - 1
```
peterschwarz: Time to live should not be reduced by the preprocessor - these changes should be made by the handlers, not during what is effectively a parsing phase.

validator/tests/test_client_request_handlers/test_submit_handlers.py
```diff
@@ -52,6 +54,7 @@ def test_batch_submit_without_wait(self):
 
         self.assertEqual(self.status.OK, response.status)
 
+    @unittest.expectedFailure
```
aludvik: This should be replaced by a unit test for the preprocessor to ensure deserialization failures are handled correctly.

validator/tests/test_client_request_handlers/test_submit_handlers.py
```diff
@@ -41,6 +42,7 @@ def setUp(self):
             store=store,
             tracker=tracker)
 
+    @unittest.expectedFailure
```
aludvik: This should not be disabled.

validator/sawtooth_validator/state/client_handlers.py
```diff
@@ -1094,3 +1102,10 @@ def _respond(self, request):
         return self._wrap_response(
             endpoint=self._gossip.endpoint,
             peers=sorted(peers, key=lambda peer: peer.endpoint))
+
+
+def client_batch_submit_request_preprocessor(message_content_bytes):
```
boydjohnson: Is this a full implementation of this preprocessor? It doesn't seem to be storing the parsed protobuf back in the dispatcher.

peterschwarz:
@boydjohnson With the current implementation, since every handler is isolated and can be reordered, you would need to either a) conventionally added it first, or b) every handler would need to possibly parse the content.

validator/sawtooth_validator/state/client_handlers.py
nick-drozd: The preprocessors are side effect-free. The result is stored at https://github.com/hyperledger/sawtooth-core/pull/1489/files#diff-5b210e75615ded71391b5b742dbc779eR204

boydjohnson: I think what is important that I didn't realize, is that `content=future.result(timeout)` in the defined function that gets called when the future resolves.

validator/sawtooth_validator/networking/dispatch.py
```diff
@@ -127,16 +137,21 @@ def remove_send_last_message(self, connection):
                          connection)
 
     def dispatch(self, connection, message, connection_id):
-        if message.message_type in self._msg_type_handlers:
-            priority = self._priority.get(message.message_type, Priority.LOW)
+        message_type = message.message_type
+        if message_type in self._msg_type_handlers:
+            priority = self._priority.get(message_type, Priority.LOW)
```
aludvik: This change is non-functional and not related to the commit message. Should be in a separate commit or dropped.

validator/sawtooth_validator/state/client_handlers.py
```diff
@@ -1094,3 +1102,10 @@ def _respond(self, request):
...
+    request = client_batch_submit_pb2.ClientBatchSubmitRequest()
+    request.ParseFromString(message_content_bytes)
```
aludvik: This still needs to correctly handle deserialization errors and respond back the client if that happens. This will require modification to the preprocessor callback (https://github.com/hyperledger/sawtooth-core/pull/1489/files#diff-5b210e75615ded71391b5b742dbc779eR202) to do something like: https://github.com/hyperledger/sawtooth-core/blob/master/validator/sawtooth_validator/networking/dispatch.py#L190

nick-drozd:
Still lacks client response and unit tests -- I will add them after the LR run.

nick-drozd:
Two LR networks, one without message preprocessors and one with, were run with a timer measuring how long it took for a gossip message to pass through all its handlers (see https://github.com/hyperledger/sawtooth-core/pull/1527). 

Here is the gossip throughput for the first few hours of the run without preprocessors:

screen shot 2018-03-20 at 5 43 14 pm

Here is the gossip throughput for the run with preprocessors:

screen shot 2018-03-20 at 5 43 58 pm

Note the units on the y-axes -- the gossip-handling time goes from taking several minutes to several seconds.

Now, this can't be called an unqualified success. Due to the vagaries of the workload generator, the run with preprocessors didn't have any invalid transactions, while the run without did. The differences between the two graphs was probably influenced by this. Still, it seems likely that the preprocessors had a significant positive impact on gossip message throughput.

validator/sawtooth_validator/networking/dispatch.py
```diff
@@ -160,15 +175,90 @@ def add_handler(self, message_type, handler, executor, priority=None):
         if priority is not None:
             self._priority[message_type] = priority
 
+    def set_preprocessor(self, message_type, preprocessor, executor):
```
aludvik: The type expected for preprocessor should be clearly documented here in a docstring. I would expect it to be something like a function `fn(message_content: bytes) -> PreprocessorResult` where `PreprocessorResult` informs the dispatcher what to do next and wraps the processed message content.

validator/sawtooth_validator/state/client_handlers.py
```diff
@@ -1094,3 +1104,20 @@ def _respond(self, request):
...
+
+def client_batch_submit_request_preprocessor(message_content_bytes):
+    request = client_batch_submit_pb2.ClientBatchSubmitRequest()
+
+    try:
+        request.ParseFromString(message_content_bytes)
+    except DecodeError:
+        LOGGER.error('ClientBatchSubmitRequest failed to deserialize')
+
+        return HandlerResult(
```
aludvik: Returning an undefined type in one case and a HandlerResult in another case is not a good interface to support. The return type should be consistent, which probably means creating a new PreprocessorResult that optionally wraps the processed content and signals to the Dispatcher what to do next.

validator/sawtooth_validator/networking/dispatch.py
```diff
@@ -374,6 +463,19 @@ class HandlerStatus(enum.Enum):
...
 
 
+class PreprocessorResult(HandlerResult):
+    def __init__(self, content=None, status=None,
+                 message_out=None, message_type=None):
+        """
+        :param content: the content returned if preprocessing is successful
+        :param status HandlerStatus: informs the dispatcher on how to proceed
+        :param message_out protobuf Python class:
+        :param message_type: validator_pb2.Message.* enum value
```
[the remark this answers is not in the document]
boydjohnson: Speaking with @nick-drozd offline, these docstrings are consistent with other docstrings in the module, but not other parts of the code base. If we feel it necessary to have consistent docstrings we should do that as a separate task.

validator/sawtooth_validator/gossip/gossip_handlers.py
```diff
@@ -234,30 +227,64 @@ def __init__(self, gossip, completer):
...
         else:
-            LOGGER.info("received %s, not BATCH or BLOCK",
-                        gossip_message.content_type)
+            LOGGER.info("received %s, not BATCH or BLOCK", tag)
         return HandlerResult(status=HandlerStatus.PASS)
+
+
+def gossip_message_preprocessor(message_content_bytes):
+    gossip_message = GossipMessage()
+    gossip_message.ParseFromString(message_content_bytes)
```
peterschwarz: Missing the Error handling here

validator/sawtooth_validator/gossip/gossip_handlers.py
```diff
@@ -234,30 +227,64 @@ def __init__(self, gossip, completer):
...
+        obj.ParseFromString(gossip_message.content)
+
+    content = obj, tag, gossip_message.time_to_live
+
+    return PreprocessorResult(content=content)
+
+
+def gossip_block_response_preprocessor(message_content_bytes):
+    block_response = GossipBlockResponse()
+    block_response.ParseFromString(message_content_bytes)
```
peterschwarz: Missing the error handling here

PR #1490, opened 2018-03-05
Updated PoET specs readability

askmish:

Sections edited:
 - z-test
 - Population Size and Local Mean Computation
 - Comments on Multi-user or Multi-ledger SGX Enclave Service

### DIFF
--- a/docs/source/architecture/poet.rst
+++ b/docs/source/architecture/poet.rst
@@ -589,13 +589,13 @@ Security Considerations
 Comments on Multi-user or Multi-ledger SGX Enclave Service
 ----------------------------------------------------------
 
-It is possible to use the same enclave for multiple users or ledgers by making
-username and ledgername input **Parameters** to ``generateSignUpData()`` and
-``unsealSignUpData()``. Then the sign-up tuple (username, ledgername, PPK, PSK,
-MCID) is sealed to disk, with username and ledgername used to generate the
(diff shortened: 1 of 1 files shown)


docs/source/architecture/poet.rst
```diff
@@ -590,7 +590,7 @@ Comments on Multi-user or Multi-ledger SGX Enclave Service
 ----------------------------------------------------------
 
 It is possible to use the same enclave for multiple users or ledgers by making
-username and ledgername input **Parameters** to ``generateSignUpData()`` and
+username and ledgername, as input parameters to ``generateSignUpData()`` and
```
chenette: I agree that this sentence was confusing, and I'm glad you're improving it.  However, this change is grammatically incorrect ("as" doesn't work with "make").  

I suggest saying "... by using username and ledgername as input parameters...". Note that there should not be a comma before "as".

P.S. I'm OK with "use" and "using" in the same sentence, but if you don't like it, you could say "by providing username and ledger name as input parameters".

aludvik:
@askmish I have some commits on an old branch that add a bunch of information about how SGX and PoET work that we decided not to merge awhile ago. If you are working on PoET and/or working on documenting PoET, you may or may not find them useful: https://github.com/aludvik/sawtooth-core/commits/poet-primer

PR #1494, opened 2018-03-06
State verifier regression test

aludvik:


validator/sawtooth_validator/server/state_verifier.py
```diff
@@ -49,21 +49,9 @@ class ExecutionError(Exception):
...
-    memory for all blocks in the blockstore and verifies the state root
-    hashes.
-
-    Raises:
-        InvalidChainError: The chain in the blockstore is not valid.
-        ExecutionError: An unrecoverable error was encountered during batch
-            execution.
-    """
-
+def get_databases(
```
nick-drozd: This should be one line (`def get_database(bind_network, data_dir=None)`).

nick-drozd:
I ran this without 97c3696 and it failed, which is good. However, `run_docker_test` hung and did not exit. This is the same behavior I was seeing yesterday running the state verifier locally. Is there any way to fix that?

aludvik:
Updated timeout and collapsed function arguments

PR #1496, opened 2018-03-06
Override OS Keepalive for ZMQ Sockets

peterschwarz:
Override OS keep-alive, as the underlying FD's will not necessarily be closed when the ZMQ socket it is, based on the underlying default Linux    settings.  Setting this to the same value as the validator's connection timeouts and interval should reduce the amount of time a socket is left open by the OS.

Also includes some minor reordering of the close/shutdown logic for a given Send/Receive thread.

Based on contributor feedback in this issue: https://github.com/zeromq/libzmq/issues/1453


aludvik:
For the ignorant like myself:
http://www.tldp.org/HOWTO/TCP-Keepalive-HOWTO/overview.html
http://api.zeromq.org/4-2:zmq-setsockopt

PR #1497, opened 2018-03-06
Prevent concurrent modification of tracked status

peterschwarz:
Prevent the caller of `get_invalid_txn_info` from reading a possibly concurrently modified collection.

This should prevent the error where faulty values are encountered while reading through the collection of transaction infos.


peterschwarz:
Updated to add unit tests and properly prevent the concurrent modification of the data.

PR #1499, opened 2018-03-07
Parallel scheduler fixes

nick-drozd:
This PR fixes two parallel scheduler bugs and adds some generated test cases.


validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -532,6 +575,9 @@ def add_batch(self, batch, state_hash=None, required=False):
                 txn_id = txn.header_signature
                 # Update our internal state with the computed predecessors.
                 self._txn_predecessors[txn_id] = set(predecessors)
+                self._predecessor_chain.add_relationship(
+                    txn_id=txn_id,
+                    predecessors=set(predecessors))
```
peterschwarz: Does this set need to be generated twice?  Are they modified over time in different ways?

nick-drozd: Yes, `add_relationship` mutates a copy of `predecessors`. Actually, both set conversions could done away with: `_find_{input,output}_dependencies` uses lists instead of sets, when the output is immediately converted to a set, and `add_relationship` could avoid mutation with a little cleverness.

Anyway, I moved the second set conversion (which is really just a copy operation) into the body of `add_relationship`, since it's an implementation detail that the caller doesn't need to worry about.

nick-drozd:
Updated to take out most of the big yaml files and reduce the size of the remaining big one (from ~1200 lines to ~800).

PR #1501, opened 2018-03-08
Remove LGPL chardet from Sawtooth install scripts

pankajgoyal2:
Signed-off-by: pankajgoyal2 <pankaj.goyal@intel.com>


docker/sawtooth-debug-python
```diff
@@ -70,7 +68,6 @@ RUN echo "deb http://repo.sawtooth.me/ubuntu/ci xenial universe" >> /etc/apt/sou
     python3-zmq-dbg \
     python3-cffi-backend-dbg python3-apt-dbg \
     python3-netifaces-dbg \
-    libzmq-dbg python3-cchardet-dbg \
```
askmish: libzmq-dbg should stay?

TomBarnes: This PR is solely to remove chardet.  I recommend we investigate other opportunities to remove unused components separately.

TomBarnes: Confirming that only chardet is to be removed.  cchardet should not be removed, as it is a dependency of aiohttp.

pankajgoyal2:
probably I need to update the commit message for my latest push.

NOTICES
```diff
@@ -2156,4 +1646,4 @@ freely, subject to the following restrictions:
 2. Altered source versions must be plainly marked as such, and must not be
    misrepresented as being the original source code.
 
-3. This notice may not be removed or altered from any source distribution.
\ No newline at end of file
+3. This notice may not be removed or altered from any source distribution.
```
vaporos: Although this is probably a good change, this change is not related to the functional change. To make it easier to cherry-pick without conflict, this should be in a separate commit.

dcmiddle: Do you mean the newline?

vaporos: Yes

PR #1503, opened 2018-03-08
Convert all dashboard derivates to 10s/10

aludvik:
Using non_negative_derivative(1s) hurts dashboard performance without
improving accuracy since we are only reporting every 10s.

### DIFF
--- a/docker/grafana/dashboards/dashboard.json
+++ b/docker/grafana/dashboards/dashboard.json
@@ -1016,9 +1016,15 @@
                   },
                   {
                     "params": [
-                      "1s"
+                      "10s"
                     ],
                     "type": "non_negative_derivative"
+                  },
+                  {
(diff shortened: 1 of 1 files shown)


docker/grafana/dashboards/dashboard.json
```diff
@@ -2281,7 +2323,7 @@
                   },
                   {
                     "params": [
-                      "1s"
+                      "10s"
```
peterschwarz: Looks like you are missing the `/10` for this one

aludvik: The diff doesn't show this, but the dashboard already had the `/10` and was incorrect before.

peterschwarz: My mistake

PR #1504, opened 2018-03-08
Logging improvements around Auth Rejection

peterschwarz:
In order to track down the root cause of [STL-1020](https://jira.hyperledger.org/browse/STL-1020), several logging additions have been made.  This should give a better indication as to where in the peering process a messages is being sent at the wrong time (i.e. before the auth/trust handshake has been completed).

### DIFF
--- a/validator/sawtooth_validator/networking/dispatch.py
+++ b/validator/sawtooth_validator/networking/dispatch.py
@@ -264,6 +264,10 @@ class Dispatcher(InstrumentedThread):
                     correlation_id=original_message.correlation_id,
                     message_type=res.message_type)
                 try:
+                    LOGGER.warning(
+                        "Sending hang-up in reply to %s to connection %s",
+                        get_enum_name(original_message.message_type),
+                        connection_id)
                     self._send_last_message[connection](
                         msg=message,
                         connection_id=connection_id)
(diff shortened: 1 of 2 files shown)


validator/sawtooth_validator/networking/dispatch.py
```diff
@@ -264,6 +264,11 @@ def _determine_next(self, message_id, future):
                     correlation_id=original_message.correlation_id,
                     message_type=res.message_type)
                 try:
+                    LOGGER.warning("Sending hang-up in reply to %s to "
+                                   "connection %s",
+                                   get_enum_name(
+                                       original_message.message_type),
+                                   connection_id)e
```
nick-drozd: errant `e`

validator/sawtooth_validator/networking/dispatch.py
```diff
@@ -264,6 +264,11 @@ def _determine_next(self, message_id, future):
                     correlation_id=original_message.correlation_id,
                     message_type=res.message_type)
                 try:
+                    LOGGER.warning("Sending hang-up in reply to %s to "
```
nick-drozd: Can this message be on a single line? It's hard to grep for messages when they're broken up.

PR #1505, opened 2018-03-09
Rewrite Docker install procedure

chenette:
This is a major rewrite/reorg for the content, based on the previous rewrite for the AWS procedure.
Note that the actual commands have not changed, except for new Docker commands to check the logs.

### DIFF
--- a/docs/source/app_developers_guide/docker.rst
+++ b/docs/source/app_developers_guide/docker.rst
@@ -2,27 +2,34 @@
 Using Sawtooth with Docker
 **************************
 
+This procedure explains how to set up Hyperledger Sawtooth for application
+development using a multi-container Docker environment. It shows you how to
+start Sawtooth and connect to the necessary Docker containers, then walks you
+through the following tasks:
 
-This procedure walks through the process of setting up Hyperledger Sawtooth
-for application development using Docker Compose, introduces some of the basic
-Sawtooth concepts necessary for application development, and walks through
-performing the following tasks:
+ * Checking the status of Sawtooth components
+ * Configuring transaction family settings
+ * Using Sawtooth commands to submit transactions, display block data, and view
+   global state
+ * Examining Sawtooth logs
+ * Stopping Sawtooth and resetting the Docker environment
 
-* Submit transactions to the REST API
-* View blocks, transactions, and state with Sawtooth commands
-* Start and stop validators and transaction processors
+After completing this tutorial, you will be prepared for the other tutorials
+in this guide. The next tutorial introduces the XO transaction family with
+``xo`` client commands. The final set of tutorials describe how to use an SDK to
+create a transaction family that implements your application's business logic.
 
+.. important::
 
-Upon completing this tutorial, you will be prepared for the more advanced
-tutorials that guide you in performing app development tasks, such as
-implementing business logic with transaction families and writing clients
-which use Sawtooth's REST API.
+   This demo environment is required for the other tutorials in this guide.
+
 
 About the Demo Sawtooth Environment
 ===================================
 
-This tutorial describes how to use Docker to set up a single Sawtooth node
-with a validator, a REST API, and two transaction processors.
+The Docker demo environment is a single validator node that is running a
+validator, a REST API, and three transaction processors. This environment uses
+Developer mode (dev mode) consensus and serial transaction processing.
 
 .. figure:: ../images/appdev-environment-one-node-3TPs.*
    :width: 100%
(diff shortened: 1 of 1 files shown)


docs/source/app_developers_guide/docker.rst
```diff
@@ -2,696 +2,672 @@
...
 
-One of the on-chain settings is the list of supported transaction families.
-In the example below, a JSON array is submitted to the ``sawset``
-command, which creates and submits a batch of transactions containing the
-configuration change.
+.. note::
 
-The submitted JSON array tells the validator or validator network to accept
-transactions of the following types:
+   By convention, the transaction processors use a string to make the log file
```
aludvik: I would change "string" to "random string", since they are all strings

docs/source/app_developers_guide/docker.rst
```diff
@@ -2,696 +2,672 @@
...
+    in the example Compose file, ``sawtooth-default.yaml``.
 
-Log into this container by running this command from the host computer's
-terminal window:
+    +---------------+---------------------------------------+----------+----------------------+
+    | **Component** | **Container Name**                    | **Port** | **Host Name**        |
+    +===============+=======================================+==========+======================+
+    | Validator     | ``sawtooth-validator-default``        | 4004     | ``validator``        |
+    +---------------+---------------------------------------+----------+----------------------+
+    | REST API      | ``sawtooth-rest-api-default``         | 8008     |                      |
```
aludvik: missing host-name `rest-api`

docs/source/app_developers_guide/docker.rst
```diff
@@ -2,696 +2,672 @@
...
+    +---------------+---------------------------------------+----------+----------------------+
+    | Settings TP   | ``sawtooth-settings-tp-default``      |          | ``settings-tp``      |
+    +---------------+---------------------------------------+----------+----------------------+
+    | IntegerKey TP | ``sawtooth-intkey-tp-python-default`` |          | ``intkey-tp-python`` |
+    +---------------+---------------------------------------+----------+----------------------+
+    | XO TP         | ``sawtooth-xo-tp-python-default``     |          | ``xo-tp-python``     |
+    +---------------+---------------------------------------+----------+----------------------+
+    | Shell         | ``sawtooth-shell-default``            |          |                      |
+    +---------------+---------------------------------------+----------+----------------------+
```
aludvik: There is some nuances about the context in which these values have meaning. To make sense of it, you need to separate out the following:
- What you can see when interacting with the containers from the host environment
- What you can see when interacting with the other containers from within a container

From the host environment, you can refer to each of the containers with name listed under "Container Name". You cannot, however, interact with the containers using their "Host Name" from the host. So you can do things like `docker exec -it sawtooth-validator-default bash` from the host, but not things like `curl http://rest-api:8008` from the host.

The "Host Name" is only valid when containers to refer to each other (this has to do with virtual docker networks and special DNS resolution rules within containers). So if you were to connect to a container with `docker exec -it sawtooth-validator-default bash` and then _from within the container_ do `curl http://rest-api:8008`, it would work.

Perhaps the trickiest part is the ports. There is a difference between "exposed" ports and "forwarded" ports and the naming conventions are not helpful. An "exposed" port is a port that can be seen from one container to another container, but not from the host to the container. On the other hand, a "forwarded" port is one that can be seen from the host to the container, but not necessarily from one container to another. To make this even more confusing, in the `sawtooth-default.yaml` container being referenced, within the validator and the rest api, the port mentioned is both exposed and forwarded:

```
    expose: # These are the exposed ports (container-to-container)
      - 4004
    ports: # These are the forwarded ports (host-to-container)
      - "4004:4004"
```

docs/source/app_developers_guide/docker.rst
```diff
@@ -2,696 +2,672 @@
...
+============================
 
-To confirm that a validator is running and reachable from the client
-container, run this ``curl`` command as root:
+Use these steps to be sure that you can connect to the validator
+and the REST API.
 
-.. code-block:: console
+ #. To confirm that a validator is running and reachable from the client
+    container, run this ``curl`` command:
```
aludvik: This is the same as below for checking connectivity to the REST API. Is there supposed to be a different command run here? Checking connectivity to the REST API with the /blocks route also checks that the validator is running, since the REST API depends on it to get a response.

docs/source/app_developers_guide/docker.rst
```diff
@@ -2,696 +2,672 @@
...
 
-After shutdown has completed, run this command:
+This terminal window will continue to display log messages as you run commands
+in other containers.
 
-.. code-block:: console
+.. note::
 
-  % docker-compose -f sawtooth-default.yaml down
+   If you need to reset the environment environment for any reason, see
```
aludvik: typo, "environment" is repeated

docs/source/app_developers_guide/docker.rst
```diff
@@ -2,696 +2,672 @@
...
+   "Stopping Sawtooth".
```
aludvik: This should be a link

docs/source/app_developers_guide/docker.rst
```diff
@@ -2,696 +2,672 @@
...
+following command:
 
 .. code-block:: console
 
-  % docker exec -it sawtooth-shell-default bash
+  $ docker exec -it sawtooth-shell-default bash
+  root@75b380886502:/#
 
+In the following steps, the simplified prompt ``#`` is used for the commands
+that run from the terminal window for the client container. The ``$`` prompt
```
aludvik: "should be run"

docs/source/app_developers_guide/docker.rst
```diff
@@ -2,696 +2,672 @@
...
+denotes commands that should be executed from a terminal window on your host
+system. The full prompt ``root@21bb5af673a8:/#`` shows commands that run in the
+validator container.
```
aludvik: I am not a huge fan of this convention, as it is pretty arbitrary. Maybe ``root@client #`` for client, ``host $``, and ``root@validator #`` would be more meaningful?

docs/source/app_developers_guide/docker.rst
```diff
@@ -2,696 +2,672 @@
 Using Sawtooth with Docker
 **************************
 
+This tutorial explains how to set up Hyperledger Sawtooth for application
+development using a multi-container Docker environment. This step-by-step
+procedure shows you how to start Sawtooth and connect to the necessary Docker
+containers, then walks you through the following tasks:
```
aludvik: "This tutorial" followed by "This step-by-step procedure" feels awkward. Can the second sentence just start with "It shows"?

docs/source/app_developers_guide/docker.rst
chenette: Good idea! Will do.

docs/source/app_developers_guide/docker.rst
chenette: No, it's just that I didn't understand that checking connectivity to the REST API with the /blocks route also checks that the validator is running. I'll remove 3. and 4. from this step.

docs/source/app_developers_guide/docker.rst
chenette: Thanks for the explanation.  I propose adding this sentence after the table to explain about the exposed and forwarded/published ports: 

     "The validator and REST API ports are exported to other containers and
       forwarded (published) for external connections, such as from your host
       system."

docs/source/app_developers_guide/docker.rst
chenette: Fixed

docs/source/app_developers_guide/docker.rst
chenette: Fixed

docs/source/app_developers_guide/docker.rst
chenette: Fixed (along with another repeated word).

docs/source/app_developers_guide/docker.rst
chenette: Fixed

docs/source/app_developers_guide/docker.rst
chenette: Fixed

docs/source/app_developers_guide/docker.rst
chenette: Fixed.

docs/source/app_developers_guide/docker.rst
aludvik: Did you mean to say "exported" here? Or should this say "exposed"?

chenette: Fixed.

PR #1510, opened 2018-03-12
Execute dispatcher callback directly

nick-drozd:
I'm putting up this PR to test a hypothesis about why https://github.com/hyperledger/sawtooth-core/pull/1489 failed its LR run. The hypothesis is that callbacks are sometimes being executed in the dispatcher thread rather than in the intended threadpool threads. These callbacks can potentially perform expensive computations, slowing down the dispatcher and thereby slowing down everything.

### DIFF
--- a/validator/sawtooth_validator/networking/dispatch.py
+++ b/validator/sawtooth_validator/networking/dispatch.py
@@ -174,28 +174,20 @@ class Dispatcher(InstrumentedThread):
             del self._message_information[message_id]
             return
 
-        future = handler_manager.execute(connection_id, message.content)
-
         timer_tag = type(handler_manager.handler).__name__
         timer_ctx = self._get_dispatch_timer(timer_tag).time()
 
-        def do_next(future):
+        def do_next(result):
             timer_ctx.stop()
             try:
-                self._determine_next(message_id, future)
+                self._determine_next(message_id, result)
             except Exception:  # pylint: disable=broad-except
                 LOGGER.exception(
                     "Unhandled exception while determining next")
 
-        future.add_done_callback(do_next)
+        handler_manager.execute(connection_id, message.content, do_next)
 
-    def _determine_next(self, message_id, future):
-        try:
-            result = future.result(timeout=self._timeout)
-        except TimeoutError:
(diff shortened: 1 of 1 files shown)


validator/sawtooth_validator/networking/dispatch.py
```diff
@@ -174,48 +174,40 @@ def _process(self, message_id):
...
-        try:
-            res = future.result(timeout=self._timeout)
-        except TimeoutError:
-            LOGGER.exception("Dispatcher timeout waiting on handler result.")
-            raise
+        handler_manager.execute(connection_id, message.content, do_next)
 
-        if res is None:
+    def _determine_next(self, message_id, result):
+        if result is None:
```
peterschwarz: the rename of result should be a separate commit.

nick-drozd: What do you want instead? `_determine_next(self, message_id, res)`?

peterschwarz: I like the rename, I just think the rename and the move to a parameter could be two separate commits.  It was hard to see the actual change, due to the renaming clutter.

nick-drozd:
Here are before and after Yappi graphs from a network of three validators running 3000 smallbank transactions:

Before:
master-1

After:
callback-1

Overall they don't look dramatically different. The biggest difference that I can tell is that the parallel scheduler's `unschedule_incomplete_batches` goes from taking 4% of CPU time to 5.5%. This suggests that more batches are being processed than can be published, which is what we would expect if throughput really is increased.

nick-drozd:
Here are before and after committed transaction graphs from the first 12 hours of LR runs:

Before:
screen shot 2018-03-13 at 1 05 25 pm

After:
screen shot 2018-03-13 at 1 05 12 pm

(One of the nodes running this PR segfaulted, but that is likely unrelated to the changes here.)

validator/sawtooth_validator/networking/dispatch.py
nick-drozd: I pushed the renaming into the refactor commit.

PR #1511, opened 2018-03-12
Short-circuit test-state-verifier shutdown

aludvik:
There is a bug in the shutdown code for some component this depends on, which causes it to occasionally hang during shutdown. I tried a number of things to get this test passing while avoiding depending on the shutdown code to work properly, but none of them worked 100% of the time. This test is still valid, since an exception will be raised if there is a problem with state verifying before the `os._exit(0)` line is run. In other words, the `os._exit(0)` line will only be run if the test was successful.

The reason this test is failing because of the shutdown code is that the transaction execution components are being run from the same container as the test itself, which no other integration or unit test does to the same extent as here. In the other tests, only the exit code of the test container matters, so when it exits, all of the other containers are just killed without checking whether they shut down cleanly or not.

### DIFF
--- a/bin/run_tests
+++ b/bin/run_tests
@@ -274,7 +274,6 @@ test_validator() {
     run_docker_test test_namespace_permission
     copy_coverage .coverage.namespace_permission
     run_docker_test test_state_verifier --timeout 30
-    copy_coverage .coverage.state_verifier
 }
 
 test_go_sdk() {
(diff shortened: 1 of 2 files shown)


aludvik:
Created this bug to track the issue: https://jira.hyperledger.org/browse/STL-1117

nick-drozd:
I tried running this without https://github.com/hyperledger/sawtooth-core/commit/97c3696daef315e1efcd34dd9f1d58d12284aacc. The Python failed, which is good, but the Docker container hung. That's not so bad, I guess, since it still gates changes that fail the test.

PR #1512, opened 2018-03-12
Log networking irregularities as warnings

nick-drozd:
These messages indicate states that should not normally arise, e.g.
trying to unregister a connection only to find that the connection
wasn't registered in the first place. They should be logged at least
at warning level.

### DIFF
--- a/validator/sawtooth_validator/networking/handlers.py
+++ b/validator/sawtooth_validator/networking/handlers.py
@@ -212,10 +212,10 @@ class PingHandler(Handler):
         if connection_id in self._last_message:
             if time.time() - self._last_message[connection_id] < \
                     self._allowed_frequency:
-                LOGGER.debug("Too many Pings in %s seconds before "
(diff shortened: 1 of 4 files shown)


nick-drozd:
These were all seen just before an occurrence of the "Too many files open" bug.

validator/sawtooth_validator/networking/handlers.py
```diff
@@ -212,10 +212,10 @@ def handle(self, connection_id, message_content):
         if connection_id in self._last_message:
             if time.time() - self._last_message[connection_id] < \
                     self._allowed_frequency:
-                LOGGER.debug("Too many Pings in %s seconds before "
-                             "authorization is complete: %s",
-                             self._allowed_frequency,
-                             connection_id)
+                LOGGER.warning("Too many Pings in %s seconds before "
```
peterschwarz: Could you add the time difference here?  It would be helpful to know frequently node is sending the ping.

PR #1515, opened 2018-03-14
Add option to intkey-workload to set an unnecessary dependency

boydjohnson:
This differs from #1509 in that this dependency is able to be satisfied, but unnecessary, while #1509 was for dependencies that could not be satisfied.  This PR finishes STL-1017.

### DIFF
--- a/perf/intkey_workload/src/main.rs
+++ b/perf/intkey_workload/src/main.rs
@@ -125,6 +125,15 @@ fn get_arg_matches<'a>() -> ArgMatches<'a> {
                 .help("Comma separated list of u8 to make the workload reproduceable"),
         )
         .arg(
+            Arg::with_name("unnecessary")
+                .long("unnecessary")
+                .takes_value(true)
+                .number_of_values(1)
+                .default_value("0.0")
+                .value_name("UNNECESSARY")
+                .help("Probability of a transaction having a satisfiable but unnecessary depedendency"),
+        )
+        .arg(
             Arg::with_name("unsatisfiable")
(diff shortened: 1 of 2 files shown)


perf/intkey_workload/src/main.rs
```diff
@@ -209,6 +218,11 @@ fn run_load_command(args: &ArgMatches) -> Result<(), Box<Error>> {
             String::from("unsatisfiable must be a positive float in the range [0.0, 1.0]")
         })?;
 
+    let unnecessary: f32 = args.value_of("unnecessary")
+        .unwrap_or("0.0")
+        .parse()
```
peterschwarz: What happens if this is a negative float or bigger than 1?

boydjohnson: Less than zero will act like 0, and greater than 1 will act like 1, but I should put in some parsing so that something unexpected doesn't happen for the user.

boydjohnson:
```
./intkey-workload --wildcard 7.9
Value must be between 0.0 and 1.0, inclusively
```
This is what one of the error messages looks like now.

perf/intkey_workload/src/main.rs
```diff
@@ -28,9 +28,13 @@ mod intkey_addresser;
 mod intkey_iterator;
 mod intkey_transformer;
 
+use std::convert::From;
```
peterschwarz: I don't think this has to be imported

boydjohnson: rustc doesn't warn about an unneeded import.

PR #1517, opened 2018-03-14
More database optimizations

nick-drozd:
This PR contains a couple of database-related performance
improvements.

The profiling method I'm using here will be a little different from
usual. Rather than running transactions through a network, I'm running
the state verifier over a 300-block blockstore that was taken from an
actual LR network run (many of the functions in the graphs below are
called exactly 300 times). This profiles just the components involved
in transaction execution. Notably, it does not involve the block
publisher or any inter-validator communication, although does include
communcation between the validator and several transaction processors.

Here is a Yappi graph of this setup running master:

master

The first commit adds a dedicated single-key `get` method to
`LMDBNoLockDatabase`, where previously `get` was delegated to
`get_multi`. This shaves off almost 2% of CPU time from
`LMDBNoLockDatabase.get`. It also slightly reduces (by ~0.2%) the time
taken by `LMDBNoLockDatabase.update`. I'm not sure why this is; maybe
the database writes don't have to wait as long for the reads?

Here is a Yappi graph after this change:

db-get

The second commit caches the settings views obtained by the executor.
This seems to be an impactful change. For one thing, database reads
are reduced by about one third. The total CPU time taken by
`TransactionExecutorThread._execute_schedule` goes from ~18% to ~15%.

Here is a Yappi graph after this change:

config

### DIFF
--- a/validator/sawtooth_validator/database/lmdb_nolock_database.py
+++ b/validator/sawtooth_validator/database/lmdb_nolock_database.py
@@ -65,6 +65,15 @@ class LMDBNoLockDatabase(database.Database):
         with self._lmdb.begin() as txn:
             return txn.get(key.encode()) is not None
 
+    def get(self, key, index=None):
+        with self._lmdb.begin() as txn:
+            packed = txn.get(key.encode())
+
+        try:
+            return cbor.loads(packed)
+        except ValueError:
+            return None
+
     def get_multi(self, keys, index=None):
         with self._lmdb.begin() as txn:
             result = []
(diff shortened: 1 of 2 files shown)


jsmitchell:
Can you split these into two PRs, or just submit a PR with the database get() commit by itself? The settings view change will require more discussion given the need for the cache to be aware of intra-block txns which change settings and their potential effect on downstream txns in that block.

nick-drozd:
Closing in favor of https://github.com/hyperledger/sawtooth-core/pull/1522

PR #1518, opened 2018-03-14
Only request peers from connections which have completed handshake

peterschwarz:
In permissioned networks, it is possible for the network to request peers from connections which have not completed the authorization handshake.  Get Peer Messages will not be sent to connections that have not completed this handshake.

### DIFF
--- a/validator/sawtooth_validator/networking/interconnect.py
+++ b/validator/sawtooth_validator/networking/interconnect.py
@@ -735,28 +735,34 @@ class Interconnect(object):
         """
         Get stored public key for a connection.
         """
-        if connection_id in self._connections:
-            connection_info = self._connections[connection_id]
-            return connection_info.public_key
-        return None
+        with self._connections_lock:
+            try:
+                connection_info = self._connections[connection_id]
+                return connection_info.public_key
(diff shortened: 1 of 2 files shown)


validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -758,6 +758,20 @@ def get_connection_status(self, connection_id):
...
+        handshake.
+
+        Returns:
+            bool - True if the connection handshake is complete, False
+                otherwise
+        """
+        if connection_id in self._connections:
+            connection_info = self._connections[connection_id]
+            return connection_info.status == ConnectionStatus.CONNECTED
+        return False
```
aludvik: I think you need to acquire `self._connections_lock` here.

aludvik: Also, I think we prefer `try/except` to `if .. in ..`..

peterschwarz: None of those methods that look at that map use that.  Will put up a separate commit to refactor those to use a try/catch and the lock.

PR #1519, opened 2018-03-14
Make unscheduler faster

nick-drozd:
Running 3000 transactions through a 3-node network with 100 transactions per batch, `ParallelScheduler.unschedule_incomplete_batches` takes around 5% of total CPU time. Changing a certain scheduler data structure from a list to a dict reduces that practically nothing.

Here are before and after graphs from validator 1:

master-1

unscheduler-1

Here are before and after graphs from validator 2:

master-2

unscheduler-2

One could be forgiven for thinking these numbers are bogus. I certainly did at first. But I reran the test several times and the results always come out the same.

### DIFF
--- a/validator/sawtooth_validator/execution/scheduler_parallel.py
+++ b/validator/sawtooth_validator/execution/scheduler_parallel.py
@@ -877,10 +877,6 @@ class ParallelScheduler(Scheduler):
 
             self._condition.notify_all()
 
-    def _unscheduled_transactions(self):
-
-        return self._txns_available.copy()
-
     def _has_predecessors(self, txn_id):
         for predecessor_id in self._txn_predecessors[txn_id]:
             if predecessor_id not in self._txn_results:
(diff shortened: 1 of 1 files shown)


validator/sawtooth_validator/execution/scheduler_parallel.py
```diff
@@ -877,10 +877,6 @@ def set_transaction_execution_result(
 
             self._condition.notify_all()
 
-    def _unscheduled_transactions(self):
-
-        return self._txns_available.copy()
```
peterschwarz: Isn't this copy here thanks to possible concurrent modification of this list?

nick-drozd: Is the list being modified by multiple functions at once? If so, shouldn't a lock be used? I was thinking it was just so that the list could be modified during iteration. This change is only here for convenience, so I can get rid of it if need be.

nick-drozd: As far as I can tell, `_txns_available` is only used in functions with locks (except for `_remove_subsequent_result_because_of_batch_failure` and `_reschedule_if_outstanding`, but those are themselves only called from functions with locks).

@boydjohnson, is that right?

PR #1526, opened 2018-03-16
STL-1085 Go TP Reconnect

aludvik:
Previously, there were a number of problems with the shutdown and
reconnect code. Graceful shutdown would hang if the processor was
registering or if the validator never responded with an unregister
response. Reconnection could fail when a validator disconnected
because threads were blocked waiting for a response and would never
stop.

This commit solves these problems by using a ZMQ socket to signal
shutdown so it can be included in a poll with a receiving socket
and by aggresively closing sockets that worker threads depend on
so any send/receives will unblock and the worker threads can shutdown.


peterschwarz:
So you have a recommended way to test this?

aludvik:
@peterschwarz I started up a validator, rest api, go tp, and workload and messed around with killing the validator and go tp to make sure things continued to work.

PR #1532, opened 2018-03-21
Handle retry of batchlists in sawtooth perf

boydjohnson:
This change affects the rust intkey-workload tool as well as smallbank-workload.

This PR provides functionality so that every batchlist submitted gets a HTTP 202 (Accepted) status code. If a batchlist gets a non 202, it will be submitted again immediately.


boydjohnson:
I've done manual testing by starting up the `intkey-workload` and `smallbank-workload` before starting the REST api and validator. Once they have run for 20-30 seconds, I start up the REST api and validator. There are no invalid transactions in the logging output.

perf/sawtooth_perf/src/batch_map.rs
```diff
@@ -0,0 +1,171 @@
...
+    // Mark that batchlist associated with batch id has been submitted
+    // to a validator.
+    pub fn mark_submit_success(&mut self, batch_id: String) {
+        self.batches_by_id.remove(batch_id.as_str());
+    }
+
+    // Get a batchlist by id, to submit it to a validator.
+    pub fn get_batchlist_to_submit(&mut self, batch_id: String) -> Option<BatchList> {
+        self.batches_by_id
+            .clone()
```
aludvik: This seems like an expensive clone(). Is it needed since you are cloning the batchlist below?

perf/sawtooth_perf/src/batch_map.rs
```diff
@@ -0,0 +1,171 @@
...
+            .get(batch_id.as_str())
+            .map(|batchlist| batchlist.clone())
+    }
+
+    // Idempotent method for adding a BatchList
+    pub fn add(&mut self, batchlist: BatchList) {
+        let batch_id = match batchlist.batches.last() {
+            Some(batch) => Some(batch.header_signature.clone()),
+            None => None,
+        };
```
aludvik: Using `.map()` instead of `match` here is less verbose

perf/sawtooth_perf/src/batch_map.rs
```diff
@@ -0,0 +1,171 @@
...
+        match batch_id {
+            Some(batch_id) => {
+                if !self.batches_by_id.contains_key(batch_id.as_str()) {
+                    self.batches_by_id.insert(batch_id.clone(), batchlist);
+                }
+            }
+            None => {}
```
aludvik: You can use `.map_or()` here instead, and you can chain it with the previous map: http://play.integer32.com/?gist=b31eda605c1d6d61c0d5aaf41ce048ae&version=stable

perf/sawtooth_perf/src/batch_map.rs
boydjohnson: Good catch.

PR #1533, opened 2018-03-21
Specify grafana version

grkvlt:
Backport the Grafana version fix to the 1.0 release

### DIFF
--- a/docker/grafana/sawtooth-stats-grafana
+++ b/docker/grafana/sawtooth-stats-grafana
@@ -13,7 +13,7 @@
 # limitations under the License.
 # ------------------------------------------------------------------------------
 
-FROM grafana/grafana:latest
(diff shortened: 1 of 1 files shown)


vaporos:
Usually, we are backporting to a staging branch, then merging that staging branch to 1-0 after testing it on an LR network.

@peterschwarz @dcmiddle Thoughts on skipping the 1-0-staging-00 branch based on the nature of this change? Seems pretty safe but then we aren't testing the specific resulting commit id (which @dcmiddle has advocated for previousy).  I think it should probably go to 1-0-staging-00 unless there is a specific reason not to do that.

peterschwarz:
This should be re-opened against `1-0-staging-00` branch, which 1.0.3 will be produced from

PR #1534, opened 2018-03-22
Create sawtooth-default-poet compose file

dplumb94:
This file pulls hyperledger/sawtooth-..:1.0 images and starts a 5 node poet network with intkey and xo transaction processors.


chenette:
I don't know YAML, so I can't comment on the details, but all the right pieces are there.  This is what we need to document the Docker procedure for adding multiple nodes to a Sawtooth network. THANKS!!!

P.S.  Note for posterity (aka myself): In addition to the intkey and xo transaction processors, this Compose file also includes the settings-tp and poet-validator-registry-tp transaction processors. So it matches the upcoming documentation.

PR #1536, opened 2018-03-23
Add new procedure on adding a node to a Sawtooth network

chenette:
- Includes Docker and Ubuntu (but not AWS, because AWS is only 0.8).

- Still needs information on how to verify network functionality - see the last section.


chenette:
The Docker Compose file for this procedure is in PR #1534

docs/source/app_developers_guide/adding_a_node.rst
```diff
@@ -0,0 +1,373 @@
...
+      .. code-block:: none
+
+         bind = [
+           "network:tcp://127.0.0.1:8800",
+           "component:tcp://127.0.0.1:4004"
+         ]
+
+   #. Set ``endpoint`` to the IP address and port of the advertised network
+      endpoint. This is the public address that clients will use to contact this
+      validator node via the REST API.  The default value is the same as the
```
dcmiddle: `endpoint` is the publicly addressable IP and port of the validator to validator connection. it may differ from the `network` setting in `bind` when the validator is hosted behind a firewall. In that case one would be relying on the firewall to map the `endpoint` IP:Port to the `bind` `network` IP:Port.  `endpoint` is the value this validator will share with other validators when it is peering with them.

docs/source/app_developers_guide/adding_a_node.rst
```diff
@@ -0,0 +1,373 @@
...
+      bind network value.
+
+      .. code-block:: none
+
+         endpoint = "tcp://external-ip-1:8008"
```
dcmiddle: probably be more consistent to make this port 8800

docs/source/app_developers_guide/adding_a_node.rst
```diff
@@ -0,0 +1,373 @@
...
+
+#. Edit this file to change the following settings:
+
+   a. Set ``bind`` to the network and component endpoints for this node. For
+      this application development environment, can use the default values.
+
+      .. code-block:: none
+
+         bind = [
+           "network:tcp://127.0.0.1:8008"
```
dcmiddle: 8800 is default in our example toml file. Someone pipe up if we have a different default specified elsewhere.

docs/source/app_developers_guide/adding_a_node.rst
```diff
@@ -0,0 +1,373 @@
...
+           "component:tcp://127.0.0.1:4004"
+         ]
+
+   #. Set the advertised network endpoint URL. The default value is:
+
+      .. code-block:: none
+
+         endpoint = "tcp://external-ip-2:8008"
```
dcmiddle: 8800

docs/source/app_developers_guide/adding_a_node.rst
```diff
@@ -0,0 +1,373 @@
...
+
+   #. Select static peering (the default) by setting ``peers`` to the IP
+      address and port of the first validator.
+
+      .. code-block:: none
+
+         peers = ["tcp://external-ip-1:8801"]
```
dcmiddle: 8800

docs/source/app_developers_guide/adding_a_node.rst
```diff
@@ -0,0 +1,373 @@
...
+
+      .. note::
+
+         If you are adding several nodes, specify the nodes in a comma-separated
+         list, as in this example:
+
+         .. code-block:: none
+
+            peers = ["tcp://external-ip-1:8801","tcp://external-ip-3:8801",...,"tcp://external-ip-5:8801"]
```
dcmiddle: 8800

docs/source/app_developers_guide/adding_a_node.rst
```diff
@@ -0,0 +1,373 @@
...
+
+Confirm Network Functionality
+=============================
+
+#. To check whether peering has occurred on the network, submit a block query
+   to the REST API from the first validator node.
+
+   .. code-block:: none
+
+      NEED INFO (COMMANDS AND OUTPUT)
```
dcmiddle: please fill-in `todo` sections

docs/source/app_developers_guide/adding_a_node.rst
chenette: Thanks, Dan.  I'll add this info and fix the port typos that you noted below (plus other egregious typos asap).

chenette:
Pushed a fix to make Dan's corrections to the endpoint info & 8800 port numbers. This fix also reformats the the Docker container info for clarity, adds Docker & Ubuntu steps to set the transaction families on the first validator node, shows the Ubuntu start-component steps for validator node 2, and fixes several egregious typos.

docs/source/app_developers_guide/adding_a_node.rst
```diff
@@ -0,0 +1,432 @@
...
+After making the configuration changes, restart all validator nodes in the
+Sawtooth network.
+
+#. Shut down the first validator node as described in
+   :ref:`stop-sawtooth-ubuntu-label`.
+
+#. Shut down the second validator node as described in
+   :ref:`stop-sawtooth-ubuntu-label`.
+
+#. Start all Sawtooth components on the first validator node, as shown in this
```
peterschwarz: I think, for the initial validator, you only need to restart the validator process for the changes listed above to take effect.

docs/source/app_developers_guide/adding_a_node.rst
grkvlt: Also, there is a `/peers` endpoint, and CLI commands to list peers, they should be listed here

sofiter:
When I try the command 
`sawset proposal create \
> sawtooth.validator.transaction_families='[{"family": "intkey", "version": "1.0"},\
> {"family":"sawtooth_settings", "version":"1.0"}\
> {"family":"xo", "version":"1.0"}\
> {"family":"sawtooth_validator_registry", "version":"1.0"}]'

I get on both my VMs that I tried it on the erro
`Error: Unable to read key file: [Errno 2] No such file or directory: '/home/user/.sawtooth/keys/user.priv`

I have the keys in the correct directory and I used path.toml also just to be sure for the paths.

docs/source/app_developers_guide/adding_a_node.rst
chenette: Thanks, Andrew.  Question for all reviewers (or you):  How would someone use the /peers endpoint to verify peering?  Could someone provide an example for this step? (With output, if appropriate.)  Thanks in advance.

Also should I show both "sawnet peers list" and "sawtooth peer list"? 
- "sawnet peers list" displays the peers of the specified nodes
- "sawtooth peer list" displays the addresses of a specified validator’s peers

docs/source/app_developers_guide/adding_a_node.rst
chenette: Why wouldn't you start the REST API and transaction processors on the first validator node?  (The first command in this step shuts them all down.)

chenette:
I pushed a fix-up commit with reviewer corrections and related changes.

Note: I also changed the Ubuntu install procedure - I moved the user and root keygen commands to their own steps.  This should help clarify that both keygen steps are needed on the second validator node.

docs/source/app_developers_guide/adding_a_node.rst
```diff
@@ -0,0 +1,421 @@
...
+#. You can also use Sawtooth commands to show peer information.
+
+   * Use ``sawtooth peer list`` to list the peers of a particular node. For more
+     information, run ``sawtooth peer list --help``.
+
+   * Use ``sawnet peers list`` to display a complete graph of peers on the
+     network. For more information, run ``sawnet peers list --help``.
+
+#. Submit a transaction on the first validator node. This example sets the key
+   ???keyname??? to the value 999.
```
peterschwarz: keyname can be an string: e.g. "MyKey"

docs/source/app_developers_guide/adding_a_node.rst
```diff
@@ -0,0 +1,421 @@
...
+   .. code-block:: console
+
+      user@validator-1$ intkey set NEED-EXAMPLE-KEY-NAME 999
+
+#. On the second validator node, watch for this transaction to appear on the
+   blockchain.
+
+   .. code-block:: console
+
+      user@validator-2$ ??? NEED EXAMPLE COMMAND AND OUTPUT
```
peterschwarz: `intkey show --url <rest-api-url> <keyname>` will get you your example output

chenette:
Pushed another fix-up commit - this one adds "intkey set" and "intkey show" examples at the end. It also fixes minor formatting and wording issues.

docs/source/app_developers_guide/adding_a_node.rst
```diff
@@ -0,0 +1,443 @@
...
+         root@validator-0#
+
+   #. Run the following command to create and submit a batch containing the
+      new settings:
+
+      .. code-block:: console
+
+         root@validator-0# sawset proposal create --url http://rest-api:8008 \
+         --key /root/.sawtooth/keys/my_key.priv \
+         sawtooth.validator.transaction_families='[{"family": "intkey", "version": "1.0"},{"family":"sawtooth_settings", "version":"1.0"},{"family":"xo", "version":"1.0"},{"family":"sawtooth_validator_registry", "version":"1.0"}]'
```
dcmiddle: Sorry I missed this on the first go around. Do others feel this step is necessary? I think the validator will accept all txn families unless given an explicit list. Since these instructions are just for adding a second node I'd prefer to pare back anything that is optional.

docs/source/app_developers_guide/adding_a_node.rst
```diff
@@ -0,0 +1,443 @@
...
+
+#. You can connect to a validator registry container, such as
+   ``sawtooth-poet-validator-registry-tp-0``, then use the following command
+   to verify that ``poet-validator-registry-tp`` is running.
+
+   .. code-block:: console
+
+      # ps --pid 1 fw
```
dcmiddle: what is the expected output?

docs/source/app_developers_guide/adding_a_node.rst
```diff
@@ -0,0 +1,443 @@
...
+
+#. If you need to stop or reset the environment, enter CTRL-c in each
+   container to stop the Sawtooth component, then run the following command
```
dcmiddle: I think you can just do `docker-compose -f sawtooth-default-poet.yaml down`.  I don't think the ctrl-c is necessary and in fact sometimes it may not work.
Someone feel free to correct me if they have a different experience.

docs/source/app_developers_guide/adding_a_node.rst
```diff
@@ -0,0 +1,443 @@
...
+   described in :ref:`stop-sawtooth-ubuntu-label`.
+
+#. Open a new terminal window on the first validator node (the client-1 window)
+   and configure the Sawtooth settings to allow transactions from the Settings,
+   IntegerKey, XO, and Validator Registry transaction families. Run this
+   command:
+
+     .. code-block:: console
+
+        user@client-1$ sawset proposal create \
```
dcmiddle: as mentioned above, maybe we can cut this step out entirely

docs/source/app_developers_guide/adding_a_node.rst
```diff
@@ -0,0 +1,443 @@
...
+      file. For more information, see
+      :doc:`../sysadmin_guide/configuring_sawtooth/validator_configuration_file`.
+
+#. Edit ``/etc/sawtooth/validator.toml`` to change the following settings:
+
+   a. Set the ``bind`` directive to the network and component values for this
+      node.
+
+      * ``network`` specifies the validator-to-validator connection (by default,
+        ``127.0.0.1:8800``)
```
dcmiddle: @jsmitchell is there some reason we default this to loopback instead of 0.0.0.0?

docs/source/app_developers_guide/adding_a_node.rst
```diff
@@ -0,0 +1,443 @@
...
+      For this application development environment, use the default settings:
+
+        .. code-block:: none
+
+           bind = [
+             "network:tcp://127.0.0.1:8800",
+             "component:tcp://127.0.0.1:4004"
+           ]
+
+   #. Set ``endpoint`` to the publicly addressable IP and port of the
```
dcmiddle: I know I gave the words 'publicly addressable' but maybe 'routable' is a better word choice.

docs/source/app_developers_guide/adding_a_node.rst
```diff
@@ -0,0 +1,443 @@
...
+      validator-to-validator connection. By default, ``endpoint`` is the same as
+      the bind ``network`` setting: ``tcp://127.0.0.1:8800``. You must change
+      the ``endpoint`` value if the validator is hosted behind a firewall. In
```
dcmiddle: 127.0.0.1 will never work for endpoint. each computer sees that address as it's own loopback interface. I wonder if we can be more clear here? Also while I did mention you need to change endpoint for NAT'ing / Firewall if the `network` is bound to anything but a routable IP address (i.e. if it is bound to 0.0.0.0 or 127.0.0.1) endpoint needs to list the actual IP of the machine (or the IP that the firewall is advertising). This is a tricky thing to communicate without a ton of words. 
Consider just conveying that the endpoint is the routable address and suggest there's a special case for a firewalled environment and maybe we can foot note that?
Or if we can just default the 'network' to 0.0.0.0 then maybe we can skip that part of the description and just pare back to this setting ('endpoint') and the 'peers' list. 
We can have an advanced network guide in the future to help walk people through the difference between bind and endpoint.

docs/source/app_developers_guide/adding_a_node.rst
```diff
@@ -0,0 +1,443 @@
...
+      Configuration File" in the System Administration Guide. For more information
+      on these settings, run ``sawtooth-validator --help`` on a validator node.
+
+.. _install-second-val-ubuntu-label:
+
+Step 2: Install and Configure a Second Validator Node
+------------------------------------------------------
+
+#. Install and configure Sawtooth on the second node as described in
+   :doc:`ubuntu`, with one difference: **Do not** create a genesis block
```
dcmiddle: If you don't remove the sawset command above, you may also need to explain you don't do that here or explain up there that you are making a blockchain setting to global state that you only do once. (gets wordy which is why I'd rather not have it in if it's not necessary).

docs/source/app_developers_guide/adding_a_node.rst
```diff
@@ -0,0 +1,443 @@
...
+
+      .. code-block:: none
+
+         bind = [
+           "network:tcp://127.0.0.1:8800",
+           "component:tcp://127.0.0.1:4004"
+         ]
+
+   #. Set ``endpoint`` to the publicly addressable IP and port of the
+      validator-to-validator connection. By default, ``endpoint`` is the same as
```
dcmiddle: feedback as above for 'bind' and 'endpoint'

docs/source/app_developers_guide/adding_a_node.rst
chenette: The issue is that the "first" procedure restricts the set of txn families, so this step is necessary to add xo.  I think that the original proc had this step to show that restriction is possible. If that's not an important thing, I could remove the "sawset proposal create" step everywhere, which would simplify all the procedures. 

I agree that simplicity is better.  Opinons from others?

boydjohnson: Yeah, I would remove the sawset proposal for restricting transaction_families since we want to focus only on those steps that are necessary to add another node.

vaporos: I'm okay with simplifying it (not including restricting transaction families at all). We would want some page to document that feature though so we can refer to it. (Maybe we have that already in another location?)

docs/source/app_developers_guide/adding_a_node.rst
boydjohnson: CTRL-c in each container is definitely not the way to go. ctrl-c-ing the docker-compose command will stop each container and `docker-compose down` will reset the environment.

vaporos: I usually do CTRL-C in the 'docker-compose up' window and then run docker-compose down after that, which cleans things up.

docs/source/app_developers_guide/adding_a_node.rst
vaporos: Yes, 0.0.0.0 as a default did not pass security reviews.

docs/source/app_developers_guide/adding_a_node.rst
peterschwarz: See the above comment on why we can't use 0.0.0.0 as the default

docs/source/app_developers_guide/adding_a_node.rst
chenette: @vaporos - I can move the info about restricting txn families to the Transaction Families Specification. (I can find a spot in the intro text to add this information).  I suggest changing the example to show Settings, BlockInfo, and Identity (not intkey & xo), because that's a more useful/realistic example.

chenette:
Pushed another fix that substantially rewrites this section. The Ubuntu steps now have a substantial "Prerequisites" section that explain the bind and endpoint info. Also, the config file steps have been replaced with simpler command-line options. There's much more, but that's the high-level summary.

docs/source/app_developers_guide/creating_sawtooth_network.rst
```diff
@@ -0,0 +1,460 @@
...
+
+      $ sudo -u sawtooth sawtooth-poet-validator-registry-tp -v
+
+If you have additional nodes in the network, repeat this step on those nodes.
+
+
+Confirm Network Functionality
+=============================
+
+#. To check whether peering has occurred on the network, submit a block query
```
aludvik: typo I think. "block query" should be "peers query"?

docs/source/app_developers_guide/creating_sawtooth_network.rst
```diff
@@ -0,0 +1,460 @@
...
+This example output shows that ``eth0`` is a network interface that has access
+to the Internet. In this case, you could use one of the following:
+
+* If you would like the validator node to accept connections from other
+  validator nodes on the network behind ``eth0``, you could specify a
+  network bind string such as ``tcp://eth0:8800``.
+
+* If you would like the validator node to accept only connections from local
+  Sawtooth components locally, you could specify the component bind string
+  ``tcp://lo:4004``. Note that this is equivalent to ``tcp://127.0.0.1:4004``.
```
aludvik: The above sentence is awkward. I suggest, "If you would like the validator node to only accept local connections from Sawtooth components, you could specify the component bind string"

docs/source/app_developers_guide/creating_sawtooth_network.rst
chenette: Yes.  My bad.  I'll correct this in a separate "fix typos" PR very soon now.

docs/source/app_developers_guide/creating_sawtooth_network.rst
chenette: Ditto (will fix in a separate PR soon).

chenette:
@sofiter - Thanks for your comment - it prompted us to re-evaluate whether this complicated command belonged in this procedure.  As a result, we deleted the "sawnet proposal create" command from this procedure (and other procedures), because we felt there wasn't a need to limit the allowed transaction types in this application development environment. Note that information on this setting will move to the Transaction Family Specification in an upcoming PR, but isn't there yet. In the meantime, the Architecture chapter has some information on this setting.

But if you'd still like to run this command, I think there might be some missing commas in the command you show in your comment.  (Sorry - that was my error in the version that you commented on.) I think it should be something like this:

     ``sawset proposal create sawtooth.validator.transaction_families='[{"family": "intkey", "version": "1.0"},{"family":"sawtooth_settings", "version":"1.0"},{"family":"xo", "version":"1.0"},{"family":"sawtooth_validator_registry", "version":"1.0"}]'``

Also, my local expert suspects there might be a problem with your public/private keypair - he wonders if they got generated by root instead of by your username. (So they might be in the wrong location.) 

Anyway, if you'd like more help with this issue, please go to the #sawtooth channel on Rocket Chat; - the folks there can help you directly. https://chat.hyperledger.org/channel/sawtooth

PR #1537, opened 2018-03-24
Fix file permission error while starting up validator #505

yarasan:
Signed-off-by: yarasan <yarasan@users.noreply.github.com>

### DIFF
--- a/validator/sawtooth_validator/server/log.py
+++ b/validator/sawtooth_validator/server/log.py
@@ -19,6 +19,7 @@ import sys
 import os
 
 from colorlog import ColoredFormatter
+from sawtooth_validator.exceptions import LocalConfigurationError
 
 
 class LogWriter:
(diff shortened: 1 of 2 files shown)


validator/sawtooth_validator/server/log.py
```diff
@@ -74,18 +74,22 @@ def log_configuration(log_config=None, log_dir=None, name=None):
...
+            debug_handler.setLevel(logging.DEBUG)
+
+            error_handler = logging.FileHandler(log_filename + "-error.log")
+            error_handler.setFormatter(logging.Formatter(
+                '[%(asctime)s.%(msecs)03d [%(threadName)s] %(module)s'
+                ' %(levelname)s] %(message)s', "%H:%M:%S"))
+            error_handler.setLevel(logging.ERROR)
+        except IOError as e:
+            print("Error: {}".format(str(e)))
+            sys.exit(1)
```
aludvik: I don't think a `sys.exit(1)` is the correct behavior here. Take a look at where this function is used: https://github.com/hyperledger/sawtooth-core/blob/master/validator/sawtooth_validator/server/cli.py#L254. There is already a pattern for handling initialization errors, which would be better to follow. The functions in server/cli.py that depend on this should catch and handle the error, so that the same pattern used for other errors can be used for this one.

vaporos: Agree, sys.exit() should never be scattered through the code.

vaporos:
This needs a more descriptive commit message.

https://chris.beams.io/posts/git-commit/

peterschwarz:
You should squash these commits together.  Second on the better commit messages - it's part of our contributor's guide.

yarasan:
OK, I removed the older commit and the commit message is updated.

peterschwarz:
What is the `#505` in the commit message referring to?  Please remove

yarasan:
#505 refers to the Jira issue, appended to the commit message as described by the contributor's guide. Is there a problem with the format?

peterschwarz:
If you read the commit message guide, it should be in the body of the commit message, not the summary line. 

You will also need to rebase, as there have been some issues with the listing process that have been resolved since you initially submitted this PR.

validator/sawtooth_validator/server/log.py
peterschwarz: Should this throw a CliException or the like to be caught by the cli main?

peterschwarz:
Messages look good, but it doesn’t look like the `sys.exit` comments have been addressed.

yarasan:
Sys.exit was moved from log.py to cli.py, in response to earlier comments.

rberg2:
@yarasan Hello, if you can get this rebased I will get it merged in as soon as it builds, thanks!.

vaporos:
@yarasan Please rebase so that it will pass the build. If you need assistance please join rocketchat #sawtooth-core-dev.

PR #1538, opened 2018-03-26
Rewrite "Subscribing to Events" (was "Event Subscription")

chenette:
Reorganize, expand information, and clarify steps.

### DIFF
--- a/docs/source/app_developers_guide/about_event_subscriptions.rst
+++ b/docs/source/app_developers_guide/about_event_subscriptions.rst
@@ -1,0 +1,140 @@
+*************************
+About Event Subscriptions
+*************************
+
+An application can use event subscriptions to tell the validator to supply
+information about changes to the blockchain.
+
+An event subscription consists of an event type (the name of the event) and an
+optional list of filters for this event type. If an event is in one of your
+subscriptions, you will receive the event when the validator broadcasts it.
+
+An event is “in a subscription” if the event’s ``event_type`` field matches the
+subscription’s ``event_type`` field, and the event matches any filters that are
+in the event subscription. If multiple filters are included in a single
+subscription, only events that pass *all* filters will be received.
+Note, however, that you can have have multiple subscriptions at a time, so you
+will receive all events that pass the filter or filters in each subscription.
+
+
+Event Filters
+=============
+
+An event filter operates on event attributes. A filter specifies an attribute
+key, a "match string", and a filter type.
+
+  * The attribute key specifies the event attribute that you are interested in,
+    such as ``address`` or ``block_id``.
+
+  * The match string is used to compare against the event's attribute value, as
+    identified by the attribute key.
+
+  * The filter type (simple or regular expression) determines the rules for
+    comparing the match string to the event's attribute value.
+
+The filter type can either be ``SIMPLE`` (an exact match to the specified
+string) or ``REGEX`` (a match using a regular expression). In addition, the
+filter type specifies ``ANY`` (match one or more) or ``ALL`` (must match all).
+The following filter types are supported:
+
+* ``SIMPLE_ANY``:
+  The filter's "match string" must match at least one of the
+  event's attribute values. For example, this filter type with the match string
+  "``abc``" would succeed for a single event with the following attributes,
+  because it matches the attribute value ``abc``.
+
+  .. code-block:: none
+
+     Attribute(key="address",value="abc")
+     Attribute(key="address",value="def")
+
+* ``SIMPLE_ALL``:
+  The filter's "match string" must match all of the event's attribute values.
+  This filter type with the match string "``abc``" would fail with the previous
+  example, because it does not match all the attribute values.
+
+* ``REGEX_ANY``:
+  The filter's regular expression must evaluate to a match for at least one of
+  the event's attribute values. For example, this filter type with the match
+  string "``ab.``" would succeed for a single event with the following
+  attributes, because it matches the attribute value ``abc``.
+
+  .. code-block:: none
+
+     Attribute(key="address",value="abc")
+     Attribute(key="address",value="abbbc")
+     Attribute(key="address",value="def")
+
+* ``REGEX_ALL``:
+  The filter's regular expression must evaluate to a match for all of the
+  event's attribute values. This filter type with the match string "``ab.``"
+  would fail with the previous example, because it doesn't match all three
+  attribute values above. The match string ``[ad][be]*[cf]`` would succeed.
+
+
+Event Subscription Protobuf
+===========================
+
+Event subscriptions are submitted and serviced over a ZMQ socket using the
+validator's messaging protocol.
+
+An event subscription is represented with the following protobuf messages, which
+are defined in ``sawtooth-core/protos/events.proto``.
+
+.. code-block:: python
+
+    message EventSubscription {
+      string event_type = 1;
+      repeated EventFilter filters = 2;
+    }
+
+    message EventFilter {
+      string key = 1;
+      string match_string = 2;
+
+      enum FilterType {
+          FILTER_TYPE_UNSET = 0;
+          SIMPLE_ANY = 1;
+          SIMPLE_ALL = 2;
+          REGEX_ANY  = 3;
+          REGEX_ALL  = 4;
+        }
+        FilterType filter_type = 3;
+    }
+
+A ``ClientEventsSubscribeRequest`` envelope is used to submit subscription
+requests and receive the responses.
+
+.. code-block:: protobuf
+
+  message ClientEventsSubscribeRequest {
+      repeated EventSubscription subscriptions = 1;
+      // The block id (or ids, if trying to walk back a fork) the subscriber last
+      // received events on. It can be set to empty if it has not yet received the
+      // genesis block.
+      repeated string last_known_block_ids = 2;
+  }
+
+The validator responds with a ``ClientEventsSubscribeResponse`` message that
+specifies whether the subscription was successful.
+
+.. code-block:: protobuf
+
+  message ClientEventsSubscribeResponse {
+      enum Status {
+           OK = 0;
+           INVALID_FILTER = 1;
+           UNKNOWN_BLOCK = 2;
+      }
+      Status status = 1;
+      // Additional information about the response status
+      string response_message = 2;
+  }
+
+When subscribing to events, an application can optionally request "event
+catch-up" by sending a list of block IDs along with the subscription. For more
+information, see :ref:`event-catch-up-label`.
+
+
+.. Licensed under Creative Commons Attribution 4.0 International License
+.. https://creativecommons.org/licenses/by/4.0/
--- a/docs/source/app_developers_guide/about_events.rst
+++ b/docs/source/app_developers_guide/about_events.rst
@@ -1,0 +1,161 @@
+*********************
+About Sawtooth Events
+*********************
+
+Sawtooth events occur when blocks are committed --- that is, the validator
+broadcasts events when a commit operation succeeds --- and are not persisted in
+state. Each transaction family can define the events that are appropriate for
+its business logic.
+
+An event has three parts:
+
+* Event type (the name of the event)
+* Opaque payload whose structure is defined by the event type
+* List of attributes
+
+An attribute is a key-value pair that contains transparent metadata about the
+event. The key is the name of the attribute, and the value is the specific
+contents for that key. The same key can be used for multiple attributes in an
+event.
+
+It is important to define meaningful event attributes, so that the attributes
+can be used to filter event subscriptions. Note that although attributes are not
+required, an event filter cannot operate on an event without attributes. For
+more information, see :doc:`about_event_subscriptions`.
+
+Events are represented with the following protobuf message:
+
+.. code-block:: protobuf
+
+   message Event {
+     // Used to subscribe to events and servers as a hint for how to deserialize
+     // event_data and what pairs to expect in attributes.
+     string event_type = 1;
+
+     // Transparent data defined by the event_type.
+     message Attribute {
+       string key = 1;
+       string value = 2;
+     }
+     repeated Attribute attributes = 2;
+
+     // Opaque data defined by the event_type.
+     bytes  data = 3;
+   }
+
+The ``event_type`` field (the name of the event) is used to determine how opaque
+(application-specific) data has been serialized and what transparent event
+attributes to expect.
+
+For more information, see
+:doc:`/architecture/events_and_transactions_receipts`.
+
+
+Event Namespace
+===============
+
+By convention, event names use the transaction family as a prefix, such as
+``xo/create`` for a create event from the XO transaction family.
+
+Core Sawtooth events are prefixed with ``sawtooth/``. The core events are:
+
+  * ``sawtooth/block-commit``
+
+  * ``sawtooth/state-delta``
+
+sawtooth/block-commit
+---------------------
+
+A ``sawtooth/block-commit`` event occurs when a block is committed. This event
+contains information about the block, such as the block ID, block number, state
+root hash, and previous block ID. It has the following structure:
+
+.. code-block:: protobuf
+
+   Event {
+     event_type = "sawtooth/block-commit",
+     attributes = [
+       Attribute { key = "block_id", value = "abc...123" },
+       Attribute { key = "block_num", value = "523" },
+       Attribute { key = "state_root_hash", value = "def...456" },
+       Attribute { key = "previous_block_id", value = "acf...146" },
+     ],
+   }
+
+sawtooth/state-delta
+--------------------
+
+A ``sawtooth/state-delta`` occurs when a block is committed. This event contains
+all state changes that occurred at a given address for that block. This event
+has the following structure:
+
+.. code-block:: python
+
+     Event {
+       event_type = "sawtooth/state-delta",
+       attributes = [Attribute { key = "address", value = "abc...def" }],
+       event_data = <bytes>
+     }
+
+Note that the addresses that match the filter are in the attributes. Changed
+values are part of the event data.
+
+
+Example: An Application-specific Event
+======================================
+
+The XO transaction family could define an ``xo/create`` event that is sent when
+a game has been created. The following examples show a simple ``xo/create``
+event in several languages.
+
+Python example:
+
+   .. code-block:: python
+
+      context.add_event(
+          "xo/create", {
+              'name': name,
+              'creator': signer_public_key
+      })
+
+Go example:
+
+   .. code-block:: go
+
+    attributes := make([]processor.Attribute, 2)
+    attributes = append(attributes, processor.Attribute{
+      Key:   "name",
+      Value: name,
+    })
+    attributes = append(attributes, processor.Attribute(
+      Key:   "creator",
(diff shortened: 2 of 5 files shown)


docs/source/app_developers_guide/about_events.rst
```diff
@@ -0,0 +1,123 @@
+*********************
+About Sawtooth Events
+*********************
+
+Sawtooth events occur "on block boundaries" - that is, when blocks are
```
aludvik: Putting quotes around ""on block boundaries"" seems makes it seem like you are being intentionally vague. If that language is vague or confusing, just say "Sawtooth events occur when blocks are committed". Also, "occur" feels like the wrong word here. I prefer "are broadcasted" or something similar to indicate that this is when you will receive the event, not necessarily when it happened, although they are usually the same or closely linked.

docs/source/app_developers_guide/about_events.rst
```diff
@@ -0,0 +1,123 @@
...
+committed - and are not persisted in state. Each transaction family can define
+the events that are appropriate for its business logic.
+
+Events are represented with the following protobuf message:
+
+.. code-block:: python
```
aludvik: This isn't actually a python code block and so the comments look weird. It is probably fine to use python since the highlighting looks mostly accurate, but maybe try changing the comments to `#` so that the highlighting is more accurate.

docs/source/app_developers_guide/about_events.rst
```diff
@@ -0,0 +1,123 @@
...
+       Attribute { key = "state_root_hash", value = "def...456" },
+       Attribute { key = "previous_block_id", value = "acf...146" },
+     ],
+   }
+
+sawtooth/state-delta
+--------------------
+
+A ``sawtooth/state-delta`` occurs when a block is committed. This event
+contains a single change in state at a given address. It has the following
```
aludvik: This may have changed, but I remember this returned all state changes that occurred in this block, not a single change.

docs/source/app_developers_guide/about_events.rst
```diff
@@ -0,0 +1,123 @@
...
+
+   .. code-block:: javascript
+
+      NEED HELP
+
+Rust example:
+
+   .. code-block:: rust
+
+      NEED HELP
```
aludvik: Need to fill these in.

docs/source/app_developers_guide/about_event_subscriptions.rst
```diff
@@ -0,0 +1,137 @@
+**************************
+About Events Subscriptions
```
aludvik: typo "Event"

docs/source/app_developers_guide/about_event_subscriptions.rst
```diff
@@ -0,0 +1,137 @@
...
+**************************
+
```
aludvik: I would add a little motivation here to answer "Why do I care what an event subscription is?" Something like "Event subscriptions are how you tell a validator what information you would like to be notified about."

docs/source/app_developers_guide/about_event_subscriptions.rst
```diff
@@ -0,0 +1,137 @@
...
+An event subscription consists of an event type (the name of the event) and an
+optional list of filters for this event type.
+
+An event is “in a subscription” if the event’s ``event_type`` field matches the
```
aludvik: I would add  something like "If an event is in one of your subscriptions, you will receive the event when it is broadcast." before this.

docs/source/app_developers_guide/about_event_subscriptions.rst
```diff
@@ -0,0 +1,137 @@
...
+subscription’s ``event_type`` field, and the event matches any filters that are
+in the event subscription. If multiple filters are included in a subscription,
+only events that pass *all* filters will be received.
```
aludvik: Suggest adding that "However, you may have multiple subscriptions at a time, and will receive all events in any of these subscriptions."

docs/source/app_developers_guide/about_events.rst
```diff
@@ -0,0 +1,123 @@
...
+     // Used to subscribe to events and servers as a hint for how to deserialize
+     // event_data and what pairs to expect in attributes.
+     string event_type = 1;
+
+     // Transparent data defined by the event_type.
+     message Attribute {
+       string key = 1;
+       string value = 2;
+     }
+     repeated Attribute attributes = 2;
```
aludvik: Attributes are important to understand when subscribing to events. I know you are linking to the architecture section which talks about them, but you can't really understand filters without understanding attributes. Maybe add some high-level introduction here or in subscriptions?

docs/source/app_developers_guide/about_event_subscriptions.rst
```diff
@@ -0,0 +1,137 @@
...
+following contents:
+
+  * A string describing the filter (the filter name)
+
+  * The filter type (simple or regular expression) and match type (any or all)
+
+An event filter consists of a key and a "match string". Filters apply only to
+the event attributes that have the same key as the filter. The filter's match
+string is compared against the attribute's value according to the type of
+event filter.
```
aludvik: This is confusing because it says that a filter is composed of two parts twice, and the two parts are different between the two times they are described.

Filters operate on event attributes and are actually composed of three parts:
- The "key" of the attribute that the filter applies to
- The "match string" that the value of the attribute is compared against
- The "filter type" which is used to decide the rules for comparison against the match string and the attribute value.

docs/source/app_developers_guide/about_event_subscriptions.rst
```diff
@@ -0,0 +1,137 @@
...
+* ``SIMPLE_ALL``:
+   The filter's "match string" must match all of the event's
+   attribute values. This filter type with the match string "``abc``"
+   would fail with the previous example, because it does not match all the
+   attribute values.
+
+* ``REGEX_ANY``:
+   The filter's regular expression must evaluate to a match for
+   one or more of the event's attribute values. For example, this filter type
+   with the match string "``[ad][be][cf]``" would succeed for an event with the
```
aludvik: typo? "``[ad][be]*[cf]``"

docs/source/app_developers_guide/about_event_subscriptions.rst
```diff
@@ -0,0 +1,137 @@
...
+   "``[ad][be][cf]``" would fail with the previous example, because it doesn't
+   match all three attribute values. The match string ``[ad][be]*[cf]`` would
+   succeed.
+
+
+Event Subscription Protobuf
+===========================
+
+Event subscriptions are submitted and serviced over a ZMQ socket as messages
+using the protocol buffers format.
```
aludvik: I would say "over a ZMQ socket using the Validator's messaging protocol" since it is more complex than just protobuf

docs/source/app_developers_guide/about_event_subscriptions.rst
```diff
@@ -0,0 +1,137 @@
...
+          SIMPLE_ANY = 1;
+          SIMPLE_ALL = 2;
+          REGEX_ANY  = 3;
+          REGEX_ALL  = 4;
+        }
+        FilterType filter_type = 3;
+    }
+
+A ``ClientEventsSubscribeRequest`` envelope is used to submit subscription
+request and receive the responses.
```
aludvik: typo "requests"

docs/source/app_developers_guide/about_event_subscriptions.rst
```diff
@@ -0,0 +1,137 @@
...
+
+  message ClientEventsSubscribeRequest {
+      repeated EventSubscription subscriptions = 1;
+      // The block id (or ids, if trying to walk back a fork) the subscriber last
+      // received events on. It can be set to empty if it has not yet received the
+      // genesis block.
+      repeated string last_known_block_ids = 2;
+  }
+
+The validator responds with a ``ClientEventSubscribeResponse`` message that
```
aludvik: typo "ClientEventsSubscribeResponse" (I know that 's' is terrible but it is more accurate.)

aludvik:
Made it through the first two sections. Will take a look at the rest in a bit.

docs/source/app_developers_guide/zmq_event_subscription.rst
```diff
@@ -0,0 +1,309 @@
+********************************
+Using ZMQ to Subscribe to Events
+********************************
+
+Client applications can subscribe to Hyperledger Sawtooth events using
+`ZMQ <http://zeromq.org>`_ and protobuf messages. The general subscription
+process is as follows:
+
+#. Construct a subscription that includes the event type and an optional
```
aludvik: suggest "and 0 or more filters."

docs/source/app_developers_guide/zmq_event_subscription.rst
```diff
@@ -0,0 +1,309 @@
...
+
+The following steps assume that the XO transaction family has a ``create`` event
+that is sent when a game has been created, as in this example:
+
+.. code-block:: python
+
+     context.add_event(
+         "xo/create", {
+             'name': name,
+             'creator', signer_public_key
```
aludvik: typo, comma should be colon

docs/source/app_developers_guide/zmq_event_subscription.rst
```diff
@@ -0,0 +1,309 @@
...
+
+.. code-block:: python
+
+    subscription = EventSubscription(
+        event_type="sawtooth/state-delta",
+        filters=[
+            # Filter to only addresses in the "xo" namespace using a regex
+            EventFilter(
+                key="address",
+                match_string="xo.*",
```
aludvik: This is not the xo namespace. The xo namespace is `hashlib.sha512('xo'.encode("utf-8")).hexdigest()[0:6]` which comes out as `5b7349`, so the value here should be `match_string="5b7349.*"`

docs/source/app_developers_guide/zmq_event_subscription.rst
```diff
@@ -0,0 +1,309 @@
...
+    # Receive the response
+    resp = socket.recv_multipart()[-1]
+
+    # Parse the message wrapper
+    msg = Message()
+    msg.ParseFromString(resp)
+
+    # Validate the response type
+    if msg.message_type != CLIENT_EVENTS_SUBSCRIBE_RESPONSE:
+        print("Unexpected message type")
```
aludvik: This should probably also have a `return`

docs/source/app_developers_guide/zmq_event_subscription.rst
```diff
@@ -0,0 +1,309 @@
...
+
+    # Parse the response
+    response = ClientEventsSubscribeResponse()
+    response.ParseFromString(msg.content)
+
+    # Validate the response status
+    if response.status != ClientEventsSubscribeResponse.OK:
+      print("Subscription failed: {}".format(response.response_message))
```
aludvik: This should probably also have a `return`

docs/source/app_developers_guide/zmq_event_subscription.rst
```diff
@@ -0,0 +1,309 @@
...
+An event originates from a specific block. That is, an event is sent to the
+subscriber only when the block is committed and state is updated. As a result,
+events can be treated as output from processing and committing blocks.
+
+An application can subscribe to both ``sawtooth/block-commit`` and
+``sawtooth/state-delta`` events to match state changes with the block
+in which the changes occurred.
+
+All lists of ``block-commit`` events received from the validator will contain a
+``block-commit`` event for the block that the events came from.
```
aludvik: Is there a typo here? I don't know how to interpret this sentence.

docs/source/app_developers_guide/zmq_event_subscription.rst
```diff
@@ -0,0 +1,309 @@
...
+
+.. important::
+
+   For forking networks, we recommend subscribing to ``block-commit`` events
+   in order to watch for network forks and react appropriately. Without
+   a subscription to ``block-commit`` events, there is no way to determine
+   whether a fork has occurred.
```
aludvik: Suggest adding, "It may also be a good idea to only react to events after some number of blocks have been committed on the given fork, when there is some confidence that you are not going to have to revert those changes because you switched to a different fork."

docs/source/app_developers_guide/zmq_event_subscription.rst
```diff
@@ -0,0 +1,309 @@
...
+
+This example submits an unsubscribe request.
+
+.. code-block:: python
+
+    # Construct the request
+    request = ClientEventsUnsubscribeRequest()
+
+    # Construct the message wrapper
+    correlation_id = "123" # This must be unique for all "in-flight requests
```
aludvik: typo, rogue double-quote

docs/source/app_developers_guide/zmq_event_subscription.rst
peterschwarz: There's no typo, it's describing the contents of a block commit event. It might need a "contain only a single ``block-commit`` event" as part of its sentence.

docs/source/app_developers_guide/about_event_subscriptions.rst
chenette: Nope.  I was trying to show a REGEX filter that would match some but not all of the following attribute values.  The * makes it match them all, which is not what I want.

Given my info, is there still a typo in my regex?  I'm a bit of a regex newbie.

chenette: (Never mind - I deleted my previous comment because you're right and I'm wrong.)

chenette: The end of the following sentence is where the error is.  It should say '...because it matches the attributes "abc" and "def".'

docs/source/app_developers_guide/about_events.rst
chenette: @aludvik - The architecture section doesn't explain attributes (if it does, I couldn't find the information).  Could you provide a short, high-level summary for this document?  Many thanks!

docs/source/app_developers_guide/zmq_event_subscription.rst
chenette: To a tech writer, "return" means a carriage return (aka new line or line break). I'm guessing that's not what you mean.  Could you supply an example "return" for this so-not-a-Python-programmer?  Thanks!

docs/source/app_developers_guide/zmq_event_subscription.rst
chenette: Same comment as above.  Thanks again!

docs/source/app_developers_guide/zmq_event_subscription.rst
chenette: I propose changing this sentence to:

 "All lists of ``block-commit`` events received from the validator will contain only a single ``block-commit`` event for the block that the events came from."

Does that sound correct?

chenette: Unless this is supposed to mean "no matter how many events were generated when a block is committed, each block-commit event received from the validator is a single event"?  Because I don't know how to interpret this sentence, either.  I'm confused by "All lists of block-commit events" - if a list always contains only a single event, it's not really list.

chenette:
I pushed a fix-up commit that makes reviewer corrections and adds Go/JavaScript/Rust code examples for the xo/create event.

docs/source/app_developers_guide/zmq_event_subscription.rst
chenette: Done.

docs/source/app_developers_guide/zmq_event_subscription.rst
aludvik: Lol, it's literally just a return. Sorry for the confusion.
+    if msg.message_type != CLIENT_EVENTS_SUBSCRIBE_RESPONSE:
+        print("Unexpected message type")
+        return

docs/source/app_developers_guide/zmq_event_subscription.rst
aludvik:
```
+    if response.status != ClientEventsSubscribeResponse.OK:
+      print("Subscription failed: {}".format(response.response_message))
+      return
```

docs/source/app_developers_guide/about_event_subscriptions.rst
aludvik: There is still a typo in the regex. `[]` is "Used to indicate a set of characters" that you are accepting. `*` is used to indicate you want to accept 0 or more instances of some set of characters.

What you have says "accept any 3 character string where the first character is a or d, the second character is b or e, and the third character is c or f". What I think you want is "accept any string where the first character is a or d, the last character is c or f, and there are 0 or more characters in the middle that are either b or e". This is `[ad][be]*[cf]`.

Here is a little script that compares the two:
```
import re

def test_str(r, s):
  if r.match(s) is None:
    print("%s didn't match" % s)
  else:
    print("%s matched" % s)

def test_all(match_string):
  print("testing %s" % match_string)
  tests = [
    "abc",
    "abbbc",
    "def",
    "ac",
  ]

  r = re.compile(match_string)
  for t in tests:
    test_str(r, t)

test_all("[ad][be][cf]")
test_all("[ad][be]*[cf]")
```

```
~ » python3 testre.py
testing [ad][be][cf]
abc matched
abbbc didn't match
def matched
ac didn't match
testing [ad][be]*[cf]
abc matched
abbbc matched
def matched
ac matched
```

docs/source/app_developers_guide/about_events.rst
aludvik: ugh, you are right. Something like this:

An event is composed of three parts: the event type, which is the name of the event, an opaque payload that has a structure defined by the event type, and a list of attributes. An attribute is a key, value pair that contains some transparent meta data about the event. The key is the name of the attribute and the value is some specific value for that key. There can be many attributes with the same key in one event. Defining meaningful attributes on events is important because they can be used to filter event subscriptions.

docs/source/app_developers_guide/about_event_subscriptions.rst
aludvik: Gah, I misunderstood what you were trying to say here. You were using the right regex for what you were trying to do.

chenette: I've simplified the first regex expression to make my intent for the examples clearer.  I changed it to "ab." so that it matches only "abc" but not "abbbc" or "def".  Then, in the REGEX_ALL blurb, I said that "ab." fails but "[ad][be]*[cf]" passes.

chenette:
Pushed a fix to describe event attributes and add "return" lines to example code. I believe (hope) that this version addresses all previous reviewer comments.

docs/source/app_developers_guide/zmq_event_subscription.rst
chenette: Done

docs/source/app_developers_guide/zmq_event_subscription.rst
chenette: Done

chenette:
Pushed another fix-up commit - this one adds "intkey set" and "intkey show" examples at the end. It also fixes minor formatting and wording issues.

chenette:
Deleted previous comment (it was for a different PR). Sorry.

PR #1539, opened 2018-03-26
Add setting transaction dependencies in smallbank-workload tool

boydjohnson:


perf/smallbank_workload/src/smallbank_tranformer.rs
```diff
@@ -28,20 +29,66 @@ use sawtooth_sdk::signing;
...
     signer: &'a signing::Signer<'a>,
+    signature_by_customer_id: HashMap<u32, String>,
 }
 
 impl<'a> SBPayloadTransformer<'a> {
     pub fn new(signer: &'a signing::Signer) -> Self {
-        SBPayloadTransformer { signer: signer }
+        SBPayloadTransformer {
+            signer: signer,
+            signature_by_customer_id: HashMap::new(),
```
aludvik: Can you give this a more meaningful name? Something that gives a hint about why you want to track this mapping?

boydjohnson:
I'm going to write some unit tests for this.

perf/smallbank_workload/src/smallbank_tranformer.rs
```diff
@@ -28,20 +29,66 @@ use sawtooth_sdk::signing;
...
+        }
+    }
+
+    fn add_signature_if_create_account(
+        &mut self,
+        payload: &SmallbankTransactionPayload,
+        signature: String,
+    ) {
```
aludvik: It feels like this function is too specific/knows too much about the types being passed. I think something like `track_customer_signature(customer, signature)` would be clearer, and then the function using this can handle checking whether it is a create account and getting the customer id.

perf/smallbank_workload/src/smallbank_tranformer.rs
```diff
@@ -28,20 +29,66 @@ use sawtooth_sdk::signing;
...
+        if payload.get_payload_type() == SmallbankTransactionPayload_PayloadType::CREATE_ACCOUNT {
+            self.signature_by_customer_id
+                .insert(payload.get_create_account().get_customer_id(), signature);
+        }
+    }
+
+    fn get_all_signatures(&self, customer_ids: Vec<u32>) -> Vec<String> {
```
aludvik: I think this function name is also confusing, since it isn't actually getting all signatures and it isn't clear from the name which signatures you are actually getting. Maybe something like `get_signatures_of_customers()`?

perf/smallbank_workload/src/smallbank_tranformer.rs
aludvik: I think this map should actually be encapsulated by a separate object which exposes a simple interface for tracking dependencies for different customers.

aludvik:
Per side-band conversation, it probably makes sense to wrap that map in a struct that abstracts the dependency tracking. That struct should also be reusable across different workloads (intkey and smallbank) since it is a common thing to do. The methods should be something like `add_dependency(key, signature)`, `get_dependencies(keys)`, and `remove_dependency(key)`.

perf/smallbank_workload/src/smallbank_tranformer.rs
```diff
@@ -80,3 +143,30 @@ impl<'a> SBPayloadTransformer<'a> {
...
+where
+    T: Eq + Hash,
+{
+    pub fn new() -> DependencyTracker<T> {
+        DependencyTracker {
+            signature_by_id: HashMap::new(),
+        }
+    }
+
+    pub fn get_dependency_for_id(&self, id: &T) -> Option<&String> {
```
aludvik: Just call this `get_dependency`

perf/smallbank_workload/src/smallbank_tranformer.rs
```diff
@@ -80,3 +143,30 @@ impl<'a> SBPayloadTransformer<'a> {
...
+        self.signature_by_id.get(id)
+    }
+
+    pub fn add_signature_for_id(&mut self, id: T, signature: String) {
```
aludvik: Call this `add_dependency()`

perf/smallbank_workload/src/smallbank_tranformer.rs
```diff
@@ -28,20 +31,76 @@ use sawtooth_sdk::signing;
...
+
+    fn add_signature_if_create_account(
+        &mut self,
+        payload: &SmallbankTransactionPayload,
+        signature: String,
+    ) {
+        if payload.get_payload_type() == SmallbankTransactionPayload_PayloadType::CREATE_ACCOUNT {
+            self.dependencies
+                .add_signature_for_id(payload.get_create_account().get_customer_id(), signature);
+        }
```
aludvik: It doesn't seem like this needs to be a separate function. It just adds an extra layer of indirection.

perf/smallbank_workload/src/smallbank_tranformer.rs
boydjohnson: Makes sense, since the id is in the signature.

perf/smallbank_workload/src/smallbank_tranformer.rs
boydjohnson: So, as I see it,
```
# 1
add_signature(..., id: T, signature: String)
get_dependency(..., id: T)

#2
add_signature(..., id: T, signature: String)
get_signature(..., id: T)

#3
add_dependency(..., id: T, signature: String)
get_dependency(..., id: T)
```
I think I like #2 the most as it introduces the least amount of indirection. Although the case for #1 is that when it is `add_signature` the signature is actually the signature for the payload that introduces `id` and `get_dependency` gets the dependency for any payload that uses `id`. I'm less clear on the advantages of #3.

perf/smallbank_workload/src/smallbank_tranformer.rs
boydjohnson: It names the piece of code with a description of what it does. Without it there would be no explanation for what that piece of code did. There are certainly pieces of code here that don't have a name, and could, but it is moderately helpful to provide a name for this since it isn't directly useful for making this transaction.

perf/smallbank_workload/src/smallbank_tranformer.rs
aludvik: 3 makes the most sense to me. I think it is preferable over 1, because there is strong conceptual link between the two methods. I think it is preferable over 2, because you are tracking dependencies, not signatures. In our architecture, dependencies are specified by giving the transaction signature, but we could easily be using something else (say for example, if we used a hash of the payload as the transaction id instead of the signature).

Can you elaborate on your definition of indirection? I define indirection to mean roughly "having code in a separate function/method/class". Typically, this is for the purposes of maintaining a consistent level of abstraction in different portions of the code and to allow reuse of code. With that definition, I don't see how any of the above options have different levels of indirection, but I am interested to hear what your definition is.

perf/smallbank_workload/src/smallbank_tranformer.rs
aludvik: Hmm, that is a a fair point. This may just be a matter of taste. _In my opinion_, the if block is straightforward. Since the code doesn't need to be reused, I see moving the code behind a function as making the code more difficult to understand because you have to track down the function. I think a comment would be just as helpful to clarify what is going on.

```
+.       // Adds a dependency for the customer id if this is a create account transaction
+        if payload.get_payload_type() == SmallbankTransactionPayload_PayloadType::CREATE_ACCOUNT {
+            self.dependencies
+                .add_dependency(payload.get_create_account().get_customer_id(), signature);
+        }
```

That being said, looking at my reasons for indirection below, this function does allow you to maintain the same level of abstraction in the function where it is used.

Also, with whatever we decide on for the dependency tracker method names, this function should follow suit.

perf/smallbank_workload/src/smallbank_tranformer.rs
boydjohnson: I guess I don't have a great definition of indirection, so I won't try to make one up. What I was trying to say was that using `_dependency` adds a layer of meaning that might not always be correct, whereas `get_signature` and `add_signature` is the plain meaning of those methods.

I am fine having pretty bedrock architectural decisions of Sawtooth affect the nature of a piece  of code that creates Sawtooth Transacions.

perf/smallbank_workload/src/smallbank_tranformer.rs
boydjohnson: I think I am going to call the DependencyTracker, SignatureTracker but have the instance still be dependencies. So other uses on the Transformer class will refer to dependency.

boydjohnson:
@aludvik Thanks for the stimulating discussion. I will be reading `Code Complete` and `Refactoring`. I don't think I can be persuaded that more changes are needed.

PR #1545, opened 2018-03-29
Update Policy.Type to EntryType in identity cli

askmish:


askmish:
@vaporos I'm aware of that.
btw, updated commit message.
And here's more context:
This is not a typo. This change was left out during 1.0 release changes in identity.proto.

vaporos:
For the commit message, how about:

'Update Policy.Type to EntryType in identity cli

This change was missed during API stabilization change prior to 1.0.  Fixes
an exception which was being thrown when running XXX.
'

Where XXX is the CLI command that was failing.

PR #1547, opened 2018-03-29
Fix default bind port in rest-api cli help text 

askmish:


chenette:
@vaporos - FYI, see https://jira.hyperledger.org/browse/STL-1119. Thanks, @askmish, for fixing this problem!

PR #1551, opened 2018-03-30
Fix section titles and organization in App Dev Guide

chenette:


peterschwarz:
Need to resolve conflicts

PR #1553, opened 2018-03-30
Fix validator peering on Azure bug

arthurgreef:
Validator peering fails when using Azure endpoints such as
tcp://hashblock3.westus2.cloudapp.azure.com:8800 because the 'lo'
network interface characters are characters in 'cloudapp'. This
fix test that the network interface string is not equal to the

### DIFF
--- a/docker/compose/sawtooth-default-poet.yaml
+++ b/docker/compose/sawtooth-default-poet.yaml
@@ -27,9 +27,6 @@ services:
   validator-0:
     image: hyperledger/sawtooth-validator:1.0
     container_name: sawtooth-validator-default-0
-    volumes:
-      - ../poet-settings.sh:/etc/sawtooth/config/poet-settings.sh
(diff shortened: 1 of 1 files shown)


validator/sawtooth_validator/networking/handlers.py
```diff
@@ -82,21 +83,31 @@ def handle(self, connection_id, message_content):
...
-                             message.endpoint)
-
-                connection_response = ConnectionResponse(
-                    status=ConnectionResponse.ERROR)
-                return HandlerResult(
-                    HandlerStatus.RETURN_AND_CLOSE,
-                    message_out=connection_response,
-                    message_type=validator_pb2.Message.
-                    AUTHORIZATION_CONNECTION_RESPONSE)
+        match = re.search(r'\/\/(.*)\:', message.endpoint, re.I)
```
vaporos: It might be good to break the into a function we could write a few unit tests against (as to not get regressions).

If we did that, we might have "if not is_valid_endpoint(....):" instead of the "if error is True" below.

rberg2:
Hello.
We have reworked the permissions on our build server, if you rebase, this commit should build.

PR #1554, opened 2018-04-01
Fix for ZMQ send failures in REST API

kulkarniamol:
Added a simple exponential backoff & retry mechanism to handle ZMQ send failures.

### DIFF
--- a/rest_api/sawtooth_rest_api/messaging.py
+++ b/rest_api/sawtooth_rest_api/messaging.py
@@ -25,8 +25,36 @@ from zmq.asyncio import Context
 
 from sawtooth_rest_api.protobuf.validator_pb2 import Message
 
-
 LOGGER = logging.getLogger(__name__)
+
+
+class _Backoff:
+    """Implements a simple backoff mechanism.
+    """
+
+    def __init__(self, max_retries=3, interval=100, error=Exception()):
+        self.num_retries = 0
+        self.max_retries = max_retries
+        self.interval = interval
+        self.error = error
+
+    async def do_backoff(self, err_msg=" "):
+        if self.num_retries == self.max_retries:
+            LOGGER.warning("Failed sending message to the Validator. No more "
+                           "retries left. Backoff terminated: %s",
+                           err_msg)
+            raise self.error
+
+        self.num_retries += 1
+        LOGGER.warning("Sleeping for %s ms after failed attempt %s of %s to "
+                       "send message to the Validator: %s",
+                       str(self.num_retries),
+                       str(self.max_retries),
+                       str(self.interval / 1000),
+                       err_msg)
+
+        await asyncio.sleep(self.interval / 1000)
+        self.interval *= 2
 
 
 class _MessageRouter:
@@ -161,10 +189,22 @@ class _Sender:
             content=message_content,
             message_type=message_type)
 
-        try:
-            await self._socket.send_multipart([message.SerializeToString()])
-        except asyncio.CancelledError:
-            raise
+        # Send the message. Backoff and retry in case of an error
+        # We want a short backoff and retry attempt, so use the defaults
+        # of 3 retries with 200ms of backoff
+        backoff = _Backoff(max_retries=3,
+                           interval=200,
+                           error=SendBackoffTimeoutError())
+
+        while True:
+            try:
+                await self._socket.send_multipart(
+                    [message.SerializeToString()])
+                break
+            except asyncio.CancelledError:
+                raise
+            except zmq.error.Again as e:
+                await backoff.do_backoff(err_msg=repr(e))
 
         return await self._msg_router.await_reply(correlation_id,
                                                   timeout=timeout)
@@ -176,6 +216,14 @@ class DisconnectError(Exception):
 
     def __init__(self):
         super().__init__("The connection was lost")
+
+
+class SendBackoffTimeoutError(Exception):
+    """Raised when the send times out.
+    """
+
+    def __init__(self):
+        super().__init__("Timed out sending message over ZMQ")
 
 
 class ConnectionEvent(Enum):
--- a/rest_api/sawtooth_rest_api/exceptions.py
+++ b/rest_api/sawtooth_rest_api/exceptions.py
@@ -97,6 +97,14 @@ class ValidatorDisconnected(_ApiError):
                'Try your request again later.')
 
 
+class SendBackoffTimeout(_ApiError):
+    api_code = 19
+    status_code = 408
+    title = 'Send timed out'
+    message = ('Sending message to validator timed out. Retry limit reached. '
+               'Try your request again later.')
+
+
 class ValidatorResponseInvalid(_ApiError):
     api_code = 20
     status_code = 500
--- a/rest_api/sawtooth_rest_api/route_handlers.py
+++ b/rest_api/sawtooth_rest_api/route_handlers.py
@@ -30,6 +30,7 @@ from sawtooth_rest_api.protobuf.validator_pb2 import Message
 import sawtooth_rest_api.exceptions as errors
 import sawtooth_rest_api.error_handlers as error_handlers
 from sawtooth_rest_api.messaging import DisconnectError
+from sawtooth_rest_api.messaging import SendBackoffTimeoutError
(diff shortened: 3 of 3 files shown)


rest_api/sawtooth_rest_api/messaging.py
```diff
@@ -25,10 +26,28 @@
...
+
+    def __init__(self):
+        self.num_retries_left = 3
+        self.min_interval = 0.1
+        self.max_interval = 1.0
+        self.interval = self.min_interval
+
+    async def do_backoff(self):
+        if not self.num_retries_left:
+            LOGGER.warning("No more retries left. Backoff terminated...")
```
vaporos: This looks like a duplicate of the other logging message below which starts with "Retry limit reached". Suggest removing this one and doing all logging outside this class.

rest_api/sawtooth_rest_api/messaging.py
```diff
@@ -25,10 +26,28 @@
...
+            raise DisconnectError()
+
+        self.interval = random.randrange(self.interval, self.max_interval)
```
vaporos: There is probably no need for a random interval here, as the goal of this back off is to avoid bombarding the validator with requests, not to avoid submitting at the same time as another component.

rest_api/sawtooth_rest_api/messaging.py
```diff
@@ -25,10 +26,28 @@
...
-
 LOGGER = logging.getLogger(__name__)
 
 
+class _Backoff:
+    """Implements a simple exponential backoff mechanism.
+    """
+
+    def __init__(self):
+        self.num_retries_left = 3
```
vaporos: Is this decremented somewhere?

rest_api/sawtooth_rest_api/messaging.py
```diff
@@ -161,10 +180,21 @@ def __init__(self, socket, msg_router):
...
+        # Send the message. Backoff and retry in case of an error
+        backoff = _Backoff()
+
+        while True:
+            try:
+                await self._socket.send_multipart(
+                    [message.SerializeToString()])
+                break
+            except asyncio.CancelledError:
+                LOGGER.warning('ZMQ connection dropped. Backing off.')
```
vaporos: This should at minimum contain the backoff attempt and waiting period in the logging message.

Is it factual that "asyncio.CancelledError" is always equivalent to a ZMQ connection being dropped? If not (which seems possible), this would be a misleading logging statement in those other circumstances.

rest_api/sawtooth_rest_api/messaging.py
```diff
@@ -161,10 +180,21 @@ def __init__(self, socket, msg_router):
...
+                try:
+                    await backoff.do_backoff()
+                except DisconnectError:
+                    LOGGER.warning('Retry limit reached for ZMQ. Exiting.')
```
vaporos: This information should be logged on the same line above. Really there should only be one logging statement per backoff event.

This is also misleading, as this condition was not generated by ZMQ itself, though it might be interpreted as such when viewing it in the log.

rest_api/sawtooth_rest_api/messaging.py
kulkarniamol: Looking at the comment below to include the retry number and backoff time, I think it'll be better to move all logging into this class and remove the logs outside.

rest_api/sawtooth_rest_api/messaging.py
kulkarniamol: The interval increases with each retry. The interval value is selected from range that goes from the previous interval value to the max. We keep backing off for slightly longer periods with each retry.

rest_api/sawtooth_rest_api/messaging.py
kulkarniamol: I'll remove the log messages from here. That should address all these concerns.

rest_api/sawtooth_rest_api/messaging.py
peterschwarz: This value should be a parameter, as well.  It can be a keyword param, with a default value.

rest_api/sawtooth_rest_api/messaging.py
```diff
@@ -25,10 +26,28 @@
...
+    def __init__(self):
+        self.num_retries_left = 3
+        self.min_interval = 0.1
+        self.max_interval = 1.0
+        self.interval = self.min_interval
+
+    async def do_backoff(self):
+        if not self.num_retries_left:
+            LOGGER.warning("No more retries left. Backoff terminated...")
+            raise DisconnectError()
```
peterschwarz: It's not likely that it's disconnected, here. We get notified of disconnection via the monitor socket.  This should be a timeout exception, which should effect the resulting http status - probably an HTTP status 408

rest_api/sawtooth_rest_api/messaging.py
peterschwarz: Why not just use a predictable increase, like `* 2`?

rest_api/sawtooth_rest_api/messaging.py
peterschwarz: It indicates that the loop has been shutdown, not disconnect.

rest_api/sawtooth_rest_api/messaging.py
peterschwarz: This won't ultimately report the error back to the user, via an HTTP error.

kulkarniamol:
Updated based on review comments. Please review again.

rest_api/sawtooth_rest_api/messaging.py
```diff
@@ -161,10 +182,21 @@ def __init__(self, socket, msg_router):
...
+                           interval=200,
+                           error=SendBackoffTimeoutError())
+
+        while True:
+            try:
+                await self._socket.send_multipart(
+                    [message.SerializeToString()])
+                break
+            except asyncio.CancelledError:
+                try:
```
peterschwarz: No need for this nested try catch block

rest_api/sawtooth_rest_api/messaging.py
```diff
@@ -161,10 +182,21 @@ def __init__(self, socket, msg_router):
...
+                    await backoff.do_backoff()
```
peterschwarz: This back off should be done only when `zmq.error.Again` is thrown, not at a `asyncio.CancelledError`.  The latter is thrown when the loop is shutdown and the underlying future is canceled.

rberg2:
Hello.
We recently changed the way we whitelist committers that are allowed to run builds, if you rebase Jenkins should build this PR.

rest_api/sawtooth_rest_api/messaging.py
```diff
@@ -161,10 +182,20 @@ def __init__(self, socket, msg_router):
...
+                           error=SendBackoffTimeoutError())
+
+        while True:
+            try:
+                await self._socket.send_multipart(
+                    [message.SerializeToString()])
+                break
+            except asyncio.CancelledError:
+                raise
+            except zmq.error.Again:
```
vaporos: If this contains an interesting string (which seems possible... this is generated by EAGAIN, so it might be very generic), then it should be logged with the other info (see above).

rest_api/sawtooth_rest_api/messaging.py
```diff
@@ -161,10 +182,20 @@ def __init__(self, socket, msg_router):
             content=message_content,
             message_type=message_type)
 
-        try:
-            await self._socket.send_multipart([message.SerializeToString()])
-        except asyncio.CancelledError:
-            raise
+        # Send the message. Backoff and retry in case of an error
+        backoff = _Backoff(max_retries=3,
```
vaporos: Do we have a reason we would only retry 3 times?  Is there a total time we are trying to stay under?

Similarly, why was "interval=200" chosen? If there is reasoning behind these numbers, it could be added to the comment above this statement.

kulkarniamol:
@peterschwarz @vaporos Updated and rebased. Please review the PR. Thanks!

peterschwarz:
You need to fix up some lint errors.  Take a look at the build output on Jenkins

PR #1558, opened 2018-04-04
Wrap Validator Python main with Rust main

aludvik:


aludvik:
Fixed an intermittent test failure because of an inappropriate sleep

PR #1562, opened 2018-04-10
Add rustfmt to build process

nick-drozd:


nick-drozd:
Per a suggestion of @vaporos, broke the `rustfmt` commit up into multiple commits at top-level boundaries for ease of backporting. Also added a comment about `rustfmt` exit codes to `run_lint_rust`.

PR #1564, opened 2018-04-10
BACKPORT: Add client_max_size to rest api config

peterschwarz:
Before the REST API defaulted to only allowing a
request body size of 1MB. Now the size limit can be
configured with the config file or using the command
line. If there is not a size configured there will be
no limit on the request body size.

### DIFF
--- a/docs/source/sysadmin_guide/configuring_sawtooth/rest_api_configuration_file.rst
+++ b/docs/source/sysadmin_guide/configuring_sawtooth/rest_api_configuration_file.rst
@@ -46,6 +46,17 @@ The ``rest_api.toml`` configuration file has the following options:
 
     timeout = 900
 
+- ``client_max_size`` = `value`
+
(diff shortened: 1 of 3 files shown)


dcmiddle:
Can the limitless default cause a denial of service for that rest service?

dcmiddle:
@agunde406 can you give context on the default ^

agunde406:
The default was chosen because the rest api has been written to be very permissive. For example, there is no authorization scheme or CORS handling. From talking with @delventhalz, the rest-api should not be exposed without modification (such as setting a response body limit). For example, for production, it should be put behind a proxy.

dcmiddle:
Thanks for the context @agunde406. The deployment model should indeed be that each participating company decides whether or not to _internally_ host a REST API instance. That implies a security model where certain attacks are out of scope, but it should still be resilient to bad requests from 'internal' users.

peterschwarz:
This now depends on #1580

PR #1565, opened 2018-04-11
Rust SDK: Handle send errors on disconnect

aludvik:
When the stream is shutting down, route receives a DisconnectedError and
then tries to forward this error on to all of its channles. However,
these channels tend to already have been cleaned up, causing a send
error. This commit handles this send error by logging it, rather than
panicing. This panic was seen during shutdown.

### DIFF
--- a/sdk/rust/src/messaging/zmq_stream.rs
+++ b/sdk/rust/src/messaging/zmq_stream.rs
@@ -177,18 +177,22 @@ impl InboundRouter {
             Ok(message) => {
                 let mut expected_replies = self.expected_replies.lock().unwrap();
                 match expected_replies.remove(message.get_correlation_id()) {
-                    Some(sender) => sender.send(Ok(message)).unwrap(),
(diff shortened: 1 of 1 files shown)


aludvik:
This is an example of the thread where the router was running doing a panic on shutdown:
```
   Compiling sawtooth_sdk v0.1.0 (file:///Users/ludvik/git/sawtooth-core/sdk/rust)
    Finished dev [unoptimized + debuginfo] target(s) in 7.64 secs
     Running target/debug/deps/sawtooth_sdk-9f084f67474c56e7

running 1 test
start()
block()
message()
peer()
stop()
ALL CHECKS PASSED
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: "SendError(..)"', libcore/result.rs:945:5
test consensus::zmq_driver::tests::test_zmq_driver ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 12 filtered out
```

@grkvlt I believe this fixes the seth-rpc shutdown error you were seeing

PR #1568, opened 2018-04-11
Fix new pycodestyle issues

nick-drozd:
A new version of `pycodestyle` was released yesterday (4/10/18). It introduced a new rule that binary operators should come after a line break rather than before:

```
# No: operators sit far away from their operands
income = (gross_wages +
          taxable_interest +
          (dividends - qualified_dividends) -
          ira_deduction -
          student_loan_interest)

# Yes: easy to match operators with operands
income = (gross_wages
          + taxable_interest
          + (dividends - qualified_dividends)
          - ira_deduction
          - student_loan_interest)
```

(See https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator)

It also fixed a bug where docstrings' first lines weren't being checked for length.

### DIFF
--- a/validator/sawtooth_validator/journal/publisher.py
+++ b/validator/sawtooth_validator/journal/publisher.py
@@ -231,15 +231,15 @@ class _CandidateBlock(object):
         is already in the pending queue.
         :param batch: the batch to check
         """
-        return (self._block_store.has_batch(batch.header_signature) or
-                batch.header_signature in self._pending_batch_ids)
+        return (self._block_store.has_batch(batch.header_signature)
+                or batch.header_signature in self._pending_batch_ids)
 
     def _is_txn_already_committed(self, txn, committed_txn_cache):
         """ Test if a transaction is already committed to the chain or
         is already in the pending queue.
         """
-        return (self._block_store.has_batch(txn.header_signature) or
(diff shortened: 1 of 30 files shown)


validator/sawtooth_validator/journal/publisher.py
```diff
@@ -763,10 +763,12 @@ def on_check_publish_block(self, force=False):
                         and self._pending_batches):
                     self._build_candidate_block(self._chain_head)
 
-                if self._candidate_block and (
-                    force or
-                    self._candidate_block.has_pending_batches()) and \
-                        self._candidate_block.check_publish_block():
+                # there is no good way to format this
```
agunde406: While this may be true, I don't think we need to call it out with a comment.

nick-drozd: Do you think there should be no comment at all? My thought was that the thing looks bizarre and warranted a comment pointing out that it isn't a mistake.

agunde406: Either no comment or a comment explaining why it has to look like that.

nick-drozd: Okay, I took it out.

PR #1569, opened 2018-04-12
Add Rust clippy to build process

nick-drozd:
`clippy` is a Rust linting tool. Whereas `rustfmt` is a strictly a
formatter, `clippy` offers opinions about how code should be written.
Some of these suggestions can be considered "style improvements", like
replacing a single-pattern `match` with `if let` and getting rid of
unnecessary `return` statements. 

Other suggestions can actually impact performance, like getting rid of
unnecessary type / ownership converions (`.iter()`, `.clone()`,
`.into()`, etc). One of the more sweeping changes it proposes is to
use pass-by-reference instead of pass-by-value where possible. This
supposedly results in fewer allocations (and hence better performance)
and also makes figuring out ownership simpler in some cases.

Overall, `clippy` seems like a very useful tool, and it exposes a lot
of language features that we aren't currently using. Certainly I've
learned a lot about Rust from looking at it.

#### Downsides

1. `clippy` is actually implemented as a set of compiler plugins.
   These plugins rely on non-stable compiler internals, and
   consequently it can only be run on the nightly toolchain. This does
   not mean that we have to switch to nightly; the stable and nightly
   toolchains can be used side by side, and apparently it is not an
   uncommon practice to install nightly for the sole purpose of using
   `clippy`. The drawback here is that sometimes the nightly compiler
   breaks, and so one day our builds could fail without warning. I
   don't know if that's a common occurrence or a rare one.
   
2. Adding `clippy` to the build process seems to add a lot of time
   (look at the build times for this PR). I don't know if there's a
   way to speed it up.

I don't think we have to commit to using it forever. We could try it
for a week or two and then get rid of it if it turns out to be too
much of a hassle.


nick-drozd:
`clippy` / Rust CI discussions:
* https://dev.to/cad97/great-rust-ci-1fk6
* http://lorenzo.villani.me/2017/04/25/rust-and-clippy/
* https://news.ycombinator.com/item?id=13323796

nick-drozd:
Also, the commits are structured right now to make review easier, but they should probably be squashed somehow before going in.

nick-drozd:
Commits that get rid of unnecessary stuff:
* lifetimes: https://github.com/hyperledger/sawtooth-core/pull/1569/commits/d51e502af03b13391c5a1a8429525d950338d02a, https://github.com/hyperledger/sawtooth-core/pull/1569/commits/706f569639cf715035d79f5dd29d8faa9a74ff75
* `let`s: https://github.com/hyperledger/sawtooth-core/pull/1569/commits/586422b02f5f1c7adc141aa50d05f8b97e6cdedf
* `.iter()`: https://github.com/hyperledger/sawtooth-core/pull/1569/commits/47c867868d377fdc9a4a0e7f9e6a4f42caa3117d
* `.into()`: https://github.com/hyperledger/sawtooth-core/pull/1569/commits/c848c78e202e6644e654742089f45e267afeb490
* `.clone()`: https://github.com/hyperledger/sawtooth-core/pull/1569/commits/0ee54628f908ec81fa90ea6279b2ea1bd6460751
* `to_string()`: https://github.com/hyperledger/sawtooth-core/pull/1569/commits/c463076fa0fca26b113a8b0c426cabfcfd7648de
* `return`: https://github.com/hyperledger/sawtooth-core/pull/1569/commits/7ce5768a0a959526bb8a501d30c7404ee8b6bc41

nick-drozd:
Commits that affect correctness (maybe):

* Replace lossy casts with lossless casts: https://github.com/hyperledger/sawtooth-core/pull/1569/commits/5e674188462c49480d20e9dc455f087b855861e9
* Remove a stray semicolon (I'm actually not sure how this was compiling before): https://github.com/hyperledger/sawtooth-core/pull/1569/commits/fabdcf4695b4babac22362e6442d7f9f127c4394

nick-drozd:
Commits that affect performance (maybe):

* Cut redundant closures: https://github.com/hyperledger/sawtooth-core/pull/1569/commits/e316ce812b41690f282f8e2c9d735d2799635558
```
-        val.ok().map(|v| Vec::from(v))
+        val.ok().map(Vec::from)
```
* Replace eager-evaluating `ok_or` with lazy-evaluating `ok_or_else`: https://github.com/hyperledger/sawtooth-core/pull/1569/commits/09b482dbf0516373ac1da00275ae4a5ed88251c8
```
-            .ok_or(DatabaseError::NotFoundError(format!(
-                "Block not found: {}",
-                block_id
-            )))?;
+            .ok_or_else(|| DatabaseError::NotFoundError(format!("Block not found: {}", block_id)))?;
```
* Replace single-character string split with single character split: https://github.com/hyperledger/sawtooth-core/pull/1569/commits/2ceb81f7ff05d880df500d3923ba5e7fcd359456
```
-            let split: Split<&str> = s.split(",");
+            let split: Split<char> = s.split(',');
```

nick-drozd:
The biggest change is replacing pass-by-value with pass-by-reference (https://github.com/hyperledger/sawtooth-core/pull/1569/commits/c74736c3245343e5e7f944a6784e4c7904f7dd9b, https://github.com/hyperledger/sawtooth-core/pull/1569/commits/a0eb427e5c772084589da74f1ebd524632776c41). This supposedly requires fewer allocations, and also allows for doing away with some conversion functions:
```
-    pub fn mark_submit_success(&mut self, batch_id: String) {
-        self.batches_by_id.remove(batch_id.as_str());
+    pub fn mark_submit_success(&mut self, batch_id: &str) {
+        self.batches_by_id.remove(batch_id);
```
```
-        timed_batch_iterator.mark_submit_success(batch_id1.clone());
-        timed_batch_iterator.mark_submit_success(batch_id3.clone());
+        timed_batch_iterator.mark_submit_success(&batch_id1);
+        timed_batch_iterator.mark_submit_success(&batch_id3);
```
```
-fn load_yaml_array(yaml_str: Cow<str>) -> Result<Cow<Vec<Yaml>>, PlaylistError> {
-    let mut yaml =
-        try!(YamlLoader::load_from_str(yaml_str.as_ref()).map_err(PlaylistError::YamlInputError));
+fn load_yaml_array(yaml_str: &str) -> Result<Cow<Vec<Yaml>>, PlaylistError> {
+    let mut yaml = try!(YamlLoader::load_from_str(yaml_str).map_err(PlaylistError::YamlInputError));
```

sdk/rust/src/processor/handler.rs
```diff
@@ -278,11 +278,11 @@ impl TransactionContext {
     /// * `addresses` - the addresses to fetch
     pub fn delete_state(
         &mut self,
-        addresses: Vec<String>,
+        addresses: &[String],
```
aludvik: Technically, this is a breaking API change. The Rust SDK is considered stable yet, so this is probably fine.

nick-drozd:
New changes:

1. Ran `rustfmt` on Smallbank Rust, since I forgot to add it to the last PR.
2. Took out the breaking API change in the Rust SDK and disabled that warning locally.
3. Per request of @peterschwarz, separated the Rust lint Jenkins stage from the Python and Go linting.
4. Added a missing docstring comment slash in the Rust SDK.

PR #1570, opened 2018-04-12
Fix infinte loop for set & delete context state(STL-1105)

rranjan3:
Validator was getting into infinite loop when trying to set & delete context state in one go. Set items were being handled before the deletion which was causing some expected data to be deleted. This threw an error which kept the batch from being cleaned up. This further caused the scheduler to pick up the same batch again and again after the schedule interval(infinte loop).
Now carrying out the iteration over **set_items** post iteration over **delete_items**.


askmish:
Please add sign-off to your commit message.

peterschwarz:
Rebase one more time - we had to remove a linting step.

rranjan3:
Thanks for pointing it out. Rebased once again.

validator/sawtooth_validator/state/merkle.py
```diff
@@ -251,6 +251,11 @@ def update(self, set_items, delete_items=None, virtual=True):
             if path != '':
                 parent_address = path[:-TOKEN_SIZE]
                 path_branch = path[-TOKEN_SIZE:]
+                # Check to re-instate the path_map for an address which
+                # shares it's prefix both in set_items & delete_items
+                if path_map.get(parent_address) is None:
+                    path_map[parent_address] = {"v": None, "c": {}}
+                    path_map[parent_address]['c'].update({path_branch: None})
```
peterschwarz: I don't think this update is necessary

validator/tests/test_merkle_trie/tests.py
```diff
@@ -140,6 +140,77 @@ def test_merkle_trie_update(self):
             with self.assertRaises(KeyError):
                 self.get(address, ishash=True)
 
+    def test_merkle_trie_update_same_address_space(self):
```
peterschwarz: Could you add a doc string describing the steps of the test?  As we're currently porting this to rust, it would be good to know what the test intent is

rranjan3: I have added some comments to help make more legible.

validator/tests/test_merkle_trie/tests.py
```diff
@@ -140,6 +140,89 @@ def test_merkle_trie_update(self):
...
                 self.get(address, ishash=True)
 
+    # This UT is similar to update except that it will
+    # ensure that there are no index errors in path_map
+    # within update function in case there are addresses
+    # within set_items & delete_items which have a common
+    # prefix(of any length).
+
+    # A merkle tree is cretaed with some intial values
+    # which is then updated(set & delete).
```
dcmiddle: Great explanatory comment. Can you please move it under the method declaration with the triple quote docstring style like this: https://github.com/hyperledger/sawtooth-core/blob/master/validator/tests/test_config/tests.py#L28-L48

swt2c:
Can this please be backported to 1.0 also?

peterschwarz:
The underlying implementation has changed in master, so this PR should be closed and reopened against `1-0-staging-01` as a backport PR.   I have added a similar test to #1609 to demonstrate that the new implementation does not demonstrate the bug.

PR #1573, opened 2018-04-17
Keyserver fallback

askmish:
If primary keyserver - keyserver.ubuntu.com is down, fallback to secondary pool.sks-keyserver and continue. Both keyservers would use port 80 to communicate, to allow communication in proxy environments.

This addresses a common exception scenario with keyservers.


dcmiddle:
@askmish why p80 and not just the pool (i.e. hkp://pool.sks-keyservers.net)?

askmish:
More info here: https://sks-keyservers.net/overview-of-pools.php

> p80.pool.sks-keyservers.net
> This is a pool containing only servers available on port 80 (needs to be used as hkp://p80.pool.sks-keyservers.net:80)

PR #1574, opened 2018-04-17
More logging when validator claims blocks too frequently

benoitrazet:
Add logging the public key of the validator when claiming blocks too frequently.

### DIFF
--- a/consensus/poet/core/sawtooth_poet/poet_consensus/poet_block_publisher.py
+++ b/consensus/poet/core/sawtooth_poet/poet_consensus/poet_block_publisher.py
@@ -536,9 +536,11 @@ class PoetBlockPublisher(BlockPublisherInterface):
                 block_cache=self._block_cache,
                 poet_enclave_module=poet_enclave_module):
             LOGGER.info(
-                'Reject building on block %s: Validator is claiming blocks '
+                'Reject building on block %s: '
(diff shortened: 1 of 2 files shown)


consensus/poet/core/sawtooth_poet/poet_consensus/poet_block_publisher.py
```diff
@@ -536,9 +536,11 @@ def initialize_block(self, block_header):
                 block_cache=self._block_cache,
                 poet_enclave_module=poet_enclave_module):
             LOGGER.info(
-                'Reject building on block %s: Validator is claiming blocks '
-                'too frequently.',
-                block_header.previous_block_id[:8])
+                'Reject building on block %s: '
+                'Validator is claiming blocks too frequently. '
```
peterschwarz: This is a good addition.  You can perhaps streamline this message as:

`'Validator (signing public key: %s) is claiming blocks too frequently.'`

benoitrazet: Originally I had something like that, but It failed in `test_consensus` that checked specifically for the string `'Validator is claiming blocks too frequently'`. Then I thought I would extend the string such that I don't have to change the test. Okay, I am going to change the message and the test accordingly. Thanks! @peterschwarz

PR #1577, opened 2018-04-17
Rustify validator CLI

nick-drozd:
The CLI arguments are parsed in main.rs and then JSON-serialized and
passed over into Python. There's no special reason to use JSON; it's
just a quick and simple way to get data over the wall.

### DIFF
--- a/validator/Cargo.toml
+++ b/validator/Cargo.toml
@@ -6,6 +6,7 @@ authors = ["Intel Corporation"]
 [dependencies]
 clap = "2"
 cpython = "0.2"
+log = { version = "0.4", features = ["std"] }
 
 [[bin]]
 name = "sawtooth-validator"
--- a/validator/Cargo.toml
+++ b/validator/Cargo.toml
@@ -4,6 +4,7 @@ version = "1.1.0"
 authors = ["Intel Corporation"]
 
 [dependencies]
+clap = "2"
 cpython = "0.2"
 
 [[bin]]
--- a/validator/sawtooth_validator/server/cli.py
+++ b/validator/sawtooth_validator/server/cli.py
@@ -15,7 +15,6 @@
 
 import logging
 import sys
-import argparse
 import os
 from urllib.parse import urlparse
 import pkg_resources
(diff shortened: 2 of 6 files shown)


validator/src/server/cli.rs
```diff
@@ -0,0 +1,209 @@
...
+        .arg(
+            Arg::with_name("endpoint")
+                .short("E")
+                .long("endpoint")
+                .takes_value(true)
+                .help("specifies the advertised network endpoint URL"),
+        )
+        .arg(
+            Arg::with_name("seeds")
+                .short("S")
```
aludvik: Should be lowercase s to match existing args

validator/Cargo.toml
```diff
@@ -4,7 +4,9 @@ version = "1.1.0"
 authors = ["Intel Corporation"]
 
 [dependencies]
+clap = ">=2.29.0"
```
vaporos: This should be simply clap = "2".

validator/sawtooth_validator/server/cli.py
```diff
@@ -15,7 +15,7 @@
 
 import logging
 import sys
-import argparse
+import json
```
vaporos: Suggest we not use JSON for any serialization between components. While it may be easy, it is not good in that it does create a contract between them. Protobuf would be better.  Avoiding serialization completely would be idea.

validator/sawtooth_validator/server/cli.py
```diff
@@ -184,53 +90,26 @@ def load_validator_config(first_config, config_dir):
...
+        self.permissions = None
+        self.opentsdb_username = None
+        self.opentsdb_password = None
 
-    init_console_logging(verbose_level=verbose_level)
 
-    if opts.network_auth:
-        opts.network_auth = {"network": opts.network_auth}
+def main(opts_string):
+    opts = _Opts(opts_string)
```
vaporos: Reducing this to arbitrary key/value pairs in not acceptable as it removes the definition of the object. This would be a regression to what we had in 0.7.

validator/src/server/cli.rs
```diff
@@ -0,0 +1,209 @@
...
+        "peering": matches.value_of("peering"),
+        "peers": parse_comma_separated_args("peers", &matches),
+        "scheduler": matches.value_of("scheduler"),
+        "seeds": parse_comma_separated_args("seeds", &matches),
+        "verbose": matches.value_of("verbose"),
+    }).to_string()
+}
+
+fn parse_args<'a>() -> ArgMatches<'a> {
+    let app = App::new(DISTRIBUTION_NAME)
```
vaporos: Any reason not to use the macro syntax? It is much more concise.

peterschwarz: The macro syntax does not support `possible_values`

validator/src/server/cli.rs
```diff
@@ -0,0 +1,254 @@
...
+ * ------------------------------------------------------------------------------
+ */
+
+use clap::{App, Arg, ArgMatches};
+use cpython::{PyDict, Python};
+
+const DISTRIBUTION_NAME: &str = "sawtooth-validator";
+const VERSION: &str = env!("CARGO_PKG_VERSION");
+
+pub fn get_cli_args(py: &Python) -> PyDict {
```
aludvik: This should take `matches` as an argument and return PyResult<PyDict> instead of unwrapping. The caller can then do `.map_err(|err| err.print(*py))` to get a nice python stack trace if there was a problem. Doing an unwrap on a PyResult causes very opaque errors.

validator/src/server/cli.rs
```diff
@@ -0,0 +1,226 @@
...
+            if binding.starts_with("component") {
+                bind_component = Some(binding.splitn(2, ':').skip(1).collect::<Vec<_>>()[0]);
+            }
+        }
+    };
+
+    (bind_component, bind_network)
+}
+
+fn parse_comma_separated_args(name: &str, matches: &ArgMatches) -> Option<Vec<String>> {
```
peterschwarz: Can this also return a `Option<Vec<&'a str>>`, like the following function?

nick-drozd: We need to go from a list of possibly comma-separated args like `["a,b,c", "d", "e"]` to a list of separated args `["a", "b", "c", "d", "e"]`. In Python this was done by creating a new list and then splitting the args and appending them one by one. My idea was to join all the strings with a comma and then split on the comma. Unfortunately `join` (I think) returns `String`, so that's what the function has to return.

Is there a more elegant way to do it? I don't like the return type either, but I didn't want to spend a lot of time on it since the function gets called just twice.

nick-drozd:
Added Rust logging

PR #1581, opened 2018-04-20
sawtooth batch submit with --wait throws a stack trace

mithunshashidhara:
sawtooth batch submit with --wait throws a stack trace.
this changeset contains fix for this issue.

Signed-off-by: mithun shashidhara <mithunx.shashidhara@intel.com>

### DIFF
--- a/validator/sawtooth_validator/state/client_handlers.py
+++ b/validator/sawtooth_validator/state/client_handlers.py
@@ -551,7 +551,10 @@ def _format_batch_statuses(statuses, batch_ids, tracker):
            client_batch_submit_pb2.ClientBatchStatus.INVALID:
             invalid_txns = tracker.get_invalid_txn_info(batch_id)
             for txn_info in invalid_txns:
-                txn_info['transaction_id'] = txn_info.pop('id')
+                try:
+                    txn_info['transaction_id'] = txn_info.pop('id')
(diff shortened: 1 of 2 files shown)


validator/sawtooth_validator/state/client_handlers.py
```diff
@@ -551,7 +551,10 @@ def _format_batch_statuses(statuses, batch_ids, tracker):
            client_batch_submit_pb2.ClientBatchStatus.INVALID:
             invalid_txns = tracker.get_invalid_txn_info(batch_id)
             for txn_info in invalid_txns:
-                txn_info['transaction_id'] = txn_info.pop('id')
+                try:
```
peterschwarz: This looks like a separate fix - you should separate this out into its own with a specific message describing the fix.

mithunshashidhara: please look at the comment (log) i have added in
https://jira.hyperledger.org/browse/STL-858
the ERROR messages were seen on console where we run docker-compose up.
when i run the command
sawtooth batch submit -f poet-settings.batch --url http://rest-api:8080 --wait 3

fix in both files are related to same bug.

kulkarniamol:
@askmish pls review this once again.

PR #1584, opened 2018-04-20
Add stdeb.cfg for validator

rbuysse:
Allows inclusion of non-python dependencies. In this case, libpython3.5
is required to call validator python code from the rust wrapper.


aludvik:
Tested locally

PR #1590, opened 2018-04-24
Improve local development of Sawadm on OS X

peterschwarz:
Adds a few minor changes for improving the local development experience on OS X.  Namely:

- Limiting the size of the LMDB db in the tests
- Selecting the `fs::MetadataExt` based on os

### DIFF
--- a/adm/src/blockstore.rs
+++ b/adm/src/blockstore.rs
@@ -242,7 +242,9 @@ mod tests {
 
         let blockstore_path = &path_config.data_dir.join(config::get_blockstore_filename());
 
-        let ctx = LmdbContext::new(blockstore_path, 3, None)
+        // Set the file size to 10MB, so as to support file systems that do
+        // not support sparse files.
+        let ctx = LmdbContext::new(blockstore_path, 3, Some(10 * 1024 * 1024))
             .map_err(|err| DatabaseError::InitError(format!("{}", err)))
             .unwrap();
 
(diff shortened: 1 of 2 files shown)


nick-drozd:
Can't test, as I'm still getting
```
cargo:warning=../c/loader.c:22:10: fatal error: 'openssl/bio.h' file not found
cargo:warning=#include <openssl/bio.h>
cargo:warning=         ^
cargo:warning=1 error generated.
```

peterschwarz:
You still need to have the right openssl lib installed

adm/src/blockstore.rs
```diff
@@ -242,7 +242,7 @@ mod tests {
 
         let blockstore_path = &path_config.data_dir.join(config::get_blockstore_filename());
 
-        let ctx = LmdbContext::new(blockstore_path, 3, None)
+        let ctx = LmdbContext::new(blockstore_path, 3, Some(10 * 1024 * 1024))
```
vaporos: A comment explaining 10 * 1024 * 1024 may be appropriate.

adm/src/commands/keygen.rs
```diff
@@ -18,7 +18,10 @@
 use std::ffi::CString;
 use std::fs::{metadata, OpenOptions};
 use std::io::prelude::*;
+#[cfg(target_os="linux")]
 use std::os::linux::fs::MetadataExt;
+#[cfg(target_os="macos")]
```
vaporos: Maybe this should be "not Linux", so there is at least a fallback (might be relevant for FreeBSD, for example).

PR #1592, opened 2018-04-25
Fix validator peering on Azure

arthurgreef:
Validator peering fails when using Azure endpoints such as
tcp://hashblock3.westus2.cloudapp.azure.com:8800 because the 'lo'
network interface characters are characters in 'cloudapp'. This
fix test that the network interface string is not equal to the

### DIFF
--- a/validator/sawtooth_validator/networking/handlers.py
+++ b/validator/sawtooth_validator/networking/handlers.py
@@ -15,6 +15,9 @@
 import logging
 import os
 import time
+
+from urllib.parse import urlparse
+
 import netifaces
 
 from sawtooth_signing import create_context
@@ -58,6 +61,27 @@ class ConnectHandler(Handler):
     def __init__(self, network):
         self._network = network
 
+    @staticmethod
+    def is_valid_endpoint_host(interfaces, endpoint):
+        """
+        An enpoint host name is valid if it is a URL and if the
+        host is not the name of a network interface.
+        """
+        result = urlparse(endpoint)
+        hostname = result.hostname
+        if hostname is None:
+            LOGGER.debug("Endpoint is not a valid URL : %s", endpoint)
+            return False
+        else:
+            for interface in interfaces:
+                if interface == hostname:
+                    LOGGER.debug("Endpoint host cannot be the "
+                                 "same as a network interface. '%s': %s",
+                                 interface,
+                                 endpoint)
+                    return False
+        return True
+
     def handle(self, connection_id, message_content):
         """
         A connection must use one of the supported authorization types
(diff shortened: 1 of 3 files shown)


validator/sawtooth_validator/networking/handlers.py
```diff
@@ -15,8 +15,12 @@
 import logging
 import os
 import time
+
+from urllib3.util import url
```
peterschwarz: This needs to be an explicit dependency.  It might be a transient dependency, which could be the reason it works, but that may only be due to a core dependency requiring it currently.

arthurgreef: Hi - can you tell me what to change? I do not know what I need to do at this point.

peterschwarz: Would it be possible to make this change using the built-in urllib?

arthurgreef:
Changed to use urllib.

arthurgreef:
One of the reasons that  this is a difficult PR to complete is because I do not know what a valid endpoint should be? Is it just a name or does it need to be a valid URL. A validator test like the one below just uses the string "endpoint". I would imagine that this is not a valid endpoint but if it is then the uriparse function returns 'endpoint' as a path and not as a hostname. Should we change the tests so that all endpoints are URL's? 

def test_connect(self):
        """
        Test the ConnectHandler correctly responds to a ConnectionRequest.
        """
        connect_message = ConnectionRequest(endpoint="endpoint")
        roles = {"network": AuthorizationType.TRUST}

peterschwarz:
The test should specify a correct endpoint: e.g. `tcp://test-endpoint:8800`

arthurgreef:
I have pushed a revision. Thanks.

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

________________________________

validator/sawtooth_validator/networking/handlers.py
```diff
@@ -58,6 +61,27 @@ class ConnectHandler(Handler):
     def __init__(self, network):
         self._network = network
 
+    @staticmethod
+    def is_valid_endpoint_host(interfaces, endpoint):
```
nick-drozd: Does this need to take an `interfaces` argument? It looks like there are two calls made to this function, and they both specify the same interfaces. Having `is_valid_endpoint_host` construct the list of interfaces itself would be simpler for callers. If it does need to be specified, maybe the signature could be `(endpoint, interfaces=None`), with a `None` arg defaulting to `["*", ".".join(["0", "0", "0", "0"])] + netifaces.interfaces()`.

(My motivation is that I'm rewriting validator `main` and that task would be easier with the interface stuff cut out.)

validator/sawtooth_validator/networking/handlers.py
```diff
@@ -58,6 +61,27 @@ class ConnectHandler(Handler):
...
+        """
+        An enpoint host name is valid if it is a URL and if the
+        host is not the name of a network interface.
+        """
+        result = urlparse(endpoint)
+        hostname = result.hostname
+        if hostname is None:
+            LOGGER.debug("Endpoint is not a valid URL : %s", endpoint)
```
nick-drozd: These `debug` statements should be `warning` or `error`, or else they might not be properly reported at validator startup.

peterschwarz: These messageslevels are fine, but there should be the appropriate error message at the point of usage of the function, not at such a low-level point.  The usage of this function has two wildly different effects: one is fatal, the other is merely a warning (and perfectly recoverable, as far as the caller is concerned)

nick-drozd: Yeah, leaving the logging at the call sites would be better.

validator/sawtooth_validator/networking/handlers.py
peterschwarz: I wouldn't worry about this.  It's irrelevant to this PR

peterschwarz:
This also needs a rebase - we've had a bunch of dependencies update that cause some build failures without some corrections in our code.

peterschwarz:
Closing in favor of #1642

PR #1596, opened 2018-04-26
Replace Python Merkle Trie Implementation with Rust FFI calls

peterschwarz:
Replaces the internals of the Python implementation of the Merkle Trie with FFI calls to the rust implementation.

### DIFF
--- a/validator/sawtooth_validator/database/native_lmdb.py
+++ b/validator/sawtooth_validator/database/native_lmdb.py
@@ -1,0 +1,44 @@
+import ctypes
+from enum import IntEnum
+
+from sawtooth_validator.ffi import LIBRARY
+
+
+class NativeLmdbDatabase(object):
+    def __init__(self, path, _size=1024**4):
+        self._db_ptr = ctypes.c_void_p()
+
+        c_path = ctypes.c_char_p(path.encode())
+        c_size = ctypes.c_size_t(_size)
+        res = LIBRARY.call(
+            'lmdb_database_new', c_path, c_size, ctypes.byref(self._db_ptr))
+        if res == ErrorCode.Success:
+            return
+        elif res == ErrorCode.NullPointerProvided:
+            raise TypeError("Path cannot be null")
+        elif res == ErrorCode.InvalidFilePath:
+            raise TypeError("Invalid file path {}".format(path))
+        elif res == ErrorCode.InitializeContextError:
+            raise TypeError("Unable to initialize LMDB Context")
+        elif res == ErrorCode.InitializeDatabaseError:
+            raise TypeError("Unable to initialize LMDB Database")
+        else:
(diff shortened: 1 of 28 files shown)


validator/sawtooth_validator/database/native_lmdb.py
```diff
@@ -0,0 +1,44 @@
+import ctypes
+from enum import IntEnum
+
+from sawtooth_validator.ffi import LIBRARY
+
+
+class NativeLmdbDatabase(object):
```
aludvik: Why "Native"?

validator/sawtooth_validator/state/merkle.py
```diff
@@ -307,11 +166,114 @@ def addresses(self):
...
+            "merkle root was not a valid hash")
+    elif res == ErrorCode.InvalidAddress:
+        raise KeyError(
+            "Address was not valid ")
+    elif res == ErrorCode.StopIteration:
+        raise StopIteration()
+    elif res == ErrorCode.Unknown:
+        raise ValueError("An unknown error occurred")
+    else:
+        raise ValueError("An unknown error occurred: {}".format(res))
```
aludvik: This error code to exception mapping seems like it would be common enough to warrant providing a function in sawtooth_validator.ffi to simplify it. It might also be a good idea to define a base error type that includes some common exceptions like `Success` and `NullPointer`.

validator/sawtooth_validator/state/merkle.py
```diff
@@ -307,11 +166,114 @@ def addresses(self):
...
+        pass
+
+
+def _prepare_byte_result():
+    return (ctypes.POINTER(ctypes.c_uint8)(), ctypes.c_size_t(0))
+
+
+def _from_c_bytes(c_data, c_data_len):
+    # pylint: disable=invalid-slice-index
+    return bytes(c_data[:c_data_len.value])
```
aludvik: The above two functions could also go into the FFI module.

validator/sawtooth_validator/state/merkle.py
```diff
@@ -39,45 +33,26 @@ def _encode(value):
...
+            _libexec('merkle_db_new_with_root', database.pointer,
+                     init_root, ctypes.byref(self._merkle_db_ptr))
+        else:
+            _libexec('merkle_db_new', database.pointer,
+                     ctypes.byref(self._merkle_db_ptr))
 
-        if node["v"] is not None:
-            yield (path, _decode(node["v"]))
+    def __del__(self):
+        # check that it has not been deleted, nor is it no
```
aludvik: think there is some text missing here

validator/sawtooth_validator/state/merkle.py
peterschwarz: I didn't combine these into a common one because there's no way to do the same on the Rust side of things.  The error codes are just enums.

PR #1600, opened 2018-04-28
Move validator cli.main logic to Rust main

nick-drozd:
`sawtooth_validator.server.cli.main` contains logic for displaying validator startup information as well as configuring and starting the validator and related components. This PR moves all logic into the Rust validator `main` when doing so is simple, and into small, focused helper functions when it isn't (an example of the former is the `init_errors` check and an example of the latter is the `netifaces` interface check).

The approach here is to create the path and validator configs as Python objects and then pass them in as needed to helper Python functions, manipulating the results in Rust. The `cpython` library is used a lot in the process, so several Rust macros are introduced to cut down on repetitive error handling.


validator/sawtooth_validator/server/cli.py
```diff
@@ -150,31 +144,78 @@ def main(path_config, validator_config, verbosity):
...
+    # Medium security risk.
+    interfaces = ["*", ".".join(["0", "0", "0", "0"])]
+    interfaces += netifaces.interfaces()
+
+    for interface in interfaces:
+        if interface in endpoint:
+            LOGGER.error("Endpoint must be set when using %s", interface)
+            return True
+
+    return False
```
aludvik: Careful, this code is being modified here: https://github.com/hyperledger/sawtooth-core/pull/1592

validator/src/main.rs
```diff
@@ -13,131 +13,108 @@ fn main() {
     let gil = Python::acquire_gil();
     let py = &mut gil.python();
 
+    macro_rules! py_unwrap {
+        ($expr:expr) => {
+            $expr.map_err(|err| err.print(*py)).unwrap()
+        }
+    }
```
aludvik: Prefer generic functions to macros when possible:
```
fn py_unwrap<T>(py: Python, result: PyResult<T>) -> T {
  result.map_err(|err| err.print(py)).unwrap()
}
```

The above macro makes assumptions about the names of variables outside its scope (`py` in this case) which is not a good pattern.

PR #1604, opened 2018-04-30
Add Rust wrapper to Python transaction execution

boydjohnson:
This PR adds two traits, ExecutionPlatform and Scheduler, and implements them both with the Python transaction processing capabilities already in master.

### DIFF
--- a/validator/sawtooth_validator/execution/executor.py
+++ b/validator/sawtooth_validator/execution/executor.py
@@ -416,12 +416,12 @@ class TransactionExecutor(object):
             first_state_root = self._context_manager.get_first_root()
 
         if self._scheduler_type == "serial":
-            return SerialScheduler(
+            scheduler = SerialScheduler(
                 squash_handler=self._context_manager.get_squash_handler(),
                 first_state_hash=first_state_root,
                 always_persist=always_persist)
         elif self._scheduler_type == "parallel":
-            return ParallelScheduler(
+            scheduler = ParallelScheduler(
                 squash_handler=self._context_manager.get_squash_handler(),
                 first_state_hash=first_state_root,
                 always_persist=always_persist)
@@ -430,6 +430,9 @@ class TransactionExecutor(object):
             raise AssertionError(
                 "Scheduler type must be either serial or parallel. Current"
                 " scheduler type is {}.".format(self._scheduler_type))
+
+        self.execute(scheduler=scheduler)
+        return scheduler
 
     def check_connections(self):
         self._executing_threadpool.submit(self._check_connections)
--- a/validator/sawtooth_validator/execution/executor.py
+++ b/validator/sawtooth_validator/execution/executor.py
@@ -410,6 +410,11 @@ class TransactionExecutor(object):
     def create_scheduler(self,
                          first_state_root,
                          always_persist=False):
+
+        # Useful for a logical first state root of ""
+        if not first_state_root:
+            first_state_root = self._context_manager.get_first_root()
+
         if self._scheduler_type == "serial":
             return SerialScheduler(
                 squash_handler=self._context_manager.get_squash_handler(),
--- a/validator/src/execution/py_executor.rs
+++ b/validator/src/execution/py_executor.rs
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2018 Intel Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ------------------------------------------------------------------------------
+ */
+
+use cpython;
+use cpython::ObjectProtocol;
+
+use execution::execution_platform::ExecutionPlatform;
+
+use scheduler::Scheduler;
+use scheduler::py_scheduler::PyScheduler;
+
+pub struct PyExecutor {
+    executor: cpython::PyObject,
+}
+
+impl PyExecutor {
+    pub fn new(executor: cpython::PyObject) -> Result<PyExecutor, cpython::PyErr> {
+        Ok(PyExecutor { executor })
+    }
+}
+
+impl ExecutionPlatform for PyExecutor {
+    fn create_scheduler(&self, state_hash: &str) -> Result<Box<Scheduler>, cpython::PyErr> {
+        let py = unsafe { cpython::Python::assume_gil_acquired() };
+
+        let scheduler = self.executor
+            .call_method(py, "create_scheduler", (state_hash,), None)
+            .expect(
+                "no method create_scheduler on sawtooth_validator.execution.py_executor.PyExecutor",
+            );
+        Ok(Box::new(PyScheduler::new(scheduler)))
+    }
+}
--- a/validator/src/scheduler/mod.rs
+++ b/validator/src/scheduler/mod.rs
@@ -19,6 +19,30 @@ use proto::events::Event;
 use proto::transaction_receipt::StateChange;
 
 mod execution_result_ffi;
+use batch::Batch;
+
+pub trait Scheduler {
+    /// Add a batch to the scheduler, optionally specifying that the transactions
+    /// in this batch and each of the batches in order up to this one should produce a
+    /// Merkle root specified by expected_state_hash. If the two roots are equal, the results of
+    /// the block are written to the database, otherwise not.
+    fn add_batch(
+        &mut self,
+        batch: Batch,
+        expected_state_hash: Option<&str>,
+        required: bool,
+    ) -> Result<(), SchedulerError>;
+
+    /// Signal to the scheduler that it can finish it's work and no
+    /// more batches will be handed to it.
+    fn finalize(&mut self, unschedule_incomplete: bool) -> Result<(), SchedulerError>;
+
+    /// Cancel the scheduling of the supplied transactions as they are no longer needed.
+    fn cancel(&mut self) -> Result<(), SchedulerError>;
+
+    /// Ask if the ExecutionResults are ready, optionally blocking until they become available.
+    fn complete(&mut self, block: bool) -> Result<Option<ExecutionResults>, SchedulerError>;
+}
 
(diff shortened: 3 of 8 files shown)


peterschwarz:
Separated out the generated proto source generation into #1606

boydjohnson:
This PR is now built on #1606 so that the Protobuf generation happens in a simpler way.

validator/src/scheduler/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+    pub error_data: Vec<u8>,
+}
+
+#[derive(Debug)]
+pub enum SchedulerError {
+    /// The scheduler transition is not allowed by the Finite State Machine.
+    FSMError(String),
+    /// There was an error caused by the FFI interface.
+    FFIError(String),
+    Error(String),
```
aludvik: This error should be broken up into more concrete errors

integration/sawtooth_integration/tests/test_py_executor.rs
```diff
@@ -0,0 +1,98 @@
...
+
+use self::intkey_rust::start;
+use self::intkey_workload::intkey_iterator::IntKeyIterator;
+use self::intkey_workload::intkey_transformer::IntKeyTransformer;
+use self::sawtooth_perf::batch_gen::SignedBatchIterator;
+use self::sawtooth_sdk::signing;
+use self::sawtooth_sdk::messages::batch::Batch;
+use self::sawtooth_validator::execution::execution_platform::{ExecutionPlatform, NULL_STATE_HASH};
+use self::sawtooth_validator::execution::py_executor::PyExecutor;
+use self::sawtooth_validator::proto::batch;
```
aludvik: I do not believe you need `self::` here

integration/sawtooth_integration/tests/test_py_executor.rs
```diff
@@ -0,0 +1,98 @@
+/*
```
aludvik: We should follow this pattern for these kinds of tests (not quite integration but not quite unit): https://doc.rust-lang.org/stable/rust-by-example/testing/integration_testing.html. You should use `[dev-dependencies]` to specify libraries that are only required by testing (intkey here).

sdk/examples/intkey_rust/src/lib.rs
```diff
@@ -0,0 +1,18 @@
...
+use sawtooth_sdk::processor::TransactionProcessor;
+
+use handler::IntkeyTransactionHandler;
+
+pub fn start(endpoint: &str) {
+    let handler = IntkeyTransactionHandler::new();
+    let mut processor = TransactionProcessor::new(endpoint);
+    processor.add_handler(&handler);
+    processor.start();
+}
```
aludvik: I understand why you need to create a lib.rs file in intkey rust. I think it makes more sense from the perspective of intkey rust that the library only publish the IntkeyTransactionHandler and that the test handle starting up the processor.

validator/sawtooth_validator/execution/py_executor.py
```diff
@@ -0,0 +1,178 @@
...
+        component_dispatcher.start()
+        component_service.start()
+
+        self._executor = transaction_executor
+        self._context_manager = context_manager
+
+        self._component_service = component_service
+        self._component_dispatcher = component_dispatcher
+
+        self._schedulers = {}
```
aludvik: What is your plan for integrating this into the existing validator? Some of these things need to be used by other components in the system and I don't think it makes sense to create them here.

validator/src/scheduler/mod.rs
boydjohnson: I'll work on breaking up Error into more meaningful errors.

integration/sawtooth_integration/tests/test_py_executor.rs
boydjohnson: Thanks for the link.

validator/sawtooth_validator/execution/py_executor.py
boydjohnson: My thinking was that we could create a client instance of Interconnect and Dispatcher. I think that would work.

integration/sawtooth_integration/tests/test_py_executor.rs
boydjohnson: This way allows for there not to be a src directory.

boydjohnson:
@aludvik I found that the only place that I was using `SchedulerError::Error` was to mean a Rust Python mismatch where a value is None, or information isn't there that was expected to be there. So I made `SchedulerError::PythonRustMismatch(String)`. I changed what was exported from `intkey_rust` to be just the IntkeyTransactionHandler. The other changes requested didn't seem possible in a reasonable time. I couldn't get `sawtooth_validator` to build with the new dev-dependencies.

validator/sawtooth_validator/execution/py_executor.py
aludvik: You still need a solution for how you bring in objects that you don't create in this constructor. For example, the global_state_db. Maybe you can create Interconnect and Dispatcher here, but that is a departure from the current pattern and requires a user facing change (they must now also configure the client bind endpoint. fwiw I think client and component should have been separate from the beginning).

validator/sawtooth_validator/execution/py_executor.py
```diff
@@ -0,0 +1,178 @@
...
+
+        if not state_root:
+            state_root = self._context_manager.get_first_root()
+
+        scheduler = self._executor.create_scheduler(
+            self._context_manager.get_squash_handler(),
+            state_root,
+            force_persist)
+
+        self._executor.execute(scheduler=scheduler)
```
aludvik: Why did we choose to move execute here instead of exposing it an a method on the Executor interface?

validator/src/scheduler/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+#[derive(Debug)]
+pub enum SchedulerError {
+    /// The scheduler transition is not allowed by the Finite State Machine.
+    FSMError(String),
+    /// There was an error caused by the FFI interface.
+    FFIError(String),
+    /// There was a mismatch that wasn't a Python error, but the Rust should
+    /// treat as an error because, for example, a value was None when it shouldn't be,
+    /// or information was missing.
+    PythonRustMismatch(String),
```
aludvik: FFIError and PythonRustMismatch are not good error types for Scheduler as they leak details about how a particular implementation of this trait works.

If there are things that are technically possible based on the library we are using but should never happen in practice, it is probably okay to just panic on them instead of propagating the error. I suspect these errors may go away if you use that pattern. For example, if we create and hold onto a python object in the Rust constructor, that object's type should never change. So if later we go to call a method on that object and that method does not exist, we should probably just panic at that point because either some memory has been corrupted or there is a serious programming error (like the python code changed and we forgot to update it in the Rust code).

validator/sawtooth_validator/execution/py_executor.py
boydjohnson: I don't see why global_state_db can't be instantiated inside the Transaction Execution python code. That is the only portion of the code that should write to state. There may be other parts of the code that read from state. I don't see how to pass Interconnect and the dispatcher over the FFI, so having a breaking Api change seems like it is the only possibility. And it kind of makes sense, too.

validator/sawtooth_validator/execution/py_executor.py
boydjohnson: That method would take in Rust schedulers?

validator/src/scheduler/mod.rs
boydjohnson: I agree that some of the python errors -> Rust errors should just fail, but I'm sure there are python errors that should be propagated up that wouldn't normally be part of the Rust Api.

aludvik: Right, but we need an error type that makes sense for these that isn't implementation specific.

validator/sawtooth_validator/execution/py_executor.py
aludvik: Yeah, it would look just like the current execute method. I am just wondering why you chose not to expose it.

boydjohnson:
This work isn't currently going to be used.

boydjohnson:
After talking with @peterschwarz, we may be using this.

validator/src/execution/py_executor.rs
```diff
@@ -0,0 +1,72 @@
...
+
+impl PyExecutor {
+    pub fn new(
+        context_manager: cpython::PyObject,
+        executor: cpython::PyObject,
+    ) -> Result<PyExecutor, cpython::PyErr> {
+        let py = unsafe { cpython::Python::assume_gil_acquired() };
+        Ok(PyExecutor {
+            pyobj: cpython::PyModule::import(py, "sawtooth_validator.execution.py_executor")
+                .unwrap()
```
aludvik: When unwrapping a PyResult like this, use `expect("helpful context")` to make debugging easier in the future. I think if you just unwrap, Rust just gives the line where unwrap is implemented in the std library instead of the line where the error occurred.

This comment also applies to other PyResult unwraps in this PR.

validator/sawtooth_validator/execution/py_executor.py
boydjohnson: So I don't think the block publisher or block validator want a scheduler that isn't executing. The other possibility is that there would be `create_scheduler(state_root, force_persist)` and `execute(scheduler)` but I don't know if that serves a purpose and adds one more method to the interface. Do you see a reason to separate create_scheduler and execute?

validator/src/execution/py_executor.rs
boydjohnson: My experience is that .unwrap() gives `ImportError: unable to import "sawtooth_validator.execution.py_executor"` which is at least as helpful as anything I would write. I think the `.expect("helpful content")` is helpful for reading through the code and understanding why unwrapping is ok. I will update to `expect` but because it allows for more understanding in the code. The debugging messages might be a wash.

validator/sawtooth_validator/execution/py_executor.py
aludvik: I do not see I reason, but I think we should understand why the execute method was there in the first place before we remove it.

validator/sawtooth_validator/execution/executor.py
```diff
@@ -432,6 +432,9 @@ def create_scheduler(self,
                 "Scheduler type must be either serial or parallel. Current"
                 " scheduler type is {}.".format(self._scheduler_type))
 
+        self.execute(scheduler=scheduler)
```
peterschwarz: This seems like quite a side effect

aludvik:
Please place the unscheduled batch ids commit in a separate PR so that it can be reviewed independently of the already approved PR.

boydjohnson:
@aludvik Done. I will put the difference of what was reviewed before and what I would like now up in a separate PR.

PR #1606, opened 2018-05-01
Rust validator protos

peterschwarz:
Add build.rs, which includes code generation based on the files in protos.  Generates the source into `validator/src/proto` and generates a `mod.rs` file with each generated src file referenced in it.

Additionally, the source is only generated again when there are changes to the proto definition files.

### DIFF
--- a/validator/Cargo.toml
+++ b/validator/Cargo.toml
@@ -10,10 +10,15 @@ log = { version = "0.4", features = ["std"] }
 libc = ">=0.2.35"
 lmdb-zero = ">=0.4.1"
 cbor-codec = "0.7"
+protobuf = "1.4"
 rust-crypto = "0.2"
 
 [dev-dependencies]
 rand = "0.4"
+
+[build-dependencies]
+protoc-rust = "1.4"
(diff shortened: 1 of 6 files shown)


validator/Cargo.toml
```diff
@@ -10,11 +10,16 @@ log = { version = "0.4", features = ["std"] }
 libc = ">=0.2.35"
 lmdb-zero = ">=0.4.1"
 cbor-codec = "0.7"
+protobuf="1.4"
```
jsmitchell: add spaces before and after the =

validator/build.rs
```diff
@@ -0,0 +1,167 @@
...
+        .expect("file stem")
+        .to_str()
+        .expect("utf-8")
+        .to_owned();
+
+    let file_path = path.to_str().expect("utf-8").to_owned();
+
+    let file = fs::File::open(path).expect("Unable to open file");
+    let last_modified = file.metadata()
+        .expect("file to have metadata")
```
jsmitchell: how do these strings read in an error if the expect fails?

PR #1608, opened 2018-05-01
Refactor Block Publisher to support Consensus Engine prototyping

aludvik:
This PR makes some initial strides towards supporting the consensus engine api, although it does not get all the way there. It also cleans up the block publisher code to make it easier to understand when this is easy to do.

### DIFF
--- a/validator/sawtooth_validator/journal/publisher.py
+++ b/validator/sawtooth_validator/journal/publisher.py
@@ -696,10 +696,6 @@ class BlockPublisher(object):
                     self._pending_batches.update_limit(len(chain_head.batches))
                     self._pending_batches.rebuild(
                         committed_batches, uncommitted_batches)
-                    try:
-                        self.initialize_block(self._chain_head)
-                    except ConsensusNotReady:
-                        self._log_consensus_state()
 
         # pylint: disable=broad-except
         except Exception:
--- a/validator/sawtooth_validator/journal/publisher.py
+++ b/validator/sawtooth_validator/journal/publisher.py
@@ -756,19 +756,8 @@ class BlockPublisher(object):
                         return
 
                     if result.block:
-                        blkw = BlockWrapper(result.block)
-                        LOGGER.info("Claimed Block: %s", blkw)
-                        self._block_sender.send(
-                            blkw.block,
-                            keep_batches=result.injected_batches)
-                        self._blocks_published_count.inc()
-
-                        # We built our candidate, disable processing until
-                        # the chain head is updated. Only set this if
-                        # we succeeded. Otherwise try again, this
-                        # can happen in cases where txn dependencies
-                        # did not validate when building the block.
-                        self.on_chain_updated(None)
+                        self.publish_block(
+                            result.block, result.injected_batches)
 
         # pylint: disable=broad-except
         except Exception:
@@ -786,6 +775,18 @@ class BlockPublisher(object):
             result.last_batch)
 
         return result
+
+    def publish_block(self, block, injected_batches):
+        blkw = BlockWrapper(block)
+        LOGGER.info("Claimed Block: %s", blkw)
+        self._block_sender.send(blkw.block, keep_batches=injected_batches)
+        self._blocks_published_count.inc()
+
+        # We built our candidate, disable processing until the chain head is
+        # updated. Only set this if we succeeded. Otherwise try again, this can
+        # happen in cases where txn dependencies did not validate when building
+        # the block.
+        self.on_chain_updated(None)
 
     def has_batch(self, batch_id):
         with self._lock:
--- a/validator/sawtooth_validator/journal/publisher.py
+++ b/validator/sawtooth_validator/journal/publisher.py
@@ -573,7 +573,7 @@ class BlockPublisher(object):
         else:
             return []
 
-    def _build_candidate_block(self, previous_block):
+    def initialize_block(self, previous_block):
         """ Build a candidate block and construct the consensus object to
         validate it.
         :param previous_block: The block to build on top of.
@@ -688,7 +688,7 @@ class BlockPublisher(object):
                     self._pending_batches.update_limit(len(chain_head.batches))
                     self._pending_batches.rebuild(
                         committed_batches, uncommitted_batches)
-                    self._build_candidate_block(chain_head)
+                    self.initialize_block(chain_head)
 
         # pylint: disable=broad-except
         except Exception as exc:
@@ -718,7 +718,7 @@ class BlockPublisher(object):
                 if (self._chain_head is not None
                         and not self._building()
                         and self._pending_batches):
-                    self._build_candidate_block(self._chain_head)
+                    self.initialize_block(self._chain_head)
 
                 if (self._building()
                         and (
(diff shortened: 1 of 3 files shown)


validator/sawtooth_validator/journal/publisher.py
```diff
@@ -556,24 +556,24 @@ def get_current_queue_info(self):
     def chain_head_lock(self):
         return self._lock
 
-    def _build_candidate_block(self, chain_head):
+    def _build_candidate_block(self, previous_block):
         """ Build a candidate block and construct the consensus object to
         validate it.
-        :param chain_head: The block to build on top of.
+        :param previous_block_id: The block to build on top of.
```
agunde406: previous_block_id -> previous_block

validator/sawtooth_validator/journal/publisher.py
```diff
@@ -486,9 +486,7 @@ def __init__(self,
         self._transaction_executor = transaction_executor
         self._block_sender = block_sender
         self._batch_publisher = BatchPublisher(identity_signer, batch_sender)
-        self._pending_batches = []  # batches we are waiting for validation,
-        # arranged in the order of batches received.
```
agunde406: Typo in corresponding commit message: "using a three separate" ->  "using three separate"

validator/sawtooth_validator/journal/publisher.py
```diff
@@ -812,14 +765,81 @@ def on_check_publish_block(self, force=False):
...
+
+    def append(self, batch):
+        self._batches.append(batch)
+        self._ids.add(batch.header_signature)
+        self._update_gauge()
+
+    def rebuild(self, committed=None, uncommitted=None):
+        """Recomputes the list of pending batches
+
+        :param committed_batches: Batches committed in the current chain
```
agunde406: Param names have changed, update docstring.

validator/sawtooth_validator/journal/publisher.py
```diff
@@ -812,14 +765,81 @@ def on_check_publish_block(self, force=False):
...
         return False
 
 
+class PendingBatchesPool:
+
+    def __init__(self):
+        # Ordered batches waiting to be processed
+        self._batches = list()
+        self._ids = set()
+        self._gauge = COLLECTOR.gauge('pending_batch_gauge', instance=self)
```
nick-drozd: In the current pattern, all the metrics are defined in the same place. Would it be better to leave the gauge in the publisher and pass it in as an argument?

validator/sawtooth_validator/journal/publisher.py
```diff
@@ -812,14 +765,81 @@ def on_check_publish_block(self, force=False):
...
+        if uncommitted is None:
+            uncommitted = []
+
+        committed_set = set(b.header_signature for b in committed)
+
+        previous_batches = self._batches
+
+        self.reset()
+
+        # Uncommitted and pending disjoint sets
```
nick-drozd: are disjoint sets

validator/sawtooth_validator/journal/publisher.py
```diff
@@ -812,14 +765,81 @@ def on_check_publish_block(self, force=False):
...
-            if batch_id in self._pending_batch_ids:
+            if batch_id in self._pending_batches:
                 return True
             if batch_id in self._queued_batch_ids:
                 return True
 
         return False
 
 
+class PendingBatchesPool:
```
peterschwarz: I suppose "Pool" is a bit more accurate than `Queue`

nick-drozd:
Commit message for `Refactor pending batches tracking into class`: previously done using __a__ three
separate data structures

validator/sawtooth_validator/journal/publisher.py
```diff
@@ -833,6 +833,7 @@ def update(self, still_pending, last_sent):
         last_index = self._batches.index(last_sent)
         unsent = self._batches[last_index + 1:]
         self._batches = still_pending + unsent
+        self._ids = set(b.header_signature for b in self._batches)
```
peterschwarz: Why not just replace `self._ids` with a generator?  Seems like it would be better to use `self._batches` as the source of truth, and the ids are just a view.  Secondly, making it a set where as self._batches is a list seems unnecessary or is missing a check that the should be same length - any discrepancy implies there is a duplicate batch in there.

validator/sawtooth_validator/journal/publisher.py
```diff
@@ -52,6 +53,24 @@
...
 
 
+class ConsensusNotReady(Exception):
+    """
+    Consensus is not ready to build a block.
+    """
+
+
+class NoPendingBatchesRemaining(Exception):
+    """There are pending batches remaining to be processed."""
```
aludvik: typo s/are/are not/

validator/sawtooth_validator/journal/publisher.py
```diff
@@ -59,6 +59,18 @@ class ConsensusNotReady(Exception):
     """
 
 
+class NoPendingBatchesRemaining(Exception):
```
agunde406: The name seems different from the message

validator/sawtooth_validator/journal/publisher.py
```diff
@@ -744,12 +742,30 @@ def has_batch(self, batch_id):
...
         return False
 
 
+class IncomingBatchQueue:
+    def __init__(self):
+        self._queue = queue.Queue()
+        self._ids = ConcurrentSet()
+
+    def put(self, batch):
+        self._ids.add(batch.header_signature)
```
peterschwarz: If these needs to be kept in sync, atomically, and this and `get` are used across threads, there needs to be a lock around the actions here.  Based on your commit message this doesn't guarantee they stay in sync, from a thread perspective.

validator/sawtooth_validator/journal/publisher.py
aludvik: The main benefit of using a set like this is improved lookup times. You are correct to point out that it does not correctly handle duplicate batches though, so I will add a check around that.

aludvik:
Fixed a bug I introduced (forgot to reset the pending batches pool in one place) and added a new commit to prevent blocks from being built outside the publisher thread.

validator/sawtooth_validator/journal/publisher.py
```diff
@@ -427,10 +433,10 @@ def finalize_block(self, identity_signer, pending_batches):
...
 class _PublisherLoggingStates:
     """Collects and tracks the changes in various states of the Publisher.  For
-    example it tracks 'consensus_not_ready', which denotes entering or exiting
-    of that state."""
-    def __init__(self, consensus_not_ready=False):
-        self.consensus_not_ready = False
+    example it tracks 'consensus_ready', which denotes entering or exiting of
+    that state."""
+    def __init__(self, consensus_ready=True):
+        self.consensus_ready = True
```
peterschwarz: Clearly, the old version was the same, but the parameter is not being used.

aludvik:
Completed overnight stability check of 10 node PoET network under 20 TPS workload. Network stayed in consensus and did 9.88 TPS.

PR #1609, opened 2018-05-02
Add test to ensure infinite loop fixed

peterschwarz:
In the python implementation of the Merkle Trie, there was an infinite loop that could occur if a delete was done before a set on the same namespace.

This adds a test to show the at the same bug does not occur in the rust implementation.

The original PR was #1570


boydjohnson:
This looks like the test is for a set before a delete, but your comment says that it is from a delete before a set. Could you explain?

peterschwarz:
Added a doc comment from the original PR to help explain things.

PR #1610, opened 2018-05-02
Add Python Consensus SDK

nick-drozd:
Right now this PR includes interfaces for `Engine`, `Service`, and `Driver`, and implements `ZmqService` with unit tests, but does not include the `ZmqDriver`.

The `Service` interface is slightly different from the Rust one. Specifically, it returns dictionaries instead of lists for `get_{blocks,settings,state}`. This seems to me like a more convenient way to access the data. I can change it to match Rust, or change Rust to match this.

### DIFF
--- a/sdk/python/sawtooth_sdk/consensus/zmq_driver.py
+++ b/sdk/python/sawtooth_sdk/consensus/zmq_driver.py
@@ -1,0 +1,150 @@
+# Copyright 2018 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


nick-drozd:
`ZmqDriver` and test added

nick-drozd:
Took out type annotations

sdk/python/sawtooth_sdk/consensus/zmq_driver.py
```diff
@@ -0,0 +1,150 @@
...
+from sawtooth_sdk.messaging.stream import Stream
+from sawtooth_sdk.protobuf import consensus_pb2
+from sawtooth_sdk.protobuf.validator_pb2 import Message
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+class ZmqDriver(Driver):
+    # pylint: disable=super-init-not-called
```
peterschwarz: I would still call the super init - event though it is an ABC with no implementation, there's no reason to avoid it.  Either that, or it doesn't seem like you need this disable comment, since you don't use it anywhere else.

nick-drozd: Took out the disable and added the super init call

PR #1613, opened 2018-05-02
C010 mounted dockerfiles

rbuysse:
This PR is a subset of the changes proposed in https://github.com/Cargill/sawtooth-rfcs/blob/c006-docker-compose-builds/text/0000-docker-compose-builds.md

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -118,8 +118,10 @@ services:
         python3 setup.py clean --all
         python3 setup.py build
         mkdir -p bin
-        cargo build
-        cp ./target/debug/sawtooth-validator bin/sawtooth-validator
+        mkdir -p lib
+        cargo build --release
+        cp ./target/release/sawtooth-validator bin/sawtooth-validator
+        cp ./target/release/libsawtooth_validator.so lib/libsawtooth_validator.so
         sawadm keygen
         sawadm genesis
         sawtooth-validator -vv \
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,0 +1,153 @@
+# Copyright 2018 Cargill Incorporated
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+version: '3.6'
+
+services:
+
+  settings-tp:
+    build:
+      context: .
+      dockerfile: families/settings/Dockerfile
+    image: sawtooth-settings-tp-local:${ISOLATION_ID}
+    volumes:
+      - ./:/project/sawtooth-core
+    container_name: sawtooth-settings-tp-local
+    depends_on:
+      - validator
+    command: |
+      bash -c "
+        bin/protogen
+        cd families/settings
+        python3 setup.py clean --all
+        python3 setup.py build
+        settings-tp -vv -C tcp://validator:4004
+      "
+    stop_signal: SIGKILL
+
+
+  intkey-tp-python:
+    build:
+      context: .
+      dockerfile: sdk/examples/intkey_python/Dockerfile
+    image: intkey-tp-python-local:${ISOLATION_ID}
+    volumes:
+      - ./:/project/sawtooth-core
+    container_name: intkey-tp-python-local
+    depends_on:
+      - validator
+    command: |
+      bash -c "
+        bin/protogen
+        cd sdk/examples/intkey_python
+        python3 setup.py clean --all
+        python3 setup.py build
+        intkey-tp-python -vv -C tcp://validator:4004
+      "
+    stop_signal: SIGKILL
+
+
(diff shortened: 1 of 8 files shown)


docker-compose.yaml
```diff
@@ -0,0 +1,155 @@
...
+      dockerfile: docker/sawtooth-shell
+    image: sawtooth-shell-local:${ISOLATION_ID}
+    volumes:
+      - ./:/project/sawtooth-core
+    container_name: sawtooth-shell-local
+    depends_on:
+      - validator
+    command: |
+      bash -c "
+        sawtooth keygen && \
```
delventhalz: This backslash is unnecessary.

docker-compose.yaml
```diff
@@ -0,0 +1,155 @@
...
+    stop_signal: SIGKILL
+
+
+  validator:
+    build:
+      context: .
+      dockerfile: validator/Dockerfile
+    image: sawtooth-validator-local:${ISOLATION_ID}
+    volumes:
+      - ./:/project/sawtooth-core
```
delventhalz: How do we feel about masking the `/target` directory so we don't dump a bunch of build artifacts onto our host machines?

As seen here:
https://github.com/hyperledger/sawtooth-supply-chain/blob/master/docker-compose.yaml#L70

validator/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+
+RUN curl -OLsS https://github.com/google/protobuf/releases/download/v3.5.1/protoc-3.5.1-linux-x86_64.zip \
+ && unzip protoc-3.5.1-linux-x86_64.zip -d protoc3 \
+ && rm protoc-3.5.1-linux-x86_64.zip
+
+RUN curl https://sh.rustup.rs -sSf > /usr/bin/rustup-init \
+ && chmod +x /usr/bin/rustup-init \
+ && rustup-init -y
+
+ENV PATH=$PATH:/project/sawtooth-core/bin:/protoc3/bin:/project/sawtooth-core/bin:/root/.cargo/bin \
```
delventhalz: Why don't we build the cargo dependencies here. That way you don't have to on every `up`. It can cut like 90 seconds off of the process.

Supply chain does it here:
https://github.com/hyperledger/sawtooth-supply-chain/blob/master/processor/Dockerfile#L35

rbuysse:
The .env is in PR #1612. I expected that one would get merged first. Apologies for not making that explicit.

validator/Dockerfile
aludvik: The "mounted" images should not COPY any source code nor should they have to be rebuilt when the source changes. They should only have to be rebuilt when dependencies change. I think a better solution to this problem is to use a persistent volume that you mount when you run the container. That way you aren't caching possibly outdated manifest files or dependencies in the image itself. Something like the following if you are using plain docker:
```
docker volume create cargo-registry && docker volume create validator-target;
docker run -it -v cargo-registry:/root/.cargo/registry -v validator-target:/project/sawtooth-core/validator/target sawtooth-validator
```

docker-compose.yaml
aludvik: While I think that is a cool way to avoid cluttering up your directory locally, I think we should keep the default behavior as simple and intuitive as possible. This is the "mounted" docker-compose file so I would expect it to just mount the directory, not mount the directory except directories A, B, C, etc. Now we are asking people to dig through the compose file to see what directories really are mounted and which are masked.

It also seems like starting down this road would open the door for extending this list every time someone decided they did not want a particular directory modified locally. Seems better to have a simple default and let people customize the behavior based their preferences.

delventhalz: I think "don't mount generated files" is a pretty unambiguous rule, and does not open the door to any extensions (unless they are more generated files). 

More important than spamming repos with unusable generated files, if people are doing local work you will overwrite their generated files, causing difficult to diagnose bugs. The reverse is true as well. Running `cargo build` on your host machine will break a currently running container. I think preventing these bugs is more important than the confusion that may result if people intentionally try to use the target directory as a mounted directory. I don't understand what the use case for that would be, so why would it ever come up?

validator/Dockerfile
delventhalz: I don't see how including dependencies in the form of a `COPY`ed `Cargo.toml` is functionally different than including dependencies as hard-coded `apt-get` commands. Either way, if the dependency changes, the docker image will need to be rebuilt.

Also, it is worth pointing out that _not_ rebuilding the Dockerfile will _not_ break the build. The out-of-date cached dependencies will simply be installed when you `cargo build` during `up`. If you do not build when building the image, you will have to install _every_ dependency on `up`, so I feel like installing the dependencies in the image still has the advantage here.

I am not familiar with the commands you listed. If they make it so you can _just_ build your changes when you `up`, and not have to reinstall and rebuild all of your Rust dependencies, then I'm down.

docker-compose.yaml
aludvik: I think that what are describing is a matter of opinion. My understanding of the goal of this update to the build system is to make the way it works as simple and intuitive as possible. The goal of the "mounted" Dockerfile is to provide an environment that you can mount the source directory into and run it. The simplest and most intuitive thing that can be done to achieve this goal is to mount the source repo as-is. If there are users who have more complex needs then I think they should solve them as needed, rather than try to accommodate every need out of the box. The "accommodate everyone" philosophy is what got the build tools into the nasty state they are in now in the first place.

aludvik: In other words, there is nothing stopping you from adding that line yourself when you are running things locally, if that is the behavior you want. I just think the out of the box behavior should be as simple and intuitive as possible.

PR #1617, opened 2018-05-03
Write change log on Merkle Trie Changes

peterschwarz:
Write a change log entry for each new merkle trie that records the additions for a given root and the deletions made by a successor trie.

This includes:
- some changes to have the hashes stored internally as bytes (though this is not complete)
- Refactor in LmdbDatabase to have a DatabaseReader trait
- ChangeLogEntry protobuf message

### DIFF
--- a/validator/src/state/merkle.rs
+++ b/validator/src/state/merkle.rs
@@ -17,6 +17,7 @@
 
 use std::collections::BTreeMap;
 use std::collections::HashMap;
+use std::collections::HashSet;
 use std::collections::VecDeque;
 use std::io::Cursor;
 
@@ -31,10 +32,20 @@ use cbor::value::Text;
 use crypto::digest::Digest;
 use crypto::sha2::Sha512;
 
+use protobuf;
+use protobuf::Message;
+use protobuf::ProtobufError;
+
 use database::database::DatabaseError;
 use database::lmdb::LmdbDatabase;
+use database::lmdb::DatabaseReader;
+
+use proto::merkle::ChangeLogEntry;
+use proto::merkle::ChangeLogEntry_Successor;
 
 const TOKEN_SIZE: usize = 2;
+
+pub const CHANGE_LOG_INDEX: &str = "change_log";
 
 /// Merkle Database
 pub struct MerkleDatabase {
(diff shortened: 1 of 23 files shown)


validator/src/state/merkle.rs
```diff
@@ -660,7 +660,7 @@ mod tests {
 
     #[test]
     fn node_deserialize() {
-        let packed = from_hex("a26163a162303063616263617647676f6f64627965");
+        let packed = ::hex::decode("a26163a162303063616263617647676f6f64627965").expect("proper hex");
```
aludvik: expect() should take a message indicating what the error was, so this should say "improper hex" not "proper hex". https://doc.rust-lang.org/std/result/enum.Result.html#examples-17

validator/src/state/merkle.rs
```diff
@@ -140,7 +140,7 @@ impl MerkleDatabase {
 
         let mut batch = Vec::with_capacity(path_addresses.len());
         // initializing this to empty, to make the compiler happy
-        let mut key_hash = String::new();
+        let mut key_hash = Vec::with_capacity(0);
```
aludvik: Why not `Vec::new()`?

validator/src/state/merkle.rs
```diff
@@ -628,6 +743,20 @@ mod tests {
...
             );
 
+            let change_log: ChangeLogEntry = {
+                // check that we have a change log entry for the new root
+                let reader = db.reader().unwrap();
+                protobuf::parse_from_bytes(&reader
+                    .index_get(CHANGE_LOG_INDEX, new_root_bytes)
+                    .expect("No db errors")
+                    .expect("A change log entry"))
+                    .expect("The change log entry to have bytes")
```
aludvik: expect should take an error message, not what you expect

peterschwarz: Base on the previous comment I will update this, but in other instances I've used functions like  'expect', the whole statement reads like a sentence: Expect no db errors, expect a change log entry, etc. But the way it is printed (which I don't particularly) looks like it loses this context.

validator/src/state/merkle.rs
peterschwarz: To me, Vec::new() implies that values will be pushed on to this instance.  In this case, it's a placeholder to make the compiler happy.  What is being mutated about that variable is what it points to, not the contents of the vector.  This makes that explicit.

peterschwarz:
Corrected a grammatical error commented on in PR #1626

PR #1619, opened 2018-05-04
Beginning implementation of the BlockManager

boydjohnson:
~This PR implements `put` and `unref` for the BlockManager.~ This PR implements `put`, `unref`, `get`, and ` branch`. Further methods will be added to get blocks, view branch diffs, and persist blocks.

~I am trying to make PRs of a length that people can review them easily, and so only implemented these two methods.~

### DIFF
--- a/validator/src/block.rs
+++ b/validator/src/block.rs
@@ -14,23 +14,17 @@
  * limitations under the License.
  * ------------------------------------------------------------------------------
  */
-extern crate cbor;
-extern crate cpython;
-extern crate crypto;
-extern crate libc;
-extern crate lmdb_zero;
-extern crate protobuf;
 
-#[macro_use]
-extern crate log;
-#[cfg(test)]
-extern crate rand;
+use batch::Batch;
 
-// exported modules
-pub mod database;
-pub mod proto;
-pub mod state;
-
-pub mod batch;
-pub mod block;
-pub mod transaction;
+#[derive(Clone, Debug, PartialEq)]
+pub struct Block {
+    pub header_signature: String,
+    pub batches: Vec<Batch>,
+    pub state_root_hash: String,
+    pub consensus: Vec<u8>,
+    pub batch_ids: Vec<String>,
+    pub signer_public_key: String,
+    pub previous_block_id: String,
+    pub block_num: u64,
+}
--- a/validator/src/journal/block_manager.rs
+++ b/validator/src/journal/block_manager.rs
@@ -16,6 +16,8 @@
  */
 
 use std::collections::HashMap;
+use std::rc::Rc;
+use std::sync::RwLock;
 
 use block::Block;
 use journal::{BlockStore, NULL_BLOCK_IDENTIFIER};
@@ -35,10 +37,10 @@ struct Anchor {
 
 #[derive(Default)]
 pub struct BlockManager {
-    block_by_block_id: HashMap<String, (Block, i64)>,
-    blockstore_by_name: HashMap<String, Box<BlockStore>>,
+    block_by_block_id: Rc<RwLock<HashMap<String, (Block, i64)>>>,
+    blockstore_by_name: Rc<RwLock<HashMap<String, Box<BlockStore>>>>,
 
-    anchors: HashMap<String, Anchor>,
+    anchors: Rc<RwLock<HashMap<String, Anchor>>>,
 }
 
 /// The BlockManager maintains integrity of all the blocks it contains,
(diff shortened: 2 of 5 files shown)


boydjohnson:
I am going to update this PR with the full trait BlockManager and unimplemented!() methods for the methods of the trait that aren't implemented yet.

validator/src/coretypes/batch.rs
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ------------------------------------------------------------------------------
+ */
+
+use coretypes::Transaction;
+
+#[derive(Clone)]
```
aludvik: We probably want to derive Debug on these too

validator/src/lib.rs
```diff
@@ -31,3 +31,4 @@ pub mod coretypes;
 pub mod database;
 pub mod proto;
 pub mod state;
+pub mod storage;
```
aludvik: It's probably better to follow the existing organization rather than invite discussion over a new reorganization.

validator/src/storage/mod.rs
```diff
@@ -33,3 +30,31 @@ pub trait BlockStore {
...
     fn iter(&self) -> Box<Iterator<Item = Block>>;
 }
+
+#[derive(Debug, PartialEq)]
+pub enum BlockManagerError {
+    MissingPredecessor(String),
+    MissingInput,
+}
+
+pub trait BlockManager {
```
aludvik: Are there going to be multiple implementations of the BlockManager?

validator/src/storage/mod.rs
```diff
@@ -33,3 +30,31 @@ pub trait BlockStore {
 
     fn iter(&self) -> Box<Iterator<Item = Block>>;
 }
+
+#[derive(Debug, PartialEq)]
```
aludvik: Why the PartialEq?

validator/src/storage/unsync_block_manager.rs
```diff
@@ -0,0 +1,236 @@
...
+use coretypes::Block;
+use storage::{BlockManager, BlockManagerError, BlockStore, NULL_BLOCK_IDENTIFIER};
+
+struct Anchor {
+    pub blockstore: String,
+    pub ref_count: i64,
+    pub block_id: String,
+}
+
+pub struct UnsyncBlockManager {
```
aludvik: If BlockManager isn't a trait, we can just call this BlockManager

validator/src/storage/unsync_block_manager.rs
```diff
@@ -0,0 +1,236 @@
...
+
+    fn garbage_collect(&mut self) {
+        let blocks_to_remove: Vec<String> = self.block_by_block_id
+            .iter()
+            .filter(|&(_, &(_, count))| count <= 0)
+            .map(|(block_id, &(_, _))| block_id.to_owned())
+            .collect();
+        blocks_to_remove.iter().for_each(|block_id| {
+            self.block_by_block_id.remove(block_id);
+        });
```
aludvik: I think you can simplify this by removing the collect() just continuing to use the first iterator

validator/src/coretypes/batch.rs
boydjohnson: Sounds good.

validator/src/lib.rs
boydjohnson: Yeah, I believe you are right.

validator/src/storage/mod.rs
boydjohnson: I was thinking that the unsync and the sync versions of the block manager would both implement the trait. I agree that it is probably best to not have it be a trait.

validator/src/storage/unsync_block_manager.rs
boydjohnson: The problem with that is that `self.block_by_block_id` is being immutably borrowed in the iterator and mutably borrowed in the for_each.

validator/src/journal/block_manager.rs
```diff
@@ -0,0 +1,448 @@
...
+    pub fn new() -> Self {
+        BlockManager {
+            block_by_block_id: Rc::new(RwLock::new(HashMap::new())),
+            blockstore_by_name: Rc::new(RwLock::new(HashMap::new())),
+
+            anchors: Rc::new(RwLock::new(HashMap::new())),
+        }
+    }
+
+    fn garbage_collect(&mut self) {
```
boydjohnson: I realize now that this garbage_collect function isn't written right. For every block that is removed, the previous_block_id block should have it's ref-count reduced.

validator/src/block.rs
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ------------------------------------------------------------------------------
+ */
+
+use batch::Batch;
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct Block {
```
aludvik: I wonder if these types should have a header field containing the data that needs to be in the header.

boydjohnson:
I am going to close this because I am very close to having a full implementation. I will open a new PR when I have the full implementation, probably tomorrow.

PR #1626, opened 2018-05-08
Implement Prune Functionality in MerkleDatabase

peterschwarz:
Implements the `prune` function on the MerkleDatabase. This function prunes a single trie, using its change log, by either:

- A parent trie is pruned based on the deletions that occurred with its only successor
- A trie with no successors is pruned by deleting all of the additions

If a trie has multiple successors, it is not deleted. 

Additionally, it adds the FFI layer for python access 

Note this is built on top of #1617

### DIFF
--- a/validator/src/state/merkle.rs
+++ b/validator/src/state/merkle.rs
@@ -39,6 +39,7 @@ use protobuf::ProtobufError;
 use database::database::DatabaseError;
 use database::lmdb::DatabaseReader;
 use database::lmdb::LmdbDatabase;
+use database::lmdb::LmdbDatabaseWriter;
 
 use proto::merkle::ChangeLogEntry;
 use proto::merkle::ChangeLogEntry_Successor;
@@ -61,6 +62,8 @@ pub enum MerkleDatabaseError {
     SerializationError(EncodeError),
     ChangeLogEncodingError(String),
     InvalidRecord,
+    InvalidHash(String),
+    InvalidChangeLogIndex(String),
     DatabaseError(DatabaseError),
     UnknownError,
 }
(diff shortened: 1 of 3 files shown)


protos/merkle.proto
```diff
@@ -0,0 +1,42 @@
...
+
+option java_multiple_files = true;
+option java_package = "sawtooth.sdk.protobuf";
+option go_package = "merkle_pb2";
+
+// An Entry in the change log for a given state root.
+message ChangeLogEntry {
+    // A state root that succeed this root.
+    message Successor {
+        // A root hash of a merkle trie based of this root.
```
jsmitchell: based on or based off, not based of

peterschwarz: This is part of PR #1617

validator/src/state/merkle.rs
```diff
@@ -112,6 +115,59 @@ impl MerkleDatabase {
         })
     }
 
+    pub fn prune(db: &LmdbDatabase, merkle_root: &str) -> Result<bool, MerkleDatabaseError> {
```
aludvik: Why did you choose to make this an associated function instead of a method? Also, I am not a huge fan of returning a bool here. Can we return something more interesting like what was pruned (https://rust-lang-nursery.github.io/api-guidelines/flexibility.html#functions-expose-intermediate-results-to-avoid-duplicate-work-c-intermediate) or use a custom enum type in place of bool to convey meaning and promote flexibility: https://rust-lang-nursery.github.io/api-guidelines/type-safety.html#arguments-convey-meaning-through-types-not-bool-or-option-c-custom-type

peterschwarz: I made it an associated function, as it's not a an operation on an individual trie, but on the database as a whole.  It doesn't need to be a method of MerkleDatabase, since it doesn't need a state root, or a root node or any of that.

I can change the return type to the addresses of what was pruned, if that is more interesting.  Since it operates on one root at a time, returning was it pruned or not is all the information there is, beyond errors.

validator/src/state/merkle.rs
```diff
@@ -112,6 +115,65 @@ impl MerkleDatabase {
...
+
+            if let Some(ref mut parent_change_log) =
+                get_change_log(&db_writer, parent_root_bytes)?.as_mut()
+            {
+                let successors = parent_change_log.take_successors();
+                let new_successors = successors
+                    .into_iter()
+                    .filter(|successor| root_bytes != successor.get_successor())
+                    .collect::<Vec<_>>();
+                parent_change_log.set_successors(protobuf::RepeatedField::from_vec(new_successors));
```
aludvik: This might be doing some extra allocations that we may want to optimize in the future.

peterschwarz: Currently, `RepeatedField` from the protobuf library doesn't give us a `retain` method (like `Vec`), which would be the equivalent operation without extra allocations.

This isn't a high-frequency operation (most likely once per block), though, so I'm not sure it's going to be a big optimization point.

PR #1632, opened 2018-05-09
docs: Clarify Docker Install Instructions

danintel:
1. Change the Docker Compose link to install instructions, rather than
just a listing of recent release files.

2. Prepend "sudo" to "docker-compose" command examples
This prevents errors with this hard-to-understand message:
    `ERROR: Couldn't connect to Docker daemon at
    http+docker://localhost - is it running? `

3. Add a step under "Step 7: Connect to Each Container" to list Docker
containers.

4. Prepend "sudo" to "docker" command examples


docs/source/app_developers_guide/docker.rst
```diff
@@ -141,11 +142,24 @@ To start the Sawtooth Docker environment, perform the following tasks:
 
    .. code-block:: console
 
-     user@host$ docker-compose -f sawtooth-default.yaml up
+     user@host$ sudo docker-compose -f sawtooth-default.yaml up
```
peterschwarz: We’ve been trying to avoid requiring the user making any calls with `sudo`. As far as I know, none of the team needs to do this, and your justification for doing this is preventing an error message that has to do with a users local docker setup, which it is not this document’s responsibility to train users on how to use docker.

danintel: That effectively requires giving root access to the client user, without the auditing benefits of sudo. The error message was there for a reason and there's a reason most or all distributions require using sudo with docker.

But I will make the change you suggest and add a note assuming an insecure setup.

danintel:
Removed sudo and added note on error from not running as root or in the docker group.

TomBarnes:
If the system administrator is expected to run sawtooth nodes with root privileges, then that is a non-starter from a security perspective.

dcmiddle:
There's a balance with tutorials in saying too much or too little. Without guidance on common missteps people can get stuck, but with too much extra guidance a simple activity looks complicated and possibly takes too long.
What do people think about moving this kind of help to something like a FAQ. Not a general FAQ but something targeted at this app dev guide?

chenette:
As far as a FAQ (or a section on troubleshooting) ... This is a not-insignificant time commitment, both in creation and maintenance. 

I agree with @dcmiddle that adding more info to a step, although very helpful, makes this procedure more complicated. It's a fine line between providing helpful tips and overwhelming the procedure.  However, tech writers usually have a few tricks up their sleeves... 

How about enclosing these new Docker tips in an actual ``.. tip::`` block, so that it would be easier for the eye to skip if the reader doesn't need the info?  This is the same as a ``.. note::``, but with the word "Tip".  Details here: http://docutils.sourceforge.net/docs/ref/rst/directives.html#tip

danintel:
I like the "Tip" suggestion and I can go along with, but I have an alternative--adding two preventive steps

1. Mention at the top, under Docker install, that docker and docker-compose must run with sudo or with a user in group docker.

2. Mention, in a step before `docker-compose ... up`, to run `docker-compose ... down` if docker was not previously cleanly shutdown.

I think the FAQ idea is good and should be added to as we go along--it doesn't need to be a dedicated effort.

chenette:
@danintel - Yes, I like this alternative. Go for it.  Thanks!

danintel:
I added a ``warning`` and a ``tip`` and I think I addressed everyone's concerns. Let me know.

docs/source/app_developers_guide/docker.rst
```diff
@@ -72,11 +72,19 @@ Compose.
...
 * Linux: Install the latest versions of
   `Docker Engine <https://docs.docker.com/engine/installation/linux/ubuntu>`_
-  and `Docker Compose <https://github.com/docker/compose/releases>`_.
+  and
+  `Docker Compose <https://docs.docker.com/compose/install/#install-compose>`_.
 
 In this procedure, you will open six terminal windows to connect to the Docker
 containers: one for each Sawtooth component and one to use for client commands.
 
+.. warning::
```
vaporos: This is Linux specific and should be considered out-of-scope for this document.

PR #1635, opened 2018-05-10
Implement consensus engine publishing services

aludvik:
This PR sets up a consensus engine message handlers, a skeleton consensus proxy, and implements the consensus engine publisher-related services. It also adds a provisional new type of bind argument type "consensus" which, when passed, enables the consensus engine feature. This argument may be merged with component in the future and is intended to allow for the development of the new consensus engine feature incrementally without breaking existing features.

### DIFF
--- a/validator/sawtooth_validator/consensus/handlers.py
+++ b/validator/sawtooth_validator/consensus/handlers.py
@@ -1,0 +1,269 @@
+# Copyright 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+
+
+import logging
+
+from google.protobuf.message import DecodeError
+
+from sawtooth_validator.protobuf import consensus_pb2
+from sawtooth_validator.protobuf import validator_pb2
+
+from sawtooth_validator.networking.dispatch import Handler
+from sawtooth_validator.networking.dispatch import HandlerResult
+from sawtooth_validator.networking.dispatch import HandlerStatus
+
+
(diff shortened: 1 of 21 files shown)


sdk/examples/devmode_rust/src/main.rs
```diff
@@ -0,0 +1,156 @@
...
+impl Devmode {
+    fn new(period: time::Duration) -> Self {
+        Devmode {
+            period: period,
+            exit: Exit::new(),
+        }
+    }
+}
+
+impl Engine for Devmode {
```
nick-drozd: If we're following the transaction handler model, the engine and `main` should be in separate files.

nick-drozd:
How does the CE block publisher differ from the normal one?

nick-drozd:
Is it possible at this point to run the engine in an integration test? It could be a simple smoke test, in the sense of "turn it on and verify that it doesn't start smoking".

validator/sawtooth_validator/journal/publisher_ce.py
```diff
@@ -0,0 +1,848 @@
...
+            self._max_batches == 0
+            or len(self._pending_batches) < self._max_batches
+        )
+
+    def _check_batch_dependencies(self, batch, committed_txn_cache):
+        """Check the dependencies for all transactions in this are present.
+        If all are present the committed_txn is updated with all txn in this
+        batch and True is returned. If they are not return failure and the
+        committed_txn is not updated.
+        :param batch: the batch to validate
```
boydjohnson: Are we keeping these docstrings in rst format because the rest of the journal/ files are also?

validator/sawtooth_validator/consensus/proxy.py
```diff
@@ -18,11 +18,16 @@
 LOGGER = logging.getLogger(__name__)
 
 
+class UnknownBlock(Exception):
```
nick-drozd: This exception already exists in the SDK: https://github.com/hyperledger/sawtooth-core/blob/b66c6d0f47c323efc3a3e6d81f9a0a42d7070733/sdk/python/sawtooth_sdk/consensus/exceptions.py#L25

validator/sawtooth_validator/consensus/handlers.py
```diff
@@ -148,9 +159,18 @@ def __init__(self, proxy):
     def handle_request(self, request, response):
         try:
             self._proxy.finalize_block(request.data)
-        except EmptyBlock:
+        except KeyError:
```
nick-drozd: I think this should be `UnknownBlock` instead of `KeyError`.

PR #1636, opened 2018-05-11
Make rust sdk compatible with protoc 1.6

RyanLassigBanks:
Wasn't sure if this had been addressed in some other way or at all.

`protoc_rust` version 1.6 removes the `MessageStatic` trait requires the `customize` field, and moves `RepeatedField`. As such the rust sdk fails to build on a fresh install.


grkvlt:
As an alternative, for the `1-0` branch, I have created #1640 which pins the rust dependency version for protobuf to 1.4.4 so that the breaking changes do not cause build issues.

PR #1637, opened 2018-05-11
Implement BlockManager in Rust

boydjohnson:
This work is based off of the ideas in this RFC pull request: https://github.com/hyperledger/sawtooth-rfcs/pull/5. 

~The persist method is not yet implemented, but I wanted to get eyes on the work so far.~

### DIFF
--- a/validator/src/journal/block_manager.rs
+++ b/validator/src/journal/block_manager.rs
@@ -868,6 +868,7 @@ mod tests {
     use super::{BlockManager, BlockManagerError};
     use block::Block;
     use journal::NULL_BLOCK_IDENTIFIER;
+    use journal::block_store::InMemoryBlockStore;
 
     fn create_block(header_signature: &str, previous_block_id: &str, block_num: u64) -> Block {
         Block {
@@ -1047,5 +1048,61 @@ mod tests {
         assert_eq!(branch_diff_iter4.next(), Some(&e));
         assert_eq!(branch_diff_iter4.next(), Some(&d));
         assert_eq!(branch_diff_iter4.next(), None);
+    }
+
+    #[test]
+    fn test_persist_ref_unref() {
+        let mut block_manager = BlockManager::new();
+
+        let a = create_block("A", NULL_BLOCK_IDENTIFIER, 0);
+        let b = create_block("B", "A", 1);
+        let c = create_block("C", "A", 1);
+        let d = create_block("D", "C", 2);
+        let e = create_block("E", "D", 3);
+        let f = create_block("F", "C", 2);
+
+        let q = create_block("Q", "F", 3);
+        let p = create_block("P", "E", 4);
+
+        block_manager.put(vec![a.clone(), b.clone()]).unwrap();
+        block_manager
+            .put(vec![c.clone(), d.clone(), e.clone()])
+            .unwrap();
+        block_manager.put(vec![f.clone()]).unwrap();
+
+        let blockstore = Box::new(InMemoryBlockStore::new());
+        block_manager.add_store("commit", blockstore).unwrap();
+
+        block_manager.persist("C", "commit").unwrap();
+        block_manager.persist("B", "commit").unwrap();
+
+        block_manager.ref_block("D").unwrap();
+
+        block_manager.persist("C", "commit").unwrap();
+        block_manager.unref_block("B").unwrap();
+
+        block_manager.persist("F", "commit").unwrap();
+        block_manager.persist("E", "commit").unwrap();
+
+        block_manager.unref_block("F").unwrap();
+        block_manager.unref_block("D").unwrap();
+
+        block_manager.persist("A", "commit").unwrap();
+
+        block_manager.unref_block("E").unwrap();
+
+        assert_eq!(
+            block_manager.put(vec![q]),
+            Err(BlockManagerError::MissingPredecessor(format!(
+                "During Put, missing predecessor of block Q: F"
+            )))
+        );
+
+        assert_eq!(
+            block_manager.put(vec![p]),
+            Err(BlockManagerError::MissingPredecessor(format!(
+                "During Put, missing predecessor of block P: E"
+            )))
+        );
     }
 }
--- a/validator/src/journal/block_manager.rs
+++ b/validator/src/journal/block_manager.rs
@@ -375,8 +375,8 @@ impl BlockManager {
             .map(|blockstore| blockstore.get(vec![block_id]).nth(0).unwrap())
     }
 
-    pub fn branch(&self, tip: &str) -> Box<Iterator<Item = Block>> {
-        unimplemented!()
+    pub fn branch<'a>(&'a self, tip: &str) -> Box<Iterator<Item = &Block> + 'a> {
+        Box::new(BranchIterator::new(self, tip.into()))
     }
 
     pub fn branch_diff(&self, tip: &str, exclude: &str) -> Box<Iterator<Item = Block>> {
@@ -553,6 +553,58 @@ impl<'a> Iterator for GetBlockIterator<'a> {
     }
 }
 
+struct BranchIterator<'a> {
+    block_manager: &'a BlockManager,
+    next_block_id: String,
+    blockstore: Option<String>,
+}
+
+impl<'a> BranchIterator<'a> {
+    pub fn new(block_manager: &'a BlockManager, first_block_id: String) -> Self {
+        BranchIterator {
+            block_manager,
+            next_block_id: first_block_id,
+            blockstore: None,
+        }
+    }
+}
+
+impl<'a> Iterator for BranchIterator<'a> {
+    type Item = &'a Block;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.next_block_id == NULL_BLOCK_IDENTIFIER {
+            None
+        } else if self.blockstore.is_none() {
+            let (block_option, blockstore_name_option) = self.block_manager
+                .get_block_from_main_cache_or_blockstore_name(self.next_block_id.as_str());
+            let block = block_option.or_else(|| {
+                blockstore_name_option
+                    .map(|blockstore_name| {
+                        self.blockstore = Some(blockstore_name.into());
+                        self.block_manager
+                            .get_block_from_blockstore(self.next_block_id.clone(), blockstore_name)
+                            .expect("Blockstore referenced by anchor does not exist")
+                    })
+                    .unwrap_or(None)
+            });
+            if let Some(block) = block {
+                self.next_block_id = block.previous_block_id.clone();
+            }
+            block
+        } else {
+            let blockstore_id = self.blockstore.as_ref().unwrap();
+            let block = self.block_manager
+                .get_block_from_blockstore(self.next_block_id.clone(), blockstore_id)
+                .expect("The Blockmanager has lost a blockstore that is referenced by an anchor")
+                .expect(
+                    "The block was not in the blockstore referenced by a successor block's anchor",
+                );
+            Some(block)
+        }
+    }
+}
+
 #[cfg(test)]
 mod tests {
 
@@ -680,5 +732,28 @@ mod tests {
         assert_eq!(get_block_iter.next(), Some(&c));
         assert_eq!(get_block_iter.next(), None);
         assert_eq!(get_block_iter.next(), None);
+    }
+
+    #[test]
+    fn test_branch_in_memory() {
+        let mut block_manager = BlockManager::new();
+        let a = create_block("A", NULL_BLOCK_IDENTIFIER, 0);
+        let b = create_block("B", "A", 1);
+        let c = create_block("C", "B", 2);
+
+        block_manager
+            .put(vec![a.clone(), b.clone(), c.clone()])
+            .unwrap();
+
+        let mut branch_iter = block_manager.branch("C");
+
+        assert_eq!(branch_iter.next(), Some(&c));
+        assert_eq!(branch_iter.next(), Some(&b));
+        assert_eq!(branch_iter.next(), Some(&a));
+        assert_eq!(branch_iter.next(), None);
+
+        let mut empty_iter = block_manager.branch("P");
+
+        assert_eq!(empty_iter.next(), None);
     }
 }
--- a/validator/src/journal/block_manager.rs
+++ b/validator/src/journal/block_manager.rs
@@ -20,7 +20,7 @@ use std::collections::HashSet;
 
 use block::Block;
 use journal::NULL_BLOCK_IDENTIFIER;
-use journal::block_store::BlockStore;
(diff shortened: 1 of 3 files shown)


validator/src/journal/block_store.rs
```diff
@@ -0,0 +1,96 @@
...
+        Ok(())
+    }
+
+    fn iter(&self) -> Box<Iterator<Item = Block>> {
+        unimplemented!()
+    }
+}
+
+struct GetBlockIterator<'a> {
+    blockstore: &'a InMemoryBlockStore,
```
peterschwarz: This probably shouldn't be tied to `InMemoryBlockStore`, or should be called `InMemoryBlockStoreIterator`

validator/src/journal/block_manager.rs
```diff
@@ -0,0 +1,214 @@
...
+pub enum BlockManagerError {
+    MissingPredecessor(String),
+    MissingPredecessorInBranch,
+    MissingInput,
+    UnknownBlock,
+    UnknownBlockStore,
+}
+
+struct Anchor {
+    pub blockstore: String,
```
peterschwarz: Probably would name this `blockstore_name`

validator/src/journal/block_manager.rs
```diff
@@ -388,6 +391,152 @@ impl<'a> Iterator for BranchIterator<'a> {
...
+        }
+
+        iterator
+    }
+
+    fn block_num_difference(&self, left: &Block, right: &Block) -> i64 {
+        left.block_num as i64 - right.block_num as i64
+    }
+
+    fn walk_back_1_from_block(&self, block_id: String) -> Option<&'a Block> {
```
peterschwarz: `get_previous_block` seems like a better name

validator/src/journal/block_manager.rs
```diff
@@ -388,6 +391,152 @@ impl<'a> Iterator for BranchIterator<'a> {
...
+        let block = self.block_manager.get_block_by_block_id(block_id.as_str());
+        block
+            .map(|b| {
+                self.block_manager
+                    .get_block_by_block_id(b.previous_block_id.as_str())
+            })
+            .unwrap_or(None)
+    }
+
+    fn walk_back_n_from_block(&self, block_id: String, n: u64) -> Option<&'a Block> {
```
peterschwarz: Likewise, `get_nth_previous_block`

validator/src/journal/block_store.rs
```diff
@@ -0,0 +1,96 @@
...
+    fn put(&mut self, blocks: Vec<Block>) -> Result<(), BlockStoreError> {
+        blocks.into_iter().for_each(|block| {
+            self.block_by_block_id
+                .insert(block.header_signature.clone(), block);
+        });
+        Ok(())
+    }
+
+    fn iter(&self) -> Box<Iterator<Item = Block>> {
+        unimplemented!()
```
peterschwarz: Doesn't look like you're using the Iterator below.

boydjohnson: The iterator below is used in `get`.

validator/src/journal/block_store.rs
boydjohnson: I can probably do &'a BlockStore, because all we care about is that it is a BlockStore.

validator/src/journal/block_manager.rs
boydjohnson: Cool, Thanks!

validator/src/journal/block_manager.rs
boydjohnson: Thanks, Good suggestion.

validator/src/journal/block_manager.rs
boydjohnson: Thanks, good suggestion.

validator/src/journal/block_manager.rs
```diff
@@ -0,0 +1,303 @@
...
+        block_id: &str,
+        blockstore_name: &str,
+        external_ref_count: i64,
+        internal_ref_count: i64,
+    ) {
+        let anchor = Anchor {
+            blockstore_name: blockstore_name.to_string(),
+            external_ref_count,
+            internal_ref_count,
+            block_id: block_id.to_string(),
```
aludvik: Per previous style discussions, prefer `into()` for string conversions like this.

validator/src/journal/block_manager.rs
```diff
@@ -0,0 +1,303 @@
...
+pub enum BlockManagerError {
+    MissingPredecessor(String),
+    MissingPredecessorInBranch,
+    MissingInput,
+    UnknownBlock,
+    UnknownBlockStore,
+}
+
+#[derive(Default)]
+struct Anchors {
```
aludvik: Please add a docstring explaining why this struct exists.

validator/src/journal/block_manager.rs
```diff
@@ -0,0 +1,303 @@
...
+    pub fn put(&mut self, branch: Vec<Block>) -> Result<(), BlockManagerError> {
+        match branch.split_first() {
+            Some((head, tail)) => {
+                let predecessor_of_head_in_memory =
+                    self.block_by_block_id.contains_key(&head.previous_block_id);
+
+                let predecessor_of_head_in_any_blockstore = self.blockstore_by_name.values().any(
+                    |blockstore| blockstore.get(vec![head.previous_block_id.clone()]).count() > 0,
+                );
+                let head_is_genesis = head.previous_block_id == NULL_BLOCK_IDENTIFIER;
```
aludvik: I wouldn't use blockchain terminology here, just think of it as a tree, so this would be the root. Also, I think it should be possible to avoid doing root-node-specific checks by making the root nodes anchors.

validator/src/journal/block_manager.rs
```diff
@@ -0,0 +1,303 @@
...
+
+                if !(predecessor_of_head_in_memory || predecessor_of_head_in_any_blockstore
+                    || head_is_genesis)
+                {
```
aludvik: This check should probably be refactored behind a contains function.

validator/src/journal/block_manager.rs
```diff
@@ -0,0 +1,303 @@
...
+use std::collections::HashMap;
+
+use block::Block;
+use journal::NULL_BLOCK_IDENTIFIER;
+use journal::block_store::BlockStore;
+
+#[derive(Debug, PartialEq)]
+pub enum BlockManagerError {
+    MissingPredecessor(String),
+    MissingPredecessorInBranch,
```
aludvik: This should probably take the predecessor that was missing.

validator/src/journal/block_manager.rs
```diff
@@ -0,0 +1,303 @@
...
+
+use std::collections::HashMap;
+
+use block::Block;
+use journal::NULL_BLOCK_IDENTIFIER;
+use journal::block_store::BlockStore;
+
+#[derive(Debug, PartialEq)]
+pub enum BlockManagerError {
+    MissingPredecessor(String),
```
aludvik: Similar to below, this should add a clarifying phrase about where the predecessor was expected.

validator/src/journal/block_manager.rs
boydjohnson: I'm not sure it is possible to do without the NULL_BLOCK_IDENTIFIER check. If the first put is `C 3, D 4` and with C pointing to an unknown block B is it ok or is it an error?

aludvik: I am thinking that when you create the block manager, rather than having anchors be empty, you can initialize it with an anchor to a block with id = NULL_BLOCK_IDENTIFIER

validator/src/journal/block_manager.rs
```diff
@@ -0,0 +1,1039 @@
...
+    pub blockstore_name: String,
+    pub external_ref_count: i64,
+    pub internal_ref_count: i64,
+    pub block_num: u64,
+    pub block_id: String,
+}
+
+#[derive(Default)]
+pub struct BlockManager {
+    block_by_block_id: HashMap<String, (Block, i64, i64)>,
```
aludvik: I would use a struct here instead of a tuple so it is clearer which i64 is for internal/external.

validator/src/journal/block_manager.rs
```diff
@@ -0,0 +1,1039 @@
...
+    pub fn unref_block(&mut self, tip: &str) -> Result<(), BlockManagerError> {
+        let ref_count_block_id_result = self.lower_tip_blocks_refcount(tip);
+
+        let mut blocks_to_remove = vec![];
+
+        let mut optional_new_tip = None;
+
+        let (external_ref_count, internal_ref_count, block_id) = ref_count_block_id_result?;
+
+        if external_ref_count <= 0 && internal_ref_count <= 0 {
```
aludvik: I think this should be split up into checking whether the count is less than 0, which should be a panic, and checking if the count is 0, which should cause a purge.

validator/src/journal/block_manager.rs
aludvik: Can we use a u64 here with checked subtraction to catch when we try to unref below 0? https://doc.rust-lang.org/std/primitive.u64.html#method.checked_sub

validator/src/journal/block_store.rs
aludvik: Is this going to be implemented, or is it intentionally unimplemented?

validator/src/journal/block_manager.rs
```diff
@@ -0,0 +1,1039 @@
...
+    ///     - If branch is empty, an error is returned
+    pub fn put(&mut self, branch: Vec<Block>) -> Result<(), BlockManagerError> {
+        match branch.split_first() {
+            Some((head, tail)) => {
+                if !self.contains(head.previous_block_id.as_str()) {
+                    return Err(BlockManagerError::MissingPredecessor(format!(
+                        "During Put, missing predecessor of block {}: {}",
+                        head.header_signature, head.previous_block_id
+                    )));
+                } else {
```
aludvik: This doesn't need to be else, since you are returning above.

validator/src/journal/block_manager.rs
```diff
@@ -0,0 +1,1039 @@
...
+                            Some(previous_block_id) => {
+                                if block.previous_block_id != previous_block_id {
+                                    return Err(BlockManagerError::MissingPredecessorInBranch(
+                                        format!(
+                                            "During Put, missing predecessor of block {}: {}",
+                                            block.header_signature, block.previous_block_id
+                                        ),
+                                    ));
+                                }
+                                previous = Some(block.header_signature.clone());
```
aludvik: This clone seems avoidable

validator/src/journal/block_manager.rs
```diff
@@ -0,0 +1,1039 @@
...
+            Some((head, tail)) => {
+                if !self.contains(head.previous_block_id.as_str()) {
+                    return Err(BlockManagerError::MissingPredecessor(format!(
+                        "During Put, missing predecessor of block {}: {}",
+                        head.header_signature, head.previous_block_id
+                    )));
+                } else {
+                    let mut previous = None;
+
+                    for block in tail {
```
aludvik: This internal branch integrity check feels like a good candidate for refactoring into a separate private helper function with a descriptive name so it is easier to tell what is going on.

validator/src/journal/block_manager.rs
```diff
@@ -0,0 +1,1039 @@
...
+            }
+            None => return Err(BlockManagerError::MissingInput),
+        }
+        let blocks_not_added_yet: Vec<Block> = branch
+            .into_iter()
+            .filter(|block| {
+                !self.block_by_block_id.contains_key(&block.header_signature)
+                    && !self.blockstore_by_name.values().any(|blockstore| {
+                        blockstore.get(vec![block.header_signature.clone()]).count() > 0
+                    })
```
aludvik: Can you just use `self.contains()` here?

validator/src/journal/block_manager.rs
```diff
@@ -0,0 +1,1039 @@
...
+            })
+            .collect();
```
aludvik: Can you avoid the extra vector allocation here by chaining a for_each() here instead? The you can just do `if let Some(last_block) = branch.last()` after?

validator/src/journal/block_manager.rs
```diff
@@ -0,0 +1,1039 @@
...
+
+        if let Some((last_block, blocks)) = blocks_not_added_yet.split_last() {
+            self.block_by_block_id.insert(
+                last_block.header_signature.clone(),
+                (last_block.clone(), 0, 0),
```
aludvik: I think this needs an external reference count of 1 to start.

validator/src/journal/block_manager.rs
```diff
@@ -423,6 +423,58 @@ impl<'a> Iterator for GetBlockIterator<'a> {
...
 }
 
+struct BranchIterator<'a> {
+    block_manager: &'a BlockManager,
+    next_block_id: String,
+    blockstore: Option<String>,
+}
+
+impl<'a> BranchIterator<'a> {
+    pub fn new(block_manager: &'a BlockManager, first_block_id: String) -> Self {
```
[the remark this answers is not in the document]
boydjohnson: Ah, that makes sense, so that blocks aren't being dropped while the Iterator is iterating.

validator/src/journal/block_store.rs
boydjohnson: I didn't need it for any of the work I had done so far, so I didn't implement it. Can you talk about iterating through a blockstore that wouldn't be like `branch` or `branch_diff` where you are starting from some known tip?

validator/src/journal/block_manager.rs
boydjohnson: Just so I understand, the blocks that go in with an internal reference count of 0 as well as ones that have an internal reference count of 1 would all have an external reference count of 1 to start.

validator/src/journal/block_manager.rs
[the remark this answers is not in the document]
boydjohnson: @aludvik That will change the `get` method from `&self` to `&mut self`. does that seem ok?

validator/src/journal/block_manager.rs
```diff
@@ -0,0 +1,1110 @@
...
+    /// Starting at a tip block, if the tip block's ref-count drops to 0,
+    /// remove all blocks until a ref-count of 2 is found.
+    pub fn unref_block(&mut self, tip: &str) -> Result<(), BlockManagerError> {
+        let ref_count_block_id_result = self.lower_tip_blocks_refcount(tip);
+
+        let mut blocks_to_remove = vec![];
+
+        let mut optional_new_tip = None;
+
+        let (external_ref_count, internal_ref_count, block_id) = ref_count_block_id_result?;
```
aludvik: Is there a reason to separate these lines instead of just doing: `let (external_ref_count, internal_ref_count, block_id) = self.lower_tip_blocks_refcount(tip)?;`?

PR #1638, opened 2018-05-11
docs: Clarify Ubuntu Install Instructions

danintel:
1. Remove forward reference to `user@validator$ prompt`

2. Move instructions for creating `user@client$` window before first use.


docs/source/app_developers_guide/ubuntu.rst
```diff
@@ -181,7 +181,7 @@ window as the previous step.
 
 .. code-block:: console
 
-   user@validator$ sudo sawadm keygen
+   $ sudo sawadm keygen
```
askmish: The `user@validator` prompt is a useful indicator that the commands should be run for validator only. We're following the same standard below for the client term window. Instead of removing it, the rest of the document should adopt this standard. Especailly the genesis and this sawadm commands are very specific for validators only.

vaporos: +1 for consistency

danintel: I added use@validator to all the preceding prompts, as requested.

dcmiddle:
@askmish I believe your feedback is addressed, can you please verify and :check: if you agree. :)

askmish:
@dcmiddle I'm hoping the console prompt will be useful to distinguish which command belongs to validators, general cli, etc.

PR #1639, opened 2018-05-12
Fix intermittent test failures

aludvik:


nick-drozd:
In the check block num commit message, **passed** -> **past**.

PR #1642, opened 2018-05-14
Fix validator peering on Azure

peterschwarz:
Validator peering fails when using Azure endpoints such as tcp://hashblock3.westus2.cloudapp.azure.com:8800 because the 'lo' network interface characters are characters in 'cloudapp'. This fix test that the network interface string is not equal to the

Replaces #1592 and corrects the logging usage.

### DIFF
--- a/validator/sawtooth_validator/networking/handlers.py
+++ b/validator/sawtooth_validator/networking/handlers.py
@@ -15,6 +15,9 @@
 import logging
 import os
 import time
+
+from urllib.parse import urlparse
+
 import netifaces
 
 from sawtooth_signing import create_context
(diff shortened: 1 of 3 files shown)


validator/sawtooth_validator/networking/handlers.py
```diff
@@ -58,6 +61,22 @@ class ConnectHandler(Handler):
     def __init__(self, network):
         self._network = network
 
+    @staticmethod
+    def is_valid_endpoint_host(interfaces, endpoint):
+        """
+        An enpoint host name is valid if it is a URL and if the
```
jsmitchell: s/enpoint/endpoint/

PR #1649, opened 2018-05-16
Devmode consensus engine

nick-drozd:
This PR contains:

1. A series of commits that make minor changes to the consensus SDK (changing function signatures, renaming things, fixing bugs, etc).

2. A commit to refactor the publishing logic in the existing Devmode Python code. This isn't strictly necessary, but it's helpful to be able to see exactly what Devmode does.

3. The Devmode consensus engine. This cannot currently be run in master.


nick-drozd:
Need real error handling

sdk/examples/devmode_rust/src/engine.rs
```diff
@@ -0,0 +1,221 @@
...
+use rand;
+use rand::Rng;
+
+use sawtooth_sdk::consensus::{engine::*, service::Service};
+
+pub struct DevmodeService {
+    pub service: Box<Service>,
+}
+
+impl DevmodeService {
```
peterschwarz: What is the intention behind this service wrapper?  It looks like it's just ignoring any errors that could occur.

sdk/examples/devmode_rust/src/engine.rs
```diff
@@ -0,0 +1,221 @@
...
+        DevmodeService { service }
+    }
+
+    fn get_chain_head(&mut self) -> Block {
+        self.service
+            .get_chain_head()
+            .expect("Failed to get chain head")
+    }
+
+    fn get_block(&mut self, block_id: BlockId) -> Block {
```
peterschwarz: This should be a `&BlockId` param, since you're not consuming it.

sdk/examples/devmode_rust/src/engine.rs
```diff
@@ -0,0 +1,221 @@
...
+                    Update::BlockNew(block) => {
+                        if check_consensus(&block) {
+                            service.check_block(block.block_id);
+                        } else {
+                            service.fail_block(block.block_id);
+                        }
+                    }
+
+                    Update::BlockValid(block_id) => {
+                        let block = service.get_block(block_id.clone());
```
peterschwarz: If these service methods take references, you don't need this `clone`

sdk/rust/src/consensus/engine.rs
```diff
@@ -32,7 +32,7 @@ pub enum Update {
     BlockCommit(BlockId),
 }
 
-#[derive(Default, Debug)]
+#[derive(Clone, Default, Debug)]
```
peterschwarz: This should probably be a separate commit.

sdk/examples/devmode_rust/src/engine.rs
nick-drozd: The primary motivation was to avoid duplicating the consensus data in `finalize_block` and the vector handling in `check_block` and `get_block`. The others are just to keep things consistent.

sdk/examples/devmode_rust/src/engine.rs
nick-drozd: `Service.get_blocks` takes a vector of block ids, I think consuming them.

sdk/rust/src/consensus/engine.rs
nick-drozd: Agreed

sdk/rust/src/consensus/engine.rs
```diff
@@ -32,7 +32,7 @@ pub enum Update {
     BlockCommit(BlockId),
 }
 
-#[derive(Default, Debug, Eq, Hash, PartialEq)]
+#[derive(Clone, Default, Debug, Eq, Hash, PartialEq, PartialOrd)]
```
aludvik: These derives should go in a separate commit.

protos/consensus.proto
```diff
@@ -373,6 +373,24 @@ message ConsensusBlocksGetResponse {
   repeated ConsensusBlock blocks = 2;
 }
 
+// Retrieve consensus-related information about the chain head.
+message ConsensusChainHeadGetRequest {}
```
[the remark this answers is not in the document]
nick-drozd: added

sdk/rust/src/consensus/engine.rs
nick-drozd: separated

sdk/examples/devmode_rust/src/engine.rs
```diff
@@ -0,0 +1,293 @@
...
+        }
+
+        query_result.expect("Failed to finalize block");
+    }
+
+    fn check_block(&mut self, block_id: BlockId) {
+        debug!("Checking block {:?}", block_id);
+        self.service
+            .check_blocks(vec![block_id])
+            .expect("Failed to check block");
```
peterschwarz: In a future commit, we should fix these expects - there are many reasons this could fail, and probably shouldn't be considered a `panic` situation.

nick-drozd: Agreed. I think more experience with writing engines will help us to figure out what to do with various error situations. One solution might be to limit the possible errors that an engine can be confronted with by having the service do more error handling, e.g. resending the request if the response status is `NOT_READY`.

PR #1650, opened 2018-05-16
Implement Identity View in Rust

peterschwarz:
Implements the IdentityView with tests.  It provides views on roles and policies over a StateReader.

Completes [STL-1184](https://jira.hyperledger.org/browse/STL-1184).


peterschwarz:
Updated with doc comments.

PR #1653, opened 2018-05-17
Fix publisher batch queue bug

aludvik:


aludvik:
Passed LR1 stability checks

PR #1655, opened 2018-05-17
Update protoc to 2.0

RyanLassigBanks:

### DIFF
--- a/sdk/rust/Cargo.toml
+++ b/sdk/rust/Cargo.toml
@@ -25,7 +25,7 @@ default = ["pem"]
 pem = []
 
 [dependencies]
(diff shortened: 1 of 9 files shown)


sdk/rust/Cargo.toml
```diff
@@ -25,7 +25,7 @@ default = ["pem"]
 pem = []
 
 [dependencies]
-protobuf="1.4.1"
+protobuf="2.0"
```
vaporos: We should use protobuf = "2" here instead. If we do = "2.0", I think we explicitly prevent using the SDK with 2.1+. Probably would be okay to use = "2" in the other places as well.

PR #1659, opened 2018-05-18
Publisher pending batches pool

RyanLassigBanks:
Port of PublishPendingBatches object for rust


validator/src/journal/publisher.rs
```diff
@@ -1,4 +1,107 @@
...
+        };
+
+        self.reset();
+
+        for batch in unsent {
+            self.append(batch);
+        }
+    }
+
+    pub fn update_limit(&mut self, consumed: i32) {
```
peterschwarz: Why i32 vs usize?  You can't consume negative values.

validator/src/journal/publisher.rs
```diff
@@ -1,4 +1,107 @@
...
+    pub fn update(&mut self, mut still_pending: Vec<Batch>, last_sent: Batch) {
+        let last_index = self.batches
+            .iter()
+            .position(|i| i.get_header_signature() == last_sent.get_header_signature());
+
+        let unsent = if let Some(idx) = last_index {
+            still_pending.extend_from_slice(self.batches.clone().split_off(idx + 1).as_slice());
+            still_pending
+        } else {
+            self.batches.clone()
```
aludvik: This is going to be a very expensive clone, as this list can have as many as 100K batches in it at once. Instead, you should use `std::mem::swap` (https://doc.rust-lang.org/std/mem/fn.swap.html) to take the batch list out and replace it with an empty one.

validator/src/journal/publisher.rs
```diff
@@ -1,4 +1,107 @@
...
+
+        self.reset();
+
+        for batch in unsent {
+            self.append(batch);
+        }
+    }
+
+    pub fn update_limit(&mut self, consumed: i32) {
+        self.limit.update(self.batches.len() as i32, consumed);
```
aludvik: These should probably be isize instead of i32 since they are tracking the length of a vector which is isize.

validator/src/journal/publisher.rs
```diff
@@ -1,12 +1,124 @@
-use std::collections::VecDeque;
+use proto::batch::Batch;
+use std::collections::{HashSet, VecDeque};
+use std::mem;
+use std::slice::Iter;
+
+/// Ordered batches wating to be processed
```
boydjohnson: Hi @RyanLassigBanks , I accidentally put this on #1660 which is built off of this one. `wating` is missing an `i`: `waiting`.

RyanLassigBanks: Fixed in 1660

PR #1660, opened 2018-05-19
Incoming batch queue

RyanLassigBanks:
Rust port of `IncomingBatchQueue`

### DIFF
--- a/validator/src/journal/publisher.rs
+++ b/validator/src/journal/publisher.rs
@@ -2,8 +2,90 @@ use proto::batch::Batch;
 use std::collections::{HashSet, VecDeque};
 use std::mem;
 use std::slice::Iter;
+use std::sync::mpsc::{channel, Receiver, RecvTimeoutError, SendError, Sender};
+use std::sync::{Arc, Mutex, MutexGuard, PoisonError};
+use std::time::Duration;
 
-/// Ordered batches wating to be processed
+/// This queue keeps track of the batch ids so that components on the edge
+/// can filter out duplicates early. However, there is still an opportunity for
+/// duplicates to make it into this queue, which is intentional to avoid
+/// blocking threads trying to put/get from the queue. Any duplicates
+/// introduced by this must be filtered out later.
+pub fn make_batch_queue() -> (IncomingBatchSender, IncomingBatchReceiver) {
+    let (sender, reciever) = channel();
+    let ids = Arc::new(Mutex::new(HashSet::new()));
+    (
+        IncomingBatchSender::new(ids.clone(), sender),
+        IncomingBatchReceiver::new(ids, reciever),
+    )
+}
+
+pub struct IncomingBatchReceiver {
+    ids: Arc<Mutex<HashSet<String>>>,
+    receiver: Receiver<Batch>,
+}
+
+impl IncomingBatchReceiver {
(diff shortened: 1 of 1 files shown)


validator/src/journal/publisher.rs
```diff
@@ -1,4 +1,159 @@
...
+        })
+    }
+}
+
+pub struct IncomingBatchSender {
+    ids: Arc<HashSet<String>>,
+    sender: Sender<Batch>,
+}
+
+impl IncomingBatchSender {
```
boydjohnson: This would be accessed by the completer. I would expect that something would have an `extern "C"` on the function.

peterschwarz: Don't worry about that, we'll get it as part of building the FFI layer for these structs

validator/src/journal/publisher.rs
```diff
@@ -1,4 +1,159 @@
...
+    ids: Arc<HashSet<String>>,
+    receiver: Receiver<Batch>,
+}
+
+impl IncomingBatchReceiver {
+    pub fn new(ids: Arc<HashSet<String>>, receiver: Receiver<Batch>) -> IncomingBatchReceiver {
+        IncomingBatchReceiver { ids, receiver }
+    }
+
+    pub fn get(&mut self) -> Result<Batch, RecvError> {
```
boydjohnson: I think this signature should be `pub fn get(&mut self, timeout: Duration) -> Result<Batch, RecvError>`. and then instead of `.recv` it would be `recv_timeout`: https://doc.rust-lang.org/std/sync/mpsc/struct.Receiver.html#method.recv_timeout

validator/src/journal/publisher.rs
```diff
@@ -1,6 +1,58 @@
...
+use std::sync::mpsc::{channel, Receiver, RecvError, SendError, Sender};
+
+/// This queue keeps track of the batch ids so that components on the edge
+/// can filter out duplicates early. However, there is still an opportunity for
+/// duplicates to make it into this queue, which is intentional to avoid
+/// blocking threads trying to put/get from the queue. Any duplicates
+/// introduced by this must be filtered out later.
+pub fn make_batch_queue() -> (IncomingBatchSender, IncomingBatchReceiver) {
+    let (sender, reciever) = channel();
+    let ids = Arc::new(HashSet::new());
```
peterschwarz: This needs to be an `Arc<Mutex<_>>`, since the the set is modifiable on multiple threads

validator/src/journal/publisher.rs
```diff
@@ -1,6 +1,58 @@
...
+    sender: Sender<Batch>,
+}
+
+impl IncomingBatchSender {
+    pub fn new(ids: Arc<HashSet<String>>, sender: Sender<Batch>) -> IncomingBatchSender {
+        IncomingBatchSender { ids, sender }
+    }
+    pub fn put(&mut self, batch: Batch) -> Result<(), SendError<Batch>> {
+        if !self.ids.contains(batch.get_header_signature()) {
+            self.sender.send(batch)
```
peterschwarz: You need to insert the header signature into the ids set.

peterschwarz:
This will need to get rebased after #1659 is merged.

PR #1661, opened 2018-05-20
Add TransactionCommitCache in Rust to store transactions committed

boydjohnson:
This work is part of porting Candidate block to rust.

### DIFF
--- a/validator/src/journal/chain_commit_state.rs
+++ b/validator/src/journal/chain_commit_state.rs
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2018 Intel Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


validator/src/journal/chain_commit_state.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    pub fn remove_batch(&mut self, batch: &Batch) {
+        batch
+            .transactions
+            .iter()
+            .for_each(|txn| self.remove(txn.header_signature.as_str()));
+    }
+
+    pub fn contains(&self, transaction_id: &str) -> bool {
+        unsafe {
```
aludvik: You should minimize the scope of any unsafe blocks like this. You should be able to do:
`let py = unsafe { cpython::Python::assume_gil_acquired() };

boydjohnson: Cool, thanks.

PR #1662, opened 2018-05-21
Add ChainObserver trait

peterschwarz:
Add the ChainObserver trait, which is will be used by the chain
controller to notify components of chain updates.


peterschwarz:
This isn't the right interface needed for the publisher.

PR #1665, opened 2018-05-22
Add ExecutionResult and TxnExecution Rust data objects and python conversion

boydjohnson:


validator/src/scheduler/execution_result_ffi.rs
```diff
@@ -0,0 +1,227 @@
...
+    pub is_valid: bool,
+    pub state_hash: String,
+    pub state_changes: Vec<StateChange>,
+    pub events: Vec<Event>,
+    pub data: Vec<(String, Vec<u8>)>,
+    pub error_message: String,
+    pub error_data: Vec<u8>,
+}
+
+pub trait TryFrom<T, E>
```
aludvik: I think you should implement FromPyObject and ToPyObject for these instead.

PR #1667, opened 2018-05-22
Create Rust wrapper for metrics library

aludvik:
This is a thin wrapper around the sawtooth_validator metrics library code.

Based on #1666

### DIFF
--- a/validator/src/metrics/mod.rs
+++ b/validator/src/metrics/mod.rs
@@ -1,0 +1,179 @@
+// Copyright 2018 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


validator/src/metrics/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+        metric_name: S,
+        level: Option<Level>,
+        tags: Option<HashMap<String, String>>,
+    ) -> Timer {
+        Timer {
+            py_timer: self._create_metric("timer", metric_name, level, tags),
+        }
+    }
+
+    fn _create_metric<S: AsRef<str>>(
```
nick-drozd: I don't think the leading underscore is needed, since the method is enforced as private.

aludvik: old habits die hard

PR #1672, opened 2018-05-24
Convert Jenkins workflow to use docker-compose

rbuysse:
Includes new Dockerfiles for linting, running bandit scan, running tests.

Takes first steps towards moving some sdks to their own repos.

Fixes some previous bugs.

### DIFF
--- a/.env
+++ b/.env
@@ -13,3 +13,4 @@
 # limitations under the License.
 
 ISOLATION_ID=latest
+INSTALL_TYPE=-local
--- a/bin/run_tests
+++ b/bin/run_tests
@@ -26,9 +26,6 @@ MODULE_LIST="
     identity
     block_info
     validator
-    go_sdk
-    java_sdk
-    javascript_sdk
     python_sdk
     rust_sdk
     integration
@@ -341,9 +338,6 @@ test_rust_sdk() {
 }
 
 test_integration() {
-    run_docker_test test_xo_smoke_all
-    copy_coverage .coverage.test_xo_smoke_all
-    run_docker_test test_web_socket_subscription
     run_docker_test test_poet_smoke
     copy_coverage .coverage.test_poet_smoke
     run_docker_test test_dynamic_network --timeout 600
--- a/bin/run_tests
+++ b/bin/run_tests
@@ -265,8 +265,6 @@ test_block_info() {
 test_validator() {
     run_docker_test ./validator/tests/unit_validator.yaml
     copy_coverage .coverage.validator
-    run_docker_test test_shutdown_smoke.yaml
-    copy_coverage .coverage.test_shutdown_smoke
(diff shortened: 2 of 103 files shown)


bin/run_tests
```diff
@@ -341,11 +338,6 @@ test_rust_sdk() {
 }
 
 test_integration() {
-    run_docker_test test_xo_smoke_all
-    copy_coverage .coverage.test_xo_smoke_all
-    run_docker_test test_web_socket_subscription
-    run_docker_test test_poet_smoke
```
nick-drozd: Taking out poet smoke should be in a separate commit (perhaps even a separate PR), since it doesn't pertain to restructuring the repo.

That said, if you're culling outdated tests, I suggest also getting rid of the shutdown smoke test (    https://github.com/hyperledger/sawtooth-core/blob/647f9566c8cf104cf0dc9e16404fa432d4dfd3a8/bin/run_tests#L271). It hasn't run in quite some time (the test container starts, but the actual Python tests are skipped), and it would take a fair amount of effort to get it to work.

bin/run_tests
```diff
@@ -341,11 +338,6 @@ test_rust_sdk() {
 }
 
 test_integration() {
-    run_docker_test test_xo_smoke_all
```
nick-drozd: Is there anywhere a test like this can run? Its goal is to verify that all the xo transaction processors do the same thing (in particular, that they serialize everything the same way), which no other test does.

integration/sawtooth_integration/docker/test_peer_list.yaml
```diff
@@ -137,7 +158,10 @@ services:
     stop_signal: SIGKILL
 
   rest-api-2:
-    image: sawtooth-rest-api:$ISOLATION_ID
+    build:
+      context: ../../..
```
aludvik: I think $SAWTOOTH_CORE is already set to this path. Can you use it here as well?

.env
```diff
@@ -13,3 +13,4 @@
 # limitations under the License.
 
 ISOLATION_ID=latest
+INSTALL_TYPE=-local
```
aludvik: Should you also set SAWTOOTH_CORE here?

bin/run_tests
rbuysse: I moved the removal of these into a separate commit.

bin/run_tests
rbuysse: This effort is non-trivial since some of the SDKs are moving out of the core repository into their own repos. It's on our list of to-dos.

integration/sawtooth_integration/docker/test_peer_list.yaml
rbuysse: I'm not sure that approach works since there are a ton of compose files at different depths.

PR #1673, opened 2018-05-24
Modify ExecutionResult returned by Scheduler

boydjohnson:
This is built off of #1604. This PR is just the last commit.

### DIFF
--- a/validator/src/scheduler/mod.rs
+++ b/validator/src/scheduler/mod.rs
@@ -51,8 +51,10 @@ pub use self::execution_result_ffi::{BatchResult, TransactionResult};
 pub struct ExecutionResults {
     pub beginning_state_hash: Option<String>,
     pub ending_state_hash: Option<String>,
-    pub batch_results: Vec<(String, Vec<TxnExecutionResult>)>,
+    pub batch_results: Vec<BatchExecutionResult>,
 }
+
+pub type BatchExecutionResult = (String, Option<Vec<TxnExecutionResult>>);
 
 pub struct TxnExecutionResult {
     pub signature: String,
--- a/validator/src/scheduler/py_scheduler.rs
+++ b/validator/src/scheduler/py_scheduler.rs
@@ -41,6 +41,8 @@ impl From<cpython::PyErr> for SchedulerError {
     }
 }
 
+type CombinedOptionalBatchResult = Vec<(Vec<TransactionResult>, Option<BatchResult>, String)>;
+
 pub struct PyScheduler {
     py_scheduler: cpython::PyObject,
     batch_ids: Vec<String>,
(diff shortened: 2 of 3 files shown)


validator/src/scheduler/mod.rs
```diff
@@ -52,6 +52,7 @@ pub struct ExecutionResults {
     pub beginning_state_hash: Option<String>,
     pub ending_state_hash: Option<String>,
     pub batch_results: Vec<(String, Vec<TxnExecutionResult>)>,
+    pub unscheduled_batch_ids: Vec<String>,
```
aludvik: Instead of adding a new field, I think you can include batches that were not executed in batch_results by making it something like `Vec<(String, Option<Vec<TxnExecutionResult>>)>`. You probably want to introduce a new type at that point though, so you could either do `Vec<BatchExecutionResult>` where `struct BatchExecutionResult { batch_id: String, result: Option<Vec<TxnExecutionResult>> }` or `Vec<(String, Option<BatchExecutionResult>)>` where `type BatchExecutionResult = Vec<TxnExecutionResult>`

validator/src/scheduler/mod.rs
```diff
@@ -51,9 +51,13 @@ pub use self::execution_result_ffi::{BatchResult, TransactionResult};
...
     pub beginning_state_hash: Option<String>,
     pub ending_state_hash: Option<String>,
-    pub batch_results: Vec<(String, Vec<TxnExecutionResult>)>,
+    pub batch_results: BatchExecutionResult,
 }
 
+pub type TxnExecutionResults = Option<Vec<TxnExecutionResult>>;
+
+pub type BatchExecutionResult = Vec<(String, TxnExecutionResults)>;
+
```
aludvik: Hmm, these hide some important details that make client code more confusing, namely that `BatchExecutionResult` is a collection and that `TxnExecutionResults` can be None.

I think:
```
pub type BatchExecutionResult = (String, Option<Vec<TxnExecutionResult>);
```
gives a more intuitive interface, because then `batch_results` still looks like a collection when you read the code (which you expect because the field is plural) and

validator/src/scheduler/py_scheduler.rs
```diff
@@ -116,24 +119,29 @@ impl Scheduler for PyScheduler {
...
                         .call_method(py, "get_transaction_execution_results", (id,), None)
                         .expect("No method get_transaction_execution_results on python scheduler")
                         .extract::<cpython::PyList>(py)?
                         .iter(py)
                         .map(|r| r.extract::<TransactionResult>(py))
                         .collect::<Result<Vec<TransactionResult>, cpython::PyErr>>()?;
-                    Ok((txn_results, batch_result, id.to_owned()))
-                })
-                .collect();
+                            Ok((txn_results, Some(result), id.to_owned()))
```
aludvik: It looks like the indentation got messed up above.

validator/src/scheduler/mod.rs
boydjohnson: Sounds good.

PR #1675, opened 2018-05-24
Add Rust Candidate block

boydjohnson:
This is built off of #1673.

### DIFF
--- a/validator/src/journal/candidate_block.rs
+++ b/validator/src/journal/candidate_block.rs
@@ -36,6 +36,13 @@ pub enum BlockPublisherError {
     NoPendingBatchesRemaining,
 }
 
+pub struct FinalizeBlockResult {
+    block: Option<cpython::PyObject>,
+    remaining_batches: Vec<Batch>,
+    last_batch: Option<Batch>,
+    injected_batch_ids: Vec<String>,
+}
+
 pub struct CandidateBlock {
     block_store: cpython::PyObject,
     consensus: cpython::PyObject,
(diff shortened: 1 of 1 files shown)


validator/src/journal/candidate_block.rs
```diff
@@ -0,0 +1,107 @@
...
+        committed_txn_cache: TransactionCommitCache,
+        block_builder: cpython::PyObject,
+        max_batches: u32,
+        batch_injectors: cpython::PyObject,
+        identity_signer: cpython::PyObject,
+        permission_verifier: cpython::PyObject,
+    ) -> Self {
+        let bs = unsafe {
+            let py = cpython::Python::assume_gil_acquired();
+            block_store.clone_ref(py)
```
aludvik: Why are we cloning this refs?

validator/src/journal/candidate_block.rs
```diff
@@ -220,4 +227,180 @@ impl CandidateBlock {
...
+                    py,
+                    "check_publish_block",
+                    (self.block_builder
+                        .getattr(py, "block_header")
+                        .expect("BlockBuilder has no attribute 'block_header'"),),
+                    None,
+                )
+                .expect("consensus has no method 'check_publish_block'")
+                .extract::<bool>(py)
+                .unwrap()
```
aludvik: put this in a function please

validator/src/journal/candidate_block.rs
```diff
@@ -220,4 +227,180 @@ impl CandidateBlock {
...
+            .into_iter()
+            .filter(|(_, txns)| match txns {
+                Some(t) => !t.iter().any(|t| !t.is_valid),
+                None => false,
+            })
+            .map(|(b_id, _)| b_id)
+            .collect();
+
+        let builder = {
+            let py = unsafe { cpython::Python::assume_gil_acquired() };
```
aludvik: We should try to only call this once at the beginning of the function if it is needed because the GIL is already in scope for the entire function.

validator/src/journal/candidate_block.rs
```diff
@@ -220,4 +227,180 @@ impl CandidateBlock {
...
+            self.block_builder.clone_ref(py)
+        };
+        let mut bad_batches = vec![];
+        let mut pending_batches = vec![];
+
+        for batch in self.pending_batches.clone() {
```
aludvik: I think you can probably get rid of this clone, either by doing `.iter()` or using `std::mem::replace`

validator/src/journal/candidate_block.rs
```diff
@@ -220,4 +227,180 @@ impl CandidateBlock {
...
+            } else if valid_batch_ids.contains(&batch.header_signature) {
+                if !self.check_batch_dependencies(&batch) {
+                    debug! {
+                        "Batch {} is invalid, due to missing txn dependency",
+                        header_signature
+                    };
+                    bad_batches.push(batch.clone());
+                    pending_batches.clear();
+                    pending_batches.append(&mut self.pending_batches
+                        .clone()
```
aludvik: If you `std::mem::replace` this probably becomes easier too (doesn't require clone())

validator/src/journal/candidate_block.rs
```diff
@@ -162,4 +162,47 @@ impl CandidateBlock {
...
+                "Dropping previously committed batch: {}",
+                batch_header_signature.as_str()
+            );
+            return;
+        } else if self.check_batch_dependencies(&batch) {
+            let mut batches_to_add = vec![];
+
+            // TODO: call Poll Injector python functions
+
+            // TODO: Enforce validation rules
```
aludvik: Should these be done?

boydjohnson:
@aludvik I implemented the 2 todos in add_batch, removed the spurious clone_ref, and reduced the number of calls of `.assume_gil_acquired`. I am not sure how to reduce the clones on `self.pending_batches`. I think it is either clone every time we add a batch to the the new `pending_batches` vec or clone `self.pending_batches` to begin with. I have wondered if I should spend some time making the signature into `finalize(self)` instead of `finalize(&mut self)` since it should consume the candidate block.

PR #1684, opened 2018-05-31
Implement Chain Controller In Rust

peterschwarz:
This is a fairly large change in the sense that it replaces the implementation of the `ChainController` with a rust implementation.  It is as straight of a 1-1 port as can be managed, given the constraints of both Rust (e.g. its constraints on memory usage, sharing, etc) and FFI calls (how rust interacts with python, across that boundary).

Adds an explicit dependency on [lazy_static](https://crates.io/crates/lazy_static)

### DIFF
--- a/validator/sawtooth_validator/server/core.py
+++ b/validator/sawtooth_validator/server/core.py
@@ -294,15 +294,13 @@ class Validator(object):
             permission_verifier=permission_verifier)
 
         chain_controller = ChainController(
+            block_store=block_store,
             block_cache=block_cache,
             block_validator=block_validator,
-            state_view_factory=state_view_factory,
             chain_head_lock=block_publisher.chain_head_lock,
             on_chain_updated=block_publisher.on_chain_updated,
-            chain_id_manager=chain_id_manager,
             data_dir=data_dir,
-            config_dir=config_dir,
-            chain_observers=[
+            observers=[
                 event_broadcaster,
                 receipt_store,
                 batch_tracker,
--- a/validator/src/journal/chain.rs
+++ b/validator/src/journal/chain.rs
@@ -167,7 +167,7 @@ impl<BC: BlockCache + 'static, BV: BlockValidator + 'static, CW: ChainWriter + '
         chain_head_update_observer: Box<ChainHeadUpdateObserver>,
         observers: Vec<Box<ChainObserver>>,
     ) -> Self {
-        ChainController {
+        let mut chain_controller = ChainController {
             state: Arc::new(RwLock::new(ChainControllerState {
                 block_cache,
                 block_validator,
@@ -182,7 +182,11 @@ impl<BC: BlockCache + 'static, BV: BlockValidator + 'static, CW: ChainWriter + '
             stop_handle: Arc::new(Mutex::new(None)),
             block_queue_sender: None,
             validation_result_sender: None,
-        }
+        };
+
+        chain_controller.initialize_chain_head();
+
+        chain_controller
     }
 
     pub fn chain_head(&self) -> Option<BlockWrapper> {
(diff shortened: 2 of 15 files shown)


validator/src/journal/chain.rs
```diff
@@ -0,0 +1,683 @@
...
+    fn contains(&self, block_id: &str) -> bool;
+
+    fn put(&mut self, block: BlockWrapper);
+
+    fn get(&self, block_id: &str) -> Option<BlockWrapper>;
+}
+
+#[derive(Debug)]
+pub enum ChainReadError {
+    GeneralReadError(String),
```
aludvik: Error types do not have to be enums. If there is only one kind of error, you can use a struct, new type, or tuple struct instead:
```
// Using a tuple struct
pub struct ChainHeadError(String);
```

validator/src/journal/chain.rs
```diff
@@ -0,0 +1,683 @@
...
+    fn validate_block(&self, block: BlockWrapper) -> Result<(), ValidationError>;
+
+    fn submit_blocks_for_verification(
+        &self,
+        blocks: &[BlockWrapper],
+        response_sender: Sender<(bool, BlockValidationResult)>,
+    );
+}
+
+// This should be in a validation module
```
aludvik: Any reason not to just do this now?

validator/src/journal/chain.rs
```diff
@@ -0,0 +1,683 @@
...
+            validation_result_sender: self.validation_result_sender.clone(),
+        }
+    }
+
+    fn submit_blocks_for_verification(
+        &self,
+        block_validator: &BV,
+        blocks: &[BlockWrapper],
+    ) -> Result<(), ChainControllerError> {
+        if self.validation_result_sender.is_some() {
```
aludvik: This function will silently do nothing if the sender is none, which may make debugging more difficult in the future. Can this be an expect() or a log message to help us out in the event it becomes none for some reason?

aludvik: Same with queue block

validator/src/journal/chain_ffi.rs
```diff
@@ -0,0 +1,723 @@
...
+    block_validator: *mut py_ffi::PyObject,
+    chain_head_lock: *mut py_ffi::PyObject,
+    on_chain_updated: *mut py_ffi::PyObject,
+    observers: *mut py_ffi::PyObject,
+    data_directory: *const c_char,
+    chain_controller_ptr: *mut *const c_void,
+) -> ErrorCode {
+    if block_store.is_null() {
+        return ErrorCode::NullPointerProvided;
+    }
```
aludvik: I wrote a little macro to do this in the publisher ffi code that you could reuse:
```
macro_rules! check_null {
    ($($arg:expr) , *) => {
        $(if $arg.is_null() { return ErrorCode::NullPointerProvided; })*
    }
}
```

aludvik: https://github.com/boydjohnson/sawtooth-core/blob/block.publisher.rust/validator/src/journal/publisher_ffi.rs#L61

validator/src/journal/chain_ffi.rs
```diff
@@ -0,0 +1,723 @@
...
+
+    fn submit_blocks_for_verification(
+        &self,
+        blocks: &[BlockWrapper],
+        response_sender: Sender<(bool, BlockValidationResult)>,
+    ) {
+        let gil_guard = Python::acquire_gil();
+        let py = gil_guard.python();
+
+        let ctypes_module = py.import("ctypes").expect("Unable to import ctypes");
```
aludvik: Rather than do these imports in the function every time it is called, you can do them in the constructor for PyBlockValidator and then hold onto the imported modules as instance variables.

validator/sawtooth_validator/server/core.py
```diff
@@ -294,15 +294,13 @@ def __init__(self,
...
-            chain_id_manager=chain_id_manager,
-            data_dir=data_dir,
-            config_dir=config_dir,
-            chain_observers=[
+            block_store,
+            block_cache,
+            block_validator,
+            block_publisher.chain_head_lock,
+            block_publisher.on_chain_updated,
+            data_dir,
```
aludvik: I believe kwargs were used previously to mitigate ordering mistakes made when the constructor updates. Why did you decide to remove them?

validator/src/journal/chain.rs
peterschwarz: Validation results should be an `expect`, since it can only be run after the controller has been started.  Queue block cannot, since it is exposed to python, and therefore control of the method is beyond the scope of this PR

validator/src/journal/chain.rs
peterschwarz: The comment is from before rebasing on top of other changes, so I will look and see if it is possible now

validator/src/journal/chain_ffi.rs
peterschwarz: I will add this.

validator/src/journal/chain.rs
peterschwarz: An interesting point, but I would prefer to remain consistent in the code by using enums for now.

PR #1686, opened 2018-05-31
Implement consensus peer messaging

nick-drozd:


sdk/examples/devmode_rust/src/engine.rs
```diff
@@ -253,6 +281,35 @@ impl Engine for DevmodeEngine {
                             service.initialize_block();
                         }
 
+                        Update::PeerMessage(message, sender_id) => match message
+                            .message_type
+                            .as_ref()
+                        {
+                            "published" => {
```
peterschwarz: Any reason why these are strings as opposed to enums?

nick-drozd: Added `DevmodeMessage` enum

PR #1689, opened 2018-06-03
Add Blocks at a Configured Depth to Pruning Queue

peterschwarz:
This PR creates a pruning manager, which prunes state roots from a queue at predictable times.  It is evaluated at chain commit time. 10 percent of the queue is pruned, which allows state roots to be returned to validity, if need be before they are finally pruned. 

This PR does not solve the problem of abandoned forks, as they cannot be properly re-evaluated if a validator switches back to them. To properly cleanup state after a chain has been dropped, we will need the Block Tree Manager (See [RFC 5](https://github.com/hyperledger/sawtooth-rfcs/pull/5)).  When uncommitted blocks expire from the Block Tree Manager, their state roots can be pruned.


validator/src/journal/chain.rs
```diff
@@ -263,6 +266,19 @@ impl<BC: BlockCache + 'static, BV: BlockValidator + 'static, CW: ChainWriter + '
             let chain_head_block = new_block.clone();
             state.chain_head = Some(new_block);
 
+            state.state_prune_manager.update_queue(
+                &result
+                    .new_chain
+                    .iter()
```
nick-drozd: Would it make more sense to have the function take care of this `.iter.map.collect` stuff rather than make the caller do it?

peterschwarz: The state pruning manager operates on state root hashes, not blocks

nick-drozd:
Oh, commit `Add decimate to StatePruneManager` has `If a root is prune,` in the commit message (prune -> pruned).

peterschwarz:
Added some new commits to handle the issue where an identical leaf is produced, and then is pruned from a prior state root hash.  This approach is the most direct - scanning the change logs.

Future changes could make this more efficient by managing this with an in-memory reference count.

PR #1691, opened 2018-06-04
DO NOT MERGE - TESTING BANDIT

dcmiddle:
Add back in code that should trip bandit. Want to see if Jenkins stops
the build.


dcmiddle:
Verified that bandit errors will still stop the build. And that the bandit output is inaccessible from the Jenkins UI as reported in https://jira.hyperledger.org/browse/STL-320.

PR #1692, opened 2018-06-04
Add Peripheral Consensus Engine Objects

aludvik:
This PR adds the peripheral objects used by the consensus engine API including:
- All consensus engine API message handlers
- A Proxy object that routes requests between message handlers and internal validator objects
- A Notifier object that encapsulates sending updates to consensus engines from internal validator objects

It also implements the query services and networking services.


nick-drozd:
commit message for `Add separate consensus service and dispatcher`

interferring -> interfering

nick-drozd:
Looks good to me, but since I have commits in there, I don't think I should formally approve it.

aludvik:
Lint is broken on this but jenkins is not checking it correctly

PR #1695, opened 2018-06-05
Fixes to Rust CandidateBlock, PyScheduler, TransactionCommitCache

boydjohnson:
These fixes were found in integrating the Rust BlockPublisher with the already committed Rust and Python code. They are necessary for the Rust BlockPublisher PR.

### DIFF
--- a/validator/src/journal/chain_commit_state.rs
+++ b/validator/src/journal/chain_commit_state.rs
@@ -59,13 +59,16 @@ impl TransactionCommitCache {
     }
 
     pub fn contains(&self, transaction_id: &str) -> bool {
-        let py = unsafe { cpython::Python::assume_gil_acquired() };
+        self.committed.contains(transaction_id) || self.blockstore_has_txn(transaction_id)
+    }
 
-        self.committed.contains(transaction_id)
-            || self.blockstore
-                .call_method(py, "has_transaction", (transaction_id,), None)
-                .unwrap()
-                .extract::<bool>(py)
-                .unwrap()
+    fn blockstore_has_txn(&self, transaction_id: &str) -> bool {
+        let gil = cpython::Python::acquire_gil();
+        let py = gil.python();
+        self.blockstore
+            .call_method(py, "has_transaction", (transaction_id,), None)
+            .unwrap()
+            .extract::<bool>(py)
+            .unwrap()
     }
 }


validator/src/scheduler/execution_result_ffi.rs
```diff
@@ -47,11 +47,13 @@ impl<'source> FromPyObject<'source> for BatchResult {
     fn extract(py: cpython::Python, obj: &'source cpython::PyObject) -> cpython::PyResult<Self> {
         let state_hash = obj.getattr(py, "state_hash").unwrap();
 
-        let sh = state_hash.extract::<String>(py)?;
+        let sh = if state_hash != cpython::Python::None(py) {
```
peterschwarz: `FromPyObject` is implemented for `Option<T: FromPyObject>`.  This could just be:
```
let sh: Option<String> = cpython::extract(py)?;
```

validator/src/journal/candidate_block.rs
```diff
@@ -486,7 +471,31 @@ impl CandidateBlock {
                     .expect("BlockBuilder has no method 'build_block'"),
             ),
             pending_batches,
-        ))
+        )
+    }
+
+    fn consensus_not_finalize_block(&self, builder: &cpython::PyObject) -> bool {
```
peterschwarz: This function name is very confusing.

boydjohnson: Would `consensus_does_not_finalize_block` be better?

validator/src/scheduler/execution_result_ffi.rs
boydjohnson: Cool. I hadn't known that. I will change this and reevaluate my use of checking Python::None(py).

validator/src/journal/candidate_block.rs
peterschwarz: That makes more sense.  Rename it like so

boydjohnson: This is a helper method in CandidateBlock. I find it very unfortunate that the coding style of the python consensus interface leaks all the way into this helper method in the Rust candidate block. Since I put `!` in the method shouldn't there be some indication to the programmer reading it that there is a `not` check?

boydjohnson:
@aludvik I am concerned with your statement `needs to be cleaned up or have summaries added`. Looking through `git log` from master there are regularly commit messages without a body and regularly without a why.

aludvik:
@boydjohnson that is fair. I am definitely guilty of leaving them off at times and have been called out for it. I think the goal should be to always include a commit message that will help reviewers understand why you are doing something and I have been trying to let people know in general when I find that missing.

As an example, consider "Add check for None python state_hash in BatchResult". Looking at just that commit and the summary, I would have no context to understand why this is a thing that should be done or if it is correct or not. Another sentence below the summary stating "Python can also pass a None object instead of a String" makes it easier to understand why that was added later on.

boydjohnson:
I am opening this PR again.

PR #1696, opened 2018-06-05
keygen: Fix Public Key Permission in Rust

danintel:
Change permission of public key so it is readable by all.
That is, change from 0640 (rw-r-----) to 0644 (rw-r--r--).


dcmiddle:
@danintel I see the pubkey commit, is there a privkey commit to address the group readable aspect mentioned in your PR note?
```Change permission of private key so it is not group-readable.
That is, change from 0640 (rw-r-----) to 0600 (rw-------).```

danintel:
@dcmiddle  No--I removed the privkey commit and changed the commit message.  The old commit message remains in this thread, so I just edited it out to reduce confusion.

I will handle privkey permissions later.

PR #1702, opened 2018-06-06
Add regexp subscription to state-delta Events in REST API 

benoitrazet:
This PR proposes a slight extension to state-delta Event subscription through the REST API. The extension allows a user to indicate a  regexp during subscription, such that the REST API would return the addresses (in the state-delta changes) that match the regexp. Before this PR it was already possible to indicate a list of prefixes to match the addresses of interest.

The extension proposed is minimal and would be particularly useful for app developers or users with:
- addresses of interest filtered by substring (as opposed to sole prefix)
- do not have direct access to a validator with full-fledged events through ZMQ.

### DIFF
--- a/rest_api/sawtooth_rest_api/state_delta_subscription_handler.py
+++ b/rest_api/sawtooth_rest_api/state_delta_subscription_handler.py
@@ -16,6 +16,7 @@
 import asyncio
 import logging
 import json
+import re
 import aiohttp
 from aiohttp import web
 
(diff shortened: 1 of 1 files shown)


vaporos:
Two questions:

1) is this backward compatible?
2) for the other maintainers, is this an acceptable API to commit to moving forward (keeping in mind we are committed to backward compatibility of the REST API)

benoitrazet:
Thanks @vaporos for the feedback.
I squashed the two commits into one commit and added the PR description in the commit.
To answer your question 1), yes, this PR is backward compatible.

benoitrazet:
If the feature enhancement proposed in this PR is seen as beneficial, then I'm willing do the same enhancement to the `get_block_deltas` for Missed Events (to make things a little more uniform), and update the documentation accordingly. I could do the documentation first if that would help people deciding whether or not this PR is worth pursuing or not. I'm completely fine either way.

TomBarnes:
Security implications?  Is the regex expression complete open? Can a limit be placed on execution time? I understand the intent is to find trival matches, but if it completely open-ended, then its giving an unscrupulous (or inexperienced) user a lot of power.
https://www.owasp.org/index.php/Regular_expression_Denial_of_Service_-_ReDoS 
https://stackoverflow.com/questions/4289923/in-which-languages-is-it-a-security-hole-to-use-user-supplied-regular-expression
https://snyk.io/blog/redos-and-catastrophic-backtracking/
https://stackoverflow.com/questions/12841970/how-can-i-recognize-an-evil-regex
https://www.owasp.org/index.php/Regular_Expression_Security_Cheatsheet

benoitrazet:
Thanks @dcmiddle 
Very good point @TomBarnes . I'll think about a good solution. Notice that the same problem is in the event system of the validator https://github.com/hyperledger/sawtooth-core/blob/master/validator/sawtooth_validator/server/events/subscription.py#L142 but might not be as catastrophic because different people have access to different components of the system, the rest_api  being more public facing than the event system.

PR #1707, opened 2018-06-06
Rust BlockPublisher

boydjohnson:


validator/src/journal/chain_head_lock.rs
```diff
@@ -0,0 +1,54 @@
...
+    publisher: BlockPublisherImpl,
+}
+
+impl ChainHeadLock {
+    pub fn new(publisher: BlockPublisherImpl) -> Self {
+        ChainHeadLock { publisher }
+    }
+
+    pub extern "C" fn acquire(&self) -> ChainHeadGuard {
+        eprintln!("{} {}: locking chain head", file!(), line!());
```
aludvik: Delete debugging statement

validator/src/journal/chain_head_lock.rs
```diff
@@ -0,0 +1,54 @@
...
+pub struct ChainHeadLock {
+    publisher: BlockPublisherImpl,
+}
+
+impl ChainHeadLock {
+    pub fn new(publisher: BlockPublisherImpl) -> Self {
+        ChainHeadLock { publisher }
+    }
+
+    pub extern "C" fn acquire(&self) -> ChainHeadGuard {
```
aludvik: Does this need to be extern "C"?

validator/src/journal/chain_head_lock_ffi.rs
```diff
@@ -0,0 +1,95 @@
...
+
+#[repr(u32)]
+#[derive(Debug)]
+pub enum ErrorCode {
+    Success = 0,
+    NullPointerProvided = 0x01,
+}
+
+#[no_mangle]
+pub extern "C" fn chain_head_lock_lock(
```
aludvik: This function should stay consistent with the C library function naming scheme, it should be renamed `chain_head_lock_acquire()` to match the method name.

validator/sawtooth_validator/journal/chain.py
```diff
@@ -59,8 +58,7 @@ def __init__(
             ctypes.py_object(block_store),
             ctypes.py_object(block_cache),
             ctypes.py_object(block_validator),
-            ctypes.py_object(chain_head_lock),
-            ctypes.py_object(on_chain_updated),
+            chain_head_lock.pointer,
```
aludvik: I would suggest changing this commit message to something like:
```
Pass Rust ChainHeadLock pointer to ChainController

Previously the chain controller and block publisher shared a Python RLock(). Now that
the BlockPublisher is in Rust and is using a RwLock, this lock needs to be passed to the
ChainController.
```

validator/src/journal/chain.rs
```diff
@@ -87,8 +88,13 @@ pub trait ExternalLock: Send + Sync {
...
 
-pub trait ExternalLockGuard: Drop {
-    fn release(&self);
+pub trait ExternalLockGuard {
+    fn notify_on_chain_updated(
+        &self,
+        chain_head: Option<BlockWrapper>,
+        committed_batches: Vec<Batch>,
+        uncommitted_batches: Vec<Batch>,
+    );
```
aludvik: I believe you can remove the ExternalLock and ExternalLockGuard traits now

validator/src/journal/publisher.rs
```diff
@@ -16,13 +16,664 @@
...
+        BlockPublisherState {
+            transaction_executor,
+            chain_head,
+            candidate_block,
+            pending_batches,
+        }
+    }
+}
+
+pub struct BlockPublisherImpl {
```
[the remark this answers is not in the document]
boydjohnson: Yeah, I used Impl because that was the structure where most of the methods were going to be on. I think SyncBlockPublisher is a good name.

validator/src/journal/chain_head_lock.rs
boydjohnson: I don't think so.

validator/src/journal/chain.rs
boydjohnson: Yeah, I think that is correct.

validator/src/journal/chain_head_lock.rs
peterschwarz: We've been pretty consistent about using that, though I think it has become optional in recent versions of Rust

peterschwarz: In other words, if we want to get rid of it, it should be across the board (i.e. in a separate PR)

peterschwarz: Oh, wait, nevermind, I mis-read where this was being commented on.

aludvik: :)

PR #1708, opened 2018-06-07
Poet engine skeleton

nick-drozd:


sdk/examples/poet_python/sawtooth_poet/main.py
```diff
@@ -0,0 +1,79 @@
...
+DISTRIBUTION_NAME = 'sawtooth-poet'
+
+
+def parse_args(args):
+    parser = argparse.ArgumentParser(
+        formatter_class=argparse.RawTextHelpFormatter)
+
+    parser.add_argument(
+        '-C', '--connect',
+        default='tcp://localhost:5005',
```
dcmiddle: Do engines talk on the component port or is this 5005 going to be the new standard for a consensus port?

sdk/examples/poet_python/sawtooth_poet/engine.py
```diff
@@ -0,0 +1,168 @@
...
+            LOGGER.info('Failed consensus check: %s', block)
+            self._fail_block(block.block_id)
+
+    def _handle_valid_block(self, block_id):
+        block = self._get_block(block_id)
+
+        self._chain_head = self._get_chain_head()
+
+        LOGGER.info(
+            'Choosing between chain heads -- currnent: %s -- new: %s',
```
dcmiddle: typo current

sdk/examples/poet_python/sawtooth_poet/engine.py
```diff
@@ -0,0 +1,168 @@
...
+                continue
+            except exceptions.InvalidState:
+                return None
+
+        return block_id
+
+    def _check_publish_block(self):
+        return True
+
+    def start(self, updates, service, chain_head, peers):
```
dcmiddle: What is `service`? Please add a doc string on these params.

sdk/examples/poet_python/sawtooth_poet/main.py
aludvik: This is provisional endpoint that will be added to the validator. We may merge it in with the component endpoint to simplify deployment, but I don't think we should commit to that until we understand how it will affect performance: https://github.com/hyperledger/sawtooth-core/pull/1692/commits/9f723237ab0ac315a4784202f73eda98582f2d58

sdk/examples/poet_python/sawtooth_poet/engine.py
nick-drozd: `Service` is a component analogous to a transaction processor's `Context`. Engines use services to make calls to the validator through methods like `commit_block`, `broadcast`, etc. Docstrings are available from the abstract base classes:

`Engine`: https://github.com/hyperledger/sawtooth-core/blob/master/sdk/python/sawtooth_sdk/consensus/engine.py
`Service`: https://github.com/hyperledger/sawtooth-core/blob/master/sdk/python/sawtooth_sdk/consensus/service.py

sdk/examples/poet_python/sawtooth_poet/main.py
dcmiddle: should this port be 5050 then?

sdk/examples/poet_python/sawtooth_poet/engine.py
dcmiddle: Thanks for the clarifications and links.
Which component will be responsible for calling start? (I don't think we have any design docs on this yet)

sdk/python/sawtooth_sdk/consensus/zmq_driver.py
```diff
@@ -67,7 +67,7 @@ def start(self, endpoint):
                 break
 
             try:
-                message = self._stream.receive().result(0.1)
+                message = self._stream.receive().result(10)
```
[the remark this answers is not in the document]
nick-drozd: It seems to work with the higher number but not with lower number. I don't really understand what the number is doing, so I don't have any insight into why, or whether there is a more suitable number.

sdk/examples/poet_python/sawtooth_poet/engine.py
nick-drozd: `Driver` is analogous to `TransactionProcessor` and `Engine` is analogous to `TransactionHandler`. When the driver is started in `main` (https://github.com/nick-drozd/sawtooth-core/blob/4d4181115d139b6547a29568adac030f7bfe8a31/sdk/examples/poet_python/sawtooth_poet/main.py#L73), it will in turn start the engine.

sdk/examples/poet_python/sawtooth_poet/main.py
aludvik: yep

sdk/python/sawtooth_sdk/consensus/zmq_driver.py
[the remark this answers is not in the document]
peterschwarz: Increasing the time so that it works seems like a valid commit message

PR #1712, opened 2018-06-12
Update test scripts to build images before running tests

rbuysse:
Without this modification building test images counts against the test
timeout. Depending on cache status and network latency, this can cause
false test failures.


aludvik:
Does this always force building now?

PR #1716, opened 2018-06-13
Package REST API config example

rbuysse:


rbuysse:
Will include this in a larger config packaging PR.

rbuysse:
Will include this in a larger config packaging PR.

PR #1720, opened 2018-06-14
Validator synchronization issue after transaction processor is restarted

mithunshashidhara:
When running multiple validators, if transaction processor is restarted
on one of the nodes, the validator on that node stops processing future blocks
even after the transaction processor is up.

A stopped TP is removed after it fails to respond to ping.
The TP removal procedure was not getting carried out properly due to checks of
alive_threads and _open_futures. By removal of these checks TP restart
keeps validators in sync.

Resolves: STL-1115

Signed-off-by: mithunshashidhara <mithunx.shashidhara@intel.com>

### DIFF
--- a/validator/sawtooth_validator/execution/executor.py
+++ b/validator/sawtooth_validator/execution/executor.py
@@ -351,10 +351,10 @@ class TransactionExecutorThread:
                 {signature: fut}
 
     def remove_broken_connection(self, connection_id):
+        self._processor_manager.remove(connection_id)
         if connection_id not in self._open_futures:
             # Connection has already been removed.
             return
-        self._processor_manager.remove(connection_id)
         futures_to_set = [
             self._open_futures[connection_id][key]
             for key in self._open_futures[connection_id]
@@ -477,8 +477,7 @@ class TransactionExecutor:
 
     def _remove_broken_connection(self, connection_id):
         for t in self._alive_threads:
-            if not t.is_done():
-                t.remove_broken_connection(connection_id)
+            t.remove_broken_connection(connection_id)
 
     def execute(self, scheduler):
         self._remove_done_threads()


mithunshashidhara:
corrected file permissions, please check.

validator/sawtooth_validator/execution/executor.py
```diff
@@ -348,9 +348,6 @@ def _send_and_process_result(self, content, connection_id, signature):
                 {signature: fut}
 
     def remove_broken_connection(self, connection_id):
-        if connection_id not in self._open_futures:
```
peterschwarz: This guard still seems like it could be necessary.

vaporos:
@agunde406 should probably review this change.

validator/sawtooth_validator/execution/executor.py
mithunshashidhara: with this check in place the issue is not resolved. i tested again today (on 1.0.4) without this check  issue reported in this bug persists.

dcmiddle: @mithunshashidhara it's possible that the actual bug then is in the management of that list. I don't think we want to remove a check if we don't understand why it's resolving a symptom.

mithunshashidhara: there is no need to send response to TP as it has been stopped.

agunde406: Can you explain how this was tested? Was it tested with both schedulers? I am concerned with the fact that futures are not being resolved and no InternalError is being returned. As such the txn will not be retryed and the serial scheduler will hang waiting on that future. I will try to test this as well but I am curious what your testing strategy was.

mithunshashidhara: switch to 1.0.4, run "./bin/build_all -t python" to build docker images. goto directory docker/compose/  "docker-compose -f sawtooth-local.yaml up" . copy my test script to docker/compose/  and run the script https://jira.hyperledger.org/secure/attachment/15726/STL-1115_repro
with my changes you should see 110,120,130,140,150.
without changes you will see inconsistent result. today i tested with both schedulers and things are fine with my changes.

agunde406:
Upon local testing, there are actually two bugs wrapped up in this code. The first is that you were correct that `self._processor_manager.remove(connection_id)` was not being correctly called for all connections. 

The second bug is that if a transaction processor dies mid-transaction execution the internal error causes the transaction to be resubmitted when there is no transaction processor handler and causes the validator to hang. More works need to be done to correctly handle the different types of internal errors (transaction processor dies, workers full in multithreaded transaction processor such in go, etc)

Your current code does not fix the second issue and removes the needed handling of the future before it reaches the second bug, which would cause the serial scheduler to hang waiting on a future result (would also hang completion of a block in the parallel sense). As such I would suggest you change the code in remove_broken_connection to the following:

```
def remove_broken_connection(self, connection_id):
        if connection_id in self._open_futures:
            futures_to_set = [
                self._open_futures[connection_id][key]
                for key in self._open_futures[connection_id]
            ]

            response = processor_pb2.TpProcessResponse(
                status=processor_pb2.TpProcessResponse.INTERNAL_ERROR)
            result = FutureResult(
                message_type=validator_pb2.Message.TP_PROCESS_RESPONSE,
                content=response.SerializeToString(),
                connection_id=connection_id)
            for fut in futures_to_set:
                fut.set_result(result)
                self._future_done_callback(fut.request, result)

        self._processor_manager.remove(connection_id)
```

This will fix the case that the transaction processor is manually stopped/started. A new bug/story should be opened around the second issue, I'll work to get that created.

mithunshashidhara:
@agunde406 i tested with the changes proposed by you, this bug is not resolved and i saw key error exception on validator logs. please test with the steps described  in my prev post. 

i agree with you that this function has issues, that is the reason i chose to remove most of its functionality. 

my goal is to resolve this bug, i understand you want to retain this function as much as possible. to make it a win-win for both of us please see my new change set.

agunde406:
Here is the new bug I created for the case where a tp is currently processing a transaction when it crashes. https://jira.hyperledger.org/browse/STL-1404

mithunshashidhara:
Thanks for approving this PR, a second approval is needed for this PR to be merged. Please do the needful.

PR #1722, opened 2018-06-15
Modify get_ and set_state for multiple addresses

ltseeley:
Modifies the get_state and set_state methods for TransactionContext in
the Rust SDK to take in a list of addresses to get/set instead of just
one address. This aligns with the functionality of the other SDKs. Also
modifies existing uses of SDK (smallbank_rust, inkey_rust, and xo_rust)
to call methods properly.


bridger-herman:
Could you please update the commit message to reflect the changes to the XO Rust example as well?

PR #1723, opened 2018-06-18
Change consensus proxy state_get to handle address prefixes

nick-drozd:
Previously it only accepted a list of fully specifed addresses. Now it
can accept lists containing both full addresses and address prefixes.

Signed-off-by: Nick Drozd <drozd@bitwise.io>


nick-drozd:
Added another minor fix to the CSDK.

PR #1724, opened 2018-06-18
Poet modifications

nick-drozd:
These are a few small changes to prepare the PoET code for use by the consensus engine.


peterschwarz:
The use of sawtooth_validator protobufs can be removed when we no longer include poet as part of the validator process.

PR #1727, opened 2018-06-18
Package example configs backport

rbuysse:


rberg2:
Adding @vaporos and @dcmiddle for backport approval.

PR #1730, opened 2018-06-18
Remove JS SDK files from core repo

delventhalz:
The JS SDK is being moved to its own repo. One half of that is removing those files from sawtooth-core. This PR removes all JS SDK components, as well as updating various scripts, compose files, and documentation which reference the removed Javascript components.


delventhalz:
The sister PR to this is [Add SDK files from core repo](https://github.com/hyperledger/sawtooth-sdk-javascript/pull/2) on `hyperledger/sawtooth-sdk-javascript`

docker/sawtooth-dev-javascript
```diff
@@ -14,8 +14,7 @@
 # ------------------------------------------------------------------------------
 
 # Description:
-#   Builds an image to be used when developing in JavaScript. The default CMD is to run
-#   build_javascript.
+#   Builds an image to be used when developing in JavaScript.
```
peterschwarz: Is there any need for this docker file to remain?

delventhalz: Yes. It is used by the web socket integration test:
https://github.com/hyperledger/sawtooth-core/blob/master/integration/sawtooth_integration/docker/test_web_socket_subscription.yaml#L90

delventhalz:
Updated `ci/sawtooth-build-docs` so curl is still installed. This should hopefully fix the build issues:
https://github.com/hyperledger/sawtooth-core/pull/1730/files#diff-1340a1a8adaaee39b9aeaa59673e13b8R82

rberg2:
I have been trying to get this one rebuilt since the move to JJB. I think if this one can be rebased to pick up the apt-key changes this will succeed.

PR #1739, opened 2018-06-20
C021 improve linting times

rbuysse:


peterschwarz:
On master, the lint stage takes ~5 min, with building the lint requirements at ~17.5 min.  On this PR, the lint stage takes 17min, and the build lint requirements stage takes ~19 min.  This doesn't seem like an improvement.

PR #1744, opened 2018-06-25
Add own peer info to consensus engine startup data

aludvik:


protos/consensus.proto
```diff
@@ -81,8 +81,11 @@ message ConsensusRegisterResponse {
   }
 
   Status status = 1;
+
+  // Startup Info
   ConsensusBlock chain_head = 2;
   repeated ConsensusPeerInfo peers = 3;
+  ConsensusPeerInfo own_peer_info = 4;
```
peterschwarz: Can we come up with a better name for this?  `validator_peer_info` or just `peer_info` would be better

aludvik: Why are those better?

peterschwarz: How about `local_peer_info`?

PR #1749, opened 2018-06-28
Use Random Number instead of Timestamp for Nonce

danintel:
Timestamps have a greater chance of collision, as use of
parallel transaction processing increases (either through
threads or multiple processes).

### DIFF
--- a/sdk/examples/noop_python/sawtooth_noop/client_cli/create_batch.py
+++ b/sdk/examples/noop_python/sawtooth_noop/client_cli/create_batch.py
@@ -30,7 +30,7 @@ LOGGER = logging.getLogger(__name__)
 class NoopPayload(object):
     def __init__(self):
         self.nonce = binascii.b2a_hex(random.getrandbits(
-            8 * 8).to_bytes(8, hbyteorder='little'))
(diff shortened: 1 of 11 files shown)


sdk/examples/noop_python/sawtooth_noop/client_cli/create_batch.py
```diff
@@ -31,7 +30,7 @@
 class NoopPayload(object):
     def __init__(self):
         self.nonce = binascii.b2a_hex(random.getrandbits(
-            8 * 8).to_bytes(8, hbyteorder='little'))
+            8 * 8).to_bytes(8, byteorder='little'))
```
vaporos: This is a different type of change than the rest of the commit; suggest it be in a different commit with an alternate explanation if it's not an accident.

danintel: As requested, I reintroduced the hbyteorder typo.

Then I separated the s/hbyteorder/byteorder/ fix into a separate fix.

danintel:
Rebased as this PR is waiting in the queue.

PR #1755, opened 2018-07-02
Add sleep after not ready registration

aludvik:
Add a sleep after receiving a not ready response from the validator.
This will prevent the consensus engine from overwhelming the
validator while trying to register before the validator has a received a
genesis block.

### DIFF
--- a/sdk/rust/src/consensus/zmq_driver.rs
+++ b/sdk/rust/src/consensus/zmq_driver.rs
@@ -38,6 +38,8 @@ use std::time::Duration;
 
 const REGISTER_TIMEOUT: u64 = 300;
 const SERVICE_TIMEOUT: u64 = 300;
+const INITAL_RETRY_DELAY: Duration = Duration::from_millis(100);
+const MAX_RETRY_DELAY: Duration = Duration::from_secs(3);
 
 /// Generates a random correlation id for use in Message
 fn generate_correlation_id() -> String {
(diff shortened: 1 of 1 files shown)


sdk/rust/src/consensus/zmq_driver.rs
```diff
@@ -197,6 +197,7 @@ pub fn register(
                         break;
                     }
                     ConsensusRegisterResponse_Status::NOT_READY => {
+                        ::std::thread::sleep(::std::time::Duration::from_millis(1500));
```
vaporos: This would be better done as an exponential backoff up to a limit. Why 1500?

peterschwarz: The current value was due to testing, while trying to find the deadlocks with the validator changes.  I agree an exponential backoff would make more sense here.

sdk/rust/src/consensus/zmq_driver.rs
```diff
@@ -200,7 +202,12 @@ pub fn register(
                     }
                     ConsensusRegisterResponse_Status::NOT_READY => {
                         thread::sleep(retry_delay);
-                        retry_delay = retry_delay * 2;
+                        if retry_delay < MAX_RETRY_DELAY {
```
peterschwarz: You could just use `retry_delay = ::std::cmp::min(retry_delay * 2, MAX_RETRY);` and avoid the big if block here

PR #1761, opened 2018-07-05
Prevent pruning of leaves via duplicate log

peterschwarz:
This commit introduces the idea of a duplicate log. The duplicate log is a record of the count of additional collisions at a given hash key.

When node is written to the database, if an entry already exists for it, a count is updated in the duplicate log.   When a node is deleted, the count is decremented.  Once there are no duplicates, the count is removed entirely.  Nodes may be fully deleted at this point.\\

This fixes the issue that can occur with state pruning where a fork produces a duplicate hash, whose original value was produced outside of the state pruning window.  Those values would be deleted when the fork was prune, which would cause errors validating transactions that still expected the value to be in state.

### DIFF
--- a/validator/src/state/merkle.rs
+++ b/validator/src/state/merkle.rs
@@ -49,6 +49,8 @@ use state::StateReader;
 const TOKEN_SIZE: usize = 2;
 
 pub const CHANGE_LOG_INDEX: &str = "change_log";
+pub const DUPLICATE_LOG_INDEX: &str = "duplicate_log";
+pub const INDEXES: [&'static str; 2] = [CHANGE_LOG_INDEX, DUPLICATE_LOG_INDEX];
 
 /// Merkle Database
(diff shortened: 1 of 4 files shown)


validator/src/state/merkle.rs
```diff
@@ -559,6 +561,54 @@ fn write_change_log(
...
+        if let Some(ref_count) = db_writer.index_get(DUPLICATE_LOG_INDEX, key)? {
+            from_bytes(ref_count)
+        } else {
+            0
+        },
+    )
+}
+
+fn to_bytes(num: u64) -> [u8; 8] {
+    unsafe { ::std::mem::transmute(num) }
```
aludvik: This should standardize on a memory representation before transmuting: https://doc.rust-lang.org/nightly/src/core/num/mod.rs.html#3706-3708 Same with from_bytes()

PR #1756, opened 2018-07-02
Modify Chain Controller and Block Publisher to use Consensus Proxy

peterschwarz:
This PR combines the changes necessary to modify the validator to support consensus engines implemented with the SDK API's.

This will require the PR's for Poet and DevMode engines to properly pass tests.  Will update and rebase  as soon as those PR's are available.

PR #1771, opened 2018-07-13
Implement Consensus Engine API in Validator

aludvik:
This is a rebase of #1756 with one additional commit added.

### DIFF
--- a/validator/packaging/validator.toml.example
+++ b/validator/packaging/validator.toml.example
@@ -29,6 +29,7 @@
 bind = [
   "network:tcp://127.0.0.1:8800",
   "component:tcp://127.0.0.1:4004"
+  "consensus:tcp://127.0.0.1:5050"
 ]
 
 # The type of peering approach the validator should take. Choices are 'static'
--- a/validator/sawtooth_validator/consensus/proxy.py
+++ b/validator/sawtooth_validator/consensus/proxy.py
@@ -138,7 +138,26 @@ class ConsensusProxy:
             self._get_blocks([block_id])[0].get_state_view(
                 self._state_view_factory)
 
-        return _map_with_none(state_view.get, addresses)
+        result = []
+
+        for address in addresses:
+            # a fully specified address
+            if len(address) == 70:
+                try:
+                    value = state_view.get(address)
+                except KeyError:
+                    value = None
+
+                result.append((address, value))
+                continue
+
+            # an address prefix
+            leaves = state_view.leaves(address)
+
+            for leaf in leaves:
+                result.append(leaf)
+
+        return result
 
     def _get_blocks(self, block_ids):
         try:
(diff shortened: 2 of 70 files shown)


aludvik:
This PR introduces a memory leak. Merging is pending diagnosing and resolving the leak.

validator/packaging/validator.toml.example
```diff
@@ -29,6 +29,7 @@
 bind = [
   "network:tcp://127.0.0.1:8800",
   "component:tcp://127.0.0.1:4004"
+  "consensus:tcp://127.0.0.1:5050"
```
jsmitchell: s/5050/5005

validator/sawtooth_validator/consensus/proxy.py
```diff
@@ -70,27 +70,28 @@ def summarize_block(self):
...
         self._block_publisher.cancel_block()
 
-    # Using chain controller
     def check_blocks(self, block_ids):
+        for block_id in block_ids:
+            if block_id.hex() not in self._block_cache:
+                raise UnknownBlock(block_id.hex())
+
+    def get_block_statuses(self, block_ids):
+        """Returns a list of tuples of (lblock id, BlockStatus) pairs.
```
jsmitchell: s/lblock/block/

validator/src/journal/chain_ffi.rs
```diff
@@ -283,14 +375,20 @@ pub extern "C" fn chain_controller_chain_head(
...
-        *block = chain_head.to_py_object(py).steal_ptr();
+        let controller = (*(chain_controller
+            as *mut ChainController<PyBlockCache, PyBlockValidator>))
+            .light_clone();
+
+        let chain_head = py.allow_threads(move || controller.chain_head());
+
+        // This is relying on the BlockWrapper being backed by a PyObject and `into_py_object()`
+        // not incrementing the reference count on the PyObject when passing up to Python. If these
+        // changes are invalidated, memory leaks may occure to BlockWrappers with incorrect
```
jsmitchell: s/occure/occur/

validator/src/journal/publisher.rs
```diff
@@ -589,35 +516,27 @@ impl BlockPublisher {
...
-        let mut block_publisher = self.publisher.clone();
+        let block_publisher = self.publisher.clone();
         builder
             .spawn(move || {
-                let mut now = Instant::now();
-                let check_period = { block_publisher.check_publish_block_frequency };
                 loop {
                     // Receive and process a batch
-                    match batch_rx.get(check_period) {
+                    match batch_rx.get(Duration::from_millis(100)) {
```
jsmitchell: What's the consequence of this 100 millisecond timeout?

peterschwarz: The timeout allows the loop to check to see if the process has been flagged to exit.  

A possibly better solution would be to use exit channels that we can select on, but this requires a discussion about channel libraries, since this is not yet supported in the standard library.

validator/packaging/validator.toml.example
aludvik: This is not a typo, the default port is 5050: https://github.com/hyperledger/sawtooth-core/blob/master/validator/sawtooth_validator/config/validator.py#L34

PR #1774, opened 2018-07-17
Fix new python lint from Pylint 2.0

aludvik:


aludvik:
Made a small mistake in converting to using the in pattern for comparing against multiple possible values.

aludvik:
@peterschwarz updated

PR #1775, opened 2018-07-17
Add BlockManager ffi functions

boydjohnson:
~This only adds functions that are used by the completer and proxy on
the block cache.~


boydjohnson:
While evaluating the use of block_cache in the BlockValidator it seems necessary to expose a full iterator for branch_diff in the BlockManager. I will reopen this when that work is done.

boydjohnson:
This PR is built on #1793 and only the last 3 commits are of this PR.

PR #1779, opened 2018-07-19
Implement Peering Updates

aludvik:
A couple of peering updates implementations were missed that are needed by Raft.

### DIFF
--- a/validator/sawtooth_validator/consensus/notifier.py
+++ b/validator/sawtooth_validator/consensus/notifier.py
@@ -41,7 +41,7 @@ class ConsensusNotifier:
         self._notify(
             validator_pb2.Message.CONSENSUS_NOTIFY_PEER_CONNECTED,
             consensus_pb2.ConsensusNotifyPeerConnected(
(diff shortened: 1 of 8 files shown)


validator/sawtooth_validator/consensus/notifier.py
```diff
@@ -28,20 +29,22 @@ class ConsensusNotifier:
...
         self._service = consensus_service
+        self._registered_engines = ConcurrentSet()
 
     def _notify(self, message_type, message):
-        futures = self._service.send_all(
-            message_type,
-            message.SerializeToString())
-        for future in futures:
-            future.result()
+        if self._registered_engines:
```
peterschwarz: For this statement to work, you need to implement `__bool__` on `ConcurrentSet` - otherwise this is always true (the default `bool` operation is `None==False`, `!None == True`)

ltseeley: When `__bool__` isn't defined, Python calls `__len__` instead: https://docs.python.org/3/reference/datamodel.html#object.__bool__

PR #1781, opened 2018-07-19
Build fixups

rberg2:
Tags are required to get the dev version number and JJB multibranch
fetches without tags.


vaporos:
LGTM

PR #1783, opened 2018-07-20
Update "Genesis Operation" section of Journal (in Arch doc)

chenette:
Correct and clarify to describe current architecture of genesis batch and
block creation; add links to relevant topics in other sections.

### DIFF
--- a/docs/source/architecture/journal.rst
+++ b/docs/source/architecture/journal.rst
@@ -370,22 +370,35 @@ it.
 A delay is employed in the checking loop to ensure that there is time for the
 batch processing to occur.
 
+
 Genesis Operation
 =================
 
-The Journal supports Genesis operation. This is the action of creating a root of
-the chain (the Genesis block) when the block store is empty. This operation is
-necessary for bootstrapping a validator network with the desired consensus
-model, any deployment-specific configuration settings, as well as any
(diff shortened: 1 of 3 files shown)


docs/source/architecture/journal.rst
```diff
@@ -394,84 +407,96 @@ The file contains a protobuf-encoded list of batches:
...
+   genesis block) to the file ``block-chain-id`` in the validator’s data
+   directory. The blockchain ID is used to mark this validator's genesis block
+   as the correct one for the blockchain.
+
+Genesis Block Processing
+------------------------
+
+To complete the genesis operation process and commit the genesis block, all
+necessary transaction processors must be running. At a minimum, the Settings
+transaction processor (:doc:`../cli/settings-tp`) is required.
```
aludvik: This isn't true. You can deploy a validator without a Settings transaction processor if no settings are being set or you could deploy it with your own implementation of a transaction processor to handle on-chain settings. I would change "At a minimum, the Settings transaction processor (:doc:`../cli/settings-tp`) is required." to "If any transactions in the genesis block modify settings, the Settings transaction processor (:doc:`../cli/settings-tp`) or an alternative implementation should be among those started."

PR #1785, opened 2018-07-20
Cleanup workload package CI

aludvik:
Update the package metadata and add intkey workload. This make it easier for other repos to run tests that use these workload packages.


aludvik:
Missed the copy-debs.yaml modification

PR #1786, opened 2018-07-22
Update Journal chapter in Architecture guide

chenette:
- Redefine "journal" as the group of validator components that handle the
  processing of batches and proposed blocks.

- Update picture to show the current architecture (add client box next to gossip box, remove
  consensus pieces). Added the XML source for the picture, which was
  created with draw.io. 

- Reorganize subsections to match content and picture.

- Rewrite to remove obsolete and inconsistent information.

- Replace "The Consensus Interface" (obsolete) with a generic summary
  of the consensus interface.

- Remove "Genesis Operation"; (its update is in a separate PR).


docs/source/architecture/journal.rst
```diff
@@ -2,476 +2,196 @@
...
 *******
 
-The Journal is responsible for maintaining and extending the blockchain for the
-validator. This responsibility involves validating candidate blocks, evaluating
-valid blocks to determine if they are the correct chain head, and generating
-new blocks to extend the chain.
+The `journal` is the group of validator components that work together to handle
+batches and proposed blocks. These components are responsible for completing
+published blocks, publishing batches into blocks to extend the chain, and
+validating proposed blocks to determine if they are the correct chain head.
```
aludvik: Suggest changing "to determine if they are the correct chain head" to "to determine if they should be considered for the new chain head", since the consensus engine decides what the next chain head should be.

docs/source/architecture/journal.rst
```diff
@@ -2,476 +2,196 @@
...
 
-The Journal is the consumer of Blocks and Batches that arrive at the validator.
-These Blocks and Batches arrive via interconnect, either through the gossip
-protocol or the REST API. The newly-arrived Blocks and Batches are sent to the
-Journal, which routes them internally.
+Blocks and batches arrive via interconnect, either through the gossip protocol
+or the REST API. The processing of these blocks and batches is handled in
```
aludvik: change "or the REST API" to "or from client requests" since clients may submit batches directly to the validator, and the way this is phrased makes it sound like the REST API is required."

docs/source/architecture/journal.rst
```diff
@@ -2,476 +2,196 @@
...
+is high.
 
-If a block is not present in the BlockCache, it will look in the BlockStore for
-the block. If it is not found or the lookup fails, the block is unknown to the
-system. If the block is found in the BlockStore it is loaded into the
-BlockCache and marked as valid. All blocks in the BlockStore are considered
-valid.
+This approach is flexible enough to work with different consensus algorithms.
+Sawtooth includes a *consensus interface* that communicates the components
+in the journal.
```
aludvik: The last part of this sentence doesn't make sense. Maybe you meant "that defines communication between the consensus and components in the journal."

docs/source/architecture/journal.rst
```diff
@@ -2,476 +2,196 @@
...
-mean (a measure of the total amount of time spent waiting) to determine the
-valid fork. Consensus algorithms with finality, such as PBFT, will only ever
-produce blocks that extend the current head. These algorithms will never have
-forks to resolve. The ForkResolver for these algorithms with finality will
-always select the new block that extends the current head.
-
-The ChainController
-===================
+The Completer checks for dependencies, ensures that the previous block exists,
+and makes sure that the batches exist in the Block Store or BlockCache.
```
aludvik: space between "Block" and "Store"

docs/source/architecture/journal.rst
```diff
@@ -2,476 +2,196 @@
...
-batch processing to occur.
+The ChainController
+===================
 
-Genesis Operation
-=================
+The ChainController is responsible for maintaining and extending the
+blockchain for the validator. This responsibility involves validating proposed
+blocks, evaluating valid blocks to determine if they are the correct chain
+head, and generating new blocks to extend the chain.
```
aludvik: The ChainController does not "generating new blocks to extend the chain". Also, consensus handles "evaluating valid blocks to determine if they *should be" the new chain head" and the chain controller just handles making it so if consensus decides it should be.

docs/source/architecture/journal.rst
```diff
@@ -2,476 +2,196 @@
...
-genesis-time transactions for an application's Transaction Family.
+The ChainController determines which chain the validator is currently on and
+coordinates any change-of-chain activities that need to happen.
 
-Genesis Batch Creation
-----------------------
+The ChainController is designed to be able to handle multiple block validation
+activities simultaneously. For instance, if multiple forks form on the network,
+the ChainController can process blocks from all of the competing forks
+simultaneously. This allows progress to be made even when several deep forks are
```
aludvik: This is false. The ChainController is incapable of validating blocks concurrently as it is written today. It can only validate one block at a time.

docs/source/architecture/journal.rst
```diff
@@ -2,476 +2,196 @@
...
+BlockStore for the block. If it is not found or the lookup fails, the block is
+marked as unknown. If the block is found in the BlockStore, it is loaded into
+the BlockCache and marked as valid.
 
-* The genesis.batch file exists
-* There is no block specified as the chain head
+The BlockCache keeps blocks that are currently relevant, tracked by the last
+time the block was accessed. Periodically, blocks that have not been accessed
+recently are purged from the block cache, but only if none of the other blocks
+in the BlockCache reference those blocks as predecessors.
```
aludvik: I am not really sure why we are documenting the BlockCache and BlockStore. These are internal data structures that are in the process of being replaced.

docs/source/architecture/journal.rst
chenette: Oops - I dropped a word by mistake. I meant "that communicates WITH the components in the journal."  I'm trying to avoid any hints of the upcoming consensus engine/consensus proxy for now.

docs/source/architecture/journal.rst
peterschwarz: But it is accurate.

chenette: We're documenting it because the picture shows these two items. (And because it was documented before.)  I vote that we leave the information here for now.

chenette:
I changed my initial comment to reflect the current contents of this PR, and to note that "Genesis Operation" (update in a separate PR) didn't move to a separate file.

PR #1790, opened 2018-07-24
Change queue limit multiplier

peterschwarz:
Change the queue limit multiplier to a constant and bump it from 2 to 10


validator/src/journal/publisher.rs
```diff
@@ -884,7 +886,7 @@ impl QueueLimit {
     pub fn get(&self) -> usize {
         // Limit the number of items to 2 times the publishing average.  This
```
jsmitchell: Make sure you also update the comment

PR #1793, opened 2018-07-25
Use a single store of reference counts in the BlockManager

boydjohnson:


boydjohnson:
@aludvik It is an optimization. We could always look up the block and and get it's previous block id. It is nice not to in `unref` where we can walk back, only accessing RefCounts.

aludvik:
Okay, thanks. For that matter, is including the block id in RefCount necessary, since you already have it as the key in the map? We should weigh keeping additional fields in RefCount against the cost of making sure they stay in sync. Mapping the block id to the RefCount, and then duplicating the id and the previous id, introduces an opportunity for those fields to fall out of sync (there is nothing stopping you from accidentally mapping a block id to a RefCount with a different block id).

peterschwarz:
Not having the block id's in the RefCount doesn't solve the problem of having things "fall out of sync".  You could still have a block id point to the wrong ref count, except there would be no way to know without those ids.

PR #1794, opened 2018-07-25
Add configuration settings for network, component and signature thread pool workers

peterschwarz:
This PR adds configuration settings for the validator's network, component, and signature thread pool max workers.  This allows for easier tuning of these settings during performance testing, without having to run custom builds of the validator.

Note: These settings are config file only.

### DIFF
--- a/validator/packaging/validator.toml.example
+++ b/validator/packaging/validator.toml.example
@@ -67,6 +67,19 @@ minimum_peer_connectivity = 3
 # The maximum number of peers that will be accepted.
 maximum_peer_connectivity = 10
 
+# The following settings are for tuning the thread allocation in several of the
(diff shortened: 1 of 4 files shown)


validator/packaging/validator.toml.example
```diff
@@ -67,6 +67,19 @@ minimum_peer_connectivity = 3
 # The maximum number of peers that will be accepted.
 maximum_peer_connectivity = 10
 
+# The following settings are for tuning the thread allocation in several of the
+# thread pools used by the validator.  These settings should be modified by
```
chenette: I suggest stronger, simpler wording: 

"Only advanced users should change these settings."
     or
"Do not change these settings unless you are an experienced Sawtooth user."

PR #1800, opened 2018-07-27
Increment Refcount on Branch/Branch Diff iteration

peterschwarz:
In order to insure that a branch or branch-diff iterator will not be modified during iteration, the tip block of the branches in question will be ref'ed with the `BlockManager`.  When the iterator is dropped, the respective blocks will be unref'ed.

Additionally:

- Added more tests
- Fixed GetBlockIterator to skip missing block ids
- Replaced owned Strings for block ids with string references


validator/src/journal/block_manager.rs
```diff
@@ -557,20 +557,59 @@ impl Iterator for GetBlockIterator {
...
     next_block_id: String,
     blockstore: Option<String>,
 }
 
 impl BranchIterator {
     fn new(state: Arc<RwLock<BlockManagerState>>, first_block_id: String) -> Self {
+        let next_block_id = {
+            let mut block_manager = state
+                .write()
+                .expect("No log holder should have poisoned the lock");
```
aludvik: The argument should state the error that occurred, not the expected behavior. Also there is a typo.

validator/src/journal/block_manager.rs
```diff
@@ -557,20 +557,59 @@ impl Iterator for GetBlockIterator {
...
+            match block_manager.ref_block(&first_block_id) {
+                Ok(_) => first_block_id.clone(),
+                Err(BlockManagerError::UnknownBlock) => NULL_BLOCK_IDENTIFIER.to_string(),
+
+                Err(err) => {
+                    error!(
+                        "Unable to ref block at {}: {:?}; ignoring",
+                        &first_block_id, err
+                    );
+                    NULL_BLOCK_IDENTIFIER.to_string()
```
aludvik: Handling this without panic'ing makes sense in the context of the existing API for the branch_diff iterator, but it raises the question whether the return type here should be `Result<Self, Error>` so we can return an error when the tip isn't found, rather than returning an empty iterator.

validator/src/journal/block_manager.rs
```diff
@@ -557,20 +557,59 @@ impl Iterator for GetBlockIterator {
...
             blockstore: None,
         }
     }
 }
 
+impl Drop for BranchIterator {
+    fn drop(&mut self) {
+        let mut block_manager = self.state
+            .write()
+            .expect("No log holder should have poisoned the lock");
```
aludvik: The argument should state the error that occurred, not the expected behavior. Also there is a typo.

validator/src/journal/block_manager.rs
```diff
@@ -583,7 +581,7 @@ impl BranchIterator {
         };
         BranchIterator {
             state,
-            initial_block_id: first_block_id.clone(),
+            initial_block_id: next_block_id.clone(),
```
aludvik: Is this a typo, or should it be in a different commit?

peterschwarz: Yeah, that looks like it should be in the previous commit, as should the changes to drop

PR #1802, opened 2018-07-31
Fix Python & JS link inconsistencies in SDK API Reference cover page

chenette:

### DIFF
--- a/docs/source/sdks.rst
+++ b/docs/source/sdks.rst
@@ -7,9 +7,11 @@ Python
 ------
 
 .. toctree::
+   :maxdepth: 1
 
-   sdks/python_sdk/processor.rst
-   sdks/python_sdk/sawtooth_signing.rst
+   Transaction Processor <sdks/python_sdk/processor.rst>
+
+   Signing <sdks/python_sdk/sawtooth_signing.rst>
 
 Go
 --
@@ -21,11 +23,13 @@ Go
 
 Javascript
 ----------
-- `Javascript SDK Reference
+
+- `Transaction Processor and Signing
   <https://sawtooth.hyperledger.org/docs/js-sdk/releases/latest/>`__
 
 Rust
 ----
+
 - `Transaction Processor
   <rust_sdk/doc/sawtooth_sdk/processor/index.html#http://>`__
 - `Signing


docs/source/sdks.rst
```diff
@@ -6,10 +6,9 @@ SDK API Reference
 Python
 ------
 
-.. toctree::
+- `Transaction Processor <sdks/python_sdk/processor.html>`_
 
-   sdks/python_sdk/processor.rst
-   sdks/python_sdk/sawtooth_signing.rst
+- `Signing <sdks/python_sdk/sawtooth_signing.html>`_
```
rberg2: Is it expected at this point that these links don't work in the build artifact? https://build.sawtooth.me/job/Sawtooth-Hyperledger/job/sawtooth-core/job/PR-1802/1/artifact/docs/build/html/sdks.html

docs/source/sdks.rst
```diff
@@ -6,10 +6,9 @@ SDK API Reference
 Python
 ------
 
-.. toctree::
+- `Transaction Processor <sdks/python_sdk/processor.html>`_
 
-   sdks/python_sdk/processor.rst
```
peterschwarz: The python docs should not have changed, since we didn't move those.  The existing links work in current master

chenette: Agreed.  The Python docs are auto-generated, and they're still in the same place.  I had to change the extension from "rst" to "html" when I removed the toc construct, but that was just to get the Sphinx docs to generate the correct (same) URLs for the python docs.  

The links still point to the published SDK files on master:  
https://sawtooth.hyperledger.org/docs/core/nightly/master/sdks/python_sdk/processor.html
https://sawtooth.hyperledger.org/docs/core/nightly/master/sdks/python_sdk/sawtooth_signing.html

docs/source/sdks.rst
chenette: The build artifacts have a few glitches, so I'm not surprised.  (See my response to Peter's comment for more info.)

docs/source/sdks.rst
peterschwarz: When are they generated?

chenette: They seem to be generated during a doc build, but I don't know anything about the exact mechanics and timing.  Perhaps @vaporos can answer this question?

chenette: I have learned that my change to `docs/source/conf.py` is a bad idea, because it turns off the auto-generation of the python SDK docs.  However, if I remove the detailed contents list for the Python SDK, the doc build fails because`sdks/python_sdk/processor.rst` and `sdks/python_sdk/sawtooth_signing.rst` aren't referenced in a TOC.  

Sphinx has an `:orphan:` metadata directive that would ignore this error, but I can't add it to the files because they're auto-generated.  I will need help to figure out where (and how) these auto-generated files are created:  

https://sawtooth.hyperledger.org/docs/core/nightly/master/sdks/python_sdk/processor.html

https://sawtooth.hyperledger.org/docs/core/nightly/master/sdks/python_sdk/sawtooth_signing.html

chenette: More importantly, I need to find out if it's possible to insert or prepend an `:orphan:` line in these auto-generated files.

chenette:
I updated this PR to fix the Python entries without breaking the auto-generated Python SDK docs. (The original PR added files to the exclude section of `docs/source/conf.py`, which was the wrong thing to do.) The Python SDK docs are once again being generated correctly.

PR #1805, opened 2018-07-31
Refactor block store access

aludvik:


aludvik:
Added a commit to also remove block cache from batch injectors.

PR #1806, opened 2018-07-31
Integrate BlockManager into Validator

boydjohnson:


validator/sawtooth_validator/journal/block_validator.py
```diff
@@ -251,83 +299,105 @@ def _check_chain_head_updated(self, chain_head, block):
...
-            self._validate_batches_in_block(blkw, prev_state_root)
+            execution_results, num_txns = \
+                self._validate_batches_in_block(block, prev_state_root)
 
-            blkw.status = BlockStatus.Valid
+            return BlockValidationResult(
+                execution_results=execution_results,
+                block_id=block.header_signature,
+                num_transactions=num_txns,
+                status=1)
```
aludvik: We should continue using an enum for block statuses here.

validator/sawtooth_validator/journal/block_manager.py
```diff
@@ -61,6 +61,13 @@ def put(self, branch):
                  self.pointer,
                  ctypes.py_object(branch))
 
+    def __contains__(self, item):
+        try:
+            next(self.get([item]))
+            return True
+        except StopIteration:
+            return False
+
```
aludvik: This is a somewhat non-intuitive implementation. Can you add a comment around the fact that it is depending on `get()` to return either exactly 1 or exactly 0 items?

validator/sawtooth_validator/consensus/proxy.py
```diff
@@ -187,10 +201,9 @@ def state_get(self, block_id, addresses):
...
-                for block_id in block_ids
-            ]
-        except KeyError:
-            raise UnknownBlock()
+        for block_id in block_ids:
+            if block_id.hex() not in self._block_manager:
+                raise UnknownBlock()
+        return [b for b in self._block_manager.get([
+            block_id.hex() for block_id in block_ids
+        ])]
```
aludvik: I think this could be done more efficiently by constructing the list to return first, and then checking its length against the length of the input block_ids to determine if a block was missing.

validator/src/journal/publisher.rs
```diff
@@ -173,15 +173,15 @@ impl SyncBlockPublisher {
             .pending_batches
             .rebuild(Some(committed_batches), Some(uncommitted_batches));
 
-        if let Some(block_id) = previous_block_id {
-            self.initialize_block(state, &block_id)
+        if let Some(prev) = previous_block_id {
```
aludvik: The previous_block_id variable should be renamed to reflect that the thing it points to is a Block and not a block id.

validator/tests/test_consensus/tests.py
```diff
@@ -284,29 +284,29 @@ def test_cancel_block(self):
...
+        self._mock_block_manager["56"] = "block0"
+        self._mock_block_manager["78"] = "block1"
         self._proxy.check_blocks(block_ids)
 
         with self.assertRaises(UnknownBlock):
             self._proxy.check_blocks([bytes([0x00])])
 
     def test_commit_block(self):
-        self._mock_block_cache["34"] = "a block"
+        self._mock_block_manager["34"] = "a block"
```
aludvik: I am confused by how this works, since ConsensusProxy expects to be able to call `block_manager.get(["34"])` and get back an iterator, but this fails with `TypeError: unhashable type: 'list'` when I try it manually. I think you need to make the mock act more like the real thing.

validator/sawtooth_validator/journal/block_validator.py
```diff
@@ -13,6 +13,9 @@
 # limitations under the License.
 # ------------------------------------------------------------------------------
 
+# pylint disable: inconsistent-return-statements
```
aludvik: Which functions does this apply to? Can it be moved to just those functions, instead of across the entire file?

validator/sawtooth_validator/journal/block_validator.py
```diff
@@ -86,8 +98,8 @@ def __init__(self,
                  thread_pool=None):
         """Initialize the BlockValidator
         Args:
-            block_cache: The cache of all recent blocks and the processing
-                state associated with them.
+            block_manager: The componenet that stores blocks and maintains
```
aludvik: typo s/componenet/component/

validator/sawtooth_validator/journal/block_validator.py
```diff
@@ -106,7 +118,13 @@ def __init__(self,
         Returns:
             None
         """
-        self._block_cache = block_cache
+
+        self._block_manager = block_manager
+        self._block_store = block_store
+        self._chain_head_fn = None
+        self._block_validity_fn = None
+        self._block_validity_cache = _Cache(100)
```
aludvik: This 100 is meaningless without implementing the capacity limit, so should be dropped until that is implemented.

validator/sawtooth_validator/journal/event_extractors.py
```diff
@@ -14,13 +14,26 @@
...
+from sawtooth_validator.protobuf.block_pb2 import BlockHeader
 from sawtooth_validator.protobuf.events_pb2 import Event
 from sawtooth_validator.protobuf.transaction_receipt_pb2 import StateChangeList
 
 
 class BlockEventExtractor(EventExtractor):
     def __init__(self, block):
-        self._block = block
+
+        if isinstance(block.header, bytes):
```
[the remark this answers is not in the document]
boydjohnson: Thanks, good catch.

validator/src/journal/publisher_ffi.rs
```diff
@@ -380,7 +381,7 @@ pub extern "C" fn block_publisher_on_chain_updated(
     let mut publisher = unsafe { (*(publisher as *mut BlockPublisher)).clone() };
     py.allow_threads(move || {
         publisher.publisher.on_chain_updated_internal(
-            chain_head,
+            chain_head.block(),
```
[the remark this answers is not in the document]
boydjohnson: That method is only used in the BlockPublisher unit tests. I haven't removed the use of blockwrappers from those tests, because this is just a chain controller PR.

validator/src/journal/publisher.rs
```diff
@@ -238,7 +233,7 @@ impl SyncBlockPublisher {
                     "get_setting",
                     (
                         "sawtooth.publisher.max_batches_per_block",
-                        previous_block.block().state_root_hash.clone(),
+                        previous_block.state_root_hash.clone(),
```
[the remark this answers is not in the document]
boydjohnson: Cool, thanks!

validator/sawtooth_validator/journal/block_validator.py
boydjohnson: Nice catch. I will fix this on my next push of updates.

validator/tests/test_consensus/tests.py
boydjohnson: I changed to have a MockBlockManager class that implements `get(block_ids)` but also __setitem__ so that this still works. The reason  why that is nice is so you can specify the header_signature separate from the "block". In the case of the one you called out, the string "a block" doesn't have an attribute `header_signature` and it save me some work in constructing blocks that do.

validator/src/journal/publisher_ffi.rs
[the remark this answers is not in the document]
peterschwarz: It will get fixed in the publisher PR, so no worries

validator/sawtooth_validator/journal/block_validator.py
boydjohnson: Thanks. As discussed I will modify the ChainController/BlockValidator/BlockScheduler interaction so that this cache is not necessary.

validator/sawtooth_validator/journal/block_validator.py
```diff
@@ -13,6 +13,9 @@
 # limitations under the License.
 # ------------------------------------------------------------------------------
 
+# pylint: disable=inconsistent-return-statements
```
aludvik: Which functions does this apply to? Can it be moved to just those functions, instead of across the entire file?

validator/sawtooth_validator/journal/block_validator.py
aludvik: It looks like you implemented the capacity? Are you keeping this in the PR?

boydjohnson: No, I have a fix that I will push soon that undoes this cache. This was the bandaid "cache".

validator/sawtooth_validator/journal/block_validator.py
boydjohnson: I can't figure out how to apply this to just the functions. I could leave the line numbers that it applies to in a comment. I thought that since we are re-writing this in Rust very soon, I shouldn't spend too much time on it.

validator/src/journal/publisher.rs
```diff
@@ -469,6 +479,9 @@ impl SyncBlockPublisher {
         let mut candidate_block = None;
         mem::swap(&mut state.candidate_block, &mut candidate_block);
         if let Some(mut candidate_block) = candidate_block {
+            self.block_manager
+                .unref_block(&candidate_block.previous_block_id())
+                .expect("Unable to unref block that was ref'ed during initialize block");
```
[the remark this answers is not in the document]
peterschwarz: Added a commit to address this.

PR #1808, opened 2018-08-02
Integrate BlockManager in BlockPublisher

peterschwarz:
Replaces the use of the block cache in the BlockPublisher with the BlockManager.  Additionally, it refs the previous block when it initializes a block, and unref's the same when the block is canceled.

This is based on #1806


peterschwarz:
These changes have been added to #1806

PR #1809, opened 2018-08-02
Enable devmode liveness test

aludvik:


boydjohnson:
@aludvik When I ran this test locally it would produce blocks too quickly, and so only 1 batch would be in every block. I set the min_wait_time to 3 and the max_wait_time to 5. I had good results with that.

PR #1812, opened 2018-08-03
Fix rust lint

eloaverona:
A new update of rustfmt broke the rust lint tests.
This PR:
- Updates bin/run_lint_rust and bin/run_lint_validator to correctly run rustfmt 
- Fix all new lint errors


boydjohnson:
Do we know why this wasn't being linted previously?

aludvik:
> It seems that rustfmt is not working? It's breaking the rust lint tests.
I'm getting this error:  `Unrecognized option: 'write-mode'` . (there's a build in Jenkins that also failed with this error)
On the file `bin/run_lint_rust` on line 42 `diff=$(cargo fmt -- --write-mode=diff)`. It seems that now instead of  `--write-mode=diff`it should be `--check`. But that is not working on my local machine either.
Source about write-mode and check: https://github.com/rust-lang-nursery/rustfmt#checking-style-on-a-ci-server

@eloaverona from chat

PR #1813, opened 2018-08-07
Fix timestamp validation in BlockInfo to match the documentation

eugene-babichenko:
BlockInfo was validating timestamps as follows:

```python
if (timestamp - now) > tolerance:
```

However, such verification leads to incorrect validation of blocks when downloading old blocks. To make it match the documentation which says that 

> The timestamp in the new BlockInfo message must be less than the peer’s measured local time, adjusted to UTC, plus a network time synchronization tolerance that is greater than or equal to zero.

I changed the verification to this:

```python
if timestamp > now + tolerance:
```

### DIFF
--- a/families/block_info/sawtooth_block_info/processor/handler.py
+++ b/families/block_info/sawtooth_block_info/processor/handler.py
@@ -46,7 +46,7 @@ def validate_hex(string, length):
 
 def validate_timestamp(timestamp, tolerance):
     now = time.time()
-    if (timestamp - now) > tolerance:
+    if timestamp > now + tolerance:
         raise InvalidTransaction(
             "Timestamp must be less than local time."
             " Expected {0} in ({1}-{2}, {1}+{2})".format(


peterschwarz:
The existing code works as expected, and in fact, is equivalent to what's in the documentation functionally.  If now is in the future, the left side of the comparison will be negative, which will still have the clause evaluate to `False`.  

It works, as-is, in catch-up mode (i.e. with old blocks) - in fact, the current line of code fixes an issue where the previous version did not, so we've addressed this problem in the past.

peterschwarz:
I do want to say thanks for the PR, though.  One thing to note, that in future PR's, make sure you commit using `git commit -s`: we require signed-off commits.

Please refer to the [Contributor's Guide](https://sawtooth.hyperledger.org/docs/core/releases/latest/community/contributing.html).

peterschwarz:
Closing: See above comments

PR #1814, opened 2018-08-08
Fixes for some of the clippy lint

boydjohnson:
This PR fixes some of the clippy lint in areas that don't have a PR up right now.

### DIFF
--- a/validator/src/metrics/mod.rs
+++ b/validator/src/metrics/mod.rs
@@ -104,9 +104,9 @@ impl MetricsCollectorHandle {
 
         let py_level = self
             .py_metrics
-            .get(py, into_level_str(level.unwrap_or(Default::default())))
+            .get(py, into_level_str(level.unwrap_or_default()))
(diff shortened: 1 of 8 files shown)


validator/src/metrics/mod.rs
```diff
@@ -104,9 +104,9 @@ impl MetricsCollectorHandle {
 
         let py_level = self
             .py_metrics
-            .get(py, into_level_str(level.unwrap_or(Default::default())))
+            .get(py, into_level_str(level.unwrap_or_default()))
             .expect("Failed to get metric level");
-        let py_tags: PyDict = tags.unwrap_or_else(|| HashMap::new()).into_py_object(py);
+        let py_tags: PyDict = tags.unwrap_or_else(HashMap::new).into_py_object(py);
```
agunde406: this looks like a different change then what is in the commit message? Atleast a comment should be added to the commit about this change.

PR #1815, opened 2018-08-08
Port of ChainCommitState to Rust

boydjohnson:


validator/src/journal/chain_commit_state.rs
```diff
@@ -21,6 +21,32 @@ use cpython;
...
+pub struct ChainCommitState<BS: IndexedBlockStore> {
+    block_store: BS,
+    ancestor: Option<Block>,
+    uncommitted_batch_ids: Vec<String>,
+    uncommitted_txn_ids: Vec<String>,
+}
+
+fn check_no_duplicates(ids: &[&str]) -> Option<String> {
+    for (i, id1) in ids.iter().enumerate() {
+        if ids[0..i].contains(id1) || ids[i + 1..ids.len()].contains(id1) {
```
peterschwarz: seems like you can just check `i..ids.len()`.  If you are iterating from `0` on, you don't need to look at the slice before, you've already checked those as duplicates.

validator/src/journal/chain_commit_state.rs
```diff
@@ -104,6 +104,122 @@ impl<BS: IndexedBlockStore> ChainCommitState<BS> {
...
+
+    pub fn validate_no_duplicate_batches(
+        &self,
+        batch_ids: &[&str],
+    ) -> Result<(), ChainCommitStateError> {
+        if let Some(batch_id) = check_no_duplicates(batch_ids) {
+            return Err(ChainCommitStateError::DuplicateBatch(batch_id));
+        }
+        for id in batch_ids {
+            if self.uncommitted_batch_ids.contains(&(*id).into()) {
```
peterschwarz: You should have to do this: you are turning a reference into a `String` and then a `&str`

validator/src/journal/chain_commit_state.rs
```diff
@@ -21,6 +21,32 @@ use cpython;
...
+}
+
+pub struct ChainCommitState<BS: IndexedBlockStore> {
+    block_store: BS,
+    ancestor: Option<Block>,
+    uncommitted_batch_ids: Vec<String>,
+    uncommitted_txn_ids: Vec<String>,
+}
+
+fn check_no_duplicates(ids: &[&str]) -> Option<String> {
```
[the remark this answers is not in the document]
boydjohnson: I don't see any extra indentation.

Yes, it returns the id of a duplicate.

validator/src/journal/chain_commit_state.rs
boydjohnson: This could be written as a double for loop, but is written with a for loop and contains.

`for (i, id) in ["1", "2", "1"]`

and then `if ["1"].contains("1") || ["1"].contains("1")` for the first loop through.

validator/src/journal/chain_commit_state.rs
boydjohnson: Yeah, I agree it looks a little funny. if I just dereference the `&&str` I get this error
```error[E0308]: mismatched types
   --> src/journal/chain_commit_state.rs:156:50
    |
156 |             if self.uncommitted_txn_ids.contains(*id) {
    |                                                  ^^^ expected struct `std::string::String`, found str
    |
    = note: expected type `&std::string::String`
               found type `&str`
```

boydjohnson:
I updated the signatures of the 2 validation functions that operate on `ids` to `Vec<String>`. I think this or `&[String]` would be correct.

validator/src/journal/chain_commit_state.rs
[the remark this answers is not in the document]
peterschwarz: Sorry, I meant "Intention"

validator/src/journal/block_store.rs
```diff
@@ -38,7 +38,21 @@ pub trait BlockStore: Sync + Send {
     fn iter<'a>(&'a self) -> Result<Box<Iterator<Item = Block> + 'a>, BlockStoreError>;
 }
 
-#[derive(Default)]
+/// A block store that has fast lookups for transactions and batches
+pub trait IndexedBlockStore: BlockStore {
```
aludvik: I think this makes more sense as a separate trait rather than an extension of the BlockStore trait. I would suggest:
```
pub trait BatchIndex {
  fn contains(&self, id: &str) -> bool;
  fn block_by_id(&self, id: &str) -> Option<Block>;
}
```
We will want this kind of indexing when extend batch and transaction indexing to the BlockManager so duplicates and dependencies can be checked correctly.

PR #1817, opened 2018-08-09
Fix state bad head Internal server error [STL-1345].

shresthichauhan:
Return JSON response with error code 60. Head id added 
to list_state handler function in rest api route handlers.

Signed-off-by: shresthichauhan <shresthix.chauhan@intel.com>

### DIFF
--- a/rest_api/sawtooth_rest_api/route_handlers.py
+++ b/rest_api/sawtooth_rest_api/route_handlers.py
@@ -262,7 +262,7 @@ class RouteHandler:
             paging: Paging info and nav, like total resources and a next link
         """
         paging_controls = self._get_paging_controls(request)
-
+        self._get_head_id(request)
         head, root = await self._head_to_root(request.url.query.get(
             'head', None))
         validator_query = client_state_pb2.ClientStateListRequest(


shresthichauhan:
Thank you so much for your review comments.

Please find actual result returned with this additional line. This is to demonstrate bug is resolved.

after_fix_json_response_60

As we know state head is block ID that is 128 character hex-strings. So fix will return JSON response with error code 60 for bad Head parameter. 

Sure will add a unit test case in this PR and will inform you about the same.

rest_api/sawtooth_rest_api/route_handlers.py
```diff
@@ -262,7 +262,7 @@ def __init__(
             paging: Paging info and nav, like total resources and a next link
         """
         paging_controls = self._get_paging_controls(request)
-
+        head_id=self._get_head_id(request)
```
peterschwarz: This should be pushed down into the `_head_to_root` function - this particular error can occur in the `fetch_state` function as well.

Likewise, as that head_id is unused, I suspect the build is failing due to lint

shresthichauhan: Thanks for suggesting _head_to_root function.

PR specific to list_state function (for endpoint '/state'). Please refer "start_rest_api" in rest_api.py, fetch_state for '/state/{address}' endpoint.

As per my understanding, ClientStateListRequest is only function where head_id not pass as parameter in comparison to other functions (ClientTransactionListRequest, ClientBatchListRequest and ClientBlockListRequest).

message ClientStateListRequest {
	string state_root = 1;
	string address = 3;
	ClientPagingControls paging = 4;
	repeated ClientSortControls sorting = 5;
}

message ClientBlockListRequest {
	string head_id = 1;
	repeated string block_ids = 2;
	ClientPagingControls paging = 3;
	repeated ClientSortControls sorting = 4;
}

Adding head_id in ClientStateListRequest, impact many other functionality. Please suggest.

clientblocklistrequest

rest_api/sawtooth_rest_api/route_handlers.py
```diff
@@ -262,7 +262,7 @@ def __init__(
             paging: Paging info and nav, like total resources and a next link
         """
         paging_controls = self._get_paging_controls(request)
-
+        self._get_head_id(request)
```
peterschwarz: The return value should be passed to `_head_to_root`.  This now hides the fact that you're calling that to validate the input.  A future developer may come along and delete that line, thinking its extraneous.

PR #1818, opened 2018-08-09
Port block scheduler to Rust

boydjohnson:
This Rust block scheduler will be used once the BlockValidator is ported
to Rust. At that point the Python BlockScheduler will be removed.

### DIFF
--- a/validator/src/journal/block_scheduler.rs
+++ b/validator/src/journal/block_scheduler.rs
@@ -1,0 +1,207 @@
+/*
+ * Copyright 2018 Intel Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ------------------------------------------------------------------------------
+ */
+
+use std::collections::{HashMap, HashSet};
+use std::sync::{Arc, Mutex};
+
+use block::Block;
+use journal::block_validator::BlockStatusStore;
+use journal::block_wrapper::BlockStatus;
+use journal::{block_manager::BlockManager, NULL_BLOCK_IDENTIFIER};
+use metrics;
+
+lazy_static! {
+    static ref COLLECTOR: metrics::MetricsCollectorHandle =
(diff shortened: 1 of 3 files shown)


validator/src/journal/block_scheduler.rs
```diff
@@ -0,0 +1,260 @@
...
+}
+
+pub struct BlockScheduler<F: Fn(&str) -> BlockStatus> {
+    state: RwLock<BlockSchedulerState<F>>,
+}
+
+impl<F: Fn(&str) -> BlockStatus> BlockScheduler<F> {
+    pub fn new(block_manager: BlockManager) -> Self {
+        BlockScheduler {
+            state: RwLock::new(BlockSchedulerState {
```
peterschwarz: Does this struct need to be guarded by a lock, or does the owning structure (i.e. the future port of the Block validator) need to manage the a lock anyway?

validator/src/journal/block_scheduler.rs
```diff
@@ -0,0 +1,260 @@
...
+struct BlockSchedulerState<F: Fn(&str) -> BlockStatus> {
+    pub block_manager: BlockManager,
+    pub block_validity_fn: Option<F>,
+    pub pending: HashSet<String>,
+    pub processing: HashSet<String>,
+    pub descendants: HashMap<String, Vec<Block>>,
+}
+
+impl<F: Fn(&str) -> BlockStatus> BlockSchedulerState<F> {
+    fn set_block_validity_fn(&mut self, func: F) {
```
peterschwarz: Secondly, who is providing this method? Are you going to run in to a deadlock with other locks?

validator/src/journal/block_scheduler.rs
boydjohnson: In the python version the BlockScheduler has its own lock.

validator/src/journal/block_scheduler.rs
boydjohnson: That method comes from the ChainController. The same as is used now in the BlockScheduler.

boydjohnson:
I'm going to put this PR on hold until I prototype the BlockValidator in Rust using this BlockScheduler.

validator/src/journal/block_scheduler.rs
```diff
@@ -0,0 +1,260 @@
...
+use journal::block_manager::BlockManager;
+use journal::block_wrapper::BlockStatus;
+use metrics;
+
+lazy_static! {
+    static ref COLLECTOR: metrics::MetricsCollectorHandle =
+        metrics::get_collector("sawtooth_validator.block_validator");
+}
+
+pub struct BlockScheduler<F: Fn(&str) -> BlockStatus> {
```
aludvik: Consider replacing this closure trait with a new, block validity specific trait. Something like:
```
pub trait BlockStatusStore {
  fn status(id: &str) -> BlockStatus;
}
```

PR #1819, opened 2018-08-10
Add SettingsViewFactory in Rust

boydjohnson:
The SettingsViewFactory produces SettingsView<MerkleDatabase> instances
with a particular merkle root.

### DIFF
--- a/validator/src/state/settings_view.rs
+++ b/validator/src/state/settings_view.rs
@@ -21,6 +21,9 @@ use crypto::digest::Digest;
 use crypto::sha2::Sha256;
 use protobuf;
 
+use database::lmdb::LmdbDatabase;
+
+use state::merkle::MerkleDatabase;
 use state::StateDatabaseError;
 use state::StateReader;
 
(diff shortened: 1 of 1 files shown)


validator/src/state/settings_view.rs
```diff
@@ -58,6 +59,24 @@ impl From<ParseIntError> for SettingsViewError {
     }
 }
 
+pub struct SettingsViewFactory {
+    merkle_db: MerkleDatabase,
```
peterschwarz: This should actually be the LmdbDatabase for the state database

validator/src/state/settings_view.rs
```diff
@@ -58,6 +59,24 @@ impl From<ParseIntError> for SettingsViewError {
...
+impl SettingsViewFactory {
+    pub fn new(merkle_db: MerkleDatabase) -> Self {
+        SettingsViewFactory { merkle_db }
+    }
+
+    pub fn create_settings_view<S: Into<String>>(
+        &mut self,
+        merkle_root: S,
+    ) -> Result<SettingsView<MerkleDatabase>, StateDatabaseError> {
+        self.merkle_db.set_merkle_root(merkle_root)?;
```
peterschwarz: A new merkle db should be created here, and then the merkle root set.

PR #1820, opened 2018-08-15
Fix lint in smallbank workload

boydjohnson:


perf/smallbank_workload/src/smallbank_transformer.rs
```diff
@@ -42,7 +42,7 @@ pub struct SBPayloadTransformer<'a> {
 impl<'a> SBPayloadTransformer<'a> {
     pub fn new(signer: &'a signing::Signer) -> Self {
         SBPayloadTransformer {
-            signer: signer,
+            signer,
```
aludvik: This should be a separate commit

boydjohnson: Thanks. Will fix that.

PR #1823, opened 2018-08-17
Fix clippy lint in validator ffi

boydjohnson:


validator/src/journal/publisher_ffi.rs
```diff
@@ -304,8 +304,6 @@ pub unsafe extern "C" fn block_publisher_summarize_block(
             *result = consensus.as_ptr();
             *result_len = consensus.as_slice().len();
 
-            mem::forget(result);
```
peterschwarz: I think this will drop the content immediately.  I think this change should be dropped.

peterschwarz: Actually, I think `consensus` is the correct thing to forget here.

boydjohnson: Yeah, you are right `mem::forget(consensus)` is right. `mem::forget(result)` is wrong.

PR #1826, opened 2018-08-22
Rewrite battleship processor in Rust

knkski:
Replaces the Python Battleship TP/CLI with a Rust version.


families/battleship/tests/test_tp_battleship.yaml
```diff
@@ -1,61 +0,0 @@
...
-    image: sawtooth-battleship-tp$INSTALL_TYPE:$ISOLATION_ID
-    volumes:
-      - $SAWTOOTH_CORE:/project/sawtooth-core
-    expose:
-      - 4004
-    command: battleship-tp-python -vv -C tcp://test-tp-battleship:4004
-    stop_signal: SIGKILL
-
-  test-tp-battleship:
-    build:
```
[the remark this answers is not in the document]
knkski: @agunde406: This test relies on Python files that are getting deleted in this PR. Are you saying we should keep those files?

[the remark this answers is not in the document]
agunde406: That is what has been done in for the other tp that have been rewritten so far.

families/battleship/Dockerfile
```diff
@@ -12,38 +12,36 @@
...
+RUN curl https://sh.rustup.rs -sSf > /usr/bin/rustup-init \
+ && chmod +x /usr/bin/rustup-init \
+ && rustup-init -y
 
 WORKDIR /project/sawtooth-core
+
+CMD cd families/battleship \
+ && echo "\033[0;32m--- Building battleship ---\n\033[0m" \
+ && cargo build --release \
+ && cp ./target/release/processor /project/sawtooth-core/bin/battleship-tp \
```
boydjohnson: I think `./target/release/processor` should be `./target/release/battleship`.

knkski: I've got the binaries under `src/bin/` as `processor` and `cli`, so unless we change those, I think this needs to be `./target/release/processor`.

families/battleship/tests/test_tp_battleship.yaml
```diff
@@ -44,16 +53,18 @@ services:
...
       - 4004
-    command: nose2-3
-        -c /project/sawtooth-core/families/battleship/nose2.cfg
-        -v
-        -s /project/sawtooth-core/families/battleship/tests
-        test_tp_battleship
     stop_signal: SIGKILL
     environment:
-        TEST_BIND: "tcp://eth0:4004"
+        TEST_BIND: "tcp://0.0.0.0:4004"
```
boydjohnson: Do we know why this change is necessary? It is generally bad form to bind to all interfaces.

families/battleship/src/game.rs
```diff
@@ -0,0 +1,337 @@
...
+
+/// Get a state address for a given game of battleship
+pub fn get_battleship_address(name: &str) -> String {
+    vec![get_battleship_prefix(), get_battleship_name_hash(name)].join("")
+}
+
+/// Hash the value and nonce for a revealed space
+/// When a player joins, we're given a hashed board of their ship locations.
+/// On each player's turn, they give us a nonce and the value of that space,
+/// e.g. `'-'` and `"g54h45hwer"`. We hash those together and compare against
```
boydjohnson: Nice explanation.

families/battleship/tests/test_tp_battleship.yaml
knkski: @boydjohnson: Not sure why I had that there, removed that change

bin/run_tests
```diff
@@ -202,6 +202,7 @@ test_block_info() {
 
 test_battleship() {
   run_docker_test ./families/battleship/battleship_unit_tests.yaml
+  run_docker_test ./families/battleship/tests/test_tp_battleship.yaml
```
agunde406: What's the reasoning for removing them and then adding them back? Instead of just not removing them?

knkski: It seems weird to me to have a commit that has tests that won't run due to missing code. Given that and the fact that there were changes to the tests, it felt easier to just re-introduce them and make those changes in a separate commit.

agunde406: I agree that it is weird that the commit has the test that will fail but it also has two downsides.  It removes the history from the original tests and makes it hard to see what the changes you made to the tests are. I would suggest not removing the tests to begin with.

bin/run_tests
```diff
@@ -196,6 +196,10 @@ test_block_info() {
   copy_coverage .coverage.block_info_injector
 }
 
+test_battleship() {
```
agunde406: Please add a description to the commit to explain the changes

families/battleship/tests/battleship_board.py
```diff
@@ -0,0 +1,194 @@
+# Copyright 2016 Intel Corporation
```
agunde406: This should be bitwise 2018 if this is a new file

families/battleship/tests/battleship_message_factory.py
```diff
@@ -0,0 +1,72 @@
+# Copyright 2017 Intel Corporation
```
agunde406: This should be bitwise 2018 if this is a new file

families/battleship/tests/test_tp_battleship.py
```diff
@@ -15,21 +15,30 @@
 
 import logging
 import json
-
 from sawtooth_sdk.protobuf.validator_pb2 import Message
-from sawtooth_processor_test.transaction_processor_test_case \
```
agunde406: Since this is significant changes, you should add a bitwise copyright underneath the intel on

knkski:
@agunde406: Updated with those changes

PR #1828, opened 2018-08-22
Add block validation checks in Rust

boydjohnson:
This PR adds code to validate the batches and state produced, on chain rules, permissions, and duplicates and dependencies.

~~This PR is built on #1827~~

### DIFF
--- a/validator/src/gossip/permission_verifier.rs
+++ b/validator/src/gossip/permission_verifier.rs
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2018 Intel Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ------------------------------------------------------------------------------
+ */
+
+use cpython::{self, ObjectProtocol};
+
+use batch::Batch;
+use transaction::Transaction;
+
+pub trait PermissionVerifier {
+    fn is_batch_signer_authorized(&self, batch: &Batch, state_root: &str, from_state: bool)
+        -> bool;
+}
+
+pub struct PyPermissionVerifier {
+    verifier: cpython::PyObject,
+}
+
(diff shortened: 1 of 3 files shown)


validator/src/gossip/permission_verifier.rs
```diff
@@ -0,0 +1,59 @@
...
+        let py = gil.python();
+
+        self.verifier
+            .call_method(
+                py,
+                "is_batch_signer_authorized",
+                (batch, state_root, from_state),
+                None,
+            )
+            .expect("PermissionVerifier has not method `is_batch_signer_authorized`")
```
jsmitchell: did you mean "has no method"?

boydjohnson: Yes, thanks. Will fix

validator/src/journal/block_validator.rs
```diff
@@ -70,3 +72,118 @@ impl<'source> FromPyObject<'source> for BlockValidationResult {
         })
     }
 }
+
+trait StateBlockValidation {
```
peterschwarz: Could you add a doc comment to these traits, so that I can differentiate between the their behaviour and output?

boydjohnson: I can add a doc comment to the traits. Their main behavior is that they return `ValidationError::BlockValidationFailure` if the block fails that check. Something that I thought of is that there could be a single `BlockValidation` trait with generic `Result<R, ValidationError>` return value. What do you think of that?

boydjohnson: The generic return type turned out to be a bad idea.

validator/src/journal/block_validator.rs
```diff
@@ -70,3 +72,125 @@ impl<'source> FromPyObject<'source> for BlockValidationResult {
...
+    fn validate_block(
+        &self,
+        block: Block,
+        previous_state_root: Option<&String>,
+    ) -> Result<BlockValidationResult, ValidationError>;
+}
+
+/// A generic block validation. Returns a ValidationError::BlockValidationFailure on
+/// validation failure. Returns generic R: Sized, for a block that passes this
```
peterschwarz: Looks like you left in the experimental return type

validator/src/journal/block_validator.rs
```diff
@@ -70,3 +72,120 @@ impl<'source> FromPyObject<'source> for BlockValidationResult {
         })
     }
 }
+
+/// A generic block validation. Returns a ValidationError::BlockValidationFailure on
+/// validation failure. It is a dependent validation if it can return
+/// ValidationError::BlockStoreUpdated and is an independent validation otherwise
+trait BlockValidation {
+    type ReturnValue;
```
peterschwarz: Neat

PR #1829, opened 2018-08-23
K8s config files

rberg2:
The PR gets the kubernetes files in the repo and preserves them in the documentation build artifacts, so the links in the kubernetes documentation @chenette is working on will work.

### DIFF
--- a/docker/kubernetes/sawtooth-kubernetes-default-poet.yaml
+++ b/docker/kubernetes/sawtooth-kubernetes-default-poet.yaml
@@ -1,0 +1,588 @@
+---
+apiVersion: v1
+kind: List
+
+items:
+
+- apiVersion: extensions/v1beta1
+  kind: Deployment
+  metadata:
+    name: sawtooth-0
+  spec:
+    replicas: 1
+    template:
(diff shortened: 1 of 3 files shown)


docker/kubernetes/sawtooth-kubernetes-default-poet.yaml
```diff
@@ -0,0 +1,588 @@
...
+  spec:
+    replicas: 1
+    template:
+      metadata:
+        labels:
+          name: sawtooth-0
+      spec:
+        containers:
+          - name: sawtooth-intkey-tp-python
+            image: hyperledger/sawtooth-intkey-tp-python:1.0
```
chenette: Does this specify an image with Sawtooth 1.0.x (not master)?  The new Kubernetes procedure is going into master, so I was assuming that the Sawtooth environment would also be the master version (not 1.0.x).

rberg2: we don't publish nightly docker images from the master branch. The docker images get published during the release.

PR #1832, opened 2018-08-27
Resolve all Rust clippy lint

boydjohnson:
~~I only made changes that wouldn't change an API, which left some lint still in the package.~~


boydjohnson:
I am working right now on fixing more clippy lint in sawtooth-core that will become part of this PR.

validator/src/journal/chain_ffi.rs
```diff
@@ -520,7 +520,7 @@ impl BlockStore for PyBlockStore {
                 .and_then(|blkw| blkw.getattr(py, "block"))
                 .map_err(|py_err| {
                     pylogger::exception(py, "Unable to call block_store.get_blocks", py_err);
-                    BlockStoreError::Error(format!("Unable to get blocks"))
+                    BlockStoreError::Error("Unable to get blocks".into())
```
peterschwarz: Again, we have a third way of `&str` to `String`, here. Let's pick one.

boydjohnson: I went with .into() for changing an &str into String.

PR #1833, opened 2018-08-28
C023 fix rust versioning

rbuysse:


aludvik:
Can you provide some explanation for what this is fixing?

rbuysse:
@aludvik Every time a rust package is built, it's versioned with the version number from the Cargo.toml file. This is undesirable. This PR modifies the installed Dockerfiles for the rust components so that packages are versioned with the output of bin/get_version, matching the rest of the packages.

PR #1835, opened 2018-08-28
Complete integration of Rust BlockValidator

boydjohnson:
~~This PR is built on #1828 .~~


validator/src/journal/block_manager.rs
```diff
@@ -430,6 +430,14 @@ impl BlockManager {
         Box::new(GetBlockIterator::new(Arc::clone(&self.state), block_ids))
     }
 
+    pub fn get_block_from_blockstore(
```
peterschwarz: block seems redundent; `get_from_blockstore` would be fine

validator/src/journal/block_validator.rs
```diff
@@ -217,10 +220,14 @@ where
         transaction_index: T,
         view_factory: StateViewFactory,
     ) -> Self {
-        let (tx, rcv) = channel::<(Block, Sender<BlockValidationResult>)>();
+        let (tx1, rcv1) = channel::<(Block, Sender<BlockValidationResult>)>();
+        let (tx2, rcv2) = channel::<(Block, Sender<BlockValidationResult>)>();
         BlockValidator {
-            tx,
-            rx: Some(rcv),
+            tx1,
```
peterschwarz: Why not just a vec of tuples?

boydjohnson: Makes sense. Then there can be a configurable number of background threads.

validator/src/journal/block_manager.rs
boydjohnson: Sounds good.

PR #1850, opened 2018-09-07
Fix typos & misc errors in App Dev multi-node proc

chenette:
Correct intro description of the multi-node environment: Change "five nodes"  
to "multiple nodes" (because the Ubuntu proc creates two nodes)  

Make Settings and Validator Registry terms consistent with the rest of the doc 

Clarify that the Ubuntu multi-node procedure starts with an existing  single-node environment                                                       

In Ubuntu proc, add curly braces to "yourname" in the user key path & file name, 
to clarify it's not a literal 

Replace placeholder "xxx" text with links to the single-node K8s proc 

Fix format problems, a bad step number (K8s proc), and several typos

### DIFF
--- a/docs/source/app_developers_guide/creating_sawtooth_network.rst
+++ b/docs/source/app_developers_guide/creating_sawtooth_network.rst
@@ -99,7 +99,7 @@ processing and static peering. However, it has the following differences:
   Software Guard Extensions (SGX) to implement a leader-election lottery system.
   PoET simulator provides the same consensus algorithm on an SGX simulator.
 
-* An additional transaction processor, Validator Registry, which handles PoET
+* An additional transaction processor, PoET Validator Registry, handles PoET
   settings for a multiple-node network.
 
 Prerequisites
@@ -371,7 +371,7 @@ processing and static peering. However, it has the following differences:
   Software Guard Extensions (SGX) to implement a leader-election lottery system.
   PoET simulator provides the same consensus algorithm on an SGX simulator.
 
-* An additional transaction processor, Validator Registry, which handles PoET
+* An additional transaction processor, PoET Validator Registry, handles PoET
   settings for a multiple-node network.
 
 .. _prereqs-multi-ubuntu-label:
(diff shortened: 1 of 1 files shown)


docs/source/app_developers_guide/creating_sawtooth_network.rst
```diff
@@ -78,7 +78,7 @@ Docker: Start a Multiple-node Sawtooth Network
 
 In this procedure, you will use a Docker Compose file that creates a new
 application development environment with five validator nodes and four
-transaction processors (Settings, IntegerKey, XO, and PoET Validator Registry).
+transaction processors (Settings, IntegerKey, XO, and Validator Registry).
```
aludvik: I do not think this is a good change to make, as it will be unclear to those with a strong knowledge of Sawtooth that this "Validator Registry" is a PoET specific transaction processor. If consistency is needed, then we should make the change in the spec.

docs/source/app_developers_guide/creating_sawtooth_network.rst
```diff
@@ -538,7 +538,9 @@ in :doc:`ubuntu`.
 
    .. note::
 
-      This network requires ``settings-tp`` and ``poet-validator-registry-tp``.
+      This network requires the Settings transaction processor, ``settings-tp``,
+      and the Validator Registry transaction processor,
```
aludvik: Suggest also using "PoET Validator Registry" here.

docs/source/app_developers_guide/creating_sawtooth_network.rst
chenette: I could restore "PoET" here, but then it will be inconsistent with the rest of the Sawtooth docs. The transaction family spec calls it "Validator Registry", as does the rest of the doc set (mostly).  FYI: This was the original name of the V.R. txn family spec in the doc - this chapter never had "PoET" in the title.  For examples, see: 
https://sawtooth.hyperledger.org/docs/core/nightly/master/transaction_family_specifications.html
https://sawtooth.hyperledger.org/docs/core/nightly/master/transaction_family_specifications/validator_registry_transaction_family.html

If "PoET" is needed in the name here, it is also needed everywhere else. If so, I suggest a separate PR to make the term consistent everywhere. Note that this affects the images in the app dev guide as well as the Txn Fam spec.  I think some CLI "code" (help output) and config files will also need to change.  I suspect that the doc in the separate PoET repo might need changing, too.

chenette:
@aludvik - I pushed a fixup commit to restore "PoET" to the Validator Registry terms where I had removed it. I also added PoET to the terms that were formerly PoET-less. 

These terms are now consistent within this section.  Consistency for the rest of the doc set will require a separate PR.

PR #1851, opened 2018-09-07
Undo include - move content from separate file into topic

chenette:
Using the include directive to add contents from a separate file
causes a problem with the search index.

Searching for any content in this file will find a topic called"no title" (in brackets).
When viewed, this untitled topic isn't "in" the doc nav - it doesn't have
any Previous and Next buttons on the page.

This PR simply moves the relevant content into the topic so that
Sphinx treats the content correctly for a search.


chenette:
For an example of the problem, see:                                             
                                                                                
https://sawtooth.hyperledger.org/docs/core/nightly/master/search.html?q=registry&check_keywords=yes&area=default
                                                                                
https://sawtooth.hyperledger.org/docs/core/nightly/master/transaction_family_specifications/example_transaction_families.html?highlight=registry

docs/source/transaction_family_specifications.rst
```diff
@@ -2,7 +2,69 @@
...
+  is an extensible role- and policy-based system for defining permissions in a
+  way that can be used by other Sawtooth components.
+  Identity is written in Python.
+  The family name is ``sawtooth_identity``;
+  the associated transaction processor is ``identity-tp`` (see
+  :doc:`/cli/identity-tp`).
+
+* The :doc:`/transaction_family_specifications/integerkey_transaction_family`
+  simply sets, increments, and decrements the value of entries stored in a state
+  dictionary. IntegerKey is available in Go, Java, and JavaScript (Node.js).
```
aludvik: Does it make sense to also update this list to include Python and Rust at this time?

docs/source/transaction_family_specifications.rst
```diff
@@ -2,7 +2,69 @@
...
+  The :doc:`sawset command </cli/sawset>` provides an example CLI client.
+
+  .. note::
+
+    In a production environment, you should always run a transaction processor
+    that supports the Settings transaction family.
+
+* The :doc:`/transaction_family_specifications/smallbank_transaction_family`
+  provides a cross-platform workload for comparing the performance of
+  blockchain systems. Smallbank is written in Go.
```
aludvik: Same with this, it is also written in Rust now.

docs/source/transaction_family_specifications.rst
chenette: I would rather handle changes to the "available languages" list in a separate PR, so that this PR is focused on fixing the search problem.

docs/source/transaction_family_specifications.rst
chenette: Perhaps that separate PR should remove the list(s) of specific languages from this topic so that it doesn't get out of date.  In any case, I'd prefer to do this type of change in a separate PR.

PR #1852, opened 2018-09-10
Bumper updates

rberg2:


ci/sawtooth-rest-api
```diff
@@ -35,6 +35,7 @@ RUN echo "deb [arch=amd64] http://repo.sawtooth.me/ubuntu/bumper/stable xenial u
  && apt-get update \
  && apt-get install -y -q \
     python3-sawtooth-rest-api \
+    python3-sawtooth-sdk \
```
aludvik: Why is this needed? If the rest api depends on the sdk and is not being installed, that is a sign that something about the dependency resolution for the rest api package is broken and should be fixed there.

dcmiddle:
@vaporos ok with you? Think I need your thumbs up to merge to 1.1 branch.

PR #1853, opened 2018-09-11
Modify Duplicates and Dependencies check to decrease time that persist lock must be held

boydjohnson:
This PR does 2 main things
- uses a lock to guarantee that the block store stays stable during duplicates and dependencies check that relies on the block store
- simplifies the code that was in ChainCommitState, bringing in some of that functionality to the block manager - in particular `contains_any_transactions(branch_head_id, transaction_ids)` and the corresponding batch id function.

### DIFF
--- a/validator/src/journal/block_manager.rs
+++ b/validator/src/journal/block_manager.rs
@@ -17,8 +17,8 @@
 
 #![allow(unknown_lints)]
 
-use std::collections::HashMap;
-use std::iter::Peekable;
+use std::collections::{HashMap, HashSet};
+use std::iter::{FromIterator, Peekable};
 use std::sync::{Arc, RwLock};
 
 use block::Block;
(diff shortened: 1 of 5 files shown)


validator/src/journal/block_manager.rs
```diff
@@ -398,13 +402,158 @@ impl BlockManagerState {
...
+
+    pub fn contains_transaction(
+        &self,
+        block_id: &str,
+        id: &str,
+    ) -> Result<bool, BlockManagerError> {
+        let _lock = self
+            .persist_lock
+            .read()
+            .expect("The persist RwLock is poisoned");
```
aludvik: Can we just acquire a read lock on block_by_block_id?

validator/src/journal/block_manager.rs
```diff
@@ -398,13 +402,158 @@ impl BlockManagerState {
...
+    fn transaction_index_contains(
+        &self,
+        index: &Box<IndexedBlockStore>,
+        block_id: &str,
+        id: &str,
+    ) -> Result<bool, BlockManagerError> {
+        let index = index.as_ref();
+        if TransactionIndex::contains(index, id)? {
+            let block = TransactionIndex::get_block_by_id(index, id)?
+                .expect("BlockStore updated during transaction index check");
```
aludvik: These expects are a concern, as they imply that it is possible for the store to be updated concurrently in-between calls. You can eliminate this one by eliminating the call to contains() and just using get_block_by_id() directly. The second one requires holding the lock on `block_by_block_id` during the duration of this call to ensure it, and the underlying blockstores, are not modified while looking up the transaction.

validator/src/journal/block_manager.rs
```diff
@@ -398,13 +402,158 @@ impl BlockManagerState {
...
+            .expect("The persist RwLock is poisoned");
+        if let Some((_, store)) = self
+            .state
+            .blockstore_by_name
+            .read()
+            .expect("Blockstore RW Lock is poisoned")
+            .iter()
+            .find(|(_, store)| store.get(&[block_id]).map(|res| res.count()).unwrap_or(0) > 0)
+        {
+            self.transaction_index_contains(store, block_id, id)
```
aludvik: I don't think there is any guarantee here that block_id won't be dropped from the store in between when the above get is called and this line executes. I think you need to hold the lock around `block_by_block_id` for the duration of this function.

validator/src/journal/block_manager.rs
```diff
@@ -398,13 +402,158 @@ impl BlockManagerState {
...
 }
 
 impl BlockManager {
     pub fn new() -> Self {
         BlockManager::default()
     }
 
+    fn transaction_index_contains(
+        &self,
+        index: &Box<IndexedBlockStore>,
```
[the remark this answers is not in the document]
boydjohnson: I changed it, because clippy told me to, but squashed it into the wrong commit.

validator/src/journal/block_manager.rs
```diff
@@ -398,13 +402,158 @@ impl BlockManagerState {
...
+        block_id: &str,
+        id: &str,
+    ) -> Result<bool, BlockManagerError> {
+        let index = index.as_ref();
+        if TransactionIndex::contains(index, id)? {
```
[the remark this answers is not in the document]
aludvik: I think if you make this function take `&TransactionIndex` that problem goes away.

validator/src/journal/block_manager.rs
boydjohnson: Yeah, this could be clearer. `if let Some(block) = TransactionIndex::get_block_by_id(index, id)` is how I should go.

validator/src/journal/block_manager.rs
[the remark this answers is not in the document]
boydjohnson: Yeah, I think that is right. It could be &TransactionIndex and &BatchIndex for the signature. I would like to make this function generic on which index is being used. That would be a win in reducing code.

validator/src/journal/block_manager.rs
boydjohnson: There is a deadlock that would happen. During splitting up the locks task, I made sure to have an precedence of locks, so that if an overlap of scope happened as happens in `put` and `unref`, `references_by_block_id` is locked first and then `block_by_block_id`, so there is always the same ordering of locks. In the `contains_*` functions, there is a call to `.branch` which calls `ref` which locks the references_by_block_id`, so we would have a `block_by_block_id` then `references_by_block_id` locking order, and a recipe for a deadlock.

aludvik: I think using a separate empty lock is actually more confusing than acquiring the lock around block_by_block_id. Acquiring the lock around block_by_block_id is a hard guarantee that nothing is going to be modifying which blocks are stored, whereas acquiring this empty lock is a soft guarantee that nothing is going to be modifying which blocks are being persisted, provided we always acquire the persist lock before persisting. Using an empty lock side steps the way Rust locks are designed to work. We are able to get away with it here, because we have the trait bound on BlockStore that it is thread-safe already. It might be cleaner to wrap BlockStore in an RwLock and remove the Sync trait bound, which I think would be the equivalent of this that doesn't require using an empty lock.

validator/src/journal/block_manager.rs
```diff
@@ -398,13 +402,200 @@ impl BlockManagerState {
...
+                .fold(vec![], |mut arr, b| {
+                    for transaction in &b.transactions {
+                        arr.push(&transaction.header_signature)
+                    }
+                    arr
+                })
+                .into_iter(),
+        );
+        let compare_ids: HashSet<&String> = HashSet::from_iter(ids.iter().cloned());
+        !block_ids.is_disjoint(&compare_ids)
```
aludvik: Losing the ability to tell which batch or transaction was the duplicate seems like a significant loss of insight. Can we use a different HashSet method that retains this information? For example, we could do an intersection and check whether the first returned value from the iterator is None: https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.intersection

validator/src/journal/block_manager.rs
```diff
@@ -398,13 +402,200 @@ impl BlockManagerState {
...
+                        return Ok(true);
+                    }
+                }
+            }
+            Ok(false)
+        }
+    }
+
+    fn block_contains_any_transactions(&self, block: &Block, ids: &[&String]) -> bool {
+        let block_ids: HashSet<&String> = HashSet::from_iter(
```
aludvik: Is transaction_ids a better name for this?

validator/src/journal/block_manager.rs
```diff
@@ -398,13 +402,200 @@ impl BlockManagerState {
...
+                    arr
+                })
+                .into_iter(),
+        );
+        let compare_ids: HashSet<&String> = HashSet::from_iter(ids.iter().cloned());
+        !block_ids.is_disjoint(&compare_ids)
+    }
+
+    fn block_contains_any_batches(&self, block: &Block, ids: &[&String]) -> bool {
+        let block_ids = HashSet::from_iter(block.batches.iter().map(|b| &b.header_signature));
```
aludvik: Is batch_ids a better name for this?

validator/src/journal/block_manager.rs
aludvik: After spending some time thinking about it, I am okay with using persist_lock here. I think there are probably better solutions, but I can't think of what they are and this should work.

validator/src/journal/block_manager.rs
boydjohnson: Yeah, I think this is a good idea. We could also produce all the duplicated ids using intersection.

validator/src/journal/block_manager.rs
boydjohnson: Yes, thanks!.

validator/src/journal/block_manager.rs
boydjohnson: Yes, Thanks!

boydjohnson:
I conducted an 8 hour test where 2 validators were producing duplicate batches in their blocks. This .md file explains the results. No duplicate batches were included in the chain, but 2 blocks that did not have duplicates were found from these validators.


[DuplicatesValidation.txt](https://github.com/hyperledger/sawtooth-core/files/2381308/DuplicatesValidation.txt)

validator/src/journal/block_manager.rs
```diff
@@ -398,13 +402,152 @@ impl BlockManagerState {
...
+        index: &IndexedBlockStore,
+        block_id: &str,
+        id: &str,
+    ) -> Result<bool, BlockManagerError> {
+        if let Some(block) = TransactionIndex::get_block_by_id(index, id)? {
+            let head = index
+                .get(&[block_id])?
+                .next()
+                .expect("BlockStore updated during transaction index check");
+            if block.block_num <= head.block_num {
```
peterschwarz: Why not just `Ok(block.block_num <= head.block_num)`

validator/src/journal/block_manager.rs
```diff
@@ -398,13 +402,152 @@ impl BlockManagerState {
...
+        index: &IndexedBlockStore,
+        block_id: &str,
+        id: &str,
+    ) -> Result<bool, BlockManagerError> {
+        if let Some(block) = BatchIndex::get_block_by_id(index, id)? {
+            let head = index
+                .get(&[block_id])?
+                .next()
+                .expect("BlockStore updated during batch index check");
+            if block.block_num <= head.block_num {
```
peterschwarz: Same here

validator/src/journal/block_manager.rs
```diff
@@ -398,13 +402,152 @@ impl BlockManagerState {
...
+        &self,
+        index: &IndexedBlockStore,
+        block_id: &str,
+        id: &str,
+    ) -> Result<bool, BlockManagerError> {
+        if let Some(block) = TransactionIndex::get_block_by_id(index, id)? {
+            let head = index
+                .get(&[block_id])?
+                .next()
+                .expect("BlockStore updated during transaction index check");
```
peterschwarz: Why is it ok to Panic here, instead of propagating an error?

validator/src/journal/chain_commit_state.rs
```diff
@@ -53,171 +43,78 @@ fn check_no_duplicates(ids: &[String]) -> Option<String> {
...
 
-    pub fn validate_no_duplicate_batches(
-        &self,
-        batch_ids: Vec<String>,
-    ) -> Result<(), ChainCommitStateError> {
-        if let Some(batch_id) = check_no_duplicates(batch_ids.as_slice()) {
+    for batch_id in batch_ids {
+        if block_manager
+            .contains_batch(branch_head_id, &batch_id)
+            .map_err(|err| ChainCommitStateError::Error(format!("{:?}", err)))?
```
peterschwarz: Is there not a `From<_>` implementation for this error type?  Since you're using `map_err` and a `?`and doing the same thing later, it might make sense, .

validator/src/journal/block_manager.rs
boydjohnson: Yeah, that seems good.

validator/src/journal/block_manager.rs
boydjohnson: So, this had originally been part of a larger function that I broke apart. It isn't right to panic here, it is right to panic in contains_any_transaction because the persist_lock is there and the persist_lock guarantees that there was not an update to the block store.

boydjohnson: Actually looking at it a second time, I think it is right to panic here as the only thing to be done by the caller would be to panic. I am going to leave a note on the docstring for the function what its requirements are.

PR #1854, opened 2018-09-12
Master bumper updates

rberg2:
Forward port of changes made to the 1.1 branch.

### DIFF
--- a/ci/sawtooth-block-info-tp
+++ b/ci/sawtooth-block-info-tp
@@ -29,7 +29,7 @@ FROM ubuntu:xenial
 
 LABEL "install-type"="repo"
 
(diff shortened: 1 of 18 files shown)


rest_api/stdeb.cfg
```diff
@@ -0,0 +1,2 @@
+[DEFAULT]
+Depends3: python3-sawtooth-sdk
```
aludvik: Why do we suddenly need to add this file, if the SDK is already listed in setup.py?

rberg2: `python3-sawtooth-rest-api` is no longer getting added as a dependency in the deb package. This file seems to be the standard way to add dependencies to the deb package when they are not added automatically.

PR #1857, opened 2018-09-14
Fix block manager restarts

aludvik:


aludvik:
Rebased on master, again

validator/src/journal/block_manager.rs
```diff
@@ -269,16 +299,41 @@ impl BlockManagerState {
...
-        Ok(())
+
+        let blockstore_by_name = self
+            .blockstore_by_name
+            .read()
+            .expect("Acquiring blockstore by name read lock; lock poisoned");
+
+        let mut block = None;
+
+        for (_, store) in blockstore_by_name.iter() {
```
boydjohnson: I would like to see this code-block as finding the store (if it exists) that contains the block. So `blockstore_by_name.iter().find(|(_, store)| store.get(&[block_id]).next().is_some())` or something like that. And then you can do direct expression assignment to block instead of having `mut block`, iteration, and break.

validator/src/journal/block_manager.rs
```diff
@@ -388,10 +443,29 @@ impl BlockManagerState {
...
+
+        let mut stores = self
+            .blockstore_by_name
             .write()
-            .expect("Acquiring blockstore write lock; lock poisoned")
-            .insert(store_name.into(), store);
+            .expect("Acquiring blockstore write lock; lock poisoned");
+
+        if let Some(head) = store.iter().expect("Failed to get store iterator").next() {
+            if !references_by_block_id.contains_key(&head.header_signature) {
```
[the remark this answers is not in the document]
aludvik: I was under the impression that next() just gets the first block from the blockstore. Is that incorrect?

validator/src/journal/block_manager.rs
aludvik: That requires getting the block from the store twice, which is why I chose not to do it.

validator/src/journal/block_manager.rs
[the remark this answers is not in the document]
boydjohnson: No you are right. I saw the iter and didn't look too closely.

validator/src/journal/block_manager.rs
boydjohnson: `filter_map(|(_, store)| store.get(&[block_id]).next()).next()` may be what is needed.

aludvik: Nice, I think that works

aludvik:
@boydjohnson updated

PR #1858, opened 2018-09-15
Implement the ZMQ_PROBE_ROUTER flag response.

CarvalhoLeonardo:
Code to answer empty ZeroMQ messages and honor the probing on the listening external socket.

Implement #STL-1446

### DIFF
--- a/validator/sawtooth_validator/networking/interconnect.py
+++ b/validator/sawtooth_validator/networking/interconnect.py
@@ -345,9 +345,15 @@ class _SendReceive:
                 if self._socket.getsockopt(zmq.TYPE) == zmq.ROUTER:
                     zmq_identity, msg_bytes = \
                         yield from self._socket.recv_multipart()
-                    self._received_from_identity(zmq_identity)
-                    self._dispatcher_queue.put_nowait(
-                        (zmq_identity, msg_bytes))
+                    if msg_bytes == b'':
+                        # send ACK for connection probes
+                        LOGGER.debug("ROUTER PROBE FROM %s", zmq_identity)
+                        self._socket.send_multipart(
+                            [bytes(zmq_identity), msg_bytes])
+                    else:
+                        self._received_from_identity(zmq_identity)
+                        self._dispatcher_queue.put_nowait(
+                            (zmq_identity, msg_bytes))
                 else:
                     msg_bytes = yield from self._socket.recv()
                     self._last_message_time = time.time()


aludvik:
Hello and thank you for the PR! Can you say a little more here and/or in your commit message about why we want to implement support for responding to router probes and more generally the patterns that require it?

peterschwarz:
You also need to sign your commit.  Do a `git commit -s --amend`, update the commit message with more information, per @aludvik's request, and re-push your branch.

aludvik:
@CarvalhoLeonardo The DCO check requires that the commit message have "Signed-off-by: Full Name <email@address.com>" as the last line. It is separate/in addition to from the GPG signing.

Can you please also post a technical explanation of why we want to support TPs that connect to the validator with a ROUTER socket? I have seen some discussion in chat, but reproducing it here would be useful for reviewers. Specifically, why is the ROUTER-ROUTER pattern required and why is the ROUTER-DEALER model insufficient? With the Go SDK, I also used IPC to fan out within the TP, but I did (left to right from validator to worker thread) `ROUTER <-(tcp)-> DEALER <-(poll and switch) -> ROUTER <-(ipc)-> DEALER`. This is a good starting point to see what I mean in detail: https://github.com/hyperledger/sawtooth-sdk-go/blob/master/processor/processor.go#L162

CarvalhoLeonardo:
I've re-sent the commit, sorry about that.

So, my rationales, besides implementing a very useful flag on the protocol, are:
 - Turn easier to implement symmetrical topologies. The Java SDK has classes that can be used to implement clients or a validator without changing anything;
- Using ROUTER <-> ROUTER, the topology completely decouple the messages for any particular worker. I don't know Go, but seeing the code, there is a map to hold references to the worker and a correlation ID being worked on. With the internal fanning proposed in the Java SDK, that isn't necessary at all. That way, long lived actions wouldn't stop a thread, giving the code a real async processing. This [class](https://github.com/CarvalhoLeonardo/sawtooth-sdk-java/blob/aab00d6bf040db695ddd152cc1e6a2ddf66ea380/sdk-transaction-processor/src/main/java/sawtooth/sdk/reactive/tp/messaging/ReceivingHandler.java#L72) holds the interesting bit.
- This approach was taken from [Freelancer](https://rfc.zeromq.org/spec:10/FLP/) and [Majordomo]((https://rfc.zeromq.org/spec:7/MDP/) protocols, that provides full async work, that I think is of interest to the project as a whole.

Hope this helps!

validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -345,9 +345,14 @@ def _receive_message(self):
                 if self._socket.getsockopt(zmq.TYPE) == zmq.ROUTER:
                     zmq_identity, msg_bytes = \
                         yield from self._socket.recv_multipart()
-                    self._received_from_identity(zmq_identity)
-                    self._dispatcher_queue.put_nowait(
-                        (zmq_identity, msg_bytes))
+                    if msg_bytes == b'':
+			# send ACK for connection probes
```
dcmiddle: indentation alignment

validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -345,9 +345,14 @@ def _receive_message(self):
...
+                        LOGGER.debug("ROUTER PROBE FROM %s",zmq_identity)
```
dcmiddle: How often will these logging messages be generated?

CarvalhoLeonardo: Every probe request, i.e., once per TP connection.

validator/sawtooth_validator/networking/interconnect.py
CarvalhoLeonardo: Strange, locally it looks ok...

image

vaporos: The line is using tabs instead of spaces.

dcmiddle:
@CarvalhoLeonardo Looks like DCO is failing. Maybe because of that last merge commit? You could rebase that out.
e.g.
git fetch upstream master #or whatever you call this remote to github/hyperledger
git rebase -i upstream/master
git push -f <yourfork> <thisbranch>

CarvalhoLeonardo:
> git fetch upstream master

Hm, now I understood why there weren't any Rust code...

Thanks.

PR #1859, opened 2018-09-17
Add queue to consensus notifier

ltseeley:
Adds a threadsafe queue to the ConsensusNotifier class in the validator
to ensure that all messages intended for the consensus engine get sent.
Previously, messages to be sent to the engine before it was registered
were simply dropped, which would result in missed BlockNew and
BlockValid notifications when a new node started up on an existing
network.


boydjohnson:
Closing in favor of #1864

PR #1864, opened 2018-09-19
 Send missed BlockNew msgs to consensus on startup

aludvik:


aludvik:
The FFI for this is actually broken and needs to be fixed.

aludvik:
New PR #1866

PR #1865, opened 2018-09-19
Update sgx guide for bumper

rberg2:
* add steps to install and start / restart poet-engine
 * add validator configuration for consensus port
 * change configuration file code blocks to ini to prevent odd syntax
highlighting


docs/source/sysadmin_guide/configure_sgx.rst
```diff
@@ -139,6 +139,7 @@ Install Sawtooth
     $ sudo apt-get update
     $ sudo apt-get install -y -q \
```
chenette: Let's remove -q so that this command is consistent with the non-SGX install command.

PR #1866, opened 2018-09-19
 Send missed BlockNew msgs to consensus on startup 

aludvik:
This PR fixes a bug with the validator where it fails to send BlockNew updates for all blocks since the chain head when the engine connects, if the engine connects after the validator has received some blocks.


validator/sawtooth_validator/consensus/handlers.py
```diff
@@ -110,7 +110,7 @@ def handle_request(self, request, response):
 
         if startup_info is None:
             response.status = consensus_pb2.ConsensusRegisterResponse.NOT_READY
-            return
+            return None
```
peterschwarz: Is `None` the correct thing to return here?  Shouldn't it be a `HandlerStatus`?

validator/sawtooth_validator/consensus/handlers.py
```diff
@@ -144,6 +144,29 @@ def handle_request(self, request, response):
...
+
+    @property
+    def request_type(self):
+        return validator_pb2.Message.CONSENSUS_REGISTER_REQUEST
+
+    def handle(self, connection_id, message_content):
+        forks = self._proxy.forks()
+
+        if not forks:
+            return None
```
peterschwarz: What will happen on the consensus side when it's waiting on the future to resolve for this message.  Does it get a response?

validator/sawtooth_validator/consensus/handlers.py
```diff
@@ -85,10 +83,13 @@ def handle(self, connection_id, message_content):
         except DecodeError:
             response.status = response.BAD_REQUEST
         else:
-            self.handle_request(request, response)
+            handler_status = self.handle_request(request, response)
+
+        if handler_status is None:
```
peterschwarz: Why not just return this from the `handle_request` implementations?  The the type is always consistently a `HandlerStatus`.  Returning `None` looks confusing, and adds the complexity here.

validator/src/journal/chain.rs
```diff
@@ -531,6 +531,63 @@ impl<
...
+            .state
+            .read()
+            .expect("No lock holder should have poisoned the lock");
+
+        match state.block_manager.ref_block(head) {
+            Err(BlockManagerError::UnknownBlock) => {
+                return None;
+            }
+            Err(err) => {
+                error!("Unexpected error occurred: {:?}", err);
```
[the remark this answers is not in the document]
aludvik: UnknownBlock is the only error variant that can be returned from ref_block(), but since we have not properly defined which methods return which errors, this is the only thing that makes sense to do.

For the BlockManager, we have one error type `BlockManagerError` for all methods. IMO, a better pattern is to define a new error type for each method so that you can be more specific about what the errors are for each method. As an example `ref_block()` would have:

```
pub enum BlockManagerRefError {
  UnknownBlock,
}
```

validator/src/journal/chain.rs
```diff
@@ -531,6 +531,63 @@ impl<
...
+            .into_iter()
+            .flat_map(|fork_head: &String| {
+                state
+                    .block_manager
+                    .branch_diff(fork_head, head)
+                    .expect("Fork not found, but should be referenced")
+                    .map(|block| Some(block))
+            }).collect();
+
+        for i in 0..forks.len() {
```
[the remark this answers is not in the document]
aludvik: It's removing any duplicates from the list. I'll move it behind a function. I can't use the `dedup()` method because I need to preserve the original order.

validator/sawtooth_validator/consensus/handlers.py
aludvik: I can make that change.

validator/src/journal/chain.rs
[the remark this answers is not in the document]
peterschwarz: Why does order matter, after you've flattened them?

validator/sawtooth_validator/consensus/handlers.py
peterschwarz: Shouldn't this return a `HandlerStatus`?

validator/src/journal/chain.rs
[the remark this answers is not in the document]
aludvik: The contract between the validator and consensus engine is that the validator will send new blocks to the engine consecutively and increasing for the block height, on a given fork. I had intended this function to return blocks in an order that was consistent with that contract, but now that you are questioning it, I see that I actually got the order reversed. Thinking about it more though, if I sort the list in increasing order of block height, I think it still satisfies that contract and I can just use dedup then.

validator/sawtooth_validator/consensus/handlers.py
aludvik: It should return HandlerResult

aludvik:
This change only affects the validator's behavior when the engine restarts. When this happens, the validator will send updates for all blocks it has in memory that are not on the current chain. It sends the block in the order that the engine expects to receive them. From the engines perspective, these blocks will appear to have just arrived at the validator. You are correct that this is an unbounded number of blocks.

If the validator restarts, all the in memory blocks will be lost and so there will be no blocks to send to the engine (other than chain head and any new blocks that arrive).

PR #1867, opened 2018-09-20
Resolve formatting errors due to rustfmt update

boydjohnson:


boydjohnson:
https://github.com/hyperledger/sawtooth-core/blob/1-1/docker/lint This is the dockerfile that controls the linting dependencies. I think the toolchain could get pegged to solve this problem. @dcmiddle , is that your preference?

dcmiddle:
Since there's a memory leak fix dependent on newer rust, I'm onboard with merging this set of linting changes. In a subsequent 1-1 PR let's consider pegging our toolchain for this release. I believe that's what we did with 1.0 to avoid spurious build failures as tooling updates.

PR #1868, opened 2018-09-20
Update non-SGX install/config/run proc in Sys Admin Guide

chenette:
Update and correct existing procedures; add missing steps and procedures; clarify and expand the explanatory text.  

- Group the non-SGX procedures in a new section so that you can tell what's non-SGX vs SGX
- Clarify and format the install procedure
- Expand the "Starting Sawtooth" procedure to show the actual service commands and add the PoET simulator service
- Move the key-generation steps into their own procedure ; add more info to clarify user vs validator keys and key file location
- Create a new genesis block procedure with the correct steps for a system with PoET simulator consensus
- Reorganize the non-SGX sections to be in the correct order

Also:
- Add an overview/intro to the Sys Admin Guide
- Reorganize other sections: Move SGX section after the "generic" sections (for non-SGX and SGX systems), put consensus section at the end because it's reference doc

### DIFF
--- a/docs/source/sysadmin_guide/creating_genesis_block.rst
+++ b/docs/source/sysadmin_guide/creating_genesis_block.rst
@@ -1,0 +1,145 @@
+**************************
+Creating the Genesis Block
+**************************
+
+.. important::
+
+   Use this procedure **only** for the first validator node on a Sawtooth
+   network. Skip this procedure if you are creating a node that will join an
+   existing Sawtooth network.
+
+The first node in a new Sawtooth network must create the genesis block to
+initialize the Settings transaction processor and specify the consensus
+mechanism. The settings in the genesis block enable other nodes to join the
+network and use these on-chain settings.
+
+Before you start the first Sawtooth node, use this procedure to create and
+submit the genesis block.
+
+.. note::
+
+   This procedure uses the same validator key for all commands that require a
+   key. In theory, some of these commands could use a different key, but
+   configuring multiple keys is a complicated process. Describing how to use
+   multiple keys to create the genesis block is outside the scope of this guide.
+
+1. Ensure that the required user and validator keys exist:
+
+   .. code-block:: console
+
+      $ ls ~/.sawtooth/keys/
+      {yourname}.priv    {yourname}.pub
+
+      $ ls /etc/sawtooth/keys/
+      validator.priv   validator.pub
+
+   If these key files do not exist, create them as described in
+   :doc:`generating_keys`.
+
+#. Become the ``sawtooth`` user. In the following commands, the prompt
+   ``[sawtooth@system]`` shows the commands that must be executed as
+   ``sawtooth``.
+
+   .. code-block:: console
+
+      $ sudo -u sawtooth -s
+      [sawtooth@system]$ cd /tmp
+
+#. Create a batch to initialize the Settings transaction family in the genesis
+   block.
+
+   .. code-block:: console
+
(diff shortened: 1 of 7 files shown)


docs/source/sysadmin_guide/systemd.rst
```diff
@@ -2,87 +2,110 @@
 Running Sawtooth as a Service
 *****************************
 
-When installing Sawtooth using apt-get, *systemd* units are added for the
-following components. These can then be started, stopped, and restarted using
-the *systemctl* command:
+When you installed Sawtooth with ``apt-get``, ``systemd`` units were added for
+the Sawtooth components (validator, REST API, transaction processors, and
+consensus engines). This procedure describes how to use the ``systemctl``
+command to start, stop, and restart Sawtooth  components as ``systemd``
```
aludvik: extra space between Sawtooth and components

docs/source/sysadmin_guide/creating_genesis_block.rst
```diff
@@ -0,0 +1,131 @@
...
+      sawtooth.consensus.algorithm=poet \
+      sawtooth.poet.report_public_key_pem="$(cat /etc/sawtooth/simulator_rk_pub.pem)" \
+      sawtooth.poet.valid_enclave_measurements=$(poet enclave measurement) \
+      sawtooth.poet.valid_enclave_basenames=$(poet enclave basename)
+
+   This is a complicated command. Here's an explanation of the options:
+
+   ``--key /etc/sawtooth/keys/validator.priv``
+         Signs the proposal with this node's validator key. By default, only the
+         ``sawtooth`` user on this node can change on-chain settings. A later
```
aludvik: "By default, only the ``sawtooth`` user on this node can change on-chain settings."

I do not think this is correct. Permissions to change settings are controlled by whoever owns the key. It would be more accurate to say "Only the key passed to `sawset genesis` may be used to change on-chain settings".

docs/source/sysadmin_guide/creating_genesis_block.rst
```diff
@@ -0,0 +1,131 @@
...
+         Adds a simulated enclave basename to the blockchain. The PoET
+         Validator Registry uses this value to check signup information.
+
+#. Create a batch to register the first Sawtooth node with the PoET Validator
+   Registry transaction processor. Without this command, the validator would not
+   be able to publish any blocks.
+
+   .. code-block:: console
+
+      [sawtooth@system]$ poet registration create --key /etc/sawtooth/keys/validator.priv -o poet.batch
```
aludvik: It is important that the key used here is the same key that the genesis validator uses. Might be worth noting.

chenette: I'm concerned that putting this note here might add confusion (because it doesn't explain which keys don't HAVE to use the same genesis-node validator key). I talked to @aludvik and we agreed to put a note at the beginning of this procedure instead. My proposed wording is:

    "This procedure uses the same validator key for all commands that require a   
     key. In theory, some of these commands could use a different key, but        
     configuring multiple keys is more complicated. Describing how to use         
     multiple keys is outside the scope of this guide."

chenette:
Reviewers: I pushed several fixup commits that make the changes from @aludvik and fix a few other minor issues that I noticed.  The corrections are in these commits: 

- Expand + procedurize Start Sawtooth Services topic  …  343d023                                                                            
- Add new Create Genesis Block procedure for non-SGX process  …  4394853
- Add parent proc for Sawtooth with PoET simulator  …  6c96bd1

PR #1872, opened 2018-09-24
 Use openssl bindings for pem functionality in SDK 

knkski:
Moves to `openssl` bindings to handle de/serialization of private keys to/from the PEM format, and removes old PEM loading convenience function written in C.

As per conversation with @aludvik, it's preferable to have each language use their respective openssl wrappers instead of replacing the old code with e.g. a C/Rust convenience wrapper that exports PEM to/from functionality.


aludvik:
I think you can also remove the `cc` build dependency from Cargo.toml

PR #1874, opened 2018-09-25
Add new proc to set allowed txn types (Sys Admin Guide)

chenette:
Reviewers: Please review only the file setting_allowed_txns.rst.  (The first 7 commits are in PR #1868.)  See this commit: https://github.com/hyperledger/sawtooth-core/pull/1874/commits/e28bd1a209ff673ba5502ccdb1e0440c2a51234d

This new procedure is based on existing steps in the App Dev Guide - see the Ubuntu section in "Creating a Sawtooth Network". 
https://sawtooth.hyperledger.org/docs/core/nightly/master/app_developers_guide/creating_sawtooth_network.html#configure-txn-procs-ubuntu-label

### DIFF
--- a/docs/source/sysadmin_guide/setting_allowed_txns.rst
+++ b/docs/source/sysadmin_guide/setting_allowed_txns.rst
@@ -1,0 +1,80 @@
+************************************************
+Setting the Allowed Transaction Types (Optional)
+************************************************
+
+By default, a validator accepts transactions from any transaction processor.
+However, Sawtooth allows you to limit the types of transactions that can be
+submitted.
+
+This setting, ``sawtooth.validator.transaction_families``, improves the
+Sawtooth network's security by ignoring any unrecognized transaction processors.
+It is an on-chain setting, which means that the change is submitted on one node;
+the other nodes in the network apply the settings change when they receive the
+block with this transaction. Note that the
+:doc:`Settings transaction processor <../transaction_family_specifications/settings_transaction_family>`
+is required to handle on-chain configuration settings.
+
+In this procedure, you will configure the validator network to limit the
+accepted transaction types to those from the Identity, IntegerKey, Settings, and
+PoET Validator Registry transaction processors.
+
+.. important::
+
+   For the environment described in this guide, you  **must** run this procedure
+   on the same node that created the genesis block, because the ``sawset
+   proposal create`` command requires the validator key that was generated on
+   that node.
+
+#. Open a terminal window on the "genesis node" (the Sawtooth node that created
+   the genesis block in a previous procedure).
+
+#. Use the ``sawset`` command to create and submit a batch of transactions that
+   changes the allowed transaction types.
+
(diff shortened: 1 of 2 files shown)


docs/source/sysadmin_guide/setting_allowed_txns.rst
```diff
@@ -0,0 +1,79 @@
+************************************************
+Setting the Allowed Transaction Types (Optional)
+************************************************
+
+By default, a validator accepts transactions from any transaction processor.
+However, Sawtooth allows you to limit the types of transactions that can be
+submitted. This on-chain setting increases the Sawtooth network's security by
+ignoring any unrecognized transaction processors.
+
+This configuration change is done once, usually on the same node that created
```
aludvik: This change can happen many times, and not just in genesis.

peterschwarz: Agree with @aludvik that this is not correct. It is often set in the genesis block initially, but can also be changed at anytime by submitting a transaction with the new value for the setting.  These changes maybe made as transaction processors are added to a network or there are version changes in the existing transaction processors.

chenette: I'll fix the wording.  I was trying to explain that this is an on-chain setting so only one node needs to run this command, but I got all the words wrong. I'll push a fix soon with an improvement.

docs/source/sysadmin_guide/setting_allowed_txns.rst
```diff
@@ -0,0 +1,80 @@
...
+#. Open a terminal window on the "genesis node" (the Sawtooth node that created
+   the genesis block in a previous procedure).
+
+#. Use the ``sawset`` command to create and submit a batch of transactions that
+   changes the allowed transaction types.
+
+   .. code-block:: console
+
+      # sawset proposal create --key /etc/sawtooth/keys/validator.priv \
+      sawtooth.validator.transaction_families='[{"family":"sawtooth_identity", "version":"1.0"}, {"family": "intkey", "version": "1.0"}, {"family":"sawtooth_settings", "version":"1.0"}, {"family":"sawtooth_validator_registry", "version":"1.0"}]'
```
rberg2: There are spaces in the intkey chunk that are not in the others, but I am not sure if that matters.

docs/source/sysadmin_guide/setting_allowed_txns.rst
```diff
@@ -0,0 +1,80 @@
+************************************************
```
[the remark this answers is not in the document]
chenette: I will move this section to after the Testing Sawtooth section, so that it can be near the other "optional" procedures, like the REST API proxy.

docs/source/sysadmin_guide/setting_allowed_txns.rst
chenette: Good catches - thanks!

PR #1877, opened 2018-09-26
Add proc to change off-chain config (Sys Admin Guide)

chenette:
This new procedure describes how to create and change the configuration files for the validator, REST API and (optionally) CLI configuration.

This PR also includes additional steps for the "Testing Sawtooth" procedure to check the list of peers.

### DIFF
--- a/docs/source/sysadmin_guide/off_chain_settings.rst
+++ b/docs/source/sysadmin_guide/off_chain_settings.rst
@@ -1,0 +1,324 @@
+****************************************************
+Changing Off-chain Settings with Configuration Files
+****************************************************
+
+Each Sawtooth component, such as the validator or the REST API, can have an
+optional configuration file that controls the component's behavior. By default,
+Sawtooth does not install any configuration files.  However, Sawtooth provides
+example configuration files that can be customized for your system.
+
+This procedure explains how to create and use Sawtooth configuration files to
+change the following Sawtooth settings:
+
+* Host and port values (bind endpoints) for the validator, REST API, and
+  consensus API
+
+* Peering type and peer nodes on the network
+
+* Network keys for secured communication between nodes (optional)
+
+* Scheduler type (optional)
+
+It also explains how to configure a non-default REST API URL for the Sawtooth
+commands.
+
+See :doc:`configuring_sawtooth` for detailed information on all the settings in
+each configuration file.
+
+.. note::
+
+   This procedure assumes that the configuration directory is
+   ``/etc/sawtooth/``. If your system uses a different location, change this
+   path in the commands below. For more information, see
+   :doc:`configuring_sawtooth/path_configuration_file`.
+
+
+Configure the Validator
+=======================
+
+The following steps configure the validator's networking information so that the
+validator advertises itself properly and knows where to search for peers.
+Additional steps specify the peers for this node, change the scheduler type
+(optional), and create a network key.
+
+#. Create the validator configuration file by copying the example file.
+
+   .. code-block:: console
+
+      $ sudo cp /etc/sawtooth/validator.toml.example /etc/sawtooth/validator.toml
+
+#. Use ``sudo`` to edit this file.
+
+   .. code-block:: console
+
+      $ sudo vi /etc/sawtooth/validator.toml
+
+
+#. Change the network settings for the validator.
+
+   a. Locate the ``endpoint`` setting, which specifies validator's external URL.
+
+      Replace the default interface and port (``127.0.0.1:8800``) with the
+      values for your node. You can use either the NAT values or the publicly
+      addressable IP address and port.
+
+      .. code-block:: ini
+
+         endpoint = "tcp://{external_interface}:{port}"
+
+   #. Locate the ``bind`` settings. If necessary, change these values for your
+      system. The default values are:
+
+        .. code-block:: ini
+
+           bind = [
+             "network:tcp://127.0.0.1:8800",
+             "component:tcp://127.0.0.1:4004",
+             "consensus:tcp://127.0.0.1:5050"
+           ]
+
+      * ``network`` specifies where the validator listens for communication
+        from other nodes
+
+      * ``component`` specifies where the validator listens for communication
+        from this validator's components, such as the REST API and transaction
+        processors
+
+      * ``consensus`` specifies where the validator listens for communication
+        from consensus engines
+
+      .. tip::
+
+         Make sure that all values in this setting are valid for your network.
+         If the bind interface doesn't exist, you might see a ZMQ error in the
+         ``sawtooth-validator`` systemd logs when attempting to start the
+         validator, as in this example:
+
+         .. code-block:: console
+
+            Jun 02 14:50:37 ubuntu validator[15461]:   File "/usr/lib/python3.5/threading.py", line 862, in run
+            ...
+            Jun 02 14:50:37 ubuntu validator[15461]:   File "zmq/backend/cython/socket.pyx", line 487, in zmq.backend.cython.socket.Socket.bind (zmq/backend/cython/socket.c:5156)
+            Jun 02 14:50:37 ubuntu validator[15461]:   File "zmq/backend/cython/checkrc.pxd", line 25, in zmq.backend.cython.checkrc._check_rc (zmq/backend/cython/socket.c:7535)
+            Jun 02 14:50:37 ubuntu validator[15461]: zmq.error.ZMQError: No such device
+            Jun 02 14:50:37 ubuntu systemd[1]: sawtooth-validator.service: Main process exited, code=exited, status=1/FAILURE
+            Jun 02 14:50:37 ubuntu systemd[1]: sawtooth-validator.service: Unit entered failed state.
+            Jun 02 14:50:37 ubuntu systemd[1]: sawtooth-validator.service: Failed with result 'exit-code'.
+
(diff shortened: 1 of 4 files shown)


docs/source/sysadmin_guide/testing_sawtooth.rst
```diff
@@ -32,6 +32,49 @@ to test basic Sawtooth functionality.
...
+      the validator and REST API, as in this example:
+
+      .. code-block:: console
+
+         {
+             "data": [
+             "tcp://validator-1:8800",
+           ],
+           "link": "http://rest-api:8008/peers"
+         }
```
aludvik: I think the whitespace/tabbing is messed up above, should be:
```
         {
           "data": [
             "tcp://validator-1:8800",
           ],
           "link": "http://rest-api:8008/peers"
         }
```
(data has two extra spaces)

docs/source/sysadmin_guide/off_chain_settings.rst
```diff
@@ -0,0 +1,300 @@
...
+
+      Replace the default address and port (``host1:8800``) with the values for
+      one or more nodes in your network. You can use either the NAT values or the
+      publicly addressable IP address and port.
+
+      Specify multiple nodes in a comma-separated list, as in this example:
+
+      .. code-block:: ini
+
+         seeds = ["tcp://{address1}:{port}",
```
rberg2: Seeds are used when dynamic peering is enabled and is not needed in static peering mode, where the peers conf is used.

docs/source/sysadmin_guide/off_chain_settings.rst
```diff
@@ -0,0 +1,300 @@
...
+Each Sawtooth component, such as the validator or the REST API, can have an
+optional configuration file that controls the component's behavior. By default,
+Sawtooth does not install any configuration files.  However, Sawtooth provides
+example configuration files that can be customized for your system.
+
+This procedure explains how to use Sawtooth configuration files to change the
+following Sawtooth settings:
+
+* Host and port values (bind endpoints) for the validator, REST API, and
+  consensus proxy
```
aludvik: Consensus Proxy is an internal implementation detail of the validator and should not be mentioned at this level.

docs/source/sysadmin_guide/off_chain_settings.rst
```diff
@@ -0,0 +1,300 @@
...
+             "component:tcp://127.0.0.1:4004",
+             "consensus:tcp://127.0.0.1:5050"
+           ]
+
+      - ``network`` specifies where the validator listens for communication from
+        validators on other nodes
+
+      - ``component`` specifies where the validator listens for the REST API
+
+      - ``consensus`` specifies the settings for the consensus proxy
```
aludvik: I would change this to:
```
      - ``network`` specifies where the validator listens for communication from other validator nodes
      - ``component`` specifies where the validator listens for communication from this validator’s
         components, such as the REST API and transaction processors
      - ``consensus`` specifies where the validator listens for communication from consensus engines 
```
This is slightly more consistent with the wording here: https://sawtooth.hyperledger.org/docs/core/nightly/master/app_developers_guide/creating_sawtooth_network.html#prereqs-multi-ubuntu-label

And includes TPs in the component description and more accurately defines the consensus endpoint.

docs/source/sysadmin_guide/off_chain_settings.rst
```diff
@@ -0,0 +1,300 @@
...
+
+      .. tip::
+
+         If the bind interface doesn't exist, you might see a ZMQ error in the
+         ``sawtooth-validator`` systemd logs when attempting to start the
+         validator, as in this example:
```
aludvik: Consider adding "Make sure all the network device names and/or IP addresses you passed are valid."

docs/source/sysadmin_guide/off_chain_settings.rst
```diff
@@ -0,0 +1,300 @@
...
+
+      Use the format ``tcp://{hostname}:{port}`` for each peer. Specify multiple
+      peers in a comma-separated list. For example:
+
+      .. code-block:: ini
+
+         peers = ["tcp://node1:8800", "tcp://node2:8800", "tcp://node3:8800"]
+
+   #. Find the ``seeds`` setting, which specifies the peers to use for the
+      initial connection to the validator network.
```
aludvik: `--seeds` is only useful in `dynamic` peering mode, and it specifies the other validators that this validator will first request peering information from in order to peer dynamically.

docs/source/sysadmin_guide/off_chain_settings.rst
chenette: Oops. I forgot to start with "(Dynamic peering only)".  Thanks for catching this.

docs/source/sysadmin_guide/off_chain_settings.rst
```diff
@@ -0,0 +1,300 @@
...
+            Jun 02 14:50:37 ubuntu systemd[1]: sawtooth-validator.service: Main process exited, code=exited, status=1/FAILURE
+            Jun 02 14:50:37 ubuntu systemd[1]: sawtooth-validator.service: Unit entered failed state.
+            Jun 02 14:50:37 ubuntu systemd[1]: sawtooth-validator.service: Failed with result 'exit-code'.
+
+#. Set the peering type and peer list (directly connected nodes) for this
+   Sawtooth node.
+
+   a. Locate the ``peering`` setting, which specifies the type of peering
+      approach the validator should take: static or dynamic. Change this setting
+      to ``dynamic`` unless static peering is required. For more information,
```
aludvik: I am not sure we want to recommend dynamic peering as the default. Raft and PBFT will require static peering. Maybe something like: "The peering type you should use depends on the network use case and consensus algorithm used. For example, public networks using an open-membership consensus algorithm should use dynamic peering while onsortium networks or networks using a fixed-membership consensus algorithm should use static peering."

docs/source/sysadmin_guide/off_chain_settings.rst
```diff
@@ -0,0 +1,300 @@
...
+
+      Specify multiple nodes in a comma-separated list, as in this example:
+
+      .. code-block:: ini
+
+         seeds = ["tcp://{address1}:{port}",
+                  "tcp://{address2}:{port}"]
+
+#. (Optional) Set the scheduler type to either ``serial`` (the default) or
+   ``parallel``.
```
aludvik: Maybe link to the arch section for more info: https://sawtooth.hyperledger.org/docs/core/nightly/master/architecture/scheduling.html#iterative-scheduling

docs/source/sysadmin_guide/off_chain_settings.rst
```diff
@@ -0,0 +1,300 @@
...
+   .. code-block:: ini
+
+      scheduler = 'parallel'
+
+#. (Optional) Set the network key to specify secured network communication
+   between nodes in the network. By default, the network is unsecured.
+
+      .. important::
+
+         The example configuration file contains generic keys. You **must**
```
aludvik: Change "generic keys" to "publicly visible example keys".

docs/source/sysadmin_guide/off_chain_settings.rst
```diff
@@ -0,0 +1,300 @@
...
+
+   .. code-block:: console
+
+      bind = ["127.0.0.1:8008"]
+
+
+#. If necessary, change the ``connect`` setting, which specifies where the
+   REST API can find this node's validator. This setting must match the
+   ``bind = [... component`` value in the validator configuration file (see
+   :ref:`change-nw-validator-label`).
```
aludvik: This is slightly more complicated and matching the string passed to the validator will not always work. If you pass a device to the validator to bind to, such as `tcp://eth0:4004`, this will need to be the public endpoint for that device on the network shared by the REST API and Validator. For example, in this compose file we bind to the eth0 device for the validator's component endpoint, but connect with the validator's name on the network "validator".

https://github.com/hyperledger/sawtooth-raft/blob/master/adhoc/node.yaml#L67
https://github.com/hyperledger/sawtooth-raft/blob/master/adhoc/node.yaml#L109

docs/source/sysadmin_guide/off_chain_settings.rst
chenette: @aludvik - Would the admin ever have to change the ``bind = [ .... consensus:` hostname and port?  (In a realistic scenario, that is.)

If yes, consensus (with or without the word "proxy") needs to be mentioned here.  If not, I'll delete this and say "don't change the consensus setting" in the `bind` step.

chenette: (From aludvik via phone): Change to "consensus API".

docs/source/sysadmin_guide/off_chain_settings.rst
chenette: @aludvik - How about this wording:

"Make sure that all values in this setting are valid for your network.  If the bind interface doesn't exist..."

aludvik: Sounds good.

docs/source/sysadmin_guide/testing_sawtooth.rst
chenette: The source file doesn't have any extra spaces before "data".  I hope Jenkins isn't inserting spurious indents in the build artifacts.

PR #1880, opened 2018-09-27
Update and edit Configuring Permissions in Sys Admin Guide

chenette:
- Write new intro
- Describe policy files and policy dir
- Note that you can lock yourself out of Identity settings if you're not careful
- Correct the sawtooth identity command name (it's not identity-tp)
- Add links to relevant info
- Fix formatting throughout (including changing indents)

Also:

- Correct the first "sawset proposal create" command to start with sudo,
  include the --key option
  Plus: Use fancy $(cat ...) syntax to provide the user's actual public key
  and add a note that this command MUST be run on the "genesis node"

- Replace the second "sawset proposal create" command to refer to
  the first one (the procedure assumes it has already been done)

### DIFF
--- a/docs/source/sysadmin_guide/configuring_permissions.rst
+++ b/docs/source/sysadmin_guide/configuring_permissions.rst
@@ -1,16 +1,36 @@
-***********************
-Configuring Permissions
-***********************
+************************************************
+Configuring Validator and Transactor Permissions
+************************************************
 
-Transactor Permissioning
-========================
+This section describes the validator and transactor permissions in Hyperledger
+Sawtooth.
 
-Overview
---------
+* `Transactor key permissioning` controls who (which users and clients) can
+  submit transactions and batches to a validator. Sawtooth provides both
+  on-chain (network-wide) and off-chain (local) settings for transactor
+  key permissioning.
+
+* `Validator key permissioning` controls which nodes can connect to the
+  Sawtooth network. These settings are configured with on-chain settings.
+
+The :doc:`Identity <../transaction_family_specifications/identity_transaction_family>`
+and :doc:`Settings <../transaction_family_specifications/settings_transaction_family>`
+transaction processors (or equivalent) are required to process the on-chain
+settings. Note that Settings includes several settings to control who can
+change on-chain settings and how many votes are required for a settings change.
+
+.. note::
+
+   Sawtooth also includes another type of transaction control that limits which
+   transaction processors can submit transactions. For more information, see
+   "Setting the Allowed Transaction Types (Optional)".
+
+Transactor Key Permissioning
+============================
 
 A running protected network needs a mechanism for limiting which transactors
 are allowed to submit batches and transactions to the validators. There are
-two different methods for defining the transactors a validator will accept.
+two different methods for defining the transactors that a validator will accept.
 
 The first method is configuring a validator to only accept batches and
 transactions from predefined transactors that are loaded from a local validator
@@ -19,7 +39,7 @@ is immutable while the validator is running. This set of permissions are only
 enforced when receiving a batch from a connected client, but not when receiving
 a batch from a peer on the network.
 
-The second method uses the identity namespace which will allow for network-wide
+The second method uses the Identity namespace which handles network-wide
 updates and agreement on allowed transactors. The allowed transactors are
 checked when a batch is received from a client, received from a peer, and
 when a block is validated.
@@ -32,17 +52,16 @@ configuration is checked.
 It is possible that in the time that the batch has been in the pending queue,
 the transactor permissions have been updated to no longer allow a batch signer
 or a transaction signer that is being included in a block. For this reason, the
-allowed transactor roles will also need to be checked on block validation.
+allowed transactor roles must also be checked on block validation.
 
 .. _Off-Chain_Transactor_Permissioning:
 
-Off-Chain Transactor Permissioning
+Off-chain Transactor Permissioning
 ----------------------------------
-Validators can be locally configured by creating a validator.toml file and
-placing it in the config directory. Adding configuration for transactor
-permissioning to the configuration shall be done in the following format:
-
-validator.toml:
+Validators can be locally configured by creating a validation configuration
+file, ``validator.toml``, and placing it in the config directory. Add
+the configuration for transactor permissioning to this file using the following
+format:
 
 .. code-block:: none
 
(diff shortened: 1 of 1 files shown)


docs/source/sysadmin_guide/configuring_permissions.rst
```diff
@@ -84,106 +121,114 @@ transactions. Each role equals a policy name that corresponds to a policy
...
-To configure on-chain roles, the signer of identity transactions needs to have
-their public key set in the Setting "sawtooth.identity.allowed_keys". Only
-those transactors whose public keys are in that setting are allowed to update
-roles and policies.
+Before you can configure on-chain roles, your public key must be set in the
+``sawtooth.identity.allowed_keys`` setting. Only those transactors whose
+public keys are in that setting are allowed to update roles and policies.
 
-.. code-block:: console
+1. Make sure that the Identity transaction processor and the REST API are
```
agunde406: Should this mentioned the Setting transaction processor to?

docs/source/sysadmin_guide/configuring_permissions.rst
```diff
@@ -84,106 +121,114 @@ transactions. Each role equals a policy name that corresponds to a policy
...
 
-default:
+``default``:
   When evaluating role permissions, if the role has not been set, the default
-  policy will be used. The policy can be changed to meet the network's
+  policy is used. The policy can be changed to meet the network's
   requirements after initial start-up by submitting a new policy with the name
   default. If the default policy has not been explicitly set, the default
-  is “PERMIT_KEY \*”.
+  is ``PERMIT_KEY \*``.
```
agunde406: I think "\" needs to be remvoed

docs/source/sysadmin_guide/configuring_permissions.rst
```diff
@@ -84,106 +121,114 @@ transactions. Each role equals a policy name that corresponds to a policy
...
-requesters below.
+connection. In this section, the entities acting in the different roles are
+referred to as "requesters".
 
 Validators are able to determine whether messages delivered to them should
 be handled or dropped based on a set of role and identities stored within the
-Identity namespace. Each requester will be identified by the public key derived
-from their identity signing key. Permission verifiers examine incoming
+Identity namespace. Each requester is identified by the public key derived
+from its Identity signing key. Permission verifiers examine incoming
```
agunde406: Did you capitalize this because of the Identity transaction processor? This is the public key used by the validator.

docs/source/sysadmin_guide/configuring_permissions.rst
```diff
@@ -208,13 +253,28 @@ rules and is evaluated in order. If the public key is denied, the connection
...
+   in the previous procedure. This example shows that the same policy can be
+   used for both transaction and validator permissioning.
+
+#. Run the following command to set the role for network to permit all:
+
+   .. code-block:: console
+
+       $ sawtooth identity role create network policy_1
+
+#. To see the role in state, run the following command:
```
agunde406: move the below code block over to match the others

docs/source/sysadmin_guide/configuring_permissions.rst
```diff
@@ -226,35 +286,39 @@ Network Roles
 The following is the suggested high-level role for on-chain validator network
 permissioning.
 
-network
+``default``:
+  When evaluating role permissions, if the role has not been set, the default
+  policy is used. The policy can be changed to meet the network's
+  requirements after initial start-up by submitting a new policy with the name
+  default. If the default policy has not been explicitly set, the default
+  is ``PERMIT_KEY \*``.
```
agunde406: Remove the "\"

docs/source/sysadmin_guide/configuring_permissions.rst
```diff
@@ -208,13 +253,28 @@ rules and is evaluated in order. If the public key is denied, the connection
...
-.. code-block:: console
+   .. code-block:: console
+
+     $ sawset proposal create sawtooth.identity.allowed_keys=02b2be336a6ada8f96881cd55fd848c10386d99d0a05e1778d2fc1c60c2783c2f4
 
-    $ sawtooth identity role create network policy_1
+#. If necessary, create a new policy.
 
-To see the role in state, run the following command:
+   This procedure assumes that ``policy_1`` already exists on-chain, as created
```
agunde406: Can you add a comment about how you could use the same policy but in practice these would be different keys and it does not really make sense to do so? For example, you would not put transaction signer keys into the network policies.

docs/source/sysadmin_guide/configuring_permissions.rst
```diff
@@ -84,106 +121,114 @@ transactions. Each role equals a policy name that corresponds to a policy
...
+#. Use the following command to add a signer's public key to this setting:
 
-Once your signer key is stored in the setting, the ``identity-tp`` command can be used
-to set and update roles and policies. Make sure that the Identity transaction
-processor and the REST API are running.
+   .. code-block:: console
 
-.. literalinclude:: ../cli/output/sawtooth_identity_policy_create_usage.out
-  :language: console
+     $ sawset proposal create sawtooth.identity.allowed_keys=02b2be336a6ada8f96881cd55fd848c10386d99d0a05e1778d2fc1c60c2783c2f4
```
rberg2: maybe we should mention exactly what key is needed here. In some of my scripts I do `sawtooth.identity.allowed_keys=$(cat /home/sawtooth/.sawtooth/keys/sawtooth.pub)`

docs/source/sysadmin_guide/configuring_permissions.rst
chenette: I shall remove the now-unnecessary backslash.

docs/source/sysadmin_guide/configuring_permissions.rst
chenette: As above, I shall remove the now-unnecessary backslash.

docs/source/sysadmin_guide/configuring_permissions.rst
chenette: Would it be correct to change this phrase to "Each requester is identified by its public key?"

docs/source/sysadmin_guide/configuring_permissions.rst
chenette: Instead of an additional note, I think we should change this procedure to not reuse `policy_1`.  Otherwise, it might imply that the same policy is being being reused on purpose.

PR #1885, opened 2018-09-28
Implement and integrate Rust block store

aludvik:


validator/src/journal/block_manager_ffi.rs
```diff
@@ -93,30 +91,22 @@ pub unsafe extern "C" fn block_manager_contains(
...
-    block_store: *mut py_ffi::PyObject,
+    commit_store: *mut c_void,
 ) -> ErrorCode {
-    check_null!(block_manager, block_store_name, block_store);
+    check_null!(block_manager, commit_store);
 
-    let gil = Python::acquire_gil();
-    let py = gil.python();
-    let py_block_store = PyBlockStore::new(PyObject::from_borrowed_ptr(py, block_store));
+    let commit_store = Box::from_raw(commit_store as *mut CommitStore);
```
peterschwarz: Why do you use `from_raw`/`into_raw` here?  Why don't you just use the pointer?

aludvik: I don’t have a good reason for not just using the pointer. I thought I needed to do from_raw/into_raw to get the clone to work right, but I was confused.

peterschwarz: There are lot's of examples of this in `chain_ffi.rs`

aludvik: The thing that threw me off is the type expected by `add_store()` is `Box<BlockStore>`, not `CommitStore`, so you can't do `(*(commit_store as *mut CommitStore))`.

aludvik: I am changing my mind again. Trying to do `(*(commit_store as *mut Box<CommitStore>))` is bad, causes a seg fault, so I think I do need the from_raw/into_raw.

PR #1889, opened 2018-10-03
Sign and verify consensus messages

ltseeley:
Updates the validator to sign and verify all consensus messages that are sent and received.


validator/sawtooth_validator/consensus/handlers.py
```diff
@@ -54,7 +54,7 @@ def __init__(
         self._response_class = response_class
         self._response_type = response_type
 
-    def handle_request(self, request, response):
+    def handle_request(self, request, connection_id):
```
peterschwarz: Looks like you replaced `response` but it is still required by other implementors, and at other call sites.

validator/sawtooth_validator/server/core.py
```diff
@@ -240,7 +241,10 @@ def __init__(self,
             max_incoming_connections=20,
             max_future_callback_workers=10)
 
-        consensus_notifier = ConsensusNotifier(consensus_service)
+        registered_engines = ConcurrentMultiMap()
```
aludvik: Can we make this a separate class and give the operations on it descriptive names? For example, `ConsensusRegistry` with methods `register_engine()`, `unregister_engine()`, `get_connection_id_by_engine_info()`, and `get_engine_info_by_connection_id()`.

validator/sawtooth_validator/gossip/signature_verifier.py
```diff
@@ -113,13 +115,39 @@ def is_valid_transaction(txn):
...
+    # validate consensus message signature
+    header = ConsensusPeerMessageHeader()
+    header.ParseFromString(message_envelope.header)
+
+    context = create_context('secp256k1')
+    public_key = Secp256k1PublicKey.from_hex(header.signer_public_key)
+    if not context.verify(message_envelope.header_signature,
+                          message_envelope.header,
+                          public_key):
+        LOGGER.debug("transaction signature invalid for txn: %s",
```
aludvik: copy pasta, "message signature invalid for message"

validator/sawtooth_validator/gossip/signature_verifier.py
```diff
@@ -150,6 +178,19 @@ def handle(self, connection_id, message_content):
...
             return HandlerResult(status=HandlerStatus.PASS)
 
+        if tag == GossipMessage.CONSENSUS:
+            if obj.signature in self._seen_cache:
+                self._msg_dropped_count.inc()
+                return HandlerResult(status=HandlerStatus.DROP)
+
+            if not is_valid_consensus_message(obj):
+                LOGGER.debug("consensus message signature is invalid: %s",
+                             obj.signature)
```
aludvik: This is already logged in `is_valid_consensus_message()`

validator/sawtooth_validator/gossip/signature_verifier.py
```diff
@@ -150,6 +178,19 @@ def handle(self, connection_id, message_content):
             self._seen_cache[obj.header_signature] = None
             return HandlerResult(status=HandlerStatus.PASS)
 
+        if tag == GossipMessage.CONSENSUS:
+            if obj.signature in self._seen_cache:
+                self._msg_dropped_count.inc()
+                return HandlerResult(status=HandlerStatus.DROP)
```
aludvik: I don't think we want to drop these duplicates yet. We're not gossiping, so duplicates should be low, and we don't have anything like a nonce in the consensus message, so it is realistic that messages will be received with identical signatures.

That being said, this probably means these messages are a denial of service vector since a malicious validator can spam the same message as fast as possible to slow down a peer. Dropping duplicates here and requiring nonces in duplicate consensus messages would mitigate this since a bad validator would need to create and sign a new message every time, which is more expensive than just resending the same message.

aludvik:
@ltseeley re conversation in rocket chat with @vaporos, this PR should be backported to 1.1 after it is merged, as it is important to the proper functioning of leader-based BFT algorithms.

validator/sawtooth_validator/concurrent/atomic.py
```diff
@@ -128,3 +128,32 @@ def get(self, key, default):
...
+        with self._lock:
+            return self._info is not None
+
+    def register_engine(self, connection_id, name, version):
+        with self._lock:
+            self._info = {
+                'connection_id': connection_id,
+                'name': name,
+                'version': version
+            }
```
aludvik: This should be a namedtuple here to lock in the fields and especially since you are returning a reference to a thing that is guarded by a lock out from under the lock. If you use a dictionary, after calling `get_engine_info()`, you could modify the dictionary without acquiring the lock. namedtuples are a lot like dictionaries, but they are immutable so it doesn’t matter if you return it out from under the lock and you can't add/remove fields.

validator/sawtooth_validator/concurrent/atomic.py
```diff
@@ -128,3 +128,32 @@ def get(self, key, default):
                 return self._dict[key].copy()
             except KeyError:
                 return default
+
+
+class ConsensusRegistry:
```
aludvik: This class should go in a new file in the `consensus/` folder. `concurrent/` is just for concurrency primitives missing from the python standard library.

PR #1893, opened 2018-10-09
Add Batches to Consensus API, Rewrite Consensus Notifier in Rust and Background

aludvik:
This PR makes a number of consensus related changes:
1. Implements New and Invalid batch notifications for consensus
2. Rewrites the ConsensusNotifier in Rust
3. Moves the ConsensusNotifier to a Background thread to avoid waiting for network acks in core components

This PR is based on #1889


validator/src/journal/publisher.rs
```diff
@@ -453,6 +457,8 @@ impl SyncBlockPublisher {
     }
 
     pub fn on_batch_received(&self, batch: Batch) {
+        self.consensus_notifier
```
peterschwarz: Should this only be called after the permission check?  The batch may not be added to the pending queue if that check fails.

sdk/rust/src/consensus/engine.rs
```diff
@@ -37,36 +37,43 @@ pub enum Update {
     Shutdown,
 }
 
-#[derive(Clone, Default, Eq, Hash, PartialEq, PartialOrd)]
-pub struct BlockId(Vec<u8>);
-impl Deref for BlockId {
-    type Target = Vec<u8>;
+macro_rules! declare_id {
```
peterschwarz: 👍

sdk/rust/src/consensus/engine.rs
```diff
@@ -37,36 +37,43 @@ pub enum Update {
...
+        impl fmt::Debug for $name {
+            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+                write!(f, "{}", hex::encode(&self.0))
+            }
+        }
+    };
 }
 
+declare_id!(BatchId);
+declare_id!(BlockId);
```
peterschwarz: What happened to `PeerId`?

sdk/rust/src/consensus/engine.rs
```diff
@@ -73,6 +75,7 @@ macro_rules! declare_id {
 
 declare_id!(BatchId);
 declare_id!(BlockId);
+declare_id!(PeerId);
```
peterschwarz: This looks like it belongs in the previous commit.

validator/src/consensus/notifier.rs
```diff
@@ -176,3 +180,106 @@ fn from_hex<T: AsRef<str>>(hex_string: T, id_name: &str) -> Vec<u8> {
...
+    BlockValid(String),
+    BlockInvalid(String),
+    BlockCommit(String),
+    BatchNew(String),
+    BatchInvalid(String),
+}
+
+#[derive(Clone)]
+pub struct BackgroundConsensusNotifier {
+    tx: Arc<Mutex<Sender<ConsensusNotification>>>,
```
peterschwarz: Sender's don't need to be in a `Mutex`

aludvik: In this context, it does because of the trait bounds on the ConsensusNotifier trait, which are there because of Sync requirements on the BlockPublisher.

validator/src/journal/publisher.rs
aludvik: Since we never mark batches that fail the permission check as invalid, yes I believe you are correct. Otherwise batches may be sent to consensus but never marked invalid or committed.

PR #1897, opened 2018-10-12
Remove noisy debug logging statement, fix typo

boydjohnson:


validator/src/journal/commit_store_ffi.rs
```diff
@@ -490,10 +490,7 @@ unsafe fn return_proto<I, O: protobuf::Message + From<I>>(
 
 fn map_database_error(err: DatabaseError) -> ErrorCode {
     match err {
-        DatabaseError::NotFoundError(s) => {
-            debug!("Not found: {}", s);
-            ErrorCode::NotFound
-        }
+        DatabaseError::NotFoundError(s) => ErrorCode::NotFound,
```
arsulegai: NotFoundError(s) here can be replaced with NotFoundError(_) ?

boydjohnson: Yeah, you are right, `s` isn't being used. I will change this.

boydjohnson:
Thanks, @arsulegai. Can you re-review?

PR #1898, opened 2018-10-12
Adding optional serde support for consensus structs

knkski:
The Pbft consensus engine introduced state management via Serde, which relies on a few structs in the core consensus code having Serde support. This PR adds Serde support to the necessary structs.

### DIFF
--- a/sdk/rust/Cargo.toml
+++ b/sdk/rust/Cargo.toml
@@ -23,6 +23,7 @@ build = "build.rs"
 default = ["pem"]
 # Add support for loading PEM encoded private keys
 pem = ["openssl"]
+with-serde = ["serde", "serde_derive"]
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


knkski:
@boydjohnson: Good point - I went with `with-serde` because I couldn't figure out a way to have `protobuf` actually generate serde code without a feature named that enabled (just having `protobuf = { version = "2", features = ["with-serde"] }` didn't seem to do it). This PR doesn't introduce serde for any protobuf objects yet, but it's possible that we would. Is it worth keeping the name as-is for future compatibility with protobuf, and/or digging deeper to see if it can be done without the `with-` prefix?

boydjohnson:
From talking with @knkski offline, it appears that we have to name our feature "with-serde", because protobuf's optional dependency "with-serde".

PR #1901, opened 2018-10-16
Sign and verify consensus messages

ltseeley:
Updates how the validator sends and receives consensus peer messages.
Previously, all messages were unsigned and unverified. This update makes
the following changes:

1. Removes GossipConsensusMessage and corresponding network handler in
favor of using GossipMessage for consensus messages; this reduces
duplication and allows for reuse of the existing signature verification
functionality used for GossipMessages.
2. Updates consensus messages so they are sent with a header signature
that is verified when the message is received.


dcmiddle:
@vaporos I think we want this for backport to 1-1.

validator/sawtooth_validator/gossip/gossip_handlers.py
```diff
@@ -245,8 +247,15 @@ def handle(self, connection_id, message_content):
             # If we already have this block, don't forward it
             if not self._completer.get_block(obj.header_signature):
                 self._gossip.broadcast_block(obj, exclude, time_to_live=ttl)
+        elif tag == GossipMessage.CONSENSUS:
+            peer_message = ConsensusPeerMessage()
+            peer_message.ParseFromString(obj.message)
+
+            self._notifier.notify_peer_message(
+                message=peer_message,
+                sender_id=bytes.fromhex(connection_id))
```
aludvik: This needs to be the public key of the sending validator. I believe you can use this to get it: https://github.com/hyperledger/sawtooth-core/blob/master/validator/sawtooth_validator/gossip/gossip.py#L200

signing/sawtooth_signing/secp256k1.py
```diff
@@ -112,12 +112,13 @@ def sign(self, message, private_key):
         except Exception as e:
             raise SigningError('Unable to sign message: {}'.format(str(e)))
 
-    def verify(self, signature, message, public_key):
+    def verify(self, signature, message, public_key, sig_is_hex=True):
```
aludvik: This is not a very ergonomic way to handle this. More ergonomic alternatives include a new method `verify_bytes()` or doing an `isinstance()` check on signature before processing it: https://docs.python.org/3/library/functions.html#isinstance

I am in favor of the isinstance check, as eventually we will want to move all signatures to bytes. When that happens, we can just remove the check.

PR #1903, opened 2018-10-17
Return Result in Engine::start

knkski:
Lets us handle startup errors in the engine more gracefully.


aludvik:
PRs should be opened against https://github.com/hyperledger/sawtooth-raft and https://github.com/hyperledger/sawtooth-pbft to make this change before this is merged.

knkski:
@aludvik: https://github.com/hyperledger/sawtooth-pbft/pull/8 and https://github.com/hyperledger/sawtooth-raft/pull/42

sdk/examples/devmode_rust/src/engine.rs
```diff
@@ -214,7 +214,7 @@ impl Engine for DevmodeEngine {
         updates: Receiver<Update>,
         service: Box<Service>,
         startup_state: StartupState,
-    ) {
+    ) -> Result<(), Error>{
```
ltseeley: Should probably be `Result<(), Error> {` instead of `Result<(), Error>{`

knkski: Good catch, updated 👍

PR #1904, opened 2018-10-17
C026 Use OpenSSL for SHA2 hashing functions in Rust

aludvik:
Use a security vetted library for cryptographic hashing.


aludvik:
Added a commit to rename the new module from `crypto` to `hashlib` so that it is clearer we are *not* depending on the rust-crypto crate, which also imports as `crypto`.

validator/src/hashlib.rs
```diff
@@ -0,0 +1,24 @@
+use hex;
+use openssl;
+
+pub fn sha256_digest_str(item: &str) -> String {
+    let mut bytes: Vec<u8> = Vec::new();
```
dcmiddle: Would it be easier just do something like
```
let hash = sha256(foo);
    hex::encode(hash)
```

aludvik: Yep, added a commit

PR #1906, opened 2018-10-17
C029 Change ubuntu base image to bionic

rbuysse:


danintel:
This PR is missing any doc updates in the sysadmin guide or app developers guide for Ubuntu. It should mention that gnupg (or gnupg2 or gnupg1) needs to be installed.

PR #1914, opened 2018-10-22
Rewrite consensus notifier in Rust and run in background thread

aludvik:
This removes an inefficiency in the core validator, where it must wait for a network ack from the engine whenever it sends it a notification.

### DIFF
--- a/validator/sawtooth_validator/consensus/notifier.py
+++ b/validator/sawtooth_validator/consensus/notifier.py
@@ -13,91 +13,105 @@
 # limitations under the License.
 # ------------------------------------------------------------------------------
 
-import hashlib
+import ctypes
+from enum import IntEnum
 import logging
 
 
-from sawtooth_validator.protobuf.block_pb2 import BlockHeader
-from sawtooth_validator.protobuf import consensus_pb2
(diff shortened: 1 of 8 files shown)


boydjohnson:
@aludvik Could you talk about the layers of Python and Rust here? Which Python class is in the Python layer?

aludvik:
@boydjohnson The interconnect service and consensus registry remain in Python and are here: https://github.com/hyperledger/sawtooth-core/pull/1914/files#diff-207e424b4836caed42eb1019a97f0dc9R31

This is wrapped by an implementation of the `NotifierService` trait: https://github.com/hyperledger/sawtooth-core/pull/1914/files#diff-5f253ff9fce18b79628534b75a9d3ea2R50

And this trait impl extends anything that is a `NotifierService` to also be a `ConsensusNotifier`. This is all to make it easier to swap out the Python service/registry later.

Finally, there is an FFI interface for those methods that still need to be called from Python: https://github.com/hyperledger/sawtooth-core/pull/1914/files#diff-bda47befe07894f4242429c11ef39015R137

aludvik:
Rebased on bionic upgrade

aludvik:
Rebased on lint fixes in master

PR #1915, opened 2018-10-23
Kube file bumper update

rberg2:


docker/kubernetes/sawtooth-kubernetes-default-poet.yaml
```diff
@@ -17,23 +17,46 @@ items:
...
+              - "if [ ! -f /poet-shared/poet-enclave-measurement ]; then \
+                    poet enclave measurement >> /poet-shared/poet-enclave-measurement; \
+                fi && \
+                if [ ! -f /poet-shared/poet-enclave-basename ]; then \
+                    poet enclave basename >> /poet-shared/poet-enclave-basename; \
+                fi && \
+                if [ ! -f /poet-shared/simulator_rk_pub.pem ]; then \
+                    cp /etc/sawtooth/simulator_rk_pub.pem /poet-shared; \
+                fi && \
+                while [ ! -f /poet-shared/validator-0/keys/validator.priv ]; do sleep 1; done && \
```
vaporos: With these sleep loops, should it print out when it sleeps for debugging purposes?

boydjohnson:
@rberg2 I approved it but just found a problem. The devmode k8s yaml file doesn't include the devmode engine. https://github.com/hyperledger/sawtooth-core/blob/master/integration/sawtooth_integration/docker/test_devmode_engine_liveness.yaml#L70 Here is an example of use of the devmode engine.

PR #1916, opened 2018-10-23
Enforce consensus settings

ltseeley:
Updates the validator to:
1. Register any engine that requests
2. Set the active engine based on the value of the on-chain settings `sawtooth.consensus.algorithm.name` and `sawtooth.consensus.algorithm.version`
3. Check all consensus messages and only accept a message if it is a registration request OR it is from the registered consensus engine
4. Send `EngineActivated` and `EngineDeactivated` updates to a consensus engine when it is activated/deactivated

### DIFF
--- a/validator/sawtooth_validator/gossip/gossip.py
+++ b/validator/sawtooth_validator/gossip/gossip.py
@@ -201,6 +201,13 @@ class Gossip:
         with self._lock:
             return self._network.connection_id_to_public_key(peer)
 
+    def get_peers_public_keys(self):
+        """Returns the list of public keys for all peers."""
+        with self._lock:
+            return [
+                self._network.connection_id_to_public_key(peer)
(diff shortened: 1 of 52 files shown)


validator/sawtooth_validator/gossip/gossip.py
```diff
@@ -201,6 +201,13 @@ def peer_to_public_key(self, peer):
         with self._lock:
             return self._network.connection_id_to_public_key(peer)
 
+    def get_peers_public_keys(self):
+        """Returns the list of public keys for all peers."""
```
boydjohnson: Is this used in this PR?

ltseeley:
An integration test would be very nice to have; we don't have one in the works just yet, but it's something to think about for the future. We don't know for sure yet how/if swapping consensus engines truly works, so that's probably something we'll have to explore then.

PR #1918, opened 2018-10-25
Bumper Backports - Validator, Consensus SDKs, Devmode

aludvik:
Backports changes that are relatively easy to change that affect stability or APIs.

Note that I think we should also backport the change in 20e2265083fdc059c6660d19c59e5c75f045ae4f as it fixes a known bug with the consensus API. However, this change depends on two components in the validator not being backported (block manager and fork cache) so I will need to rewrite part of the fix for 1-1.


aludvik:
Fixed a missing argument in a unit test.

aludvik:
Recommend closing this and merging https://github.com/hyperledger/sawtooth-core/pull/1921 to avoid the need to rebase and rebuild on Jenkins.

aludvik:
#1921 is built on this PR so going to close this and merge #1921 to avoid rebuilding on Jenkins

PR #1922, opened 2018-10-26
1-1 documentation backports

boydjohnson:


docs/source/sysadmin_guide/configure_sgx.rst
```diff
@@ -426,32 +532,39 @@ Use these commands to start the Sawtooth services:
...
     $ sudo journalctl -f \
     -u sawtooth-validator \
-    -u sawtooth-tp_settings \
+    -u sawtooth-settings-tp \
     -u sawtooth-poet-validator-registry-tp \
+    -u sawtooth-poet-engine \
     -u sawtooth-rest-api \
-    -u sawtooth-intkey-tp-python
+    -u sawtooth-intkey-tp-python \
+    -u sawtooth-identity-tp.service
```
dcmiddle: Is this a copy/paste error? (.service)

boydjohnson: I think for consistencies sake we should remove the `.service`. In my limited experience, systemctl and journalctl allow .service to be at the end.

chenette: @dcmiddle - Thanks for this catch!  The same issue is also in the non-SGX section (plus the additional error of `sawtooth-poet-engine.service`). I'll submit a PR to fix the non-SGX section.

PR #1923, opened 2018-10-29
Stabilize Consensus P2P messaging API

aludvik:
This makes one more change to the p2p messaging API, which allows consensus
engines to save messages with their signatures and enables validation of
message signatures after the fact by an engine or observer.

### DIFF
--- a/protos/consensus.proto
+++ b/protos/consensus.proto
@@ -17,37 +17,31 @@ syntax = "proto3";
 
 // --== Data Structures ==--
 
-// A consensus-related message sent between peers
-message ConsensusPeerMessage {
-  // Interpretation is left to the consensus engine implementation
-  string message_type = 1;
-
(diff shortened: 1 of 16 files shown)


protos/consensus.proto
```diff
@@ -188,8 +184,17 @@ message ConsensusNotifyAck {}
 
 // Send a consensus message to a specific, connected peer
 message ConsensusSendToRequest {
-  ConsensusPeerMessage message = 1;
-  bytes peer_id = 2;
+  // Payload to send to peer
+  //
+  // NOTE: This payload will be wrapped up in a ConsensusPeerMessage struct,
```
vaporos: Probably no need for 'NOTE:' -- that's what a comment is...

aludvik: But this is a note worthy comment.

PR #1921, opened 2018-10-26
Bumper Backports - Consensus API Bug Fix; Send in-progress blocks

aludvik:
This is based on #1918 and includes a backport of a bug fix for the consensus API. Without this fix, if an engine connects to a validator after the validator has started receiving and processing blocks, the engine will miss out on notifications about earlier blocks.

PR #1924, opened 2018-10-29
Bumper Backports - Stabilize Consensus P2P messaging API

aludvik:
This makes one more change to the p2p messaging API, which allows consensus
engines to save messages with their signatures and enables validation of
message signatures after the fact by an engine or observer.


aludvik:
This PR is based on #1921. New commits are:
12393d1d3 Stabilize p2p messaging API

aludvik:
Updated to include changes to Python SDK, which I missed somehow first time around.

aludvik:
Since I was only looking at commits that touched validator/ sdk/ or signing/ I missed the commits that actually _turned on the devmode tests_. Also added a commit to actually test that message signing and verification works.

aludvik:
Tested locally against sawtooth-pbft liveness test.

PR #1746, opened 2018-06-26
Identity tp rewrite

eloaverona:
Add identity transaction processor in rust
Adapt identity-tp in rust so it can compile to wasm and be used as a Sabre contract

PR #1926, opened 2018-10-29
Identity Transaction Processor Rewrite

knkski:
Follow-up to #1746, updates it with changes to master. Opening this PR since updating that one requires being a maintainer.

Removes the current Python-based Identity TP, and adds a Rust-based Identity TP.


knkski:
@boydjohnson: Updated to remove lint. @RyanLassigBanks would probably be the person to ask about why/if we need `bin/identity-tp-deb`.

families/identity/python/Dockerfile
```diff
@@ -1,49 +0,0 @@
-# Copyright 2018 Cargill Incorporated
```
aludvik: I think this delete should go in the remove python commit, as well as the `families/identity/python/sawtooth_identity/.keep` delete

families/identity/sawtooth_identity/Cargo.toml
```diff
@@ -0,0 +1,59 @@
...
+protobuf = "2.0"
+cfg-if = "0.1"
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+rust_crypto = {git = "https://github.com/agunde406/rust-crypto", branch="wasm_sha2"}
+sabre-sdk = {git = "https://github.com/hyperledger/sawtooth-sabre"}
+
+[target.'cfg(unix)'.dependencies]
+rust-crypto = "0.2"
+sawtooth_sdk = { path = "../../../sdk/rust" }
```
aludvik: We should pull this in from crates now: https://crates.io/crates/sawtooth-sdk

families/identity/sawtooth_identity/Cargo.toml
```diff
@@ -0,0 +1,59 @@
...
+log = "0.3"
+log4rs = "0.7"
+zmq = { git = "https://github.com/erickt/rust-zmq", branch = "release/v0.8" }
```
aludvik: We should pull this in from crates now: https://crates.io/crates/sawtooth-zmq

families/identity/sawtooth_identity/Cargo.toml
```diff
@@ -0,0 +1,59 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+
+[package]
+name = "sawtooth_identity"
```
aludvik: The binary produced should be named `identity-tp` to be consistent with the naming scheme for other core tps and to match the previous entry point. See https://github.com/hyperledger/sawtooth-core/blob/master/families/settings/sawtooth_settings/Cargo.toml#L38 for an example of how to do this.

families/identity/sawtooth_identity/Cargo.toml
knkski: It seems like we should be testing this code against the codebase that it's sitting in, instead of what's been published, as it might otherwise get out of sync with what's in master.

families/identity/protos/setting.proto
```diff
@@ -0,0 +1 @@
+../../../protos/setting.proto
```
boydjohnson: Are these necessary? Could the path to the external proto file just be hard coded in build.rs?

bin/identity-tp-deb
```diff
@@ -11,4 +13,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-# ------------------------------------------------------------------------------
+# -----------------------------------------------------------------------------
+bin=/usr/bin/sawtooth_identity
+
```
boydjohnson: I don't think this file is necessary, as there is already a `bin/identity-tp`.

families/identity/sawtooth_identity/Cargo.toml
boydjohnson: I agree with @knkski. If it is in the core repo, it should be tested against the current code, and not a release version of the code. It will be easier to see inconsistencies.

families/identity/Dockerfile
```diff
@@ -11,39 +11,38 @@
...
+    CARGO_INCREMENTAL=0
 
 WORKDIR /project/sawtooth-core
+
+CMD cd families/identity/sawtooth_identity \
+ && echo "\033[0;32m--- Building identity-tp ---\n\033[0m" \
+ && rm -rf ./bin/ \
+ && mkdir -p ./bin/ \
+ && cargo build --release \
+ && cp ./target/release/sawtooth_identity ./bin/sawtooth_identity
```
aludvik: I think this is busted and should say `cp ./target/release/identity-tp ./bin/identity-tp` and that there are some additional places where this needs to be changed, such as `bin/identity-tp`

PR #1931, opened 2018-10-30
Add get_block_by_id convenience function to Rust and Python SDKs

knkski:
Adds a `get_block_by_id` convenience method to the Rust and Python SDKs that builds on top of their respective `get_block` methods.

Motivating example: https://github.com/hyperledger/sawtooth-pbft/blob/master/src/handlers.rs#L468


knkski:
@aludvik: I checked usages of `get_blocks` in Poet, PBFT and Raft. I could not find any usages in Raft, but every usage that I could find of `get_blocks` in Poet and PBFT could be replaced by usage of `get_block_by_id`. That means we could minimize overall complexity by adding a bit of complexity here.

PR #1938, opened 2018-11-08
Fix block sync for registering consensus engines

ltseeley:
Updates the way the validator sends new blocks when a consensus engine
registers to only send valid blocks. Blocks that have not been validated
yet will be sent to the engine when they get validated.

### DIFF
--- a/validator/src/journal/chain.rs
+++ b/validator/src/journal/chain.rs
@@ -528,8 +528,8 @@ impl<BC: BlockCache + 'static, BV: BlockValidator + 'static> ChainController<BC,
         info!("Ignoring block {}", block)
     }
 
-    // Returns all blocks in forks not on the chain with the given head. If head is None, uses the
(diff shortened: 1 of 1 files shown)


validator/src/journal/chain.rs
```diff
@@ -528,8 +528,8 @@ impl<BC: BlockCache + 'static, BV: BlockValidator + 'static> ChainController<BC,
         info!("Ignoring block {}", block)
     }
 
-    // Returns all blocks in forks not on the chain with the given head. If head is None, uses the
-    // current chain head. If head is not found, returns None.
+    // Returns all valid blocks in forks not on the chain with the given head. If head is None,
+    // uses the current chain head. If head is not found, returns None.
```
peterschwarz: This comment is incorrect, but should be fixed in the associate PR against master.

PR #1946, opened 2018-11-12
Properly version settings-tp package

rbuysse:


families/settings/Dockerfile-installed-xenial
```diff
@@ -25,9 +25,11 @@ RUN apt-get update \
  && apt-get install -y \
  curl \
  gcc \
+ git \
```
peterschwarz: Why is Git needed? Is it used in a script not part of this PR?

rbuysse: The bin/get_version script runs `git describe` to determine the real version of the repository.

PR #1948, opened 2018-11-14
Increment validator's Cargo.toml version

rbuysse:


rbuysse:
The version in validator's Cargo.toml must match the version that the repository is tagged, otherwise the validator build will fail. 

Validator is an exception in this regard because it runs setup.py after the Cargo.toml version has been modified.

PR #1949, opened 2018-11-16
Verify that dependencies are not empty strings

boydjohnson:
Transaction dependencies that are empty strings cause this error to be raised in the completer, prior to this fix.

```packages/sawtooth_validator/journal/completer.py", line 368, in handle
sawtooth-validator |     self._completer.add_batch(batch)
sawtooth-validator |   File "/usr/lib/python3/dist-packages/sawtooth_validator/journal/completer.py", line 300, in add_batch
sawtooth-validator |     if self._complete_batch(batch):
sawtooth-validator |   File "/usr/lib/python3/dist-packages/sawtooth_validator/journal/completer.py", line 226, in _complete_batch
sawtooth-validator |     dependency):
sawtooth-validator |   File "/usr/lib/python3/dist-packages/sawtooth_validator/journal/block_store.py", line 308, in has_transaction
sawtooth-validator |     return self._block_store.contains_key(txn_id, index='transaction')
sawtooth-validator |   File "/usr/lib/python3/dist-packages/sawtooth_validator/database/indexed_database.py", line 122, in contains_key
sawtooth-validator |     return txn.cursor().set_key(key.encode())
sawtooth-validator | lmdb.BadValsizeError: mdb_cursor_get: MDB_BAD_VALSIZE: Unsupported size of key/DB name/data, or wrong DUPFIXED size```

### DIFF
--- a/validator/sawtooth_validator/gossip/structure_verifier.py
+++ b/validator/sawtooth_validator/gossip/structure_verifier.py
@@ -19,6 +19,7 @@ from sawtooth_validator.protobuf import client_batch_submit_pb2
 from sawtooth_validator.protobuf.batch_pb2 import BatchHeader
 from sawtooth_validator.protobuf.block_pb2 import BlockHeader
 from sawtooth_validator.protobuf import network_pb2
+from sawtooth_validator.protobuf import transaction_pb2
 from sawtooth_validator.networking.dispatch import Handler
 from sawtooth_validator.networking.dispatch import HandlerResult
 from sawtooth_validator.networking.dispatch import HandlerStatus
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: validator/sawtooth_validator/gossip/structure_verifier.py [hunk 1]
```diff
@@ -72,10 +73,33 @@ def is_valid_batch(batch):
                          "order of transactions in the batch: %s txn: %s",
                          batch.header_signature, header_txn_id)
             return False
+        txn_header = _deserialize_txn_header(txn.header)
+        for dep in txn_header.dependencies:
+            if not dep:
```
arsulegai: How about also verifying if the id here is not just random string but has right set of characters we expect in a signature?

validator/sawtooth_validator/gossip/structure_verifier.py
```diff
@@ -72,10 +73,33 @@ def is_valid_batch(batch):
...
+                LOGGER.debug("Malformed dependency in batch: %s, txn:%s",
```
ltseeley: This method could probably be more descriptive; this really just reports when a dependency is empty, not any other malformed case.

peterschwarz:
Closing in favor of #2051

PR #1968, opened 2018-12-07
Rust Lint update for stable rustfmt 1-1

agunde406:


agunde406:
Closing because rustfmt has been pinned  in 1-1 to a previous version and does not need these updates

PR #1984, opened 2018-12-19
1-1 BACKPORT: Allow subscribers to connect when no Genesis Block

peterschwarz:
This change handles the case when an event subscriber connects to the
validator before it has either generated or received the genesis block.
In that instance, chain head is `None`, and an exception is raised, then
the subscriber is left in an invalid state. This fix mitigates that
issue.

### DIFF
--- a/validator/tests/test_events/tests.py
+++ b/validator/tests/test_events/tests.py
@@ -17,6 +17,7 @@
 
 import unittest
 from unittest.mock import Mock
+from unittest.mock import MagicMock
 from uuid import uuid4
 
 from sawtooth_validator.database.dict_database import DictDatabase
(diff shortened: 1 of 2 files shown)


validator/tests/test_events/tests.py
```diff
@@ -393,6 +394,41 @@ def test_broadcast_events(self):
             validator_pb2.Message.CLIENT_EVENTS,
             event_list, connection_id="test_conn_id", one_way=True)
 
+    def test_catchup_subscriber(self):
+        """Test that catch subscriber handles the case of:
+        - no blocks (i.e. the genesis block has not been produced or received
```
jsmitchell: end paren

peterschwarz: Probably not going to correct that here, but in master

PR #1988, opened 2019-01-02
Fix Python SDK's activation procedure

ltseeley:
Fix the Python SDK's `ZmqDriver._wait_until_active` method to properly
wait for the activation message. Fixes a bug that caused the engine to
never actually process the activation.

### DIFF
--- a/sdk/python/sawtooth_sdk/consensus/zmq_driver.py
+++ b/sdk/python/sawtooth_sdk/consensus/zmq_driver.py
@@ -133,9 +133,9 @@ class ZmqDriver(Driver):
                 'Registration failed with status {}'.format(response.status))
 
     def _wait_until_active(self):
+        future = self._stream.receive()
         while True:
             try:
-                future = self._stream.receive()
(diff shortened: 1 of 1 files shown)


sdk/python/sawtooth_sdk/consensus/zmq_driver.py
```diff
@@ -133,9 +133,9 @@ def _register(self):
                 'Registration failed with status {}'.format(response.status))
 
     def _wait_until_active(self):
+        future = self._stream.receive()
         while True:
             try:
-                future = self._stream.receive()
                 message = future.result(1)
             except concurrent.futures.TimeoutError:
                 continue
```
arsulegai: Please correct if I understood it wrong, is the issue related to future variable being overwritten by a new received message while it was still waiting for earlier future response to complete?

peterschwarz: That is correct

PR #1989, opened 2019-01-03
Change default scheduler from serial to parallel

jsmitchell:
Changes the default scheduler from serial to parallel in the validator, example config files, docker compose files, and docs.


docs/source/app_developers_guide/creating_sawtooth_network.rst
```diff
@@ -90,7 +90,7 @@ The following figure shows an example network with two validator nodes:
    :align: center
    :alt: Docker: Sawtooth network with five nodes
 
-Like the single-node environment, this environment uses serial transaction
+Like the single-node environment, this environment uses parallel transaction
```
chenette: @jsmitchell - It looks like the Docker, Kubernetes, and AWS environments are still using serial processing. Is that on purpose?  For example, this PR doesn't change `docker/compose/sawtooth-default.yaml`, which is used by the single-node Docker procedure in https://github.com/hyperledger/sawtooth-core/blob/master/docs/source/app_developers_guide/docker.rst 

As a result, this change is incorrect - the Docker multi-node environment is NOT the same as the Docker single-node environment. The same is true for Kubernetes and AWS; only the Ubuntu procs use parallel processing for both the single- and multi-node environments. (However, there is no AWS multi-node procedure, and the Kubernetes multi-node proc cleverly omits this sentence so it's not an issue there.)

chenette: On the other hand, if the omission is on purpose (if we want the single-node environments to use the now-non-default serial processing), I suggest that this PR add a sentence about that situation in the Docker, AWS, and Kubernetes single-node procedures.

jsmitchell: I explicitly didn't change the AWS docs, because they refer to a built version in the marketplace. I don't believe that is updated with any regularity. 

The first commit in this PR changes the default from serial to parallel in the event that the `--scheduler` option is not provided at the command line. Since `docker/compose/sawtooth-default.yaml` doesn't provide a `--scheduler` argument, the parallel scheduler will be used.

jsmitchell: Fixed

PR #1991, opened 2019-01-07
Fix run docker test

mberkau:
When I attempted to follow the steps in BUILD.md, _run_tests_ in Step Six failed. I traced the problem down to three issues -- two of which this pull request fixes.

One, in _run_docker_test_, the call to _docker inspect_ was not being passed the actual container names created by _docker-compose_. _Docker-compose_ appends a hash value to container names created by the testing scripts (e.g. latest_test-tp-settings_1_47b754af16ac), and these full names were not being passed. The simplest way to get these full container names is to parse stdout from _docker-compose_, so _run_docker_test_ was modified to do that parsing and supply the correct names in calls to _docker inspect_.

Two, the following two docker files had an option passed to _apt-get install_ in the wrong location and that was also causing the _run_tests_ process to fail. These files were edited to have the option in the correct location.
- cli-tests.dockerfile
- python-sdk-tests.dockerfile

Three, the _run_tests_ process intermittently hangs when running the unit-python-sdk tests due to an asyncio error -- sometimes the testing process completes fine, and other times it hangs. Calls to _docker-compose -p latest -f ./sdk/python/tests/unit_python_sdk.yaml up --abort-on-container-exit_ were made manually to elicit a successful and failed test result, and these results are captured in the following file.
 - [unit-python-sdk-test_results.txt](https://github.com/hyperledger/sawtooth-core/files/2729934/unit-python-sdk-test_results.txt)

This third issue is more than I can address at this time, so I've have opened bug STL-1474 on Hyperledger Sawtooth's Jira site (https://jira.hyperledger.org/browse/STL-1474).


mberkau:
I've pulled the pull request because the errors seem to be linked to an issue with my test environment. I created a new test environment and the original/unaltered files work fine. I do still intermittently get the asyncio error noted above, even in a new clean environment, so I'm leaving STL-1474 open.

PR #1994, opened 2019-01-09
Remove invalid batches from pending batches pool

transparent-mitch:
When a block publisher cancels a block, it will now sort out any
invalid batches from the candidate block and remove them from
the pending batch pool.

This included:
 -  Adding pub invalid_batches: Vec<Batch> field to CandidateBlock
 -  Moving injected batches only check to after batch sorting in
    CandidateBlock.summarize()
 -  Adding code to SyncBlockPublisher.cancel_block() to remove
    any invalid transactions recorded by the candidate block
    from state.pending_batches before canceling the block

### DIFF
--- a/validator/src/journal/candidate_block.rs
+++ b/validator/src/journal/candidate_block.rs
@@ -70,6 +70,8 @@ pub struct CandidateBlock {
     pending_batch_ids: HashSet<String>,
     injected_batch_ids: HashSet<String>,
 
+    invalid_batch_ids: HashSet<String>,
+
     committed_txn_cache: TransactionCommitCache,
 }
 
@@ -103,6 +105,8 @@ impl CandidateBlock {
             pending_batches: vec![],
             pending_batch_ids: HashSet::new(),
             injected_batch_ids: HashSet::new(),
+
+            invalid_batch_ids: HashSet::new(),
         }
     }
 
@@ -382,11 +386,6 @@ impl CandidateBlock {
         let mut bad_batches = vec![];
         let mut pending_batches = vec![];
 
-        if self.injected_batch_ids == valid_batch_ids {
-            // There only injected batches in this block
-            return Ok(None);
-        }
-
         for batch in self.pending_batches.clone() {
             let header_signature = &batch.header_signature.clone();
             if batch.trace {
@@ -431,10 +430,17 @@ impl CandidateBlock {
                     committed_txn_cache.add_batch(&batch.clone());
                 }
             } else {
+                self.invalid_batch_ids.insert(batch.header_signature.clone());
                 bad_batches.push(batch.clone());
                 debug!("Batch {} invalid, not added to block", header_signature);
             }
         }
+
+        if self.injected_batch_ids == valid_batch_ids {
+            // There only injected batches in this block
+            return Ok(None);
+        }
+
         if execution_results.ending_state_hash.is_none() || self.no_batches_added(&builder) {
             debug!("Abandoning block, no batches added");
             return Ok(None);
@@ -530,5 +536,13 @@ impl CandidateBlock {
         } else {
             Err(CandidateBlockError::BlockEmpty)
         }
+    }
+
+    pub fn has_invalid_batches(&self) -> bool {
+        return !self.invalid_batch_ids.is_empty();
+    }
+
+    pub fn get_invalid_batch_ids(&self) -> &HashSet<String> {
+        return &self.invalid_batch_ids;
     }
 }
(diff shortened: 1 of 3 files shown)


peterschwarz:
@transparent-mitch Looks like you have some lint errors in the code: You need to run `cargo fmt` on your changes.

peterschwarz:
You should just squash that fix lint errors commit, since you are fixing an error introduced by the first commit.

validator/src/journal/candidate_block.rs
```diff
@@ -431,10 +428,17 @@ impl CandidateBlock {
                     committed_txn_cache.add_batch(&batch.clone());
                 }
             } else {
+                self.invalid_batches.push(batch.clone());
```
peterschwarz: Isn't the contents of this just the `bad_batches` vector?  If so, it seems like you are missing the above case when dependencies have failed.

validator/src/journal/publisher.rs
```diff
@@ -479,6 +479,20 @@ impl SyncBlockPublisher {
     }
 
     fn cancel_block(&self, state: &mut BlockPublisherState, unref_block: bool) {
+        let invalid_batches = if let Some(ref cb) = state.candidate_block {
+            cb.invalid_batches.clone()
```
peterschwarz: This could probably be done with out the clone, give that you don't need to pass ownership of those invalid batches.

validator/src/journal/candidate_block.rs
peterschwarz: Additionally, you could just store the batch header signature, and not the whole batch (which could be sizeable).

transparent-mitch: I was hesitant to use the `bad_batches` vector _because_ of the dependencies. I could imagine a case where you'd want to retry the batches with unmet dependencies but not those that are invalid for any other reason.

Just using headers is definitely more efficient. Good catch.

validator/src/journal/publisher.rs
transparent-mitch: You're probably right. It was a tiny bit hairy with state owning both the candidate_block and the pending_batches. This was a sure-fire way to get it done.

jsmitchell:
> More discussion is needed about how summarize_block and cancel block need to be called.

@boydjohnson @peterschwarz This seems like as good a place as any to have that discussion.

arsulegai:
@transparent-mitch this is a good start, however it works for the cases where validator would build the block. Let's also consider the cases where validator would not start building block, in Raft for example follower validator nodes don't build unless they get to be leader.

transparent-mitch:
@arsulegai When else would you check if a txn is valid? Are you suggesting the batches get sent to the TP before they are added to the block?

We've actually tried modifying Raft to have the followers build blocks as well, just to purge the invalid batches as quickly as possible. Seems to work for now. That just could be left up to the consensus designer's discretion... in which case it might make sense to just have a `purge_invalid_txns` service instead of including it in `cancel_block`.

arsulegai:
Call to build block from consensus engine for the purpose of purging invalid transactions seems workaround for me. There's scope of writing RFC and have elaborate discussion for solution. As you mentioned having a purge service is still better option than tying purge operation with cancel_block.

In addition to it improvisations can be applied such that this service will purge only those services which are complete (i.e. all dependencies are available), ready to be scheduled, but it never got added to any block recently (i.e. ``X`` number of blocks were published but this transaction was never included in them).

Given that we may explore and end up having more better ideas with RFC.

danintel:
This might fix https://jira.hyperledger.org/browse/STL-1477

ryjones:
@vaporos @peterschwarz any progress on this? looks like it got merged into a dead branch. take a look - thanks

validator/src/journal/candidate_block.rs
```diff
@@ -431,10 +430,17 @@ impl CandidateBlock {
                     committed_txn_cache.add_batch(&batch.clone());
                 }
             } else {
+                self.invalid_batch_ids.insert(batch.header_signature.clone());
```
arsulegai: Looks like line 413 to line 415 intended to remove bad batches from the pending batches?

jaromil:
Hi! is there any chance to backport this patch to the 1.2.x series so that, meanwhile the purge service RFC gets drafted and debated, we can have a working setup that does not loop over invalidtransactions? I think is worth considering while the 1.99 branch goes off a major refactoring.

peterschwarz:
This looks to have conflicts, and the PR may need to be re-opened against [hyperledger/sawtooth-lib](https://github.com/hyperledger/sawtooth-lib), as the journal code has moved there.

Otherwise, it should be re-opened against the 1-3

charalarg:
Hello everyone, i cloned the https://github.com/hyperledger/sawtooth-core/tree/1-2 branch of sawtooth-core repo, and made the changes in the two files referenced in this pr: Remmeauth/sawtooth-core#5. So building the branch with these changes, as @eugene-babichenko mentioned solves the issue in the 1.2.x sawtooth series.

PR #1997, opened 2019-01-10
Add Missing Dev Guide Ubuntu Instructions for Devmode Consensus Engine

danintel:
* Install devmode engine
* Start devmode engine
* Install and CLI names vary by Sawtooth release

### DIFF
--- a/docs/source/app_developers_guide/ubuntu.rst
+++ b/docs/source/app_developers_guide/ubuntu.rst
@@ -9,7 +9,7 @@ then walks you through the following tasks:
  * Generating a user key
  * Creating the genesis block
  * Generating a root key
- * Starting the components: validator, REST API, and
+ * Starting the components: validator, consensus engine, REST API, and
    transaction processors
  * Checking the status of the REST API
  * Using Sawtooth commands to submit transactions, display block data, and view
@@ -52,7 +52,7 @@ Settings is a reference implementation. In a production environment, you should
 always run a transaction processor that supports the Settings transaction
 family.
 
-In this procedure, you will open six terminal windows on your host system: one
+In this procedure, you will open seven terminal windows on your host system: one
 for each Sawtooth component and one to use for client commands.
 
 .. note::
@@ -113,6 +113,13 @@ stable or nightly.  We recommend using the stable repository.
    .. code-block:: console
 
       user@validator$ sudo apt-get install -y sawtooth
+
+#. Install the Sawtooth Devmode consensus engine package. Run the following
+   command:
+
+   .. code-block:: console
+
+       user@validator$ sudo apt-get install sawtooth-devmode-engine-rust
 
 #. Any time after installation, you can view the installed Sawtooth packages
    with the following command:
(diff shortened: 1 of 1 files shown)


docs/source/app_developers_guide/ubuntu.rst
```diff
@@ -52,7 +52,7 @@ Settings is a reference implementation. In a production environment, you should
 always run a transaction processor that supports the Settings transaction
 family.
 
-In this procedure, you will open six terminal windows on your host system: one
+In this procedure, you will open seven terminal windows on your host system: one
 for each Sawtooth component and one to use for client commands.
 
 .. note::
```
arsulegai: How about adding a note about starting order for consensus engine?
It can be started before validator as well, unless otherwise explicitly specified by particular consensus algorithm.

danintel: Thank you for the review.

Yes, the consensus engine can be started before, but then it doesn't connect and you don't see a connection message until after the validator started, making it harder to see problems.

Also, the consensus among Sawtooth developers is to keep this guide short and simple.

I added a note to the Sawtooth FAQ about this.

danintel:
Did a `git push -f` to see if CI builds are working now.

docs/source/app_developers_guide/ubuntu.rst
```diff
@@ -248,7 +255,34 @@ complete this procedure.
 
 .. _start-rest-api-label:
```
agunde406: look like you need to remove the label here since you redefine it below. It should probably be changed to a label for Start the DevMode Consensus Engine

danintel: Fixed. I renamed it `_start-devmode-consensus-label:`
Thanks for looking at this.

docs/source/app_developers_guide/ubuntu.rst
```diff
@@ -114,6 +114,13 @@ stable or nightly.  We recommend using the stable repository.
 
       user@validator$ sudo apt-get install -y sawtooth
 
+#. Install the Sawtooth DevMode consensus engine package. Run the following
```
chenette: The rest of the Sawtooth documentation uses "dev mode" as two words (without caps unless at the start of a sentence or in a heading).  However, this is a minor nitpick.

chenette: P.S. I noticed that the info message below says "Devmode 1.0", so one could make the case for writing "Devmode" as one word, as long as one doesn't capitalize the "M".

chenette:
@danintel - The build is failing because there's an RST syntax error on line 443. The `=====...=` line needs another `=` because  "Step 9" changed to "Step 10".

`/project/sawtooth-core/docs/source/app_developers_guide/ubuntu.rst:443: WARNING: Title underline too short.`

`Step 10: Use Sawtooth Commands as a Client`
`=========================================`

danintel:
Chenette,
Thanks!--I fixed it. I couldn't find the error but now I found the proper click sequence to drill down to it.

docs/source/app_developers_guide/ubuntu.rst
danintel: I changed it to `Devmode` as suggested. It will restart the 10 quatzillion hour build, but it still needs another reviewer in any case.

PR #1998, opened 2019-01-11
Consensus fixes

peterschwarz:
This PR has several minor fixes for various consensus proxy methods, as well as the client handlers (which has an effect on the PoET engine). 

~Probably more controversially, this removes the `panic` when a block is unref'ed in the `BlockManager` and it is already at `0`.  This panic makes the system quite unstable as a validator will crash in this situation.~

Updated to improve the panic message (and remove the controversy for a future PR).

### DIFF
--- a/validator/sawtooth_validator/consensus/proxy.py
+++ b/validator/sawtooth_validator/consensus/proxy.py
@@ -164,7 +164,10 @@ class ConsensusProxy:
         if chain_head is None:
             return None
 
-        return self._chain_controller.forks(chain_head.header_signature)
+        try:
+            return self._chain_controller.forks(chain_head.header_signature)
+        except KeyError:
+            return None
 
     # Using blockstore and state database
     def blocks_get(self, block_ids):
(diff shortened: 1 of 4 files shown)


validator/sawtooth_validator/consensus/proxy.py
```diff
@@ -189,8 +192,18 @@ def settings_get(self, block_id, settings):
...
+        try:
+            settings_view = self._settings_view_factory.create_settings_view(
+                block_header.state_root_hash)
+        except KeyError:
+            LOGGER.error(
+                'Settings from block %s requested, but root hash %s was '
+                'missing. Returning no setting values.',
+                block_id.hex(),
+                block_header.state_root_hash)
+            # The state root does not exists, which may indicate a pruned root
```
jsmitchell: s/exists/exist/

validator/sawtooth_validator/consensus/proxy.py
```diff
@@ -210,8 +223,18 @@ def state_get(self, block_id, addresses):
...
+        try:
+            state_view = self._state_view_factory.create_view(
+                block_header.state_root_hash)
+        except KeyError:
+            LOGGER.error(
+                'State from block %s requested, but root hash %s was missing. '
+                'Returning empty state.',
+                block_id.hex(),
+                block_header.state_root_hash)
+            # The state root does not exists, which may indicate a pruned root
```
jsmitchell: s/exists/exist/

peterschwarz:
I don't have any theories yet.  I think the ref's and unrefs need to be mapped out in such a way that we can follow the different threads that make these changes.  What's likely is that a component is unref'ing a block that didn't grab a reference in the first place or unref'ed twice.

PR #2001, opened 2019-01-17
Revert "Remove init of pylogger.rs from python code"

peterschwarz:
This reverts commit ba9517ec0937e0674ee561fdd6ee79d31d004f8c.

This restores proper logging in the rust code.  With the removal of this
init of pylogger, log messages from the lower layer (i.e. libsawtooth)
are not included in the logs.  This makes debugging difficult.

### DIFF
--- a/validator/src/pylogger.rs
+++ b/validator/src/pylogger.rs
@@ -40,6 +40,16 @@ pub fn set_up_logger(verbosity: u64, py: Python) {
     warn!("Started logger at level {}", verbosity_level);
 }
 
+#[no_mangle]
+#[allow(unused)]
+#[allow(private_no_mangle_fns)]
+pub extern "C" fn pylogger_init(verbosity: usize) {
+    let gil = Python::acquire_gil();
+    let py = gil.python();
+    PyLogger::init(determine_log_level(verbosity as u64), py)
+        .expect("Failed to initialize PyLogger");
+}
+
 pub fn exception(py: Python, msg: &str, err: PyErr) {
     let logger = PyLogger::new(py).expect("Failed to create new PyLogger");
     logger.exception(py, msg, err);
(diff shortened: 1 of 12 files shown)


peterschwarz:
Added a commit to fix rust lint

peterschwarz:
Added a commit to fix the doc issues

validator/src/pylogger.rs
```diff
@@ -42,12 +42,12 @@ pub fn set_up_logger(verbosity: u64, py: Python) {
...
 #[no_mangle]
 #[allow(unused)]
-#[allow(private_no_mangle_fns)]
 pub extern "C" fn pylogger_init(verbosity: usize) {
     let gil = Python::acquire_gil();
     let py = gil.python();
-    PyLogger::init(determine_log_level(verbosity as u64), py)
-        .expect("Failed to initialize PyLogger");
+    if let Err(_) = PyLogger::init(determine_log_level(verbosity as u64), py) {
+        warn!("Attempted to initialize logger twice; ignoring");
```
vaporos: Is that the only way PyLogger::init() can fail? If not, should the error text be logged?

validator/src/pylogger.rs
```diff
@@ -68,7 +68,8 @@ impl PyLogger {
     }
 
     fn init(verbosity: Level, py: Python) -> Result<(), SetLoggerError> {
-        let logger = PyLogger::new(py).unwrap();
+        let logger =
+            PyLogger::new(py).expect("Failed to instantiate Python logger; check library paths.");
```
vaporos: Avoid the period here please.

validator/src/pylogger.rs
peterschwarz: Yes, it only returns the `SetLoggerError` which is only returned if a logger has been set twice.

peterschwarz: See: https://docs.rs/log/0.4.6/log/struct.SetLoggerError.html

PR #2002, opened 2019-01-20
Don't bind to specific address

ConnorChristie:


peterschwarz:
Do we have a link to the security report?

ConnorChristie:
Would it be okay if the kubernetes deployment bound the instance to 0.0.0.0 by default instead of changing the application default? I specifically ran into this issue when using Kubernetes, trying to expose the port over a load balancer

rberg2:
In Kubernetes I run the rest-api list this: `sawtooth-rest-api -C tcp://$HOSTNAME:4004 -B 0.0.0.0:8008` then I control access to that api endpoint by using a containerPort. If you wish to expose the rest-api to the world you should follow the [Guide](https://sawtooth.hyperledger.org/docs/core/releases/latest/sysadmin_guide/rest_auth_proxy.html) that outline how to put an authenticating web server with SSL in front of the API endpoint.

ConnorChristie:
That's what I ended up doing but initially I had no clue why I wasn't able to connect through to port 8008 and only 4004 (4004 is bound to 0.0.0.0 by default). That document doesn't have anything about binding to 0.0.0.0 instead of localhost, does it?

rberg2:
The doc outlines how you would have the rest-api listen on 127.0.0.1 and a webserver listen externally in a secure way that you will not get if you put the rest-api directly on an external network interface.

danintel:
I agree about binding to localhost (127.0.0.1) instead of all (0.0.0.0). It can easily be configured in `/etc/sawtooth/validator.toml`  or `rest-api.toml` or the `sawtooth-rest-api` and `sawtooth-validator` command line options.

I didn't see it mentioned in the Sawtooth Security Audit Report though:
https://wiki.hyperledger.org/_media/security/technical_report_linux_foundation_hyperledger_december_2017_v1.0.pdf

peterschwarz:
This PR should probably be closed

PR #2016, opened 2019-02-04
Downgrade pylint dependency, typed_ast

boydjohnson:
Due to changes to a pylint dependency, Python code that uses await and
async as method names, such as asyncio.await, as in Python3.4 fails to
parse with a syntax error. This was causing pyzmq imports to fail during
linting but not during runtime.


boydjohnson:
Closing this in favor of turning off the error, as in 1-1.

PR #2019, opened 2019-02-07
Use new handler for consensus gossip messages

ltseeley:
Instead of using the `GossipBroadcastHandler` to send peer messages to the consensus engine, use a new handler `GossipConsensusHandler`. Notifying the consensus engine isn't a "broadcast" function, so it should have a different handler. Now, the `GossipBroadcastHandler` will simply pass when it encounters a peer message.

In addition to being more correct and accurate from an understandability perspective, this change allows for correct behavior of the time-to-live (TTL) functionality for gossip. Without this change, if TTL is set to <= 1, consensus messages will never be delivered to the consensus engine. With this change, TTL can be set to any value and consensus messages will be unaffected.

### DIFF
--- a/validator/sawtooth_validator/gossip/gossip_handlers.py
+++ b/validator/sawtooth_validator/gossip/gossip_handlers.py
@@ -220,11 +220,26 @@ class GossipBatchResponseHandler(Handler):
         return self._responder.get_request(batch_id)
 
 
+class GossipConsensusHandler(Handler):
+    def __init__(self, gossip, notifier):
+        self._gossip = gossip
+        self._notifier = notifier
+
+    def handle(self, connection_id, message_content):
+        obj, tag, _ = message_content
+
+        if tag == GossipMessage.CONSENSUS:
+            self._notifier.notify_peer_message(
+                message=obj,
+                sender_id=bytes.fromhex(
+                    self._gossip.peer_to_public_key(connection_id)))
+        return HandlerResult(status=HandlerStatus.PASS)
+
+
 class GossipBroadcastHandler(Handler):
-    def __init__(self, gossip, completer, notifier):
+    def __init__(self, gossip, completer):
         self._gossip = gossip
         self._completer = completer
-        self._notifier = notifier
 
     def handle(self, connection_id, message_content):
         obj, tag, ttl = message_content
@@ -246,10 +261,8 @@ class GossipBroadcastHandler(Handler):
             if not self._completer.get_block(obj.header_signature):
                 self._gossip.broadcast_block(obj, exclude, time_to_live=ttl)
         elif tag == GossipMessage.CONSENSUS:
-            self._notifier.notify_peer_message(
-                message=obj,
-                sender_id=bytes.fromhex(
-                    self._gossip.peer_to_public_key(connection_id)))
+            # We do not want to forward consensus messages
+            pass
         else:
             LOGGER.info("received %s, not BATCH or BLOCK or CONSENSUS", tag)
         return HandlerResult(status=HandlerStatus.PASS)
--- a/validator/sawtooth_validator/server/network_handlers.py
+++ b/validator/sawtooth_validator/server/network_handlers.py
@@ -37,6 +37,7 @@ from sawtooth_validator.gossip.permission_verifier import \
 from sawtooth_validator.gossip.permission_verifier import \
     NetworkConsensusPermissionHandler
 
+from sawtooth_validator.gossip.gossip_handlers import GossipConsensusHandler
 from sawtooth_validator.gossip.gossip_handlers import GossipBroadcastHandler
 from sawtooth_validator.gossip.gossip_handlers import \
     GossipMessageDuplicateHandler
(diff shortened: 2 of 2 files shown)


validator/sawtooth_validator/gossip/gossip_handlers.py
```diff
@@ -220,11 +220,26 @@ def _has_open_requests(self, batch_id):
...
+        self._gossip = gossip
+        self._notifier = notifier
+
+    def handle(self, connection_id, message_content):
+        obj, tag, _ = message_content
+
+        if tag == GossipMessage.CONSENSUS:
+            self._notifier.notify_peer_message(
+                message=obj,
+                sender_id=bytes.fromhex(
```
arsulegai: [Open Ended Question]: I don't know if these are thought through, but here are concerns around consensus peer messages
1. It should be possible to set priority of consensus peer messages through consensus engine. Which would imply apart from handling gossip messages with priority at dispatcher queue of same validator, it should be possible for other validators to process these messages faster.
2. Ability to have dynamic peering of validators, yet have consensus engine be able to talk to a peer consensus engine not directly connected. Would it be feasible to add peer info in message itself and then rely on gossip network to carry consensus peer message to destined consensus engine.

ltseeley: Those are very good questions. Just to clarify, for (1): do you mean that the consensus engine should be able to decide how high of a priority the consensus messages are? And that this means consensus messages should be separate from Gossip messages so that their priority is independent? And for (2): are you suggesting that it would be useful for some networks using a consensus engine that doesn't require full peering (or maybe full peering isn't possible) to have the broadcast functionality to relay messages to nodes that may not be directly connected? And maybe they can do it intelligently (not broadcast to everyone, just send directly to the intended peer)?

arsulegai: 1. Yes, consensus messages should be independent of other gossip messages validators exchange. This way few consensus messages can be handled with priority where immediate action is required.
2. Yes.

ltseeley: These are good ideas that are worth exploring. We're already talking about (and considering an RFC for) some way for the consensus engine to give the validator hints about how to handle messaging. We'll be sure to include these ideas!

validator/sawtooth_validator/gossip/gossip_handlers.py
```diff
@@ -246,10 +261,8 @@ def handle(self, connection_id, message_content):
             if not self._completer.get_block(obj.header_signature):
                 self._gossip.broadcast_block(obj, exclude, time_to_live=ttl)
         elif tag == GossipMessage.CONSENSUS:
-            self._notifier.notify_peer_message(
-                message=obj,
-                sender_id=bytes.fromhex(
-                    self._gossip.peer_to_public_key(connection_id)))
+            # We do not want to forward consensus messages
```
jsmitchell: Is it true that we don't want to forward consensus messages? We need to be careful here not to apply assumptions about a fully connected network to the general behavior.

ltseeley: It's been a simplifying assumption we've made up to this point because all of the engines we've implemented that use consensus messages are supposed to be fully peered (AFAIK). But it does seem like a bad assumption to make in general.

ltseeley:
> Should we go farther and make consensus messages their own thing instead of a type of gossip message? I can see pros and cons, so I'm asking the question rather than taking a position.

This is a good question. On the one hand there are advantages to handling consensus messages separately from blocks/batches since they do serve a different purpose (can give consensus messages priority, can determine broadcasting rules independently). But on the other hand, they are related in that they're both impacted by the shape of the network. I don't know that I have a strong enough opinion yet.

validator/sawtooth_validator/gossip/gossip_handlers.py
manojgop: POET is not fully peered. But POET doesn't use consensus messages now

validator/sawtooth_validator/server/network_handlers.py
```diff
@@ -242,6 +243,13 @@ def add(
         ),
         thread_pool)
 
+    # GOSSIP_MESSAGE ) Determines if this is a consensus message and notifies
+    # the consensus engine if it is
+    dispatcher.add_handler(
+        validator_pb2.Message.GOSSIP_MESSAGE,
+        GossipConsensusHandler(gossip=gossip, notifier=consensus_notifier),
+        thread_pool)
+
```
manojgop: Is it possible to assign high priority of consensus message handling. This might be required since consensus messages (eg: in case of Raft) needs to be processed at higher priority than rest of Gossip messages (block, batch) to make progress in the validator and to publish blocks sooner. For eg: In case of Raft, till all the followers gets the log entry, leader cannot commit the block. Similarly consensus messages like heart beat, election related messages has to be processed sooner compared to batch/block gossip messages. May be it might make sense to introduce a new message type like GOSSIP_CONSENSUS_MESSAGE and to set its priority high.

peterschwarz: This is a good point.

PR #2027, opened 2019-02-23
Add 2 methods for creating private key to signing contexts

nickylimjj:
Creating methods to create a private key from hex and bytes to signing 
contexts. 

Added 2 unit tests alongside.

Resolves: #STL-1231

Signed-off-by: nickylimjj <nickylimjj.nl@gmail.com>

### DIFF
--- a/signing/sawtooth_signing/core.py
+++ b/signing/sawtooth_signing/core.py
@@ -129,6 +129,22 @@ class Context(metaclass=ABCMeta):
         """
 
     @abstractmethod
+    def private_key_from_bytes(self, bytes):
+        """Generates a private key from bytes using this context.
+
+        Returns:
+            (:obj:`PrivateKey`): a privatekey
+        """
+
+    @abstractmethod
+    def private_key_from_hex(self, hex):
+        """Generates a private key from hex using this context.
+
+        Returns:
+            (:obj:`PrivateKey`): a privatekey
+        """
+
+    @abstractmethod
     def get_public_key(self, private_key):
         """Produce a public key for the given private key.
 
--- a/signing/tests/test_secp256k1_signer.py
+++ b/signing/tests/test_secp256k1_signer.py
@@ -32,6 +32,12 @@ KEY2_PRIV_HEX = \
 KEY2_PUB_HEX = \
     "039c20a66b4ec7995391dbec1d8bb0e2c6e6fd63cd259ed5b877cb4ea98858cf6d"
 
+KEY1_PRIV_BYTES = bytes.fromhex(KEY1_PRIV_HEX)
+KEY1_PUB_BYTES = bytes.fromhex(KEY1_PUB_HEX)
+
+KEY2_PRIV_BYTES = bytes.fromhex(KEY2_PRIV_HEX)
+KEY2_PUB_BYTES = bytes.fromhex(KEY2_PUB_HEX)
+
 MSG1 = "test"
 MSG1_KEY1_SIG = ("5195115d9be2547b720ee74c23dd841842875db6eae1f5da8605b050a49e"
                  "702b4aa83be72ab7e3cb20f17c657011b49f4c8632be2745ba4de79e6aa0"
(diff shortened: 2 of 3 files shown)


signing/tests/test_secp256k1_signer.py
```diff
@@ -71,6 +77,16 @@ def test_priv_to_public_key(self):
         public_key2 = context.get_public_key(priv_key2)
         self.assertEqual(public_key2.as_hex(), KEY2_PUB_HEX)
 
+    def test_creating_private_keys_from_hex_and_bytes_to_contexts(self):
+        context = create_context("secp256k1")
+        priv_key = context.private_key_from_hex(KEY1_PRIV_HEX)
+        self.assertEqual(context.get_public_key(priv_key).as_hex(),
+                            KEY1_PUB_HEX)
```
delventhalz: This spacing is a little weird. Should line up with the 'c' in `context`.

signing/tests/test_secp256k1_signer.py
```diff
@@ -71,6 +77,16 @@ def test_priv_to_public_key(self):
         public_key2 = context.get_public_key(priv_key2)
         self.assertEqual(public_key2.as_hex(), KEY2_PUB_HEX)
 
+    def test_creating_private_keys_from_hex_and_bytes_to_contexts(self):
+        context = create_context("secp256k1")
+        priv_key = context.private_key_from_hex(KEY1_PRIV_HEX)
+        self.assertEqual(context.get_public_key(priv_key).as_hex(),
```
delventhalz: This is a bit of a roundabout way to test the PrivateKey was created correctly. I think you can cut the public key and just use the private key's `as_hex` method directly, like in this test above:
https://github.com/hyperledger/sawtooth-core/blob/4c16158fea7de87b066ea550ee8f93b4e03a1b2e/signing/tests/test_secp256k1_signer.py#L56

signing/sawtooth_signing/core.py
```diff
@@ -128,6 +128,22 @@ def new_random_private_key(self):
...
+    @abstractmethod
+    def private_key_from_bytes(self, bytes):
+        """Generates a new random PrivateKey using this context.
+
+        Returns:
+            (:obj:`PrivateKey`): a privatekey
+        """
+
+    @abstractmethod
+    def private_key_from_hex(self, hex):
```
delventhalz: While I think the at least the `from_bytes` method is absolutely necessary, there is some debate about whether or not hex is too opinionated to use. My inclination is to keep it, but I am flagging this because it has been an area of discussion in the past.

signing/sawtooth_signing/core.py
```diff
@@ -128,6 +128,22 @@ def new_random_private_key(self):
             (:obj:`PrivateKey`): a random private key
         """
 
+    @abstractmethod
+    def private_key_from_bytes(self, bytes):
+        """Generates a new random PrivateKey using this context.
```
delventhalz: Let's update these docstrings so they aren't just a copy/paste of `new_random_private_key`

signing/tests/test_secp256k1_signer.py
```diff
@@ -71,6 +77,14 @@ def test_priv_to_public_key(self):
         public_key2 = context.get_public_key(priv_key2)
         self.assertEqual(public_key2.as_hex(), KEY2_PUB_HEX)
 
+    def test_creating_private_keys_from_hex_and_bytes_to_contexts(self):
+        context = create_context("secp256k1")
+        priv_key = context.private_key_from_hex(KEY1_PRIV_HEX)
+        self.assertEqual(priv_key.as_hex(), KEY1_PUB_HEX)
```
delventhalz: Copypasta: You want to assert that the private key equals `KEY1_PRIV_HEX`, not PUB

signing/tests/test_secp256k1_signer.py
```diff
@@ -71,6 +77,14 @@ def test_priv_to_public_key(self):
...
+
+        priv_key = context.private_key_from_bytes(KEY1_PRIV_BYTES)
+        self.assertEqual(priv_key.as_bytes(), KEY1_PUB_BYTES)
```
delventhalz: Same issue as above. This test will fail

nickylimjj:
I'm getting tripped up by the pylint error `E0110: Abstract class 'PrivateKey' with abstract methods instantiated (abstract-class-instantiated)`. Thoughts?

delventhalz:
@vaporos What would you say is the use case for the current algorithm-agnostic factory pattern if it has no functionality for loading existing private keys? I have yet to personally encounter a use case where I only needed new private keys, which means I have never personally used the factory. My goal here is to make the factory usable.

What if as an alternative to `from_bytes`, there was a `from_existing` method? The format of the existing private key would be algorithm specific. Perhaps it could include a second optional `format` parameter, allowing algorithms to support multiple formats if needed.

peterschwarz:
Closing this due to the signing library has been moved to https://github.com/hyperledger/sawtooth-sdk-python.

PR #2031, opened 2019-02-26
Upgrade packages to address High risk CVEs

scealiontach:
Several high risk CVEs found in the sawtooth-stats-grafana image via Anchore scanning.  These vulnerabilities will prevent this image from being deployed in a highly locked down environment.

Signed-off-by: Kevin O'Donnell <kodonnel@gmail.com>

### DIFF
--- a/docker/grafana/sawtooth-stats-grafana
+++ b/docker/grafana/sawtooth-stats-grafana
@@ -23,7 +23,7 @@ RUN apt-get update && \
 	libidn2-0 \
 	systemd \
 	udev \
-	perl-base \
+	perl-base && \
     rm -rf /var/lib/apt/lists/*
 
 COPY grafana/dashboards /etc/grafana/dashboards
--- a/docker/grafana/sawtooth-stats-grafana
+++ b/docker/grafana/sawtooth-stats-grafana
@@ -17,7 +17,13 @@ FROM grafana/grafana:4.6.3
 
 RUN apt-get update && \
     apt-get install -y \
-        curl && \
+	curl=7.38.0-4+deb8u14 \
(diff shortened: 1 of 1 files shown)


scealiontach:
[grafana.scan.prefix.txt](https://github.com/hyperledger/sawtooth-core/files/2906010/grafana.scan.prefix.txt)
Above, scan before fix

[grafana.scan.postfix.txt](https://github.com/hyperledger/sawtooth-core/files/2906011/grafana.scan.postfix.txt)
Above, scan post fix

vaporos:
There is an explicit recommendation here - https://docs.docker.com/develop/develop-images/dockerfile_best-practices/ -- which says "Avoid RUN apt-get upgrade".

scealiontach:
The reason for that best practice is generally either building in unprivieleged containers - not an issue here, or immutability/cacheability. On the immutability score, completely addressing it would require locking down the versions any installed packages.  In this case that's just curl, and the packages affected by the CVEs.  Happy to do that, agreed?

scealiontach:
[grafana.locked.versions.txt](https://github.com/hyperledger/sawtooth-core/files/2911400/grafana.locked.versions.txt)
Scan after locking versions

scealiontach:
Upgrading the grafana would be widening the scope beyond what was initially identified.  personally I would rather that be a different pull request.

vaporos:
Shouldn't this be fixed upstream? Is the problem stale locally cached images?

danintel:
At least one build error appears related to this:
`Unnecessary "elif" after "raise" (no-else-raise)`
PR #2034 https://github.com/hyperledger/sawtooth-core/pull/2034

PR #2036, opened 2019-03-07
Return an error on an empty new chain

peterschwarz:
Under certain circumstances, the validator will be told to compare chains by the consensus engine that will result in an empty new chain. If this is the case, checking index `0` will result in an index-out-of-bounds panic error.  Returning an error will be logged.

~Additionally, apply new rust format changes for the latest rust version.~


peterschwarz:
Dropped formatting commit, as rust format is pinned on 1-1

peterschwarz:
It is true that this error occurring would indicate a bug in a consensus implementation.  Ultimately, the change helps to prevent the now-outside consensus process from crashing the validator.

peterschwarz:
This code has been radically re-written in master, so the same situation does not apply there.

PR #2037, opened 2019-03-07
Filter out None when getting public keys in gossip

ltseeley:
When returning a list of public keys in the `get_peers_public_keys`
method, `Gossip` should filter out None values.

This fixes a bug where consensus activation could not occur because None
values in the list would cause it to fail.

Corresponding 1.1 backport: #2038

### DIFF
--- a/validator/sawtooth_validator/gossip/gossip.py
+++ b/validator/sawtooth_validator/gossip/gossip.py
@@ -204,9 +204,13 @@ class Gossip:
     def get_peers_public_keys(self):
         """Returns the list of public keys for all peers."""
         with self._lock:
-            return [
-                self._network.connection_id_to_public_key(peer)
-                for peer in copy.copy(self._peers)]
+            # Use a generator inside the list comprehension to filter out None
+            # values in a single pass
+            return [key for key
+                    in (self._network.connection_id_to_public_key(peer)
+                        for peer
+                        in copy.copy(self._peers))
+                    if key is not None]
 
     @property
     def endpoint(self):


validator/sawtooth_validator/gossip/gossip.py
```diff
@@ -204,9 +204,11 @@ def peer_to_public_key(self, peer):
     def get_peers_public_keys(self):
         """Returns the list of public keys for all peers."""
         with self._lock:
-            return [
-                self._network.connection_id_to_public_key(peer)
-                for peer in copy.copy(self._peers)]
+            keys = [self._network.connection_id_to_public_key(peer)
```
jsmitchell: You could do this as a generator expression within the list comprehension so that you are not iterating through the list twice (once for the list comprehension and once for the filter). May come at the cost of readability though.

validator/sawtooth_validator/gossip/gossip.py
```diff
@@ -204,9 +204,11 @@ def peer_to_public_key(self, peer):
...
+                    for peer
+                    in copy.copy(self._peers)]
+            # Make sure any None values are filtered out when returning
+            return list(filter(lambda key: key is not None, keys))
```
arsulegai: Should we address why connection ID is present but peer's public key not known part. I failed to come up with the case when this would occur, could you please explain?
If stale connection IDs can be present in gossip.py, _peers dictionary is used in other places too. There's possibility that peer_id_to_public_key() def returning None. 

I got this error when default max_peers_connectivity value (10) was not changed and statically/fully peering a network of more than 5 nodes. This may not be related to the patch here, but max_peers_connectivity setting gives impression that number of peers a node can connect to. But it turns out to be number of total connections possible by a node. In our case we have one outbound and one inbound connection with each peer.

ltseeley: @peterschwarz correct me if I'm wrong, but my understanding is that this happens when the validator hasn't yet established a connection with a node that it's configured to connect to. So when the consensus engine activation happens, gossip knows about this peer (because the validator is configured to connect to it), but the node hasn't had time to establish the connection yet, so the Interconnect doesn't have an entry for the peer yet.

validator/sawtooth_validator/gossip/gossip.py
ltseeley: Done

ltseeley:
@jsmitchell @arsulegai please see my responses to your respective comments and re-review.

validator/sawtooth_validator/gossip/gossip.py
peterschwarz: That is correct.

PR #2040, opened 2019-03-11
Temporarily remove PoET

rbuysse:


rbuysse:
The PoET repo hasn't been updated to bionic yet due to some incompatibilities with sawtooth-core master. This is preventing us from publishing nightly docker images for the shell, which is blocking the sabre bionic upgrade PR. https://github.com/hyperledger/sawtooth-sabre/pull/34

PR #2041, opened 2019-03-12
Use $SAWTOOTH_0_SERVICE_HOST to bind ports

emedina:
If $SAWTOOTH_0_SERVICE_HOST is not used, ports are not bound and it does not work.


emedina:
Forgot DCO

PR #2048, opened 2019-03-18
Use new handler for consensus gossip messages

ltseeley:
Instead of using the `GossipBroadcastHandler` to send peer messages to
the consensus engine, use a new handler `GossipConsensusHandler`.
Notifying the consensus engine isn't a "broadcast" function, so it
should have a different handler. Now, the `GossipBroadcastHandler` will
simply pass when it encounters a peer message.

Backport of PR https://github.com/hyperledger/sawtooth-core/pull/2019


ltseeley:
This is a bug fix that affects the delivery of consensus messages; without this fix, a time to live value of 1 would prevent consensus messages from being sent to the consensus engine.

PR #2049, opened 2019-03-20
Fix block manager ref counting

ltseeley:
Implements BlockRef type which is used for Ref-C and Ref-D, but not
Ref-B.

### DIFF
--- a/validator/sawtooth_validator/journal/genesis.py
+++ b/validator/sawtooth_validator/journal/genesis.py
@@ -43,7 +43,6 @@ class GenesisController:
     def __init__(self,
                  context_manager,
                  transaction_executor,
-                 completer,
                  block_manager,
                  block_store,
                  state_view_factory,
@@ -76,7 +75,6 @@ class GenesisController:
         """
         self._context_manager = context_manager
         self._transaction_executor = transaction_executor
-        self._completer = completer
         self._block_manager = block_manager
         self._block_store = block_store
         self._state_view_factory = state_view_factory
@@ -225,7 +223,6 @@ class GenesisController:
 
         LOGGER.info('Genesis block created: %s', blkw)
 
-        self._completer.add_block(block)
         self._block_manager.put([blkw.block])
         self._block_manager.persist(blkw.identifier, "commit_store")
 
--- a/validator/src/journal/block_manager.rs
+++ b/validator/src/journal/block_manager.rs
@@ -47,6 +47,42 @@ impl From<BlockStoreError> for BlockManagerError {
     }
 }
 
+/// An external block reference that is owned by and can be passed between validator components.
+/// Implements the `Drop` trait to allow for automatically decrementing the external reference
+/// count for the block ID in the block manager.
+pub struct BlockRef {
+    block_id: String,
+    block_manager: BlockManager,
+}
+
+impl BlockRef {
+    fn new(block_id: String, block_manager: BlockManager) -> Self {
+        BlockRef {
+            block_id,
+            block_manager,
+        }
+    }
+
+    pub fn block_id(&self) -> &str {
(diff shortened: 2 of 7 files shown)


validator/src/journal/block_manager.rs
```diff
@@ -47,6 +48,35 @@ impl From<BlockStoreError> for BlockManagerError {
...
+    pub fn block_id(&self) -> String {
+        self.block_id.clone()
+    }
+}
+
+impl Drop for BlockRef {
+    fn drop(&mut self) {
+        self.block_manager
+            .unref_block(self.block_id.as_str())
+            .expect("Failed to unref block");
```
vaporos: A panic in drop() is considered bad, and may cause the process to abort. This can apparently happen if a panic happens while unwinding due to another panic.

peterschwarz: Agreed. I'm not sure I even agree with the fact that unref returns a result.

 Ultimately, unref_block should not be called in any way outside of this drop call in the long run, which would mean it should be managed via static garbage collection.

ltseeley: Updated

validator/src/journal/block_manager.rs
```diff
@@ -47,6 +47,41 @@ impl From<BlockStoreError> for BlockManagerError {
...
+
+    pub fn block_id(&self) -> &str {
+        &self.block_id
+    }
+}
+
+impl Drop for BlockRef {
+    fn drop(&mut self) {
+        self.block_manager
+            .unref_block(self.block_id.as_str())
```
peterschwarz: I would perhaps add a comment on how this unref block will eventually not be callable outside of block ref.

validator/src/journal/chain.rs
```diff
@@ -498,44 +509,38 @@ impl<TEP: ExecutionPlatform + Clone + 'static, PV: PermissionVerifier + Clone +
...
-        if state
-            .block_references
-            .remove(&block.header_signature)
-            .is_none()
-        {
-            error!(
-                "Reference not found for ignored block {}",
+        match state.block_references.remove(&block.header_signature) {
+            Some(_) => info!("Ignored block {}", block),
+            None => error!(
```
peterschwarz: this should be a warning

validator/src/journal/chain.rs
```diff
@@ -498,44 +509,38 @@ impl<TEP: ExecutionPlatform + Clone + 'static, PV: PermissionVerifier + Clone +
...
-        {
-            error!(
-                "Reference not found for failed block {}",
+        match state.block_references.remove(&block.header_signature) {
+            Some(_) => {
+                self.block_validation_results
+                    .fail_block(&block.header_signature);
+                info!("Failed block {}", block);
+            }
+            None => error!(
```
peterschwarz: Same - error -> warn

validator/src/journal/block_manager.rs
```diff
@@ -47,6 +47,41 @@ impl From<BlockStoreError> for BlockManagerError {
...
+
+impl BlockRef {
+    fn new(block_id: String, block_manager: BlockManager) -> Self {
+        BlockRef {
+            block_id,
+            block_manager,
+        }
+    }
+
+    pub fn block_id(&self) -> String {
```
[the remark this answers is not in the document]
ltseeley: Updated

validator/sawtooth_validator/journal/genesis.py
```diff
@@ -43,7 +43,6 @@ class GenesisController:
     def __init__(self,
                  context_manager,
                  transaction_executor,
-                 completer,
```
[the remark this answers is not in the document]
peterschwarz: See updated commit message.

PR #2056, opened 2019-04-01
Consensus backwards compatibility fix

ltseeley:
When consensus is not set on-chain, fall back to PoET if PoET settings are set, otherwise devmode

### DIFF
--- a/adm/src/commands/genesis.rs
+++ b/adm/src/commands/genesis.rs
@@ -26,6 +26,7 @@ use protobuf::Message;
 
 use proto::batch::{Batch, BatchList};
 use proto::genesis::GenesisData;
+use proto::settings::{SettingProposal, SettingsPayload, SettingsPayload_Action};
 use proto::transaction::TransactionHeader;
 
 use config;
@@ -75,6 +76,7 @@ pub fn run<'a>(args: &ArgMatches<'a>) -> Result<(), CliError> {
         });
 
     validate_depedencies(&batches)?;
+    check_required_settings(&batches)?;
 
     let mut genesis_data = GenesisData::new();
     genesis_data.set_batches(protobuf::RepeatedField::from_vec(batches));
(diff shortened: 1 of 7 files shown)


ltseeley:
Open questions about this PR:
1. When checking for the presence of PoET settings to determine if PoET is the engine to fall back to, what settings should be checked?
2. @vaporos suggested that the genesis CLI tool should fail when the consensus settings aren't set; however, this requires knowledge of the settings TP, specifically its payload protobuf. Does that make sense to have in the genesis tool? Also, since the validator currently checks for these settings when creating the genesis block, is this necessary?
3. This PR currently updates the consensus proxy to call a method on the chain controller to update the active consensus engine; this removes the duplication of that code since it previously existed in both components. However, it probably makes more sense for this code to exist in the proxy and have the chain controller call the consensus proxy. Any thoughts on this?

danintel:
My 2 cents (I realize this is WIP),

1. This PR uses `sawtooth.consensus.poet.enclave_module_name`, but that is set for PoET SGX, not PoET CFT (aka PoET Simulator). If we're supporting PoET CFT migration, I think you want to check for either `sawtooth.poet.target_wait_time` or  `sawtooth.poet.initial_wait_time`. 

**Alternative**
You can check for `sawtooth.consensus.algorithm` = `poet` (some case insensitive comparison) for Sawtooth 1.0 migration and `sawtooth.consensus.algorithm.name` = `PoET` for Sawtooth 1.1. It's unclear if these names are case insensitive or not. If so, the appropriate case insensitive comparison is needed.

2. That would be a nice check--an earlier error or warning is better. If it's done `sawadm genesis` would have to check all the batch files passed to it. It just seems slightly hairy to go parsing into batches with protobufs and all.

On a side note, with Sawtooth 1.0 if I set, for example `sawtooth.consensus.algorithm.name` and `.version` to future-proof it for Sawtooth 1.1 migration, it errors out. That is unfortunate as it would also aid migration. It doesn't really matter though--it's water under the bridge.

3. No comment.

validator/src/journal/chain.rs
```diff
@@ -851,8 +857,33 @@ impl<TEP: ExecutionPlatform + Clone + 'static, PV: PermissionVerifier + Clone +
...
+                // If PoET settings exist, fall back to PoET
+                if settings_view
+                    .get_setting_str("sawtooth.consensus.poet.enclave_module_name", None)
+                    .map_err(|settings_err| {
+                        let err_str = format!(
+                            "Error getting sawtooth.consensus.poet.enclave_module_name from \
+                             settings view: {:?}",
+                            settings_err
+                        );
+                        warn!("{}", err_str);
```
vaporos: Will this be double logged since it is also propogated up?

ltseeley: Yep, it seems so. I'll fix that.

ltseeley: Fixed

ltseeley:
Points (2) and (3) from my original comment have been resolved. I took @danintel's suggestion for point (1) as well.

ltseeley:
@danintel what error did you see when trying to set `sawtooth.consensus.algorithm.name/version` on v1.0? I was able to do that successfully on v1.1.

danintel:
@ltseeley I forgot the exact error message. When I tried it with Sawtooth 1.0, it said the name was invalid (which is correct).
It works on v1.1, but that is expected.
The reason I want it to work on v1.0 is so someone running Sawtooth v1.0 can more easily upgrade to v1.1+ (since Sawtooth 1.1+ uses `sawtooth.consensus.algorithm.name/version` and v1.0 uses `sawtooth.consensus.algorithm` )

ltseeley:
Rebased on #2078, added commit to generate settings protobufs for Rust `sawadm`, and added genesis settings enforcement to Rust `sawadm`

adm/src/commands/genesis.rs
```diff
@@ -124,3 +126,51 @@ fn validate_depedencies(batches: &[Batch]) -> Result<(), CliError> {
     }
     Ok(())
 }
+
+fn check_required_settings(batches: &[Batch]) -> Result<(), CliError> {
```
peterschwarz: The python cli has some tests for this.  Any that can be added here?

ltseeley: Added a unit test for the updated functionality

ltseeley:
> Please add info to the documentation that explains how Sawtooth determines the default consensus.
> 
> Ideally, this information would go in the new consensus topic (but that's not available yet). In the meantime, I suggest a note in the App Dev Guide, Sys Admin Guide, and Architecture Guide.
> 
> P.S. I can provide tips on how to create a shared include file, then use Sphinx's `..include` directive to add the same content to these three guides.

Added documentation for changes to `sawadm genesis`. The consensus fallback will be documented later when the consensus documentation is revamped as per @chenette

ltseeley:
@jsmitchell @peterschwarz stories have been opened to implement your requested changes in separate PRs.

PR #2059, opened 2019-04-04
Remove futures from _SendReceive after a certain timeout.

eugene-babichenko:
Some futures produced during peering attempts were never resolved and
_SendReceive did not clean them up in any way. On a longer time scale,
this was leading to excessive memory usage by the validator service.

This commit adds the possibility to set a timeout for a Future object.
If a timeout was set for a Future, FutureCollection is able to clean
that future after a timeout when the remove_expired method is called.
This method is called periodically by _SendReceive (the timeout is equal
to connection_timeout).

Also, futures are explicitly removed when stopping an
OutboundConnection.

Signed-off-by: Yevhenii Babichenko <babichenko@482.solutions>

### DIFF
--- a/validator/sawtooth_validator/networking/future.py
+++ b/validator/sawtooth_validator/networking/future.py
@@ -35,7 +35,7 @@ class FutureTimeoutError(Exception):
 
 class Future:
     def __init__(self, correlation_id, request=None, callback=None,
-                 timer_ctx=None):
+                 timeout=None, timer_ctx=None):
         self.correlation_id = correlation_id
         self._request = request
         self._result = None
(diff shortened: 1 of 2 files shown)


jsmitchell:
@eugene-babichenko you have some lint:

10:14:59  [36mlint-python_1     |[0m ************* Module sawtooth_validator.networking.future
10:14:59  [36mlint-python_1     |[0m validator/sawtooth_validator/networking/future.py:89:8: R1705: Unnecessary "else" after "return" (no-else-return)

eugene-babichenko:
@jsmitchell Fixed that. Build is still failing because of some error with dependencies.

jsmitchell:
> @jsmitchell Fixed that. Build is still failing because of some error with dependencies.

I replayed the job and it looks good.

eugene-babichenko:
Also, it would be great to port this fix to the next 1.1.x release.

PR #2061, opened 2019-04-05
New Sys Admin doc to add and remove PBFT nodes

chenette:


docs/source/sysadmin_guide/pbft_adding_removing_node.rst
```diff
@@ -0,0 +1,163 @@
...
+******************************
+
+Membership of a Sawtooth PBFT network is controlled by the on-chain setting
+``sawtooth.consensus.pbft.members``. An administrator must update this setting
+when :ref:`adding a new node <adding-a-pbft-node-label>` or
+:ref:`removing an existing node <removing-a-pbft-node-label>`.
+
+All nodes in the network monitor this setting. When a node detects a change,
+it updates its local list of PBFT members to match the new list. When all member
+nodes have updated the setting, consensus resumes normal operation.
```
ltseeley: This last sentence isn't really accurate; none of the nodes wait until other nodes update the setting, they just update their own then continue on.

docs/source/sysadmin_guide/pbft_adding_removing_node.rst
```diff
@@ -0,0 +1,163 @@
...
+==================
+
+To add a new node to an existing PBFT network, you will install and configure
+the node, start it and wait for it to catch up with the rest of the network,
+then update ``sawtooth.consensus.pbft.members`` on an existing node.
+
+You can add several nodes at the same time.
+
+1. Install and configure Sawtooth on the new node, as described in
+   :doc:`setting_up_sawtooth_poet-sim`.
```
ltseeley: Is this supposed to be linking to section about PoET?

docs/source/sysadmin_guide/pbft_adding_removing_node.rst
```diff
@@ -0,0 +1,163 @@
...
+   Skip the procedure to create a genesis block.
+
+#. Start the new node. It will join the Sawtooth network and start the block
+   catch-up procedure, where it receives and commit the blocks that are already
```
ltseeley: commit -> commits

docs/source/sysadmin_guide/pbft_adding_removing_node.rst
```diff
@@ -0,0 +1,163 @@
...
+   on the blockchain. (For details, see the `PBFT node catch-up RFC
+   <https://github.com/hyperledger/sawtooth-rfcs/blob/master/text/0031-pbft-node-catchup.md>`_.)
+
+   The new node sends consensus messages as processes blocks, but the other
```
ltseeley: Not sure this sentence is necessary; it's true, I don't think it's necessary. I'm open to other opinions though.

docs/source/sysadmin_guide/pbft_adding_removing_node.rst
chenette: The link is correct - that section includes both PBFT and PoET steps. 

I'm aware that the file name is out of date, but we don't currently have a way to handle broken links when a file name changes. I hope to have a general fix in the future.

docs/source/sysadmin_guide/pbft_adding_removing_node.rst
chenette: Removed that sentence.

docs/source/sysadmin_guide/pbft_adding_removing_node.rst
chenette: Fixed.

docs/source/sysadmin_guide/pbft_adding_removing_node.rst
chenette: I removed the sentence.

docs/source/sysadmin_guide/pbft_adding_removing_node.rst
```diff
@@ -0,0 +1,158 @@
...
+
+   .. code-block:: console
+
+      $ cat /etc/sawtooth/keys/validator.pub
+
+   This step assumes that the validator key is stored in the default location,
+   ``/etc/sawtooth/keys``. If not, use the location specified by the ``key_dir``
+   setting (see :doc:`configuring_sawtooth/path_configuration_file`).
+
+#. On an existing member node, update ``sawtooth.consensus.pbft.members`` to
```
ltseeley: I would generalize this; "Update the setting to include the public key ... Run the following steps using a key that has permission ...; that is, a public key listed in.... Usually, the public key of the node that created the genesis block is on this list...."

It's helpful to point out that it's commonly the genesis node's key that's in this list, but could be some special key that doesn't belong to any validator.

docs/source/sysadmin_guide/pbft_adding_removing_node.rst
```diff
@@ -0,0 +1,158 @@
...
+         stall the network.
+
+      .. code-block:: console
+
+         $ sawset proposal create \
+           --key /etc/sawtooth/keys/validator.priv \
+           sawtooth.consensus.pbft.members=[previous-list, NEW-KEY]
+
+      If there are no errors, the block containing this setting change will be
+      committed quickly.
```
ltseeley: We can't really make a guarantee that it will be committed quickly. I'd just say something like: "If there are no errors, this settings change will be committed to the chain."

docs/source/sysadmin_guide/pbft_adding_removing_node.rst
```diff
@@ -0,0 +1,158 @@
...
+         Double-check each key before you run this command, because a typo could
+         stall the network.
+
+      .. code-block:: console
+
+         $ sawset proposal create \
+           --key /etc/sawtooth/keys/validator.priv \
+           sawtooth.consensus.pbft.members=[UPDATED-LIST]
+
+      If there are no errors, the block containing this setting change will be
```
ltseeley: Same as above

docs/source/sysadmin_guide/pbft_adding_removing_node.rst
```diff
@@ -0,0 +1,158 @@
...
+      committed quickly.
+
+#. Make sure that change has been committed. You can check the setting from
+   another node.
```
ltseeley: You can check the setting from any node, even one that's been removed from the list (because they still commit the block with the settings change). So I'd just say you can check the setting from any node.

docs/source/sysadmin_guide/pbft_adding_removing_node.rst
```diff
@@ -0,0 +1,158 @@
...
+
+   .. code-block:: console
+
+      $ sawtooth setting list --filter sawtooth.consensus.pbft.members
+
+   .. Note::
+
+      Until the settings change is updated on all nodes, the removed node is
```
ltseeley: I'd use "committed" instead of "updated," since it's less vague and easier to check.

docs/source/sysadmin_guide/pbft_adding_removing_node.rst
```diff
@@ -0,0 +1,158 @@
...
+      considered part of the network. If the node is are shut down too soon,
```
ltseeley: "is are"

docs/source/sysadmin_guide/pbft_updating_member_list.rst
```diff
@@ -0,0 +1,13 @@
+************************************
+PBFT Only: Updating PBFT Member List
+************************************
+
+If you are adding a new node to an existing PBFT network, you must update the
+on-chain setting ``sawtooth.consensus.pbft.members`` after the new node has
+been installed and configured.
```
danintel: Consider mentioning the setting takes effect after the containing block is published.

docs/source/sysadmin_guide/pbft_adding_removing_node.rst
```diff
@@ -0,0 +1,158 @@
...
+====================
+
+To remove an existing node from a PBFT network, you will delete the node's
+validator key from  the ``sawtooth.consensus.pbft.members`` setting, then shut
+down the removed node. You can delete several nodes at the same time.
+
+.. Note::
+
+   PBFT consensus requires a network with at least four nodes. A network with
+   fewer than four nodes will fail.
```
danintel: This is unfortunate (and unavoidable due to the algorithm) because there's also a low upper-bound of nodes (as the connections are O(n^2).

This implies to me that the genesis node must have at least four nodes. What happens if you configure, say, three nodes?

ltseeley: That's correct; at genesis and any time thereafter, there need to be at least 4 nodes. If fewer than 4 nodes are configured in the `sawtooth.consensus.pbft.members` setting, PBFT will panic because it won't be able to operate.

docs/source/sysadmin_guide/pbft_adding_removing_node.rst
```diff
@@ -0,0 +1,158 @@
...
+
+1. Update ``sawtooth.consensus.pbft.members`` to no longer include the
+   validator public key of the node you want to remove.
+
+   a. List the current PBFT member nodes:
+
+      .. code-block:: console
+
+         $ sawtooth setting list --filter sawtooth.consensus.pbft.members
```
desktophero: I think `setting` should be `settings`

docs/source/sysadmin_guide/pbft_adding_removing_node.rst
```diff
@@ -0,0 +1,158 @@
...
+   settings; that is, a node whose validator key is listed in
+   ``sawtooth.identity.allowed_keys``. Usually, the node that created the
+   genesis block is on this list. For more information, see
+   :ref:`config-onchain-txn-perm-label`.
+
+   a. List the current PBFT member nodes:
+
+      .. code-block:: console
+
+         $ sawtooth setting list --filter sawtooth.consensus.pbft.members
```
desktophero: I think `setting` should be `settings`

docs/source/sysadmin_guide/pbft_adding_removing_node.rst
chenette: You are correct.  Thanks!

docs/source/sysadmin_guide/pbft_adding_removing_node.rst
chenette: This is a lot of information to include in this step (and doesn't really "belong" to the PBFT procedure), so I'll format the info as a separate tip.  

Ideally, the Sawtooth doc should have a separate topic to explain this info, because several procedures show how to change on-chain settings.  When that happens, this tip can be replaced with a link to the general information.

docs/source/sysadmin_guide/pbft_adding_removing_node.rst
```diff
@@ -0,0 +1,158 @@
...
+All nodes in the network monitor this setting. When a node detects a change,
+it updates its local list of PBFT members to match the new list.
+
+.. _adding-a-pbft-node-label:
+
+Adding a PBFT Node
+==================
+
+To add a new node to an existing PBFT network, you will install and configure
+the node, start it and wait for it to catch up with the rest of the network,
```
rberg2: Is this where you would add the new node as a `--peer` to the existing nodes? I wonder if we should cover that part of the process too.

chenette: The other procedure describes how to configure a node as a peer on a Sawtooth network. (At least, it should - if it doesn't, that's a problem with that other procedure.)  I'd prefer to keep all the generic info in the other procedure so that this one can be focused on the PBFT-only information.

chenette: I checked - the peering information is documented in this topic (see step 4):
https://sawtooth.hyperledger.org/docs/core/nightly/master/sysadmin_guide/off_chain_settings.html

I know that a PBFT network must be "fully peered".  @ltseeley , does that mean that the admin(s) must modify the off-chain peers settings on all nodes in the network?  Or is it sufficient to list all other nodes for this new node?

In any case (and to @rberg2 's comment), does this procedure need a note to explain how to handle peering on the new node?

Also, when removing a node, does the off-chain peering setting need to be changed on the remaining nodes when a node is removed from the network?

ltseeley: It's sufficient to list all other nodes ONLY for the new node. Connections between validators are bi-directional.

chenette: I shall add info in step 1 to mention that PBFT requires full peering and to link to the procedure that describes how to change the off-chain config file. Thanks, @rberg2 for this helpful comment!

chenette: I shall also add a note in the "remove a PBFT node" procedure that you do NOT have to delete the removed node from the list of peers on other node. (You can do so if you want, but it's not required.)

chenette:
I pushed a new commit with all the corrections and improvements so far.  It includes new info on PBFT peering in the existing off-chain validator config procedure.

Also, I squashed the separate commits together, because the doc changes are all related; trying to keep them "separate" wasn't sensible or practical.

docs/source/sysadmin_guide/pbft_updating_member_list.rst
chenette: Done.

chenette:
I also moved the new Add/remove PBFT nodes procedure to later in the Sys Admin Guide, because it refers to several of the "was later now earlier" sections.

docs/source/sysadmin_guide/pbft_adding_removing_node.rst
```diff
@@ -0,0 +1,182 @@
...
+      ``sawtooth.identity.allowed_keys``. Note that this list can include user
+      keys as well as validator keys, so that changes can be made by an
+      administrator from any node. For more information, see
+      :ref:`config-onchain-txn-perm-label`.
+
+   a. List the current PBFT member nodes:
+
+      .. code-block:: console
+
+         $ sawtooth settings list --filter sawtooth.consensus.pbft.members
```
desktophero: I don't have a `members` but `peers` when I look at pbft settings (a la `sawtooth.consensus.pbft.peers`)

docs/source/sysadmin_guide/pbft_adding_removing_node.rst
```diff
@@ -0,0 +1,182 @@
...
+
+         BE VERY CAREFUL! Make sure to specify the full list of keys.
+         Double-check each key before you run this command, because a typo could
+         stall the network.
+
+      .. code-block:: console
+
+         $ sawset proposal create \
+           --key /etc/sawtooth/keys/validator.priv \
+           sawtooth.consensus.pbft.members=[previous-list, NEW-KEY]
```
desktophero: I think `members` needs to be `peers`

docs/source/sysadmin_guide/pbft_adding_removing_node.rst
```diff
@@ -0,0 +1,182 @@
...
+#. When all nodes have detected the change and updated their local copy of the
+   member list, the new member node begins to participate in the PBFT network.
+
+.. _removing-a-pbft-node-label:
+
+Removing a PBFT Node
+====================
+
+To remove an existing node from a PBFT network, you will delete the node's
+validator key from  the ``sawtooth.consensus.pbft.members`` setting, then shut
```
desktophero: `peers` not `members, I think

docs/source/sysadmin_guide/pbft_adding_removing_node.rst
```diff
@@ -0,0 +1,182 @@
...
+
+   Run the following steps on a node or as a user that has permission to change
+   on-chain settings. For more information, see the tip in
+   :ref:`adding-a-pbft-node-label`.
+
+   a. List the current PBFT member nodes:
+
+      .. code-block:: console
+
+         $ sawtooth settings list --filter sawtooth.consensus.pbft.members
```
desktophero: sed `'s/members/peers'` for the filter, based on my local env

docs/source/sysadmin_guide/pbft_adding_removing_node.rst
```diff
@@ -0,0 +1,182 @@
...
+
+         BE VERY CAREFUL! Make sure to specify the correct list of keys.
+         Double-check each key before you run this command, because a typo could
+         stall the network.
+
+      .. code-block:: console
+
+         $ sawset proposal create \
+           --key /etc/sawtooth/keys/validator.priv \
+           sawtooth.consensus.pbft.members=[UPDATED-LIST]
```
desktophero: `peers` instead of `members` please, based on local env

desktophero:
Thanks for this PR .... really helpful in how to approach this item in a new/long-living network.

The current commit shows `members` as PBFT participants for various `sawtooth` commands but `peers` is the correct item, I think. I annotated several places but not all. I didn't want to repeat the same on every spot but the `members` option in the setting property doesn't appear to be correct.

Is an update needed to get the `members` swapped to `peers` for pbft settings?

chenette:
The setting name changed from peers to members in a separate PR or two (plus related PRs to change the documentation) in the sawtooth-core and sawtooth-pbft repos. Those other PRs were merged recently, so an update should get the peer->member changes.

PR #2064, opened 2019-04-10
 Fix TxnExecutionResult.data field

ltseeley:
Fixes the `TxnExecutionResult.data` to be consistent. Before this
change, some components assumed this field was a list of
(data type,bytes) pairs, others assumed it was just a list of bytes.


ltseeley:
> The "updates .gitignore" also removes the python block info transaction family. If intentional, that deserves it's own commit and explanation.

These are the generated protobuf files that should have be ignored by the .gitignore and not checked in. Would adding details to the message of the existing commit be sufficient

vaporos:
> > The "updates .gitignore" also removes the python block info transaction family. If intentional, that deserves it's own commit and explanation.
> 
> These are the generated protobuf files that should have be ignored by the .gitignore and not checked in. Would adding details to the message of the existing commit be sufficient

Yes, it would.

ltseeley:
> > > The "updates .gitignore" also removes the python block info transaction family. If intentional, that deserves it's own commit and explanation.
> > 
> > 
> > These are the generated protobuf files that should have be ignored by the .gitignore and not checked in. Would adding details to the message of the existing commit be sufficient
> 
> Yes, it would.

Fixed

danintel:
MarLu96 1:54 AM   
Hi all, I'm trying to create a node that can deploy solidity contract with seth, but using PoET (and not devMode). Here is how I modify the docker-compose file from the seth github. It seems that when I create an account ` seth account create --nonce=0 --wait myalias ` , nothing happens like if the connection is not properly done. I have the returned message ` Error: Problem submitting account creation transaction: Transaction was submitted, but client timed out before it was committed. `

danintel 8:47 AM
@MarLu96 The problem is you are using the latest version of Seth, which is broken--it does not work. The workaround is to roll back Seth to changeset 52a0512 (Disable clippy warning about not returning Self). This forces Seth to run docker-compose with Ubuntu 16 (Xenial) and Sawtooth 1.0.
The issue I filed is https://jira.hyperledger.org/browse/STL-1509

ltseeley 9:37 AM
@danintel @MarLu96 @rajatwasan @lin2xing the issue with seth account creation has been resolved on the master branch of sawtooth-core: https://github.com/hyperledger/sawtooth-core/pull/2064

danintel:
MarLu96 1:54 AM   
Hi all, I'm trying to create a node that can deploy solidity contract with seth, but using PoET (and not devMode). Here is how I modify the docker-compose file from the seth github. It seems that when I create an account ` seth account create --nonce=0 --wait myalias ` , nothing happens like if the connection is not properly done. I have the returned message ` Error: Problem submitting account creation transaction: Transaction was submitted, but client timed out before it was committed. `

danintel 8:47 AM
@MarLu96 The problem is you are using the latest version of Seth, which is broken--it does not work. The workaround is to roll back Seth to changeset 52a0512 (Disable clippy warning about not returning Self). This forces Seth to run docker-compose with Ubuntu 16 (Xenial) and Sawtooth 1.0.
The issue I filed is https://jira.hyperledger.org/browse/STL-1509

ltseeley 9:37 AM
@danintel @MarLu96 @rajatwasan @lin2xing the issue with seth account creation has been resolved on the master branch of sawtooth-core: https://github.com/hyperledger/sawtooth-core/pull/2064

PR #2066, opened 2019-04-10
Allow re-registering consensus engines

ltseeley:
To support a consensus engine restarting independently of the validator,
the consensus registry must be able to register a conensus engine again.
When this happens, the engine will register with the same name and
version, but a different connection ID. The registry will simply remove
any old entries associated with the engine name and version before
adding the engine to the registry.

### DIFF
--- a/validator/sawtooth_validator/consensus/registry.py
+++ b/validator/sawtooth_validator/consensus/registry.py
@@ -40,6 +40,10 @@ class ConsensusRegistry:
 
     def register_engine(self, connection_id, name, version):
         with self._lock:
+            # If this engine is already registered, remove the old connection
+            self._registry = list(filter(
+                lambda e: e.name != name and e.version != version,
+                self._registry))
             self._registry.append(EngineInfo(connection_id, name, version))
 
     def activate_engine(self, name, version):


arsulegai:
There are other concerns when consensus engine restarts. For example, consensus engine is building a block and then it's restarted before finalizing the block. This will leave validator in unexpected state, and when consensus engine restarts and requests to build a new block again validator complains for the scenario.

ltseeley:
> There are other concerns when consensus engine restarts. For example, consensus engine is building a block and then it's restarted before finalizing the block. This will leave validator in unexpected state, and when consensus engine restarts and requests to build a new block again validator complains for the scenario.

This can be handled by the consensus engine; the validator will reply with an `INVALID_STATE` status if consensus attempts to initialize a block when one is already initialized, at which point consensus knows there's already a candidate block. A possible alternative would be to cancel any candidate blocks if the active consensus engine disconnects, but that's outside the scope of this PR.

> Just a thought: Instead of only removing connection ID, should there be a session with respect to each consensus engine registration/activation, and then the entire session should be ignored when a new session is created?

Can you describe what you mean by a "session"? The `EngineInfo` object essentially represents a "session": it's a connection to a consensus engine with a given name/version. When a new one registers with that same name/version, the old `EngineInfo` is removed.

arsulegai:
> > There are other concerns when consensus engine restarts. For example, consensus engine is building a block and then it's restarted before finalizing the block. This will leave validator in unexpected state, and when consensus engine restarts and requests to build a new block again validator complains for the scenario.
> 
> This can be handled by the consensus engine; the validator will reply with an `INVALID_STATE` status if consensus attempts to initialize a block when one is already initialized, at which point consensus knows there's already a candidate block. A possible alternative would be to cancel any candidate blocks if the active consensus engine disconnects, but that's outside the scope of this PR.
> 

Right! Thanks. PoET doesn't make use of it currently.

> > Just a thought: Instead of only removing connection ID, should there be a session with respect to each consensus engine registration/activation, and then the entire session should be ignored when a new session is created?
> 
> Can you describe what you mean by a "session"? The `EngineInfo` object essentially represents a "session": it's a connection to a consensus engine with a given name/version. When a new one registers with that same name/version, the old `EngineInfo` is removed.

Your answer clarified my question.

PR #2067, opened 2019-04-11
Improve consensus engine behaviour between registration and activation

peterschwarz:
This PR removes the NewBlockSync handler when an engine registered, which would send all forks to the engine, whether they had been validated or not.  This broke the contract with consensus engines (as it currently stands) and would cause problems with PoET networks during catch up.

While there is no active consensus engine, the notifications are queued, with activation messages having priority.

Instead of busy-waiting for the genesis block, this PR includes a new ChainObserver that will activate the configured engine, if those settings have changed.

Additionally, a timeout has been added on the notification send's expected result, so as not to hang threads while waiting on a response that might never come.

### DIFF
--- a/validator/sawtooth_validator/consensus/proxy.py
+++ b/validator/sawtooth_validator/consensus/proxy.py
@@ -16,6 +16,10 @@
 import hashlib
 import logging
 
+from sawtooth_validator.journal.chain import ChainObserver
+from sawtooth_validator.journal.event_extractors import \
+    ReceiptEventExtractor
+from sawtooth_validator.server.events.subscription import EventSubscription
 from sawtooth_validator.protobuf.block_pb2 import BlockHeader
 from sawtooth_validator.protobuf.consensus_pb2 import ConsensusPeerMessage
 from sawtooth_validator.protobuf.consensus_pb2 import \
@@ -27,6 +31,73 @@ LOGGER = logging.getLogger(__name__)
 
 class UnknownBlock(Exception):
     """The given block could not be found."""
+
+
+class ConsensusActivationObserver(ChainObserver):
+    """Observes chain updates and activates new engines, when the settings
+    have been updated in state, as of the commit."""
+
+    def __init__(self, consensus_registry, consensus_notifier,
+                 settings_view_factory):
+        self._registry = consensus_registry
+        self._notifier = consensus_notifier
+        self._settings_view_factory = settings_view_factory
+
+    def chain_update(self, block, receipts):
+        # Check to see if there is an active engine at all:
+        if not self._registry.has_active_engine():
+            self._update_active_engine(block)
+            return
+
+        # Otherwise, check to see if the settings changed, before updating
+        # the active engine.
+        receipt_events = ReceiptEventExtractor(receipts).extract([
+            EventSubscription(event_type="setting/update")])
+
+        for event in receipt_events:
+            if self._is_consensus_algorithm_setting(event):
+                self._update_active_engine(block)
+                break
+
+    def _update_active_engine(self, block):
+        header = BlockHeader()
+        header.ParseFromString(block.header)
+        settings_view = self._settings_view_factory.create_settings_view(
+            header.state_root_hash)
+
+        conf_name = settings_view.get_setting(
+            'sawtooth.consensus.algorithm.name')
+        conf_version = settings_view.get_setting(
+            'sawtooth.consensus.algorithm.version')
+
+        if not self._registry.is_active_engine_name_version(
+                conf_name, conf_version):
+
+            # Deactivate the old engine, if necessary
+            old_engine_info = None
+            if self._registry.has_active_engine():
+                old_engine_info = self._registry.get_active_engine_info()
+
+            self._registry.activate_engine(conf_name, conf_version)
+
+            if old_engine_info is not None:
+                self._notifier.notify_engine_deactivated(
+                    old_engine_info.connection_id)
+            self._notifier.notify_engine_activated(block)
+
+            LOGGER.info(
+                "Consensus engine %s %s activated as of block %s",
+                conf_name,
+                conf_version,
+                block.header_signature)
+
+    def _is_consensus_algorithm_setting(self, event):
+        if event.event_type != 'setting/update':
+            return False
+
+        key = event.attributes[0].value
+        return key in ('sawtooth.consensus.algorithm.name',
+                       'sawtooth.consensus.algorithm.version')
 
 
 class ConsensusProxy:
@@ -54,12 +125,15 @@ class ConsensusProxy:
 
     def activate_if_configured(self, engine_name, engine_version):
         # Wait until chain head is committed
-        chain_head = None
-        while chain_head is None:
-            try:
-                chain_head = self.chain_head_get()
-            except UnknownBlock:
-                pass
+        try:
+            chain_head = self.chain_head_get()
+        except UnknownBlock:
+            LOGGER.debug(
+                "Unable to determine if engine %s %s is configured until "
+                "chain head received",
+                engine_name,
+                engine_version)
+            return
 
         header = BlockHeader()
         header.ParseFromString(chain_head.header)
--- a/validator/sawtooth_validator/consensus/proxy.py
+++ b/validator/sawtooth_validator/consensus/proxy.py
@@ -16,6 +16,7 @@
 import hashlib
 import logging
 
+from sawtooth_validator.consensus.registry import EngineAlreadyActive
 from sawtooth_validator.journal.chain import ChainObserver
 from sawtooth_validator.journal.event_extractors import \
     ReceiptEventExtractor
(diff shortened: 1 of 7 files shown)


validator/sawtooth_validator/consensus/proxy.py
```diff
@@ -29,6 +33,68 @@ class UnknownBlock(Exception):
...
+        self._notifier = consensus_notifier
+        self._settings_view_factory = settings_view_factory
+
+    def chain_update(self, block, receipts):
+        # Check to see if there is an active engine at all:
+        if not self._registry.has_active_engine():
+            self._update_active_engine(block)
+            return
+
+        # Otherwise, check to see fi the settings changed, before updating
```
jsmitchell: s/fi/if

validator/sawtooth_validator/consensus/proxy.py
```diff
@@ -29,6 +33,68 @@ class UnknownBlock(Exception):
     """The given block could not be found."""
 
 
+class ConsensusActivationObserver(ChainObserver):
+    """Observes chain updates and activates new engines, if they have been
```
jsmitchell: activating things which have been activated is confusing

validator/sawtooth_validator/consensus/proxy.py
```diff
@@ -29,6 +33,65 @@ class UnknownBlock(Exception):
...
+        if not self._registry.has_active_engine():
+            self._update_active_engine(block)
+            return
+
+        # Otherwise, check to see fi the settings changed, before updating
+        # the active engine.
+        receipt_events = ReceiptEventExtractor(receipts).extract([
+            EventSubscription(event_type="setting/update")])
+
+        consensus_settings_changed = False
```
ltseeley: This bit can be simplified: `for ...: if self._is_consensus_algorithm_setting(event): self._update_active_engine(block); break`

validator/sawtooth_validator/consensus/proxy.py
```diff
@@ -29,6 +33,65 @@ class UnknownBlock(Exception):
...
+
+        conf_name = settings_view.get_setting(
+            'sawtooth.consensus.algorithm.name')
+        conf_version = settings_view.get_setting(
+            'sawtooth.consensus.algorithm.version')
+
+        if not self._registry.is_active_engine_name_version(conf_name, conf_version):
+            self._registry.activate_engine(conf_name, conf_version)
+
+            self._notifier.notify_engine_activated(block)
```
ltseeley: If already is an active engine, it should be notified that is deactivated (see how the chain controller currently does this)

validator/sawtooth_validator/consensus/proxy.py
```diff
@@ -54,12 +117,14 @@ def register(self, engine_name, engine_version, connection_id):
...
-        while chain_head is None:
-            try:
-                chain_head = self.chain_head_get()
-            except UnknownBlock:
-                pass
+        try:
+            chain_head = self.chain_head_get()
+        except UnknownBlock:
+            LOGGER.warning(
+                "Unable to activate engine %s %s until chain head received",
```
ltseeley: This message seems to imply that the engine should be activated; "Cannot check engine activation until chain header received" or something to that effect would be more accurate.

validator/sawtooth_validator/consensus/proxy.py
ltseeley: Did you mean "... if settings have been updated."?

validator/sawtooth_validator/consensus/proxy.py
```diff
@@ -34,9 +34,11 @@ class UnknownBlock(Exception):
 
 
 class ConsensusActivationObserver(ChainObserver):
-    """Observes chain updates and activates new engines, if they have been activated."""
+    """Observes chain updates and activates new engines, if they have been
+    activated."""
 
-    def __init__(self, consensus_registry, consensus_notifier, settings_view_factory):
+    def __init__(self, consensus_registry, consensus_notifier,
+                 settings_view_factory):
```
ltseeley: The formatting changes in the "Notify ChainObservers of genesis block" commit should be squashed into the previous commit.

validator/sawtooth_validator/consensus/proxy.py
peterschwarz: Updated the docstring

validator/sawtooth_validator/consensus/proxy.py
peterschwarz: fixed

validator/sawtooth_validator/consensus/proxy.py
```diff
@@ -29,6 +33,65 @@ class UnknownBlock(Exception):
...
+        consensus_settings_changed = False
+        for event in receipt_events:
+            consensus_settings_changed = self._is_consensus_algorithm_setting(event)
+            if consensus_settings_changed:
+                break
+
+        if consensus_settings_changed:
+            self._update_active_engine(block)
+
+    def _update_active_engine(self, block):
```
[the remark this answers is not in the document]
peterschwarz: Fixed in an new commit.

validator/sawtooth_validator/consensus/proxy.py
peterschwarz: fixed

validator/sawtooth_validator/consensus/proxy.py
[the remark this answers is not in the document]
peterschwarz: Fixed in an new commit.

validator/sawtooth_validator/consensus/proxy.py
peterschwarz: fixed

validator/sawtooth_validator/consensus/proxy.py
```diff
@@ -29,6 +33,73 @@ class UnknownBlock(Exception):
...
+        conf_name = settings_view.get_setting(
+            'sawtooth.consensus.algorithm.name')
+        conf_version = settings_view.get_setting(
+            'sawtooth.consensus.algorithm.version')
+
+        if not self._registry.is_active_engine_name_version(
+                conf_name, conf_version):
+
+            # Deactivate the old engine, if necessary
+            old_engine_info = None
```
ltseeley: This bit could be simpler by just notifying the deactivation first then activating and notifying the activation. But up to you.

validator/src/journal/chain.rs
```diff
@@ -852,74 +851,8 @@ impl<TEP: ExecutionPlatform + Clone + 'static, PV: PermissionVerifier + Clone +
...
-                    let err_str = format!("Error activating consensus engine: {:?}", reg_err);
-                    warn!("{}", err_str);
-                    ChainControllerError::ConsensusError(err_str)
-                })?;
-            self.consensus_notifier.notify_engine_activated(block);
-        } else {
-            self.consensus_notifier
-                .notify_block_commit(&block.header_signature);
-        }
+        self.consensus_notifier
```
ltseeley: This commit notification should either be before the activation change, or mutually exclusive with an activation change. Otherwise the new engine will get a block commit notification right away, which it won't expect because it didn't tell the validator to commit this block.

validator/sawtooth_validator/consensus/proxy.py
peterschwarz: Fixed

validator/sawtooth_validator/consensus/proxy.py
peterschwarz: Notifying of deactivation cannot come before changing the active engine in the registry, because sending the notification doesn't alter the registry.  So I need to get the old engine, change the active engine, and then send the messages.

validator/src/journal/chain.rs
peterschwarz: Added a commit for this.

validator/src/journal/block_manager.rs
```diff
@@ -66,6 +66,24 @@ impl BlockRef {
     pub fn block_id(&self) -> &str {
         &self.block_id
     }
+
+    pub fn get_block(&self) -> Block {
```
vaporos: BlockRef should be renamed if it's a pass-through interface to the BlockManager.

Before this change, BlockRef's primary duty was to allow holding a reference to a chain head. After this change, it's not clear what the primary job of BlockRef, as it is both pass-through and ref counting. Seems like a design regression in it's added conceptual complexity.

This change is also very tied to our current BlockRef implementation that directly holds a BlockManager. It is not clear that BlockRef should even have access to anything other than unref_block wrt BlockManager (if we exclude this new get_block function).

validator/src/journal/block_manager.rs
```diff
@@ -66,6 +66,24 @@ impl BlockRef {
...
+        self.block_manager
+            .get(&[&self.block_id])
+            .next()
+            .unwrap_or_else(|| {
+                panic!(
```
vaporos: If this code bug happens, we should definitely log it, but is it truly unrecoverable? Only the calling code could determine that, not this relatively low-level code.

validator/src/journal/block_manager.rs
ltseeley: Makes sense to me. I removed this commit and modified the chain controller to just use its own `get_block` method

PR #2070, opened 2019-04-12
Add missing dependency to local shell dockerfile

rbuysse:

### DIFF
--- a/docker/sawtooth-shell
+++ b/docker/sawtooth-shell
@@ -21,9 +21,9 @@ FROM ubuntu:bionic
 RUN apt-get update \
  && apt-get install gnupg -y
 
-RUN echo "deb http://repo.sawtooth.me/ubuntu/ci bionic universe" >> /etc/apt/sources.list \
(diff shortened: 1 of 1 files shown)


rbuysse:
Fixes the bug reported here: https://chat.hyperledger.org/channel/sawtooth?msg=E3LqqiGGvEMQrCZhs

rberg2:
also: https://jira.hyperledger.org/browse/STL-1499

danintel:
A silent change is to change the keys from the stable branch to nightly. Would this still work after release? Could we install both keys. Something like what I do:

`apt-key adv . . .--recv-keys 44FC67F19B2466EA 8AA7AF1F1091A5FD`

rbuysse:
@danintel this change brings this development dockerfile in line with the rest in the repository. Previously it was only installing non-sawtooth (cbor, colorlog, toml, etc...) dependencies from the /ci repository, but now that the python-sdk has been moved to its own repo, we need to install it from the published artifact.

We want core developers always using nightly published artifacts from the other repos so we can discover breaking changes prior to releases.

PR #2076, opened 2019-04-18
Raise and handle EngineNotRegistered

peterschwarz:
Raise an `EngineNotRegistered` exception when attempting to activate a consensus engine that has not been registered.  This will inform the callers not to do any following activities for the engine (like sending it a notification) if it hasn't yet connected.

Fixes an issue where duplicate activations were being sent to the engine due to the notifier queuing messages when there was no active engine.

### DIFF
--- a/validator/sawtooth_validator/consensus/proxy.py
+++ b/validator/sawtooth_validator/consensus/proxy.py
@@ -17,6 +17,7 @@ import hashlib
 import logging
 
 from sawtooth_validator.consensus.registry import EngineAlreadyActive
+from sawtooth_validator.consensus.registry import EngineNotRegistered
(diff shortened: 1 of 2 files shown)


validator/sawtooth_validator/consensus/proxy.py
```diff
@@ -152,6 +157,16 @@ def activate_if_configured(self, engine_name, engine_version):
                     engine_name, engine_version)
             except EngineAlreadyActive:
                 return
+            except EngineNotRegistered:
+                # The expectation is that this engine should have been
+                # registered before call this function
```
agunde406: before call -> before calling

peterschwarz: Corrected.

PR #2078, opened 2019-04-22
Build protobufs for Rust sawadm

ltseeley:
Build protobufs when building sawadm and use the locally built protobufs
instead of relying on the SDK. This allows for generating and using
protobufs independently of those provided by the SDK (with ability to
use more).


adm/build.rs
```diff
@@ -0,0 +1,171 @@
+/*
+ * Copyright 2019 Bitwise IO and Intel Corporation
```
jsmitchell: Split copyright onto two lines, like:

Copyright 2019 Bitwise IO, Inc.
Copyright 2019 Intel Corporation

ltseeley: Fixed

PR #2083, opened 2019-04-23
Add compose file for 5 node pbft network

rbuysse:

### DIFF
--- a/docker/compose/sawtooth-default-pbft.yaml
+++ b/docker/compose/sawtooth-default-pbft.yaml
@@ -1,0 +1,468 @@
+# Copyright 2019 Cargill Incorporated
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+version: '3.6'
+
+volumes:
+  pbft-shared:
+
+services:
+
+# -------------=== intkey tp ===-------------
+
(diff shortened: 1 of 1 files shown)


agunde406:
I think this needs to be rebased to pick up some rust lint fixes

docker/compose/sawtooth-default-pbft.yaml
```diff
@@ -0,0 +1,397 @@
...
+        if [ ! -e /etc/sawtooth/keys/validator.priv ]; then
+          sawadm keygen
+          mkdir -p /pbft-shared/validators || true
+          cp /etc/sawtooth/keys/validator.pub /pbft-shared/validators/validator-0.pub
+          cp /etc/sawtooth/keys/validator.priv /pbft-shared/validators/validator-0.priv
+        fi &&
+        if [ ! -e config-genesis.batch ]; then
+          sawset genesis -k /etc/sawtooth/keys/validator.priv -o config-genesis.batch
+        fi &&
+        sleep 30 &&
```
ltseeley: I'm assuming this sleep is to give the other validators time to generate their keys? If so, is there a reason it can't explicitly check for these keys and just wait til they're all there?

docker/compose/sawtooth-default-pbft.yaml
```diff
@@ -0,0 +1,397 @@
...
+
+  intkey-tp-3:
+    image: hyperledger/sawtooth-intkey-tp-python:1.1
+    container_name: sawtooth-intkey-tp-python-default-3
+    expose:
+      - 4004
+    command: intkey-tp-python -C tcp://validator-3:4004
+    stop_signal: SIGKILL
+
+  intkey-tp-4:
```
ltseeley: There are 5 each of the intkey TP, REST API, and settings TP

chenette:
Request: Could this be a five-node network instead of a four-node one?  If so, the doc changes would be simpler and the procedure less wordy.  (The current procedure for PoET has currently 5 nodes).

chenette:
Request: Could this be a five-node network instead of a four-node one? If so, the doc changes would be simpler and the procedure less wordy. (The current procedure for PoET has currently 5 nodes).

PR #2084, opened 2019-04-24
Arch doc: Add info on low-s normalized sig for txn + batch headers

chenette:
Update "Transactions and Batches" in the Sawtooth Architecture Guide: 
    
- Add  note with information on low-s normalization for the transaction and batch headers, along with info that the python3 SDK and others do this for you, but some libraries don't.
   
- Change the existing note to a tip so it'll be visually distinct (different color).

- Also: Fixed the identical header text (sphinx reports that as an error) and added bullets so "Transaction Header, Signature, and Public Keys" is easier to skim-read.

This doc change addresses the issue reported by STL-1500.

### DIFF
--- a/docs/source/architecture/transactions_and_batches.rst
+++ b/docs/source/architecture/transactions_and_batches.rst
@@ -29,8 +29,8 @@ message types:
    :caption: File: protos/transaction.proto
    :linenos:
 
-Header, Signature, and Public Keys
-----------------------------------
+Transaction Header, Signature, and Public Keys
+----------------------------------------------
 
 The Transaction header field is a serialized version of a TransactionHeader.
 The header is signed by the signer's private key (not sent with the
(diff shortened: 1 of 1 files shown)


docs/source/architecture/transactions_and_batches.rst
```diff
@@ -132,10 +149,10 @@ message types:
    :caption: File: protos/batch.proto
    :linenos:
 
-Header, Signature, and Public Keys
-----------------------------------
+Batch Header, Signature, and Public Keys
+----------------------------------------
 
-Following the pattern presented in Transaction, the Batch header field is a
+Following the pattern presented in a Transaction, the Batch header field is a
```
jsmitchell: the article is actually "the definition of the Transaction structure", not "an instance of Transaction"

chenette: Thanks.  Rewrote to clarify: "the Transaction data structure".

PR #2086, opened 2019-04-29
 Make sawadm genesis required settings optional

ltseeley:
Adds the `--ignore-required-settings` flag to the `sawadm genesis`
command that allows bypassing the required settings check. Because the
sawtooth settings transaction family is a reference implementation, it's
possible that a different settings transaction family could be used to
set the required settings at genesis. This bypass allows `sawadm
genesis` to pass in this situation. However, the genesis chain
controller will still fail if the required settings aren't in state
as a result of the genesis block, so the same requirements are enforced.

### DIFF
--- a/adm/src/main.rs
+++ b/adm/src/main.rs
@@ -110,7 +110,7 @@ fn parse_args<'a>() -> ArgMatches<'a> {
             (@arg input_file:
              +takes_value ... "file or files containing batches to add to the resulting")
             (@arg output: -o --output "choose the output file for GenesisData")
-            (@arg ignore_required_settings: --ignore-required-settings
+            (@arg ignore_required_settings: --("ignore-required-settings")
              "skip the check for settings that are required at genesis (necessary if using a
               settings transaction family other than sawtooth_settings)"))
         (@arg verbose: -v... "increase the logging level.")
--- a/adm/src/main.rs
+++ b/adm/src/main.rs
@@ -109,7 +109,10 @@ fn parse_args<'a>() -> ArgMatches<'a> {
             (about: "creates the genesis.batch file for initializing the validator")
             (@arg input_file:
              +takes_value ... "file or files containing batches to add to the resulting")
-            (@arg output: -o --output "choose the output file for GenesisData"))
+            (@arg output: -o --output "choose the output file for GenesisData")
+            (@arg ignore_required_settings: --ignore-required-settings
+             "skip the check for settings that are required at genesis (necessary if using a
+              settings transaction family other than sawtooth_settings)"))
         (@arg verbose: -v... "increase the logging level.")
     );
     app.get_matches()
--- a/adm/src/main.rs
+++ b/adm/src/main.rs
@@ -107,8 +107,8 @@ fn parse_args<'a>() -> ArgMatches<'a> {
             (@arg quiet: -q --quiet "do not display output"))
         (@subcommand genesis =>
             (about: "creates the genesis.batch file for initializing the validator")
-         (@arg input_file:
-          +takes_value ... "file or files containing batches to add to the resulting")
+            (@arg input_file:
+             +takes_value ... "file or files containing batches to add to the resulting")
             (@arg output: -o --output "choose the output file for GenesisData"))
         (@arg verbose: -v... "increase the logging level.")
     );
--- a/cli/sawtooth_cli/admin_command/genesis.py
+++ b/cli/sawtooth_cli/admin_command/genesis.py
@@ -39,12 +39,13 @@ def add_genesis_parser(subparsers, parent_parser):
         'initializing the validator.',
         epilog='This command generates a serialized GenesisData protobuf '
         'message and stores it in the genesis.batch file. One or more input '
-        'files (optional) can contain serialized BatchList protobuf messages '
-        'to add to the GenesisData. The output shows the location of this '
-        'file. By default, the genesis.batch file is stored in '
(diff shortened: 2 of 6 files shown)


ltseeley:
> This looks good. For future-proofing, will the `--ignore-required-settings` flag also ignore settings that it does not recognize?For example, if I wanted to add `sawtooth.consensus.algorithm.version` in Sawtooth 1.0 it would fail saying the setting is invalid or something like that. It would be nice to be able to add settings in the current version that will not be used until a future version.

I'm not sure I understand the question. Are you referring to setting that setting on genesis in 1.0? Or after genesis?

All the required settings check does is enforce that whatever settings are needed for the current version (for 1.2 it's only `sawtooth.consensus.algorithm.name/version`) are set at genesis when a new chain is created for that version.

danintel:
OK, forget 1.0--that was just an example.  All I want is the ability to add a setting at genesis that Sawtooth does not currently recognize. Another example, suppose Sawooth 3.0 adds `sawtooth.consensus.algorithm.foobar` and I want to add the setting an older version of Sawtooth, even though it will be unused.

ltseeley:
> OK, forget 1.0--that was just an example. All I want is the ability to add a setting at genesis that Sawtooth does not currently recognize. Another example, suppose Sawooth 3.0 adds `sawtooth.consensus.algorithm.foobar` and I want to add the setting an older version of Sawtooth, even though it will be unused.

There isn't anything that will prevent you from adding that setting. You can add any arbitrary settings at genesis. The only rules are that some settings **must** be set; currently, that's the name/version settings. The ``--ignore-required-settings`` just skips enforcing that rule.

danintel:
There was something preventing me from setting `sawtooth.consensus.algorithm.version` (and `.name`) in Sawtooth 1.0. Perhaps is just because it was expecting `sawtooth.consensus.algorithm` it errored out?

If there is nothing preventing unrecognized settings in the  general case, that is good--because the settings are supposed to be extensible anyway.

adm/src/main.rs
```diff
@@ -109,7 +109,9 @@ fn parse_args<'a>() -> ArgMatches<'a> {
             (about: "creates the genesis.batch file for initializing the validator")
             (@arg input_file:
              +takes_value ... "file or files containing batches to add to the resulting")
-            (@arg output: -o --output "choose the output file for GenesisData"))
+            (@arg output: -o --output "choose the output file for GenesisData")
+            (@arg ignore_required_settings: -i --ignore_required_settings
```
peterschwarz: I don't think there should be a short flag for this

ltseeley: Fixed

adm/src/main.rs
```diff
@@ -107,9 +107,11 @@ fn parse_args<'a>() -> ArgMatches<'a> {
             (@arg quiet: -q --quiet "do not display output"))
         (@subcommand genesis =>
             (about: "creates the genesis.batch file for initializing the validator")
-         (@arg input_file:
-          +takes_value ... "file or files containing batches to add to the resulting")
-            (@arg output: -o --output "choose the output file for GenesisData"))
+            (@arg input_file:
+             +takes_value ... "file or files containing batches to add to the resulting")
+            (@arg output: -o --output "choose the output file for GenesisData")
+            (@arg ignore_required_settings: --ignore_required_settings
```
peterschwarz: This should be `--"ignore-required-settings"` (no underscores in CLI flags)

ltseeley: Done

adm/src/main.rs
```diff
@@ -107,9 +107,11 @@ fn parse_args<'a>() -> ArgMatches<'a> {
...
         (@subcommand genesis =>
             (about: "creates the genesis.batch file for initializing the validator")
-         (@arg input_file:
-          +takes_value ... "file or files containing batches to add to the resulting")
-            (@arg output: -o --output "choose the output file for GenesisData"))
+            (@arg input_file:
+             +takes_value ... "file or files containing batches to add to the resulting")
+            (@arg output: -o --output "choose the output file for GenesisData")
+            (@arg ignore-required-settings: --ignore-required-settings
+             "skip checking for settings that are required at genesis"))
```
chenette: Wording nitpick: "skip the check for required settings"
or
"skip the check for settings that are required at genesis"

Also, would be to add a parenthetical phrase explaining why you'd do this; for example:

"skip the check for required settings (intended for backward and forward compatibility)"

... or something like that.

cli/sawtooth_cli/admin_command/genesis.py
```diff
@@ -54,6 +60,11 @@ def add_genesis_parser(subparsers, parent_parser):
         help='file or files containing batches to add to the resulting '
         'GenesisData')
 
+    parser.add_argument(
+        '--ignore-required-settings',
+        action='store_true',
+        help='skip checking for settings that are required at genesis')
```
chenette: Make the same change(s) as on line 114 of main.rs.

adm/src/main.rs
peterschwarz: It's not about backward compatibility but for supporting alternative settings transaction families.  The Sawtooth settings family is a reference implementation.

danintel: Correct. Backward compatibility would be a nice bonus.

ltseeley: Updated

cli/sawtooth_cli/admin_command/genesis.py
ltseeley: Updated

PR #2092, opened 2019-05-01
Add PBFT steps to Docker NW proc; correct existing info

chenette:
Update the App Dev Guide's multi-node procedure, "Using Docker for a Sawtooth Test Network", to add PBFT info, steps, and example information. This new info requires some corrections or rewrites to existing content. 

Also updated the docs/Makefile to copy the example PBFT yaml file to the same docs directory as the PoET yaml file.
For example, the step showing how to run `ps` in the PoET Validator Registry container now shows how to check that the REST API process is running.

This PR also corrects and clarifies existing information (text and steps) in this procedure:
    
- Add prerequisites for Docker and Docker Engine; clarify info on shutting down a single node, if running
- Clarify which container is being used in each step
- Fix error in "sawtooth proposal create" (intro text was wrong)
- Add example output to steps that needed it (for clarification)
- Correct prompts throughout
- Update obsolete terminology: validator node -> Sawtooth node (or just "node")

### DIFF
--- a/docs/source/app_developers_guide/docker_test_network.rst
+++ b/docs/source/app_developers_guide/docker_test_network.rst
@@ -3,9 +3,14 @@
 Using Docker for a Sawtooth Test Network
 ========================================
 
-In this procedure, you will use a Docker Compose file that creates a new
-application development environment with five validator nodes and four
-transaction processors (Settings, IntegerKey, XO, and PoET Validator Registry).
+This procedure describes how to use Docker to create a network of five Sawtooth
+nodes for an application development environment, using either PBFT or PoET
+consensus. (Devmode consensus is not recommended for a network.)
+
+.. include:: ../_includes/pbft-vs-poet-cfg.inc
+
+
+.. _about-sawtooth-nw-env-docker-label:
 
 About the Sawtooth Network Environment
 --------------------------------------
@@ -17,17 +22,13 @@ The following figure shows an example network with two validator nodes:
    :align: center
    :alt: Docker: Sawtooth network with five nodes
 
-Like the single-node environment, this environment uses parallel transaction
-processing and static peering. However, it has the following differences:
+.. include:: ../_includes/about-nw-each-node-runs.inc
 
-* PoET simulator consensus instead of dev mode, because dev mode's random-leader
-  consensus is not recommended for multi-node or production networks. Sawtooth
-  offers two versions of :term:`PoET consensus`. PoET-SGX relies on Intel
-  Software Guard Extensions (SGX) to implement a leader-election lottery system.
-  PoET simulator provides the same consensus algorithm on an SGX simulator.
+Like the :doc:`single-node test environment <docker>`, this environment uses
+parallel transaction processing and static peering.
 
-* An additional transaction processor, PoET Validator Registry, handles PoET
-  settings for a multiple-node network.
+
+.. _prereqs-multi-docker-label:
 
 Prerequisites
 -------------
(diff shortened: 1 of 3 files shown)


docs/source/app_developers_guide/docker_test_network.rst
```diff
@@ -212,59 +303,87 @@ setting.
...
+      root@1bee7c8d9# tail /var/log/sawtooth/settings-*-debug.log
       .
       .
       .
       [22:18:33.137 [MainThread] core DEBUG] received message of type: TP_PROCESS_REQUEST
-      [22:18:33.219 [MainThread] handler INFO] Setting setting sawtooth.validator.transaction_families changed from None to [{"family": "intkey", "version": "1.0"}, {"family":"sawtooth_settings", "version":"1.0"}, {"family":"xo", "version":"1.0"}, {"family":"sawtooth_validator_registry", "version":"1.0"}]
+      [22:18:33.219 [MainThread] handler INFO] Setting setting sawtooth.validator.transaction_families changed from None to [{"family": "intkey", "version": "1.0"}, {"family":"sawtooth_settings", "version":"1.0"}, {"family":"xo", "version":"1.0"}, ...
 
 #. Run the following command to check the setting change. You can use any
    container for this step. Also, you can specify any REST API on the network;
```
agunde406: This is confusing, you can't use any container but you can use any validator container or the shell container. I tried it a rest API container for example and it does not work.

docs/source/app_developers_guide/docker_test_network.rst
```diff
@@ -212,59 +303,87 @@ setting.
...
+        sawtooth.poet.report_public_key_pem: -----BEGIN PUBLIC KEY-----
+        MIIBIjANBgkqhki...
+        sawtooth.poet.target_wait_time: 5
+        sawtooth.poet.valid_enclave_basenames: b785c58b77152cbe7fd55ee3...
+        sawtooth.poet.valid_enclave_measurements: c99f21955e38dbb03d2ca...
+        sawtooth.publisher.max_batches_per_block: 100
+        sawtooth.settings.vote.authorized_keys: 036631291bbe87c3c9dde22...
+        sawtooth.validator.transaction_families: [{"family": "intkey", ...
+
+     ??? REVIEWERS: WHAT ABOUT ZTEST? `sawtooth.poet.ztest_minimum_win_count: 100000`
```
agunde406: ztest is not set in the docker-compose file used in this procedure.

chenette: That's my question: Why not?  It used to be set in a previous version of the file.  Was it removed on purpose?  Will the default z-test setting cause any problems if the network runs for a while?

agunde406: Setting the ztest to a high number is a hack to avoid a situation where the nodes get deadlocked because they all fail the ztest at the same time. However, this feels like a weird place to offer that information. Maybe it should be in a note or tip?

agunde406: (i do not know if we removed it on purpose or not)

docs/source/app_developers_guide/docker_test_network.rst
chenette: Fixed - changed "any container" to "the shell container or any validator container".

docs/source/app_developers_guide/docker_test_network.rst
chenette: What is the default ztest value?  (I can't find it in the current docs or the FAQ.)  Perhaps it defaults to 1000000, and that's why the setting no longer appears in `sawtooth settings list`.

chenette: Talked to @peterschwarz, who says that the default ztest value is fine for an app-dev test network. I shall remove my question from the doc and leave the example output as it is.

chenette: Removed question; marking this conversation as resolved.

docs/source/app_developers_guide/docker_test_network.rst
```diff
@@ -3,124 +3,174 @@
...
 ----------------------------------------
 
-Download the Docker Compose file for a multiple-node network,
-`sawtooth-default-poet.yaml <./sawtooth-default-poet.yaml>`_.
-Save this file in the same directory as the single-node compose file
-(``sawtooth-default.yaml``).
+Download the Docker Compose file for a multiple-node network.
 
+* For PBFT: Download
+  `sawtooth-default-pbft.yaml <./sawtooth-default-poet.yaml>`_
```
rbuysse: Link is to poet dockerfile.

You'll also need to add a step to the Makefile to copy the compose file to the correct spot. Here's how the poet compose file is copied: https://github.com/hyperledger/sawtooth-core/blob/master/docs/Makefile#L71

chenette: I fixed the borked link for the sawtooth-default-pbft.yaml file. Thanks for catching this silly error.

For the Makefile change: I assumed that this had already been done in the PR that added the PBFT yaml file.  It appears not.  Let's do this in a separate PR.

chenette: Created a separate commit for the docs/Makefile.

PR #2095, opened 2019-05-02
Fix peer configuration in PBFT docker compose

ltseeley:
Fixes the peering configuration in
`docker/compose/sawtooth-default-pbft.yaml`. Peering should only be
specified once between two validators; it was being specified in both
directions before, which lead to duplicate peering.

### DIFF
--- a/docker/compose/sawtooth-default-pbft.yaml
+++ b/docker/compose/sawtooth-default-pbft.yaml
@@ -238,11 +238,7 @@ services:
           --bind network:tcp://eth0:8800 \
           --scheduler parallel \
           --peering static \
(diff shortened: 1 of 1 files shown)


docker/compose/sawtooth-default-pbft.yaml
```diff
@@ -238,11 +238,7 @@ services:
           --bind network:tcp://eth0:8800 \
           --scheduler parallel \
           --peering static \
-          --maximum-peer-connectivity 10000 \
```
agunde406: Did you mean to delete this here but now where else?

PR #2096, opened 2019-05-06
Add Connect to REST API steps and links

chenette:
In the Docker, Kubernetes, AWS, and Ubuntu procedures:
- Add missing step to confirm REST API connectivity, if necessary.
- Rework existing "connect to Sawtooth" step to be consistent across the four
  procedures.
- Change titles so they're unique across the doc set (for clarity and because
  Sphinx complains).
- Add anchor tags for links from XO procedure.

In the "Playing with XO" procedure:
- Separate the first step (connect to Sawtooth) and expand with links to
  Docker, Kubernetes, and AWS. Text info suffices for Ubuntu.
- For "confirm REST API connectivity" step, correct or add links for the
  four procedures.

### DIFF
--- a/docs/source/app_developers_guide/aws.rst
+++ b/docs/source/app_developers_guide/aws.rst
@@ -63,15 +63,16 @@ family.
    genesis block. To learn how the typical startup process works, see
    :doc:`ubuntu`.
 
+.. _launch-sawtooth-instance-aws:
 
-Step 1: Launch a Sawtooth Instance
-==================================
+Step 1: Launch a Sawtooth Instance with AWS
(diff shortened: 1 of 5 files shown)


chenette:
I assumed that curl was always available in a container.  Is this not the case?

chenette:
@peterschwarz - I confirmed that curl is available in the Kubernetes shell container.  However, I learned that curl isn't installed by default for Ubuntu. So I added an intro sentence that says "If curl is installed, you can use this step ...".

(The Docker and AWS procedures are okay - the curl step has been there all along.)

PR #2103, opened 2019-05-10
Correct some issues with future cleanup

peterschwarz:


peterschwarz:
The main driver is the first commit, without which the validator will throw periodic exceptions due to concurrent modification of the collection.  The second commit removes the callers needs to manage the future collection externally, since its role is to hold onto those futures until they are resolved.

validator/sawtooth_validator/networking/future.py
```diff
@@ -127,24 +127,26 @@ def get(self, correlation_id):
 
     def remove(self, correlation_id):
         try:
-            del self._futures[correlation_id]
+            return self._futures.pop(correlation_id)
```
agunde406: This looks like a separate change? Should this be in the next commit?

peterschwarz: Good catch - moved it to the later commit.

validator/sawtooth_validator/networking/future.py
```diff
@@ -133,18 +133,20 @@ def remove(self, correlation_id):
...
-        correlation_ids = [
-            correlation_id
-            for correlation_id, future
-            in self._futures.items()
-            if future.is_expired()
-        ]
-        for correlation_id in correlation_ids:
-            self._futures[correlation_id].timer_stop()
-            del self._futures[correlation_id]
+        with self._lock:
```
agunde406: The commit message is confusing. `resolve` is not a method. You updated the remove_expired function. Can you redo the commit message.

PR #2117, opened 2019-05-17
Revert "Avoid duplicate peering request from same endpoint"

peterschwarz:
Reverts hyperledger/sawtooth-core#2106

#2106 introduced a race condition, that will need to be sorted out before a proper fix for this issue can be put in place.


peterschwarz:
Going to correct the DCO and where the branch should be

PR #2119, opened 2019-05-20
Use native Rust commit store in Publisher

ltseeley:
Updates the block publisher to use the native CommitStore for checking
committed transactions and batches rather than using the Python FFI.

### DIFF
--- a/validator/src/journal/chain_commit_state.rs
+++ b/validator/src/journal/chain_commit_state.rs
@@ -17,11 +17,9 @@
 
 use std::collections::HashSet;
 
-use cpython;
-use cpython::ObjectProtocol;
-
 use batch::Batch;
 use journal::block_manager::BlockManager;
+use journal::commit_store::CommitStore;
 use transaction::Transaction;
 
 #[derive(Debug, PartialEq)]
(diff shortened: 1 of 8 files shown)


validator/src/journal/chain_commit_state.rs
```diff
@@ -163,17 +160,12 @@ impl TransactionCommitCache {
...
-
-    fn blockstore_has_txn(&self, transaction_id: &str) -> bool {
-        let gil = cpython::Python::acquire_gil();
-        let py = gil.python();
-        self.transaction_committed
-            .call(py, (transaction_id,), None)
-            .expect("Call to determine if transaction is committed failed")
-            .extract::<bool>(py)
-            .unwrap()
+        // Shouldn't expect here
```
jsmitchell: Can you clarify this comment?

ltseeley: This was actually intended as a comment to myself (and I forgot to do anything about it before marking the PR as ready for review). I'm not sure how we should handle these results. It looks like these calls would fail due to a database error; are these errors recoverable? If they are, it's unclear how they should be handled by the publisher.

PR #2128, opened 2019-05-29
Use user keys for genesis and proposal in Kubernetes files

rberg2:
We want to use the user key here.


docker/kubernetes/sawtooth-kubernetes-default.yaml
```diff
@@ -63,7 +63,7 @@ items:
               - -c
               - "sawadm keygen \
               && sawtooth keygen my_key \
```
peterschwarz: This key isn’t used anymore, after this change.

rberg2: This has been addressed.

rberg2: FWIW the root cause was not using the same key for genesis and the proposal.

PR #2129, opened 2019-05-30
Disambiguate last on LmdbDatabaseReaderCursor

peterschwarz:
Disambiguate the use of `last` on `LmdbDatabaseReaderCursor` between its own method and that of Iterator. This commit renames `LmdbDatabaseReaderCursor`'s methods to `seek_first` and `seek_last` to differentiate the meaning.

This fixes the performance regression that occurs with `get_chain_head`.


peterschwarz:
Rebased on #2127

PR #2130, opened 2019-05-30
Improve consensus backwards compatibility

ltseeley:
Adds a list of additional protocols (name/version pairs) that a consensus engine can support. Also updates how consensus engine name/version settings are inferred for backwards compatibility.

### DIFF
--- a/validator/sawtooth_validator/consensus/registry.py
+++ b/validator/sawtooth_validator/consensus/registry.py
@@ -25,13 +25,14 @@ class EngineAlreadyActive(Exception):
     """The engine is already active."""
 
 
-EngineInfo = namedtuple('EngineInfo', ['connection_id', 'name', 'version'])
+EngineInfo = namedtuple(
+    'EngineInfo', ['connection_id', 'name', 'version', 'additional_protocols'])
 
 
 class ConsensusRegistry:
-    """A thread-safe construct that stores the connection_id, name, and version
-    of all registered consensus engines, and tracks which engine is currently
-    active."""
+    """A thread-safe construct that stores the connection_id, name, version,
+    and any additional supported protocols of all registered consensus engines,
+    and tracks which engine is currently active."""
 
     def __init__(self):
         self._registry = []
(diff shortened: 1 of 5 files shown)


validator/sawtooth_validator/consensus/registry.py
```diff
@@ -45,27 +46,38 @@ def has_active_engine(self):
...
             return self._active is not None
 
-    def register_engine(self, connection_id, name, version):
+    def register_engine(self, connection_id, name, version,
+                        additional_protocols):
         with self._lock:
-            # If this engine is already registered, remove the old connection
+            # If this engine is already registered, remove the old connection.
+            # If there is an existing engine that supports one of the same
+            # protocols as this engine, unregister it.
```
ltseeley: The biggest open question for me is the mechanics of this; what should we do if another engine gets unregistered? Should we notify it? What if it was the active engine, and this engine that displaced it doesn't support the active protocol?

Ultimately, this is up to the sysadmin; really you shouldn't be using overlapping protocols unless something strictly replaces another engine (e.g. an upgrade). Maybe we could enforce that in order to replace an engine, the new one must support a superset of the displaced engine's protocols?

validator/tests/test_consensus/tests.py
```diff
@@ -413,7 +413,7 @@ def __setitem__(self, key, value):
 
 class MockConsensusRegistry(Mock):
     def get_active_engine_info(self):
-        return EngineInfo('mock-id', 'mock-name', 'mock-version')
+        return EngineInfo('mock-id', 'mock-name', 'mock-version', [('n', 'v')])
```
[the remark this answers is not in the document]
ltseeley: Updated

validator/sawtooth_validator/consensus/registry.py
ltseeley: Updated so that more than one engine can support a given protocol; the engine that registered first will become active if that protocol is activated. If a registering engine supports all of the same protocols as an existing registered engine (protocols are a superset), then the existing engine is unregistered (this is deemed to be an upgrade).

validator/sawtooth_validator/consensus/registry.py
```diff
@@ -45,27 +46,46 @@ def has_active_engine(self):
...
+
+            def is_replaced(eng):
+                return (
+                    engine_handles_protocol(eng.name, eng.version, engine_info)
+                    and all([True for (n, v) in eng.additional_protocols
+                             if engine_handles_protocol(n, v, engine_info)]))
+
             self._registry = list(filter(
-                lambda e: e.name != name and e.version != version,
+                lambda e: not is_same_engine(e) and not is_replaced(e),
```
peterschwarz: Can you add some tests for these scenarios?

ltseeley: Added registry tests in a new commit

PR #2131, opened 2019-06-04
Add new About Dynamic Consensus topic; clean up terms

chenette:
Add new topic "About Dynamic Consensus"
- Summarizes admin tasks for setting consensus in a new network
   and for changing consensus on a running network.
- For each consensus type, lists requirements and on-chain settings
- Adds links to this topic from the Introduction, App Dev Guide (Ubuntu network
   procedure) and Sys Admin Guide
- Documents consensus backwards compatibility - includes information about
   consensus defaults                                                            
 
Clean up consensus terms in Sawtooth docs
- Change "dev mode" to "Devmode" to match the consensus engine name
- Make other terms consistent and coherent (in the glossary and elsewhere)

### DIFF
--- a/docs/source/architecture/journal.rst
+++ b/docs/source/architecture/journal.rst
@@ -310,8 +310,9 @@ necessary transaction processors must be running.
 If any transactions in the genesis block set or change settings, Sawtooth
 requires the `Sawtooth Settings transaction processor <../cli/settings-tp>`_
 or an equivalent implementation.
-For example, if the genesis block configures PoET consensus, this transaction
-processor is handles the transactions with PoET settings.
+For example, the genesis block specifies the :term:`consensus engine` and
+related settings, so the Settings transaction processor is required to handle
+the transactions with these settings.
 
 When the genesis block is committed, the consensus settings are stored in
 state. All subsequent blocks are processed with the configured consensus
--- a/docs/source/sysadmin_guide/about_dynamic_consensus.rst
+++ b/docs/source/sysadmin_guide/about_dynamic_consensus.rst
@@ -1,0 +1,315 @@
+.. _about-dynamic-consensus-label:
+
+***********************
+About Dynamic Consensus
+***********************
+
+Sawtooth :term:`dynamic consensus` lets you choose from a variety of consensus
+algorithms for the network.
+
+* Each consensus type has a :term:`consensus engine` that communicates with the
+  validator through the :term:`consensus API`. Each node in the network must run
+  the same consensus engine.
+
+* The validator listens for the consensus engine on the `consensus endpoint`
+  (by default, ``tcp://127.0.0.1:5050``). For more information, see
+  :doc:`/sysadmin_guide/off_chain_settings`.
+
+* The consensus type is controlled by the on-chain settings
+  ``sawtooth.consensus.algorithm.name`` and
+  ``sawtooth.consensus.algorithm.version``. These settings are required.
+
+The rest of this topic summarizes how to configure consensus and lists the
+settings and requirements for each supported consensus type:
+:ref:`PBFT <pbft-consensus-label>`,
+:ref:`PoET <poet-consensus-label>`,
+:ref:`Raft <raft-consensus-label>`, and
+:ref:`Devmode <devmode-consensus-label>`.
+
+.. note::
+
+   **Compatibility with Sawtooth 1.0**:
+   Sawtooth provides defaults for backward compatibility with previous versions
+   of Sawtooth (1.0 and earlier), when Devmode and PoET were the only supported
+   consensus algorithms.
+
+   * If ``sawtooth.consensus.algorithm.version`` is not set, the default version
+     is ``0.1``.
+
+   * If ``sawtooth.consensus.algorithm.name`` is not set, Sawtooth checks for
+     the deprecated setting ``sawtooth.consensus.algorithm``. If this setting
+     exists, Sawtooth uses the specified consensus.
+
+   * Otherwise, Sawtooth uses Devmode consensus.
+
+   **Important:** In release 1.1 and later, a consensus engine is always
+   required. The same consensus engine must be running on all nodes in the
+   network.
+
+
+Configuring Consensus for a New Network
+=======================================
+
+To configure the initial consensus, create a consensus proposal for the
+:term:`genesis block`, then start the consensus engine and any transaction
+processors that are required by that consensus type.
+
+1. The administrator of the first node uses the command ``sawset proposal
+   create`` to specify the consensus settings in the genesis block.
+   See :doc:`/sysadmin_guide/setting_up_sawtooth_poet-sim` for this procedure.
+
+#. On each node, the administrator installs and starts the consensus engine and
+   any transaction processors required for consensus (such as PoET Validator
+   Registry). See :doc:`/sysadmin_guide/systemd` for this procedure.
+
+When each node joins the network, it reads the on-chain consensus settings, then
+starts using the consensus engine to process blocks.
+
+
+Changing Consensus on a Running Network
+=======================================
+
+To change to a different consensus, start the new consensus engine on all nodes,
+plus any transaction processors that are required by that consensus type. Then
+submit a consensus change proposal.
+
+.. note::
+
+   Administrators should coordinate the change to a new consensus. For some
+   consensus types, the network could slow or become stalled if some nodes have
+   not started the new consensus engine.
+
+1. If the network has forked, stop all nodes except the one with the preferred
+   blocks.
+
+#. Each administrator installs and starts the new consensus engine and any
+   related transaction processors. See :doc:`/sysadmin_guide/installation` and
+   :doc:`/sysadmin_guide/systemd` for these procedures.
+
+#. One administrator uses the command ``sawset proposal create`` to submit a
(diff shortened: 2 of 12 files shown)


docs/source/architecture/journal.rst
```diff
@@ -310,8 +310,10 @@ necessary transaction processors must be running.
 If any transactions in the genesis block set or change settings, Sawtooth
 requires the `Sawtooth Settings transaction processor <../cli/settings-tp>`_
 or an equivalent implementation.
-For example, if the genesis block configures PoET consensus, this transaction
-processor is handles the transactions with PoET settings.
+For example, the genesis block specifies the :term:`consensus engine` and
+related settings (unless :term:`Devmode consensus` is used), so the Settings
```
ltseeley: Why "unless Devmode consensus is used"?

docs/source/sysadmin_guide/about_dynamic_consensus.rst
```diff
@@ -0,0 +1,247 @@
...
+   Sawtooth provides defaults for backward compatibility with previous versions
+   of Sawtooth (1.0 and earlier), when Devmode and PoET were the only available
+   consensus algorithms.
+
+   * If ``sawtooth.consensus.algorithm.version`` is not set, the default version
+     is ``0.1``.
+
+   * If ``sawtooth.consensus.algorithm.name`` is not set, Sawtooth checks for
+     the deprecated setting ``sawtooth.consensus.algorithm``, which was used in
+     previous releases to specify PoET consensus. If this setting exists,
```
ltseeley: This isn't 100% accurate; if this setting exists, it'll use that as the name. If this setting isn't `poet`, Sawtooth won't use PoET.

docs/source/sysadmin_guide/about_dynamic_consensus.rst
```diff
@@ -0,0 +1,247 @@
...
+  (such as PoET Validator Registry).
+
+The rest of this topic summarizes the settings and requirements for each
+available consensus type: :ref:`PBFT <pbft-consensus-label>`,
+:ref:`PoET <poet-consensus-label>`, :ref:`Raft <raft-consensus-label>`, and
+:ref:`Devmode <devmode-consensus-label>`.
+
+.. note::
+
+   **Default Consensus Settings:**
```
ltseeley: I think it would be worth adding something to the beginning of this section that the `sawtooth.consensus.algorithm.name/version` settings are required. for current versions; if you only read this, you might think what's explained below is something you could use, but you can't. Also, "Default Consensus Settings" probably isn't the best term to use, because there's not "default" consensus (as you explain below). "Backwards Compatibility of Consensus Settings" or something similar would be better.

docs/source/sysadmin_guide/about_dynamic_consensus.rst
```diff
@@ -0,0 +1,247 @@
...
+  See :doc:`/sysadmin_guide/creating_genesis_block` for more information on the
+  version and validator public keys for the PBFT member list .
+
+
+.. _poet-consensus-label:
+
+PoET Consensus
+==============
+
+:term:`PoET <PoET consensus>` is a Nakamoto-style (voting) consensus algorithm
```
ltseeley: Should be "lottery", not "voting"

docs/source/sysadmin_guide/about_dynamic_consensus.rst
```diff
@@ -0,0 +1,247 @@
...
+* Use these on-chain settings for the Devmode consensus engine:
+
+  .. code-block:: none
+
+     sawtooth.consensus.algorithm.name=Devmode
+     sawtooth.consensus.algorithm.version=0.1
+
+  .. note::
+
+     These settings are optional, because Devmode version 0.1 is the default
```
ltseeley: These settings are **not** optional; the genesis validator will fail if they are not specified in 1.2. The backwards compatibility "defaults" are only for the purpose of backwards compatibility. All new networks must explicitly have the name and version settings.

docs/source/architecture/journal.rst
danintel: I wondered if there's any DevMode-specific settings, and it seems there are:
```
sawtooth.consensus.max_wait_time
    Maximum devmode consensus wait time, in seconds
sawtooth.consensus.min_wait_time
    Minimum devmode consensus wait time, in seconds
```
Unfotunately, they don't have `devmode` in the name to make it obvious they are DevMode-specific

chenette: Thanks, @danintel - I'll add these settings to the Devmode section.

docs/source/sysadmin_guide/about_dynamic_consensus.rst
```diff
@@ -0,0 +1,247 @@
...
+  transaction processors required for consensus (such as PoET Validator
+  Registry).
+  See :doc:`/sysadmin_guide/setting_up_sawtooth_poet-sim` for this procedure.
+
+* To change consensus for a running network, administrators must coordinate the
+  changes for all nodes. Each administrator must install and start the new
+  consensus engine. Then one administrator submits a transaction to change the
+  on-chain consensus settings.
+
+  After the consensus has changed, each administrator can stop the previous
```
agunde406: It is worth mentioning that if a node is added or is catching up, they would need to have the correct consensus engines/tp connected to handle all the previous versions of consensus used to build the chain. For example if the first half of the chain used  x and then switched to y a new node would need x and y until it caught up to the other nodes.

docs/source/sysadmin_guide/about_dynamic_consensus.rst
```diff
@@ -0,0 +1,247 @@
...
+     sawtooth.consensus.algorithm.name=PoET
+     sawtooth.consensus.algorithm.version=0.1
+     sawtooth.poet.report_public_key_pem="$(cat * /etc/sawtooth/simulator_rk_pub.pem)"
+     sawtooth.poet.valid_enclave_measurements=$(poet enclave measurement)
+     sawtooth.poet.valid_enclave_basenames=$(poet enclave basename)
+
+  .. note::
+
+     For a PoET test network with less than 10 or 12 nodes, disable the
+     defense-in-depth tests (ztest) with these settings:
```
agunde406: These settings disable the z test, c test and k test not just the ztest.

docs/source/architecture/journal.rst
chenette: @danintel - Your wonderful FAQ also lists this setting for Devmode:

`sawtooth.consensus.valid_block_publishers`
    `List of public keys for allowed block publishers. For devmode`

@ltseeley - Do you know if this setting is used in Sawtooth 1.2?

ltseeley: It's not used in 1.1 or 1.2; the validator code it's used in is actually dead code. It's not in the Devmode engine either, so it looks like that's deprecated.

docs/source/sysadmin_guide/about_dynamic_consensus.rst
chenette: 💡

chenette: Would it make sense to tell admins to just keep the old engines/TPs running in case a node is added later?  (And, of course for slow nodes who aren't caught up yet.)

ltseeley: Well technically, the node that's caching up would only need x until it reaches the block where the network switched to y. To answer your question, @chenette, existing nodes won't have to keep old stuff running; when a new node is added, it will have to run the old engine + TPs.

chenette:
I pushed some fixup commits with the requested changes, plus a rewrite for the intro of "About Dynamic Consensus".  

- This topic now starts with three bullet points that summarize the consensus engine, consensus endpoint, and required consensus settings. 

- I moved the bullets on setting/changing consensus into their own subsections, "Configuring consensus for a New Network" and "Changing Consensus on a Running Network". I also expanded each paragraph into a summary of the steps (now numbered) with additional notes, tips, and links to the actual procedures.

docs/source/sysadmin_guide/about_dynamic_consensus.rst
```diff
@@ -0,0 +1,314 @@
...
+     is ``0.1``.
+
+   * If ``sawtooth.consensus.algorithm.name`` is not set, Sawtooth checks for
+     the deprecated setting ``sawtooth.consensus.algorithm``. If this setting
+     exists, Sawtooth uses the specified consensus.
+
+   * Otherwise, Sawtooth uses Devmode consensus.
+
+   **Important:** A consensus engine is always required. (There is no default
+   consensus engine.)
```
danintel: This is ambiguous to me. There is a default consensus algorithm (Devmode). But a separate consensus engine process must always be ran--and I think this statement is referring to the engine process.  Perhaps change this to something like:

> **Important:** A consensus engine is always required _to be running on all nodes_. (There is no default consensus engine.)

chenette: @danintel - How about this wording?

*Important:* In release 1.1 and later, a consensus engine is always required. The same consensus engine must be running on all nodes in the network.

danintel: Looks good. Thanks!

PR #2132, opened 2019-06-05
Finalize the schedule on Cancel of the Candidate block

peterschwarz:
When a candidate block is cancelled, it needs to finalize the schedule, as well as cancel it, so that the contexts are correctly cleaned up.

This addresses a minor memory leak of contexts.

Additionally,

Reduce cloning of transaction and batch results.


peterschwarz:
This is not a correct fix. Closing

PR #2143, opened 2019-06-11
Fix up retrieval of blocks by number.  Addresses 2 problems with the …

JAG-UK:
…original routing via get_data_by_id:

 - Certain call paths would raise the wrong exception (according to the pydoc and the expectations of callers) leading to unhandled exceptions and crashes
 - Certain call paths would pass in an int type, but the FFI call would treat it as a unicode C string, with various negative consequences

### DIFF
--- a/cli/stdeb.cfg
+++ b/cli/stdeb.cfg
@@ -1,0 +1,2 @@
+[DEFAULT]
+Depends3: python3-sawtooth-sdk


JAG-UK:
Mote that this brings along a fix to siganature_verifyer.py which I had to fix along the way.  It's technically a drive-by and could be argued to be a separate PR but I figure it'll be uncontrovertial since its a duplicate of an already accepted patch a couple of weeks ago.

Testing has been performed variously with Seth-cli, normal cURL via Seth-rpc, and go-ethereum generated code bindings from Jitsuin production code.

validator/sawtooth_validator/journal/block_store.py
```diff
@@ -85,9 +84,20 @@ def __init__(self, database):
             database.pointer,
             ctypes.byref(self.pointer))
 
-    def _get_data_by_id(self, object_id, ffi_fn_name):
+    def _get_data_by_num(self, object_id, ffi_fn_name):
         (vec_ptr, vec_len, vec_cap) = ffi.prepare_vec_result()
+        _libexec(
+            ffi_fn_name,
+            self.pointer,
+            ctypes.c_int(object_id),
```
peterschwarz: Is this the right sized value? Block numbers are `u64`

JAG-UK: Oh, darn.  I spotted that in my review and then didn't commit.  Updated to c_ulonglong.  I'll update the PR as soon as the build/test cycle is done.

peterschwarz: Since you're going to have to do a force push, could you also tidy up your commit message by following https://chris.beams.io/posts/git-commit/ ?

JAG-UK: Sure.  I can make it much shorter and more imperative.

JAG-UK: That's all ready.  Auto-Jenkins-buildn isn't working because I'm not whitelisted: I assume this will fix itself at some point.

validator/sawtooth_validator/journal/block_store.py
```diff
@@ -23,7 +23,6 @@
 from sawtooth_validator.state.merkle import INIT_ROOT_KEY
 from sawtooth_validator import ffi
 
-
```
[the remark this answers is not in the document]
JAG-UK: *facepalm* fixed.

JAG-UK:
The Jenkins build just failed again, but it's in an area I (or more importantly this PR) haven't touched - complaining about duplicate and/or missing definitions of various key structures in proto files during tests.  What gives?

rberg2:
I am not sure about that protobuf error. This branch is 86 commits behind master, so it might we worth trying a rebase.

jsmitchell:
Your "Catch up to upstream" commit is missing a signoff. You might want to instead just rebase this branch on master and then force push.

JAG-UK:
Yes I'm going to squash it all and sign it off when it passes the long build/test cycle it's currently in.  Sorry, I thought it wouldn't kick you guys again until I actually signed off.  Won't be long...

jsmitchell:
It won't even run in jenkins with that unsigned off commit in there. Better just to rebase and push, make sure everything is singed off. One of us will manually run the job, skipping the whitelist check.

JAG-UK:
OK this is no good: Locally it claimed Fastforward worked and there were no diffs to origin, but now I've pushed it's showing the entire merge with diffs.  Would you prefer that I just start again clean with the original patch?

I remain curious how being behind could have caused the last build failure anyway, but not to worry: I don't want to waste anyone's time.

jsmitchell:
No worries at all - we very much appreciate the contribution. If i were you I would do the following - branch off your current JAG-UK:master branch so you have a reference to your commits, then reset your master branch prior to your commits and rebase on sawtooth master. That should give you a branch that is current with sawtooth master. Then cherrypick your commits based on the commit ids in the other branch you made. Once you've done that, you should have an up to date branch with your new commits on top which is force pushable here.

PR #2152, opened 2019-06-12
Chime updates for 1-2

rberg2:


docs/source/_includes/create-genesis-block.inc
```diff
@@ -73,7 +73,7 @@ genesis block also includes the keys for the other nodes in the initial network.
         -o config-consensus.batch \
         sawtooth.consensus.algorithm.name=pbft \
         sawtooth.consensus.algorithm.version=VERSION \
-        sawtooth.consensus.pbft.members=[VAL1KEY,VAL2KEY,...,VALnKEY]
+        sawtooth.consensus.pbft.members='["VAL1KEY","VAL2KEY",...,"VALnKEY"]'
```
chenette: Please also change the VALxKEY stuff in the following note to:

> Replace ``'["VAL1KEY,","VAL2KEY","VAL3KEY","...,"VALnKEY"]'`` with the ...

And add a sentence to the end of that paragraph:

> Be sure to use single quotes and double quotes correctly, as shown in the example.

chenette:
One more nitpick: Could you add info about the `members` fix to the commit and PR description, please? Maybe something like "Also fixed sawtooth.consensus.pbft.members examples - each key must be double-quoted, plus single quotes for the entire string if entered on the command line."

This would be helpful for future reference (for me and other interested watchers).

docs/source/_includes/create-genesis-block.inc
rberg2: I have added this sentence to the note. I also added this sentence to the tip in docs/source/sysadmin_guide/testing_sawtooth.rst

PR #2165, opened 2019-06-18
1-2 backport: Update note on PoET SGX support to include 1.2

chenette:
Changed "1.1" to "1.1 or later". Also removed statement that PoET-SGX would be supported in Sawtooth 1.1.


agunde406:
This is a backport of #2156

PR #2170, opened 2019-06-19
1-2 backport: Update doc notes - tested on 18.04/Bionic only

chenette:
Backport of #2169 

Change the notes "tested on Ubuntu 16.04 only" to the Chime version: "tested on Ubuntu 18.04 (Bionic) only".

Also change the PoET-SGX procedure's note to say "tested with Sawtooth 1.0 on Ubuntu 16.04 only."

### DIFF
--- a/docs/source/sysadmin_guide/systemd.rst
+++ b/docs/source/sysadmin_guide/systemd.rst
@@ -4,7 +4,7 @@ Running Sawtooth as a Service
 
 .. note::
 
-    These instructions have been tested on Ubuntu 16.04 only.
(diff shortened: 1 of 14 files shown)


docs/source/sysadmin_guide/systemd.rst
```diff
@@ -4,7 +4,7 @@ Running Sawtooth as a Service
 
 .. note::
 
-    These instructions have been tested on Ubuntu 16.04 only.
+    These instructions have been tested on Ubuntu 18.04 (bionic) only.
```
rberg2: This b is lower case while the rest are upper case.

PR #2171, opened 2019-06-19
1-2 only: Change PBFT + Sabre doc links: nightly -> latest

chenette:
Change PBFT and Sabre 1.2 doc links from nightly/master to release/latest.


dcmiddle:
How does this sort of thing work when we release a new version. Do we have to go into this 1-2 branch and alter the links?

agunde406:
@dcmiddle I think we would only have to update these links if the new versions (of sabre and pbft) is no longer compatible with the 1.2 version and change it to the last compatible version. Currently the only doc link that is not nightly master for Sabre is latest since it is not at 1.0 yet.

PR #2172, opened 2019-06-19
Fix sawadm restore

peterschwarz:
Updates to protobuf broke the reading of the length-delimited backup file. This PR fixes the issue.

Additionally, adds context to many of the blockstore commands error messages.

### DIFF
--- a/adm/src/commands/blockstore.rs
+++ b/adm/src/commands/blockstore.rs
@@ -457,8 +457,32 @@ fn restore_block(source: &mut protobuf::CodedInputStream) -> Result<Option<Block
         return Ok(None);
     }
 
-    let block = protobuf::parse_from_reader(source)
-        .map_err(|err| CliError::EnvironmentError(format!("Failed to parse block: {}", err)))?;
+    source
+        .read_message()
+        .map(Some)
+        .map_err(|err| CliError::EnvironmentError(format!("Failed to parse block: {}", err)))
+}
 
-    Ok(Some(block))
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use protobuf::CodedInputStream;
+
+    #[test]
+    fn backup_and_restore() {
+        let mut buffer: Vec<u8> = vec![];
+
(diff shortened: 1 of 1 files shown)


adm/src/commands/blockstore.rs
```diff
@@ -202,11 +207,13 @@ fn run_show_command<'a>(args: &ArgMatches<'a>) -> Result<(), CliError> {
     }
     .map_err(|err| CliError::ArgumentError(format!("Error getting block: {}", err)))?;
 
-    let block_wrapper = BlockWrapper::try_from(block)
-        .map_err(|err| CliError::EnvironmentError(format!("{}", err)))?;
+    let block_wrapper = BlockWrapper::try_from(block).map_err(|err| {
+        CliError::EnvironmentError(format!("failed to great block wrapper: {}", err))
```
agunde406: Should this say "get block wrapper"

adm/src/commands/blockstore.rs
```diff
@@ -328,42 +359,54 @@ fn run_stats_command<'a>(args: &ArgMatches<'a>) -> Result<(), CliError> {
...
+        })?;
 
         loop {
             for batch in &block.batches {
                 for txn in &batch.transactions {
                     let txn_header: TransactionHeader = protobuf::parse_from_bytes(&txn.header)
-                        .map_err(|err| CliError::ParseError(format!("{}", err)))?;
+                        .map_err(|err| {
+                            CliError::ParseError(format!(
+                                "failed to parse transaction header for trensaction {}: {}",
```
agunde406: trensaction -> transaction

adm/src/commands/blockstore.rs
peterschwarz: ha - actually, it should say "create".  Fixed.

adm/src/commands/blockstore.rs
peterschwarz: Fixed

PR #2174, opened 2019-06-22
Fix retrieval of blocks by number in python block_store

JAG-UK:
Addresses 2 problems with the original routing via get_data_by_id:
 - Raise expected exception types
 - Treat block number as u64 type throughout

Signed-off-by: JAG-UK <jon.geater@gmail.com>


JAG-UK:
Hi Peter, 
Can we please move on this?  I'm falling behind the mainline again and this needs to be merged before I submit my Seth PR for eth_call and Geth support (which I think people are looking forward to.:-) )
Thanks

PR #2212, opened 2019-07-16
Docs: Fix settings + key info, add auth user proc

chenette:
- Change the Ubuntu single- and multi-node procedures to use the user's private key for the genesis block instead of the validator key. Also change the user key name to "my_key" (to match the k8s environment) and clarify the key required for settings change proposals.
    
- In other procedures, make the key-related info more generic and clarify the validator key information. 

- In the Settings transaction family spec, clarify the definition of sawtooth.settings.vote.authorized_keys and add info about voting.
    
- Add new procedure to the Sys Admin Guide for changing the authorized users in sawtooth.settings.vote.authorized_keys, "Adding Adding Authorized Users for Settings Proposals".

- Update "Adding and Removing a PBFT Node" to clarify settings-related info and steps.

### DIFF
--- a/docs/source/app_developers_guide/ubuntu.rst
+++ b/docs/source/app_developers_guide/ubuntu.rst
@@ -144,6 +144,12 @@ previous step.
    writing file: /home/yourname/.sawtooth/keys/my_key.priv
    writing file: /home/yourname/.sawtooth/keys/my_key.pub
 
+.. note::
+
+   This command specifies ``my_key`` as the base name for the key files, to be
+   consistent with the key name that is used in the example Docker and
+   Kubernetes files. By default (when no key name is specified), the
+   ``sawtooth keygen`` command uses your user name.
 
 .. _generate-root-key-ubuntu:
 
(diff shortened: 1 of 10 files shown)


docs/source/app_developers_guide/ubuntu.rst
```diff
@@ -138,11 +138,11 @@ Step 2: Generate a User Key
 Generate your user key for Sawtooth, using the same terminal window as the
 previous step.
 
-   .. code-block:: console
+.. code-block:: console
 
-      user@validator$ sawtooth keygen
-      writing file: /home/yourname/.sawtooth/keys/yourname.priv
-      writing file: /home/yourname/.sawtooth/keys/yourname.pub
+   user@validator$ sawtooth keygen my_key
```
peterschwarz: Since this `my_key` is optional, should a note be added that the default key name would be `$USER`?

chenette: Good idea. I'll add this info.

chenette: Done - added a commit with the note (in two places).

PR #2214, opened 2019-07-17
Update example kube file to use rust smallbank tp

rbuysse:


rbuysse:
Depends on PR #2213 being merged and nightly artifacts being published.

PR #2215, opened 2019-07-19
Return ACK before sending process requests

peterschwarz:
Return the TpRegisterResponse acknowledgment before adding the TP to the list of processors.  This ensures that the processor always receives the expected ack before it begins to receive any messages requesting transaction execution.

This follows a similar model to the event subscriptions.

### DIFF
--- a/validator/sawtooth_validator/execution/processor_handlers.py
+++ b/validator/sawtooth_validator/execution/processor_handlers.py
@@ -44,7 +44,7 @@ class ProcessorRegisterValidationHandler(Handler):
         if request.protocol_version > MAX_SDK_PROTOCOL_VERSION:
             ack = processor_pb2.TpRegisterResponse()
             ack.status = ack.ERROR
-            ack.protocol_version = request.protocol_version
+            ack.protocol_version = MAX_SDK_PROTOCOL_VERSION
             LOGGER.error(
                 'Validator version %s does not support the features requested'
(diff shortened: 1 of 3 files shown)


validator/sawtooth_validator/execution/processor_handlers.py
```diff
@@ -33,7 +33,46 @@
...
+    def handle(self, connection_id, message_content):
+        request = processor_pb2.TpRegisterRequest()
+        request.ParseFromString(message_content)
+
+        # Reject the request if requested version cannot be handled,
+        # validator does backward compatible support.
+        if request.protocol_version > MAX_SDK_PROTOCOL_VERSION:
+            ack = processor_pb2.TpRegisterResponse()
+            ack.status = ack.ERROR
+            # Send protocol_version of validator, so that the SDK can cross
```
jsmitchell: I realize this is moved code, but does this comment make sense in the context of an error response? It might clean it up a bit to move the ack and protocol_version lines in both sections above the if

peterschwarz: That looks like it was some copy-paste of the same message.  Likewise, I wonder if the validator should send back the version it supports in the error condition.

peterschwarz: The latter probably should be in a separate PR, though.

PR #2217, opened 2019-07-23
Add ready blocks to processing

peterschwarz:
When calling `done` on the `BlockScheduler`, add the next set of blocks (the immediate successor blocks), if any, to the `processing` list.

This corrects an issue where the blocks are scheduled again due to not being marked as in-process via the above mechanism.


arsulegai:
Also, a backport PR is required for this patch on 1-2

peterschwarz:
Will add after it is merged

PR #2221, opened 2019-07-25
Add default adapter for CLI and handle an error when it is invalid

AlexZhovnuvaty:
Signed-off-by: Alexander Zhovnuvaty <a.zhovnuvaty@gmail.com>

### DIFF
--- a/cli/sawtooth_cli/rest_client.py
+++ b/cli/sawtooth_cli/rest_client.py
@@ -28,6 +28,9 @@ class RestClient:
     def __init__(self, base_url=None, user=None):
         self._base_url = base_url or 'http://localhost:8008'
 
+        if '://' not in self._base_url.lower():
+            self._base_url = 'http://{}'.format(self._base_url)
+
         if user:
             b64_string = b64encode(user.encode()).decode()
             self._auth_header = 'Basic {}'.format(b64_string)
--- a/cli/sawtooth_cli/rest_client.py
+++ b/cli/sawtooth_cli/rest_client.py
@@ -206,6 +206,10 @@ class RestClient:
         except (requests.exceptions.MissingSchema,
                 requests.exceptions.InvalidURL) as e:
             raise CliException(e)
+        except requests.exceptions.InvalidSchema as e:
+            raise CliException(
+                ('Schema not valid in "{}": '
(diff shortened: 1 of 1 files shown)


AlexZhovnuvaty:
Observed the following error in this pull request: "continuous-integration/jenkins/pr-head — This commit cannot be built". Do not know what is the reason and how to deal with it? Please assist.

rberg2:
Hello, that was caused by a "white list" check on our build systems. I have restarted this build.

AlexZhovnuvaty:
> Could you split this into two commits? The changes are quite distinct, and then a more detailed commit message can be made for both.

I have split it into two. Pls check if current state is Ok. Otherwise pls provide best option to address this comment.

cli/sawtooth_cli/main.py
```diff
@@ -184,3 +184,7 @@ def main_wrapper():
     except:
         traceback.print_exc(file=sys.stderr)
         sys.exit(1)
+
+
+if __name__ == '__main__':
```
vaporos: This is provided by the bin/sawtooth script in the repo, or by the auto-generated wrapper script when installed. This non-executable file doesn't need it.

vaporos:
> > Could you split this into two commits? The changes are quite distinct, and then a more detailed commit message can be made for both.
> 
> I have split it into two. Pls check if current state is Ok. Otherwise pls provide best option to address this comment.

Do a rebase instead of a merge so you only have the new commits. Something like "git rebase -i upstream/master" (you will need to use --force when you push it back up).

cli/sawtooth_cli/main.py
AlexZhovnuvaty: Done. This was removed

AlexZhovnuvaty:
> > > Could you split this into two commits? The changes are quite distinct, and then a more detailed commit message can be made for both.
> > 
> > 
> > I have split it into two. Pls check if current state is Ok. Otherwise pls provide best option to address this comment.
> 
> Do a rebase instead of a merge so you only have the new commits. Something like "git rebase -i upstream/master" (you will need to use --force when you push it back up).

Done. This branch includes 2 new commits only.

PR #2228, opened 2019-08-12
Notify chain observers in all cases

peterschwarz:
Notify the chain observers that the chain head has been updated, including at `ChainController` initialization time.  This ensures that all observers are also initialized with the current chain head.

This fixes an issue that caused the tests to periodically fail when they encountered a race condition between genesis and consensus engine registration.


peterschwarz:
This isn't the correct solution for this problem.  Will re-open when I have fixed it.

PR #2229, opened 2019-08-20
Only validate block on request from consensus

ltseeley:
Updates the validator:
- Send a block to consensus as soon as it's created/received.
- Wait for a check_blocks() request from consensus to validate the
block.
- Send a BlockValid/BlockInvalid update to consensus when the block has
finished validating.

This was the original intent for consensus API behavior, and it allows
consensus to prioritize which blocks are considered.

This PR has been verified using the devmode liveness test.

### DIFF
--- a/validator/sawtooth_validator/journal/block_wrapper.py
+++ b/validator/sawtooth_validator/journal/block_wrapper.py
@@ -23,11 +23,16 @@ class BlockStatus(IntEnum):
     """
         The status of a block as the journal is concerned.
     """
-    Unknown = 0  # Block is present but not yet validated
-    Invalid = 1  # Block failed block validation.
-    Valid = 2  # Block has been validated and id valid to appear in a chain.
-    Missing = 3  # we know about the block, possibly by a successor, but
-    # we do not have it.
+    # Block is present but not yet validated
+    Unknown = 0
+    # Block failed block validation.
+    Invalid = 1
+    # Block has been validated and id valid to appear in a chain.
+    Valid = 2
+    # We know about the block, possibly by a successor, but we do not have it.
+    Missing = 3
+    # The block is currently being validated
+    InValidation = 4
 
 
 class BlockWrapper:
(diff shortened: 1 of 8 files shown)


validator/src/journal/chain.rs
```diff
@@ -639,30 +642,12 @@ impl<TEP: ExecutionPlatform + Clone + 'static, PV: PermissionVerifier + Clone +
...
             }
             BlockStatus::Invalid => {
-                let mut state = self
-                    .state
-                    .write()
-                    .expect("No lock holder should have poisoned the lock");
-
-                // Drop Ref-C: The block has been found to be invalid, and we are no longer
-                // interested in it. The invalid result will be cached for a period.
-                if state
```
peterschwarz: Don't we still need to drop this ref?  We're done with the block.

ltseeley: The ref will get dropped when consensus decides on the block (commit, ignore, or fail). Maybe consensus will always decide fail though, so we can drop it immediately? I don't have a strong opinion either way.

ltseeley: Put this drop back and added a commit to lower the logging level of the fail/ignore block methods (no need to warn on idempotent fail/ignore).

validator/sawtooth_validator/journal/block_wrapper.py
```diff
@@ -23,11 +23,16 @@ class BlockStatus(IntEnum):
...
-    Unknown = 0  # Block is present but not yet validated
-    Invalid = 1  # Block failed block validation.
-    Valid = 2  # Block has been validated and id valid to appear in a chain.
-    Missing = 3  # we know about the block, possibly by a successor, but
-    # we do not have it.
+    # Block is present but not yet validated
+    Unknown = 0
+    # Block failed block validation.
+    Invalid = 1
+    # Block has been validated and id valid to appear in a chain.
```
jsmitchell: s/id/is/ ?

PR #2230, opened 2019-08-20
Set docker-compose to build images in parallel

rberg2:
This PR adds --parallel to the docker-compose build steps in the
Jenkisnfile.


rberg2:
This reduces uncached build times from 1h30m to about 1h3m; however this also causes the console output from several docker containers to become mixed together making debugging harder.

PR #2240, opened 2019-09-17
Cache StateViews and StateView method calls; use StateViews in ContextManager instead of direct trie reads 

jsmitchell:


vaporos:
The commit messages could include more "why" explanation.

PR #2259, opened 2019-10-10
Enable heartbeats for all zmq connections

agunde406:
Before only validator to validator zmq connections had
heartbeats however if a network sat idle long enough
the zmq connection between components may disconnect.

To remain backword compatible, connections cannot be removed if
they do not respond to a PingRequest. Instead, if the argument is
reap is set to True connections that do not respond will eventually
be removed. If reap is False, the ping will still be sent but
no response is expected/required and the connection will not be
removed.


validator/sawtooth_validator/networking/handlers.py
```diff
@@ -204,7 +204,7 @@ def handle(self, connection_id, message_content):
             message_type=validator_pb2.Message.NETWORK_ACK)
 
 
-class PingHandler(Handler):
+class PingRequestHandler(Handler):
```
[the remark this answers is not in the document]
agunde406: fixed

validator/sawtooth_validator/networking/interconnect.py
```diff
@@ -244,7 +245,7 @@ def _do_router_heartbeat(self):
              if check_time - timestamp > self._heartbeat_interval]
         for zmq_identity, elapsed in expired:
             if self._is_connection_lost(
-                    self._last_message_times[zmq_identity]):
+                    self._last_message_times[zmq_identity]) & self._reap:
```
arsulegai: Bitwise operator :)

agunde406: whoops good catch. fixed.

validator/sawtooth_validator/networking/handlers.py
```diff
@@ -252,6 +252,14 @@ def handle(self, connection_id, message_content):
             message_type=validator_pb2.Message.PING_RESPONSE)
 
 
+class PingResponseHandler(Handler):
```
peterschwarz: Could you add a doc string (or in the commit message) that gives some reasoning for this handler?

agunde406: added it under hander to match the other handler doc string and updated the commit message

PR #2264, opened 2019-11-04
Add libsawtooth and stores

ltseeley:
- Add the libsawtooth crate that will provide various components used by the Sawtooth validator.
- Add the `OrderedStore` trait, which defines an interface for a key/value store that is indexed by a type with total ordering.
- Add a BTreeMap-backed implementation of the `OrderedStore` trait for an in-memory ordered store.
- Add the `TransactionReceiptStore`, which is a wrapper around an `OrderedStore` that can be used to store transaction receipts, keyed by transaction IDs, in order.
- Add the `RedisOrderedStore` struct, which is an implementation of the `OrderedStore` trait that uses a redis database.
- Add the `LmdbOrderedStore`, which is an implementation of the `OrderedStore` trait that uses an LMDB database.

### DIFF
--- a/libsawtooth/src/store/lmdb.rs
+++ b/libsawtooth/src/store/lmdb.rs
@@ -1,0 +1,349 @@
+// Copyright 2019 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::convert::TryInto;
+use std::fmt::Debug;
+use std::path::Path;
+use std::sync::Arc;
+
+use lmdb_zero as lmdb;
+use lmdb_zero::error::LmdbResultExt;
+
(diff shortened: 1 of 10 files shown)


libsawtooth/src/store/lmdb.rs
```diff
@@ -0,0 +1,345 @@
...
+            OrderedStoreError::InitializationFailed(format!("invalid filepath: {:?}", filepath))
+        })?;
+
+        let mut builder = lmdb::EnvBuilder::new().map_err(|err| {
+            OrderedStoreError::InitializationFailed(format!(
+                "failed to initialize environment builder: {}",
+                err
+            ))
+        })?;
+        builder.set_maxdbs(3).map_err(|err| {
```
vaporos: Maybe const this value.

ltseeley: Done

ltseeley:
Put modules and sub-modules behind experimental features

libsawtooth/src/store/receipt_store/mod.rs
```diff
@@ -0,0 +1,211 @@
...
+use transact::protocol::receipt::TransactionReceipt;
+
+use crate::store::OrderedStore;
+
+pub use self::error::TransactionReceiptStoreError;
+
+/// `TransactionReceiptStore` is a wrapper around an `OrderedStore` that stores
+/// `TransactionReceipt`s keyed by transaction IDs (`String`s) and indexed by `usize`s to provide
+/// ordering of transactions.
+pub struct TransactionReceiptStore(Box<dyn OrderedStore<String, TransactionReceipt, usize> + Send>);
```
peterschwarz: This should be `u64`, if it is the equivalent of block height

PR #2266, opened 2019-11-21
Add LmdbOrderedStoreIter for iterating over an LMDB ordered store

ltseeley:
Add the `LmdbOrderedStoreIter` struct that is returned by the
`LmdbOrderedStore.iter()` method. This iterator loads items into an
in-memory cache as needed and provides an iterator over the values in
the ordered store. Additionally, it provides the option to specify a
range to get only a subset of items in the database.

### DIFF
--- a/libsawtooth/Cargo.toml
+++ b/libsawtooth/Cargo.toml
@@ -24,6 +24,7 @@ description = """\
 """
 
 [dependencies]
+log = "0.4"
 lmdb-zero = { version = "0.4", optional = true }
 redis = { version = "0.13.0", default-features = false, optional = true }
 transact = "0.1"
--- a/libsawtooth/src/store/lmdb.rs
+++ b/libsawtooth/src/store/lmdb.rs
@@ -339,11 +339,15 @@ mod tests {
         let thread_id = std::thread::current().id();
         temp_db_path.push(format!("store-{:?}.lmdb", thread_id));
 
-        test_u8_ordered_store(Box::new(
-            LmdbOrderedStore::new(temp_db_path.as_path(), Some(1024 * 1024))
-                .expect("Failed to create LMDB ordered store"),
-        ));
+        let test_result = std::panic::catch_unwind(|| {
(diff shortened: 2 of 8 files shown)


libsawtooth/Cargo.toml
```diff
@@ -24,6 +24,7 @@ description = """\
 """
 
 [dependencies]
+log = "0.3.0"
```
peterschwarz: Current version is "0.4"

libsawtooth/src/store/mod.rs
```diff
@@ -23,6 +23,8 @@ pub mod receipt_store;
 pub mod redis;
 
 use std::convert::TryInto;
+pub use std::ops::Bound;
```
peterschwarz: You shouldn't have to re-export this, since it's a std library item.

libsawtooth/src/store/lmdb.rs
```diff
@@ -326,6 +308,123 @@ impl<
...
+        // Load initial values into the cache
+        if let Err(err) = cursor.reload_cache() {
+            error!("Failed to load iterator's initial cache: {}", err);
+        }
+
+        Ok(cursor)
+    }
+
+    fn reload_cache(&mut self) -> Result<(), String> {
+        let txn = lmdb::ReadTransaction::new(self.env.clone()).map_err(|err| err.to_string())?;
```
peterschwarz: Is there no way to keep the transaction for the iterator?

libsawtooth/Cargo.toml
ltseeley: Updated

libsawtooth/src/store/mod.rs
ltseeley: Fixed

libsawtooth/src/store/lmdb.rs
ltseeley: Fixed

PR #2269, opened 2019-12-13
Fix typo in block info address calculation

sergeydobrodey:


rberg2:
Hello, thanks for the correction! The [DCO](http://developercertificate.org/) check above failed because we require all commits to be signed off by the author to ensure the commit satisfies the DCO requirements. This can be done by amending the commit and signing off with this command. `git commit --amend --signoff`

PR #2280, opened 2020-01-24
Fix timestamp check in block_info for catch up

vaporos:
This is similar to 5e7315a9f0e3c8863327034c036b28e70850112c, which was
a fix for the python version of block_info. To quote that commit:

"The timecheck needs to ensure that the timestamp is only ahead of
a transaction processor's local time by the value of tolerance. The use
of absolute value enforced that this time check is within tolerence of
the local time. This fails validation in the case where a node is
catching up on a chain the may contain blocks that were published more
than time-tolerence in the past.  Correcting this to ensure the the
timestamp is no more than tolerence in the future, ensures that the
transaction can still be validated during a catch up scenario."

### DIFF
--- a/families/block_info/sawtooth_block_info/src/handler.rs
+++ b/families/block_info/sawtooth_block_info/src/handler.rs
@@ -41,10 +41,10 @@ fn validate_timestamp(timestamp: u64, tolerance: u64) -> Result<(), ApplyError> 
         .duration_since(UNIX_EPOCH)
         .expect("System time is before Unix epoch.")
         .as_secs();
-    if timestamp < (now - tolerance) || (now + tolerance) < timestamp {
+    if now + tolerance < timestamp {
         let warning_string = format!(
-            "Timestamp must be less than local time. Expected {0} in ({1}-{2}, {1}+{2})",
-            timestamp, now, tolerance
+            "Timestamp must be less than local time. Expected {0} + {1} < {2}",
+            now, tolerance, timestamp
         );
         warn!("Invalid Transaction: {}", &warning_string);
         return Err(ApplyError::InvalidTransaction(warning_string));


mtrubs:
I was able to reproduce this locally. 

1) I used this docker compose file (based mostly off those in sawtooth-core): https://gist.github.com/mtrubs/df652ed4878975cc79759cf471ed5ffa
2) once the chain was up I commit a few intkey transaction successfully
3) I then waited ~5min (as that is the default tolerance)
4) on validator-4 I deleted the contents of `/var/lib/sawtooth`
5) I then restarted validator-4

as it now needed to sync with the network from scratch it ran into this issue (note that the ~5m wait is very important to this).
```
validator-4        | [2020-01-27 16:06:19.700 WARNING  (unknown file)] [src/journal/block_validator.rs: 284] Error during block validation: BlockValidationError("During validate_on_chain_rules, error creating settings view: NotFound(\"0ca3eb7461654ddffd8cec5600d5475406a97339062470864e32e24150affbe3\")")
```

I am working to build this version of block info to assert the fix.

mtrubs:
actually this doesn't seem like quite the right fix. perhaps its the way rust handles things. still digging into what it proper.

mtrubs:
ok the problem I was seeing is that the check was still failing. I added a little extra logging locally to see what Rust was actually doing.

```
Expected 1580137139 - 1580148511 = 18446744073709540244 < 300
```

This occurs because all the values are `u64`, meaning they cannot be negative. As a result it just wraps around near the max value of `u64` and will always be greater than 300.

I adjusted the code to this and everything looks to be happy:

```
$ git diff
diff --git a/families/block_info/sawtooth_block_info/src/handler.rs b/families/block_info/sawtooth_block_info/src/handler.rs
index 102372a8..409f0931 100644
--- a/families/block_info/sawtooth_block_info/src/handler.rs
+++ b/families/block_info/sawtooth_block_info/src/handler.rs
@@ -41,10 +41,10 @@ fn validate_timestamp(timestamp: u64, tolerance: u64) -> Result<(), ApplyError>
         .duration_since(UNIX_EPOCH)
         .expect("System time is before Unix epoch.")
         .as_secs();
-    if (timestamp - now) > tolerance {
+    if now + tolerance < timestamp {
         let warning_string = format!(
-            "Timestamp must be less than local time. Expected {0} - {1} < {2}",
-            timestamp, now, tolerance
+            "Timestamp must be less than local time. MTR. Expected {0} + {1} < {3}",
+            now, tolerance, timestamp
         );
         warn!("Invalid Transaction: {}", &warning_string);
         return Err(ApplyError::InvalidTransaction(warning_string));
```

Rust can be tricky 😉

vaporos:
Updated with the corrected comparison and stringl

arsulegai:
> ok the problem I was seeing is that the check was still failing. I added a little extra logging locally to see what Rust was actually doing.
> 
> ```
> Expected 1580137139 - 1580148511 = 18446744073709540244 < 300
> ```
> 
> This occurs because all the values are `u64`, meaning they cannot be negative. As a result it just wraps around near the max value of `u64` and will always be greater than 300.
> 
> I adjusted the code to this and everything looks to be happy:
> 
> ```
> $ git diff
> diff --git a/families/block_info/sawtooth_block_info/src/handler.rs b/families/block_info/sawtooth_block_info/src/handler.rs
> index 102372a8..409f0931 100644
> --- a/families/block_info/sawtooth_block_info/src/handler.rs
> +++ b/families/block_info/sawtooth_block_info/src/handler.rs
> @@ -41,10 +41,10 @@ fn validate_timestamp(timestamp: u64, tolerance: u64) -> Result<(), ApplyError>
>          .duration_since(UNIX_EPOCH)
>          .expect("System time is before Unix epoch.")
>          .as_secs();
> -    if (timestamp - now) > tolerance {
> +    if now + tolerance < timestamp {
>          let warning_string = format!(
> -            "Timestamp must be less than local time. Expected {0} - {1} < {2}",
> -            timestamp, now, tolerance
> +            "Timestamp must be less than local time. MTR. Expected {0} + {1} < {3}",
> +            now, tolerance, timestamp
>          );
>          warn!("Invalid Transaction: {}", &warning_string);
>          return Err(ApplyError::InvalidTransaction(warning_string));
> ```
> 
> Rust can be tricky 😉

Awesome!

PR #2288, opened 2020-02-19
move optional content to end

lazaridiscom:
This doc needs to be simplified further


lazaridiscom:
closing in favor of #2289 which uses a branch.

PR #2291, opened 2020-02-19
move optional content to the end

lazaridiscom:
this doc needs to be simplified further

Signed-off-by: Lazaridis <59408072+lazaridiscom@users.noreply.github.com>


lazaridiscom:
the 2-commit-diff has "lost it". It's easier to look at each commit separately.

Not that there is much to look at.

jsmitchell:
Submitter has discorporeated.

PR #2293, opened 2020-03-09
Fix terminology: batch -> transaction

LiyanJin:


chenette:
LianJin, thank you for this correction!  However, the commit needs a DCO sign-off. You can use these instructions to add a sign-off to the commit:
https://github.com/hyperledger/sawtooth-core/pull/2293/checks?check_run_id=495368110

dcmiddle:
Thanks @chenette 
@LiyanJin you may also need to set your name and email in your git client so that the sign off flag picks up the right values. 
https://help.github.com/en/github/setting-up-and-managing-your-github-user-account/setting-your-commit-email-address

LiyanJin:
Thanks for your suggestions. @dcmiddle @chenette 
This is my first pull request and I have set my name and email address. However, I'm not familiar with the DCO sign-off. I have tried to add sign-off with reference to the instructions given by you. But it doesn't work.  Because I pushed the commit directly on the webpage, instead of pushing it up in my local cloned repository via Git. So the instruction doesn't work for me. Is there any other way to sign-off?

arsulegai:
@LiyanJin Good approach would be to clone and push with the DCO sign. You may also try this suggestion https://github.com/todogroup/gh-issues/issues/50#issuecomment-519151509

LiyanJin:
Thanks for your suggestion @arsulegai Let me try it and hope it works for me.

PR #2295, opened 2020-03-10
Fix terminology: injecting transactions into blocks-> batches

LiyanJin:
Signed-off-by: Hans Jin <hans_jin@126.com>

PR #2298, opened 2020-03-10
Fix terminology: any batch signer -> any transaction signer

LiyanJin:
Signed-off-by: Hans Jin <hans_jin@126.com>


LiyanJin:
It doesn't work for me even with s Signed-off-by in the commit

arsulegai:
I know it's not good approach, but did you leave an empty line gap between commit message and the signed-off line, also please confirm that the email id is attached to the github account.

dcmiddle:
Closing this PR in favor of #2295 which also has this commit.

LiyanJin:
@arsulegai Thanks so much for your reminder. I will pay attention to it next time.

PR #2300, opened 2020-04-06
Liyan jin patch 20200406

LiyanJin:
Signed-off-by: Hans Jin <hans_jin@126.com>
Correct some typos discovered when translating them into Chinese.

### DIFF
--- a/docs/source/architecture/events_and_transactions_receipts.rst
+++ b/docs/source/architecture/events_and_transactions_receipts.rst
@@ -312,13 +312,13 @@ the TransactionReceiptStore after a block is validated.
 
 Transaction receipts will only be stored in this off-chain store and will not
 be included in the block. Note that because a transaction may exist in multiple
-blocks at a time, the transaction receipt is stored by both transaction id and
+blocks at a time, the transaction receipt is stored with both transaction id and
 block state root hash.
 
 .. code-block:: python
 
   class TransactionReceiptStore:
-  	def put_receipt(self, txn_id, receipt):
+  	def put_receipt(self, txn_id, state_root_hash, receipt):
       	"""Add the given transaction receipt to the store. Does not guarantee
          	it has been written to the backing store.
 
@@ -329,7 +329,7 @@ block state root hash.
           	receipt (TransactionReceipt): the receipt object to store.
       	"""
 
-  	def get_receipt(self, txn_id):
+  	def get_receipt(self, txn_id, state_root_hash):
       	"""Returns the TransactionReceipt
 
       	Args:
--- a/docs/source/architecture/injecting_batches_block_validation_rules.rst
+++ b/docs/source/architecture/injecting_batches_block_validation_rules.rst
@@ -23,7 +23,7 @@ some use cases.  Two examples of inter-transaction validation are:
 BatchInjector Interface
 =======================
 
-The ``BatchInjector`` class supports injecting transactions into blocks:
+The ``BatchInjector`` class supports injecting batches into blocks:
 
 .. code-block:: python
 
--- a/docs/source/architecture/injecting_batches_block_validation_rules.rst
+++ b/docs/source/architecture/injecting_batches_block_validation_rules.rst
@@ -86,7 +86,7 @@ Validation rules use the following simple syntax:
 The following rules are defined:
 
 NofX
-  Only N transaction of transaction type X may be included in a block. The first
+  Only N transactions of transaction type X may be included in a block. The first
   argument must be an integer. The second argument is the name of a transaction
   family. For example, the string
   ``NofX:2,intkey`` means allow only two IntegerKey transactions per block.
@@ -100,9 +100,9 @@ XatY
   backwards from the last transaction in the block. The first transaction in the
   block has index 0. The last transaction in the block has index -1. If abs(Y)
   is larger than the number of transactions per block, then there would not be
-  a transaction of type X at Y and the block would be invalid. For example, the
-  string ``XatY:intkey,0`` means the first transaction in the block must be an
-  IntegerKey transaction.
+  a transaction of type X at Y and the block would be validated as invalid. For
+  example, the string ``XatY:intkey,0`` means the first transaction in the block 
+  must be an IntegerKey transaction.
 
 local
   A transaction must be signed by the same key as the block. This rule takes a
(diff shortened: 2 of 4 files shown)


docs/source/architecture/events_and_transactions_receipts.rst
```diff
@@ -312,13 +312,13 @@ the TransactionReceiptStore after a block is validated.
 
 Transaction receipts will only be stored in this off-chain store and will not
 be included in the block. Note that because a transaction may exist in multiple
-blocks at a time, the transaction receipt is stored by both transaction id and
+blocks at a time, the transaction receipt is stored with both transaction id and
```
arsulegai, line 315: The TransactionReceipt is updated upon a block commit. We may not have a scenario that at same height multiple blocks are committed. So this becomes an invalid scenario. We could have same transaction in multiple blocks but they are not stored in TransactionReceiptStore until a block is committed. That means when transaction is stored in the TransactionReceipt it only can be referred by TxnID.

put() & get() methods do not require state_root_hash. This also resonates with the latest codebase.

docs/source/architecture/events_and_transactions_receipts.rst
```diff
@@ -329,7 +329,7 @@ block state root hash.
           	receipt (TransactionReceipt): the receipt object to store.
       	"""
 
-  	def get_receipt(self, txn_id):
+  	def get_receipt(self, txn_id, state_root_hash):
       	"""Returns the TransactionReceipt
 
       	Args:
```
arsulegai, line 335: perhaps, remove the `state_root_hash` from the `Args:`

docs/source/architecture/events_and_transactions_receipts.rst
```diff
@@ -312,13 +312,13 @@ the TransactionReceiptStore after a block is validated.
```
arsulegai, line 312: This section of Transaction Receipt Store needs an update (future item). There is now support for storing invalid transactions as well under a flag. HL Sawtooth is making use of the HL Transact library for this purpose.

docs/source/architecture/injecting_batches_block_validation_rules.rst
```diff
@@ -23,7 +23,7 @@ some use cases.  Two examples of inter-transaction validation are:
 BatchInjector Interface
 =======================
 
-The ``BatchInjector`` class supports injecting transactions into blocks:
+The ``BatchInjector`` class supports injecting batches into blocks:
```
arsulegai, line 26: This section would require revision as well. Here's the latest code https://github.com/hyperledger/sawtooth-core/blob/master/validator/sawtooth_validator/journal/batch_injector.py and the comment on block_start is changed.

docs/source/architecture/permissioning_requirement.rst
```diff
@@ -257,7 +257,7 @@ Allow all transaction signers to submit transactions
 |                    |   transaction signer.                                  |
 |                    | - A network operator can configure the                 |
 |                    |   whole Sawtooth network to accept transactions signed |
-|                    |   by any batch signer.                                 |
+|                    |   by any transaction signer.                           |
```
arsulegai, line 260: There is an assumption that the batch containing this transaction is signed by a batcher that is allowed. This is applicable to current and next section.

docs/source/architecture/poet.rst
```diff
@@ -65,7 +65,7 @@ provided by other lottery algorithms. The probability of election
 is proportional to the resources contributed (in this case, resources
 are general purpose processors with a trusted execution environment).
 An attestation of execution provides information for verifying that the
-certificate was created within the enclave (and that the validator waited
+Timer was created within the enclave (and that the validator waited
```
arsulegai, line 68: `certificate` could be the right word here. Each validator has to confirm to other validators that they waited for random amount of time that they're claiming in the certificate.

@dcmiddle to confirm

docs/source/architecture/poet.rst
```diff
@@ -419,7 +419,7 @@ The server side of the validator runs the following sign-up procedure:
    a. Pass sign-up certificate of new participant (OPK, EPID Pseudonym, PPK,
       current :math:`WaitCertId_{n}` to upper layers for registration in EndPoint
       registry.
-   #. Goto 1.
+   #. Goto 1，namely waiting for a join request.
```
arsulegai, line 422: How about `s/for a join/for a new join/`?

docs/source/architecture/poet.rst
```diff
@@ -506,13 +506,14 @@ Security Considerations
 -----------------------
 
 1. :math:`T_{WT}` **motivation**: A validator has at most :math:`T_{WT}` seconds
-   to consume a WaitTimer, namely obtain a WaitCertificate on it after the
+   to consume a WaitTimer, namely generate a WaitCertificate on it after the
```
arsulegai, line 509: Validator doesn't need to run the consensus engine on it. So validator cannot generate a WaitCertificate. `obtain` sounded ok to me. PoET engine will ask SGX enclave to generate one.

docs/source/architecture/poet.rst
```diff
@@ -571,7 +572,7 @@ Security Considerations
       elapses. This again prevents a malicious user from running multiple
       instances of the enclave to create multiple WaitCertificates (each with a
       different nonce) resulting in different WaitCertId digests and selecting
-      the one that would result in the lowest duration for a new WaitTimer.
+      the one that would result in the lowest duration of its next WaitTimer.
```
arsulegai, line 575: "`for a new` WaitTimer" sounds rightly worded. The intent is to stop the scenario where multiple instances of enclave are run to get multiple WaitTimer for the same block. Not necessarily for a new (block's) WaitTimer.

docs/source/architecture/poet.rst
```diff
@@ -594,7 +595,7 @@ providing username and ledgername as input parameters to
 ``generateSignUpData()`` and ``unsealSignUpData()``. Then the sign-up tuple 
 (username, ledgername, PPK, PSK, MCID) is sealed to disk, with username and
 ledgername used to generate the filename. Anytime a user authenticates to the
-service the latter can have the enclave unseal and use the sign-up tuple from
+server the latter can have the enclave unseal and use the sign-up tuple from
```
arsulegai, line 598: I think we should leave it to `service`, a process running on a machine. Not a `server` as in web terminology.

docs/source/architecture/poet.rst
LiyanJin: I prefer 'for a new join'

docs/source/architecture/poet.rst
LiyanJin: I agree with you.

docs/source/architecture/poet.rst
LiyanJin: I agree.

arsulegai:
@LiyanJin could you please update the PR?

agunde406:
Closing due to no response on update request, please reopen if addressed. Thanks

PR #2303, opened 2020-06-10
More linting cleanup

vaporos:
Doesn't pass 'just lint' yet, but getting closer.


perf/sawtooth_perf/src/source.rs
```diff
@@ -54,10 +54,10 @@ where
             }
 
             // read the delimited length
-            let next_len = try!(self.source.read_raw_varint32());
```
peterschwarz, line 57: Well, this is clearly old code :)

PR #2309, opened 2020-06-19
Moves transaction and batch structs to libsawtooth

shannynalayna:
Involves moving both the `Transaction` and `Batch` structs to libsawtooth, and makes the necessary changes to the validator to refer to this new location. Also creates a new `PyObjectWrapper` object which, appropriately, wraps a PyObject and implements the `ToPyObject` and `FromPyObject` traits in order to function as a go-between for the Batch struct defined in the libsawtooth module and the remaining validator code that still expects to translate this struct into a PyObject.

### DIFF
--- a/libsawtooth/src/batch.rs
+++ b/libsawtooth/src/batch.rs
@@ -15,20 +15,30 @@
  * ------------------------------------------------------------------------------
  */
 
-use proto;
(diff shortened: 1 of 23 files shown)


shannynalayna:
Closing in favor of #2310

PR #2315, opened 2020-06-23
Move LmdbDatabase and MerkleDatabase to libsawtooth

agunde406:
Also moves required traits and structs to move them.


libsawtooth/Cargo.toml
```diff
@@ -54,8 +55,11 @@ lmdb-store = ["lmdb-zero", "stores"]
 receipt-store = ["stores"]
 redis-store = ["redis", "stores"]
 stores = []
-validator-internals = ["hex", "openssl", "lmdb-zero"]
+validator-internals = ["cbor-codec","hex", "openssl", "lmdb-zero"]
```
ltseeley, line 58: Missing a space here (also, lmdb-zero isn't in its alphabetical spot).

libsawtooth/src/state/merkle.rs
```diff
@@ -678,7 +676,7 @@ fn encode_and_hash(node: Node) -> Result<(Vec<u8>, Vec<u8>), StateDatabaseError>
 /// Given a path, split it into its parent's path and the specific branch for
 /// this path, such that the following assertion is true:
 ///
-/// ```
+/// ```ignore
```
ltseeley, line 679: This isn't captured in the commit message; why is this necessary?

libsawtooth/Cargo.toml
agunde406: fixed

libsawtooth/src/state/merkle.rs
agunde406: added to commit, this test was not being running in its old location but when moved was being ran. But the doc test fails, so I ignored it.

PR #2317, opened 2020-06-24
Move several modules from validator to libsawtooth

shannynalayna:
This moves a few modules over to libsawtooth, including:

* `SettingsView`
* `StateViewFactory`
* `enforce_validation_rules` function
* `PermissionVerifier` trait

### DIFF
--- a/libsawtooth/src/journal/validation_rule_enforcer.rs
+++ b/libsawtooth/src/journal/validation_rule_enforcer.rs
@@ -15,14 +15,16 @@
  * ------------------------------------------------------------------------------
  */
 
-use sawtooth::{batch::Batch, state::settings_view::SettingsView, transaction::Transaction};
+use crate::{batch::Batch, state::settings_view::SettingsView, transaction::Transaction};
 
 /// Retrieve the validation rules stored in state and check that the
 /// given batches do not violate any of those rules. These rules include:
 ///
+/// ```ignore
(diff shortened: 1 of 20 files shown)


libsawtooth/src/journal/validation_rule_enforcer.rs
```diff
@@ -15,14 +15,16 @@
  * ------------------------------------------------------------------------------
  */
 
-use sawtooth::{batch::Batch, state::settings_view::SettingsView, transaction::Transaction};
+use crate::{batch::Batch, state::settings_view::SettingsView, transaction::Transaction};
 
 /// Retrieve the validation rules stored in state and check that the
 /// given batches do not violate any of those rules. These rules include:
 ///
+/// ```ignore
```
agunde406, line 23: Why was this added?

shannynalayna: This piece was causing the docs test to fail because it seemed to be incorrectly parsing the line (it was throwing an error for N not being followed by the correct symbol)

agunde406: Can you add a comment to the commit. I had a simular issues (but it already had the `). I think its caused by the difference between 2015 and 2018 rust.

shannynalayna: Ok, yeah definitely.

PR #2320, opened 2020-06-25
Add release notes for 1.2.5

rbuysse:

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,21 @@
 # Release Notes
 
+## Changes in Sawtooth 1.2.5
+
+### sawtooth-core:
+ - Fixed issue with decoding state values which caused the validator to crash
+   when settings with long values were used. Add DecodedMerkleStateReader for
+   state views. This reader provides the CBOR-decoded values to the rust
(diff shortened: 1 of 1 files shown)


RELEASE_NOTES.md
```diff
@@ -1,5 +1,19 @@
 # Release Notes
 
+## Changes in Sawtooth 1.2.5
+
+### sawtooth-core:
+ - Add a justfile with build, clean, docker-build-doc, docker-lint, fix, lint
+   and test targets. This is currently Rust-centric as it's the easiest to put
+   into the justfile in a useful manner.
+ - Add DecodedMerkleStateReader for state views. This reader provides the
```
vaporos, line 9: Should pop this up to be the first entry in the list since it is motivating the release.

vaporos: Should probably also modify this to point out the error which is fixed. Maybe: "Fixed issue with decoding state values which caused the validator to crash when settings with long values were used. Add DecodedMerkleStateReader for state views. This reader provides the CBOR-decoded values to the rust implementations of StateView structs via a new implementation of the StateReader."

PR #2322, opened 2020-07-02
Move ChainController to libsawtooth

agunde406:


validator/src/journal/candidate_block.rs
```diff
@@ -49,7 +49,25 @@ pub struct FinalizeBlockResult {
...
+    fn summarize(&mut self, force: bool) -> Result<Option<Vec<u8>>, CandidateBlockError>;
+
+    fn finalize(
+        &mut self,
+        consensus_data: &[u8],
+        force: bool,
+    ) -> Result<FinalizeBlockResult, CandidateBlockError>;
+}
+
+pub struct SawtoothCandidateBlock {
```
vaporos, line 42: It is all sawooth... how about PyCandidateBlock?

agunde406: After offline discussion, chnage it to FFICandidateBlock

PR #2330, opened 2020-07-16
Inclusive terminology

dcmiddle:
Revise PoET doc text and Jenkins scripts to use inclusive terminology. I believe this captures all issues in sawtooth-core. I will copy the Jenkins commit to the other repos after review.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -38,10 +38,10 @@ pipeline {
     }
 
     stages {
-        stage('Check Whitelist') {
+        stage('Check User Authorization') {
             steps {
-                readTrusted 'bin/whitelist'
-                sh './bin/whitelist "$CHANGE_AUTHOR" /etc/jenkins-authorized-builders'
+                readTrusted 'bin/authorize-cicd'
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -38,10 +38,10 @@ pipeline {
     }
 
     stages {
-        stage('Check Whitelist') {
+        stage('Check User Authorization') {
             steps {
-                readTrusted 'bin/whitelist'
-                sh './bin/whitelist "$CHANGE_AUTHOR" /etc/jenkins-authorized-builders'
+                readTrusted 'bin/authorize-cicd'
+                sh './bin/authorize-cicd "$CHANGE_AUTHOR" /etc/jenkins-authorized-builders'
```
vaporos, line 44: How about 'check-authlist' as a script name?

dcmiddle: That's also good. I prefer `authorize-cicd`, though, because when you are looking in the bin directory it is explicit about what is being authorized (cicd) while using the same verb-object format and character count.

PR #2331, opened 2020-07-24
Add batch_before, batch_after and block_end events

marcopaulocova:
Add support for the batch_before, batch_after and block_end injector events.
Add support for negative values in the "XatY" and "local" rules in the setting "sawtooth.validator.block_validation_rules".
Implement missing abstract methods in the Batch Injector Factory in the python implementation.
Move injectors into a dedicated folder.

This will change the "block_validation_rules" to only be applied at the end of the block in order to avoid conflicts and multiple validations of previously approved batches.

Co-authored-by: AntonioBranco antonio.branco@voidsoftware.com
Signed-off-by: Micaelsf micael.ferreira@voidsoftware.com


vaporos:
The build is failing due to linting (and other things):

https://build.sawtooth.me/job/Sawtooth-Hyperledger/job/sawtooth-core/job/PR-2331/2/execution/node/34/log/

rbuysse:
Looks like you'll need to rebase to pick up some recently merged pylint fixes: https://github.com/hyperledger/sawtooth-core/pull/2340

validator/sawtooth_validator/journal/batch_injector.py
```diff
@@ -50,6 +50,48 @@ def block_start(self, previous_block):
         """
         raise NotImplementedError()
 
+    @abc.abstractmethod
+    def before_batch(self, previous_block, batch):
+        """Returns an ordered list of batches to inject at the beginning of the
```
agunde406, line 55: This comment does not match the method. (same with the other two added methods)

validator/src/journal/candidate_block.rs
```diff
@@ -240,36 +308,49 @@ impl CandidateBlock {
 
             // Inject blocks at the beginning of a Candidate Block
             if self.pending_batches.is_empty() {
```
agunde406, line 310: The above comment should be updated

validator/src/journal/candidate_block.rs
```diff
@@ -240,36 +308,49 @@ impl CandidateBlock {
...
+        } else {
+            debug!(
+                "Dropping batch due to missing dependencies: {}",
+                batch_header_signature.as_str()
+            );
+        }
+    }
+
+    fn handle_block_end_batches(&mut self) {
+        // Inject blocks at the end of a Candidate Block
```
agunde406, line 341: Should this say "Inject batches at the end of a Candidate Block"?

agunde406 (changes requested):

validator/sawtooth_validator/journal/batch_injector.py
micaelsfVoidSoftware: Thank you for your time, I'll fix these typos that you pointed out as soon as possible

peterschwarz (commented):
Would it be possible to add an integration test for this? Particularly testing the batch-at-end-of-block rule.

micaelsfVoidSoftware:
> Would it be possible to add an integration test for this? Particularly testing the batch-at-end-of-block rule.

Hi. Yes I can try to manage an IT for those new events batch-after/before and block-end, thanks for the review

peterschwarz (dismissed):

bin/run_tests
```diff
@@ -217,6 +226,7 @@ test_rust_sdk() {
 }
 
 test_integration() {
+    run_docker_test test_batch_injection_events
```
agunde406, line 229: Adding the test here causes it to be ran twice. It should either be added here with the coverage line, or put in the test_batch_injection_events() section, not both.

agunde406 (changes requested):

bin/run_tests
micaelsfVoidSoftware: Your are completely right, I'll remove it from integration section, thank you

peterschwarz (approved):

agunde406 (approved):

vaporos (approved):

PR #2334, opened 2020-07-31
Remove py permission verifier

peterschwarz:
This PR removes the `PyPermissionVerifier` from the validator, as the chain controller in libsawtooth uses its own Rust implementation. 

Additionally, it removes the rust gossip modules, as they are empty after this change.

(Requires a release of the sawtooth-lib to crates, before it may be merged)


peterschwarz:
Closing in favor of #2335

PR #2336, opened 2020-08-10
Update libsawtooth dependency to 0.5

ltseeley:
Updates the libsawtooth dependency to 0.5


validator/src/batch_ffi.rs
```diff
@@ -26,10 +28,9 @@ impl From<Batch> for PyObjectWrapper {
         let gil = Python::acquire_gil();
         let py = gil.python();
 
-        let batch_proto = protos::batch::Batch::from(native_batch);
         let batch_pb2 = py
             .import("sawtooth_validator.protobuf.batch_pb2")
-            .expect("unable for python to import sawtooth_validator.protobuf.batch_pb2");
+            .expect("Unable to import block_pb2");
```
agunde406, line 33: This change doesn't look intended

validator/src/journal/commit_store_ffi.rs
```diff
@@ -19,13 +19,15 @@ use std::mem;
...
 use std::slice;
 
-use protobuf;
 use sawtooth::journal::commit_store::{
     ByHeightDirection, CommitStore, CommitStoreByHeightIterator,
 };
-use sawtooth::{batch::Batch, block::Block, transaction::Transaction};
-use transact::database::error::DatabaseError;
-use transact::database::lmdb::LmdbDatabase;
+use sawtooth::transact::database::error::DatabaseError;
```
agunde406, line 25: This is missing pulling in Transaction and Batch from transact.

ltseeley: Fixed

validator/src/batch_ffi.rs
ltseeley: Fixed

PR #2350, opened 2020-10-07
Fix bug in batch tracker's chain_update method

ltseeley:
Updates the `chain_update` method for the batch tracker to use the
provided block for notifying of committed batches rather than checking
the commit store. This fixes a bug that was created by a recent change
to the chain controller; the chain controller now notifies of committed
batches before persisting the committed block to the commit store. This
resulted in the batch tracker not properly notifying of committed
batches.

The fix is to not check the commit store in `chain_update`, but rather
just use the batch IDs in the provided block. In addition to fixing the
bug, this avoids an unnecessary database call.

### DIFF
--- a/validator/tests/test_client_request_handlers/mocks.py
+++ b/validator/tests/test_client_request_handlers/mocks.py
@@ -98,7 +98,7 @@ class MockBlockStore(BlockStore):
     def clear(self):
         self.__init__(size=0)
 
-    def add_block(self, base_id, root=b'merkle_root'.hex()):
+    def make_mock_block(self, base_id, root=b'merkle_root'.hex()):
         block_id = 'b' * (128 - len(base_id)) + base_id
         head = self.chain_head
         if head:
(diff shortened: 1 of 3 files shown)


agunde406 (dismissed):

agunde406 (dismissed):

peterschwarz (approved):

validator/tests/test_client_request_handlers/mocks.py
```diff
@@ -110,20 +110,25 @@ def add_block(self, base_id, root=b'merkle_root'.hex()):
...
             consensus=b'consensus',
             state_root_hash=root)
 
         block = Block(
             header=header.SerializeToString(),
             header_signature=block_id,
-            batches=[make_mock_batch(base_id)])
+            batches=[batch])
 
-        self.put_blocks([block])
```
peterschwarz, line 126: Isn't this still required?

ltseeley: Just the way the diff shows things. It's still in the `add_block` method, not in the new `make_mock_block` method.

agunde406 (approved):

peterschwarz (approved):

PR #2352, opened 2020-10-15
Update all adm code to Rust edition 2018

isabeltomb:
Signed-off-by: Isabel Tomb <tomb@bitwise.io>

### DIFF
--- a/adm/src/blockstore.rs
+++ b/adm/src/blockstore.rs
@@ -15,12 +15,12 @@
  * ------------------------------------------------------------------------------
  */
 
-use proto::block::{Block, BlockHeader};
 use protobuf;
 use protobuf::Message;
 
-use database::error::DatabaseError;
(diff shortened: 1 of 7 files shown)


adm/src/blockstore.rs
```diff
@@ -15,12 +15,12 @@
  * ------------------------------------------------------------------------------
  */
 
-use proto::block::{Block, BlockHeader};
+use crate::proto::block::{Block, BlockHeader};
```
ltseeley, line 18: While you're at it, can you move this next to the other `use crate::...` imports?

ltseeley (approved):

adm/src/blockstore.rs
isabeltomb: Fixed

vaporos (approved):

PR #2353, opened 2020-10-16
Add 'batch' subcommand to sawadm

isabeltomb:
Add the `batch` subcommand to sawadm. The `batch` subcommand has a `list` subcommand which lists all existing batches in the current blockchain. `batch list` has optional flags for specifying the authorizer of the request, the output format of the batches, and the URL of the Validator's REST API.

### DIFF
--- a/adm/Cargo.toml
+++ b/adm/Cargo.toml
@@ -29,9 +29,11 @@ clap = ">=2.29.0"
 libc = ">=0.2.35"
 lmdb-zero = ">=0.4.1"
 protobuf = "2.0"
+sawtooth = { version = "0.6", features = ["client-rest"], optional = true }
 sawtooth-sdk = "0.4"
 serde = "1.0"
 serde_derive = "1.0"
+serde_json = { version = "1.0", optional = true }
 serde_yaml = "0.8"
 
 [build-dependencies]
@@ -58,4 +60,7 @@ experimental = [
     # The experimental feature extends stable:
     "stable",
     # The following features are experimental:
+    "client-cli",
 ]
+
+client-cli = ["sawtooth/client-rest", "serde_json"]
--- a/adm/src/commands/batch.rs
+++ b/adm/src/commands/batch.rs
@@ -1,0 +1,175 @@
+// Copyright 2020 Bitwise IO
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use clap::ArgMatches;
+use sawtooth::client::{
+    rest::RestApiSawtoothClientBuilder, Batch, SawtoothClient, SawtoothClientError as ClientError,
+    SawtoothClientError,
+};
+use serde::Serialize;
+
+use crate::err::CliError;
+
+pub fn run<'a>(args: &ArgMatches<'a>) -> Result<(), CliError> {
+    match args.subcommand() {
+        ("list", Some(args)) => run_list_command(args),
+        _ => {
+            println!("Invalid subcommand; Pass --help for usage.");
+            Ok(())
+        }
+    }
+}
+
+fn run_list_command<'a>(args: &ArgMatches<'a>) -> Result<(), CliError> {
+    let mut url = args
+        .value_of("url")
+        .unwrap_or("http://localhost:8008")
+        .to_string();
+    if !url.contains("http://") {
+        url = format!("http://{}", url);
+    }
+
+    let mut client_builder = RestApiSawtoothClientBuilder::new().with_url(&url);
+
+    if let Some(auth) = args.value_of("username") {
+        let credentials = get_credentials(auth)?;
+        client_builder = client_builder.with_basic_auth(credentials[0], credentials[1]);
+    }
+
+    let client = client_builder
+        .build()
+        .map_err(|err| CliError::EnvironmentError(format!("Failed to create client: {}", err)))?;
+
+    let batches = client.list_batches().map_err(|err| {
+        CliError::EnvironmentError(format!("Failed to retrieve batch list, {}", err))
+    })?;
+
+    let format = args.value_of("format").unwrap_or("default");
+
+    if format == "csv" {
+        let data = parse_batches_into_rows(batches)?;
+        for row in data {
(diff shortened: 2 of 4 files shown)


adm/Cargo.toml
```diff
@@ -29,9 +29,11 @@ clap = ">=2.29.0"
 libc = ">=0.2.35"
 lmdb-zero = ">=0.4.1"
 protobuf = "2.0"
+sawtooth = { version = "0.6", features = ["client-rest"] }
```
peterschwarz, line 32: This should be an optional dependency, for now, and it should be included under an experimental feature below.  Here, it can just say optional, where as when it's use, the experimental feature (possibly just called `client-cli`) can use it as such:

```
client-cli = ["sawtooth/client-rest", ...]
```

adm/src/main.rs
```diff
@@ -20,6 +20,7 @@ extern crate clap;
 extern crate libc;
 extern crate lmdb_zero;
 extern crate protobuf;
+extern crate sawtooth;
```
peterschwarz, line 23: since this is now 2018 edition, these can be omitted.

adm/src/commands/mod.rs
```diff
@@ -15,6 +15,7 @@
  * ------------------------------------------------------------------------------
  */
 
+pub mod batch;
```
peterschwarz, line 18: This should be behind the experimental feature defined in the Cargo toml, as should the uses and the clap args in main.rs.

peterschwarz (changes requested):

adm/Cargo.toml
isabeltomb: Fixed

adm/src/main.rs
isabeltomb: Fixed

adm/src/commands/mod.rs
isabeltomb: Fixed

adm/src/main.rs
```diff
@@ -17,14 +17,8 @@
 
 #[macro_use]
 extern crate clap;
-extern crate libc;
```
peterschwarz, line 20: Could you move these removals to another commit?

adm/src/main.rs
```diff
@@ -113,6 +109,17 @@ fn parse_args<'a>() -> ArgMatches<'a> {
             (@arg ignore_required_settings: --("ignore-required-settings")
              "skip the check for settings that are required at genesis (necessary if using a
               settings transaction family other than sawtooth_settings)"))
+        (@subcommand batch =>
```
peterschwarz, line 112: This should also be behind the feature

adm/src/commands/batch.rs
```diff
@@ -0,0 +1,178 @@
...
+            } else {
+                col_string += &" ".repeat(*len);
+            }
+        }
+        println!("{}", col_string);
+    }
+}
+
+#[derive(Debug, Serialize)]
+struct JsonYamlBatch {
```
peterschwarz, line 164: Maybe this could be called StructuredOutputBatch

adm/src/main.rs
isabeltomb: Fixed

adm/src/commands/batch.rs
isabeltomb: Fixed

adm/src/main.rs
isabeltomb: Moved to separate commit

adm/src/main.rs
```diff
@@ -107,5 +112,47 @@ fn parse_args<'a>() -> ArgMatches<'a> {
...
+                                "specify the user to authorize\n\
+                                request",
+                            ),
+                    )
+                    .arg(
+                        Arg::with_name("format")
+                            .long("format")
+                            .short("F")
+                            .takes_value(true)
+                            .value_name("csv,json,yaml,default")
```
ltseeley, line 152: This should be `.possible_values(...)` https://docs.rs/clap/2.33.3/clap/struct.Arg.html#method.possible_values

adm/src/main.rs
```diff
@@ -107,5 +112,47 @@ fn parse_args<'a>() -> ArgMatches<'a> {
...
+                        "identify the URL of the validator's\n\
+                            REST API (default:\n\
+                            http://localhost:8008)",
+                    ))
+                    .arg(
+                        Arg::with_name("username")
+                            .long("user")
+                            .short("u")
+                            .takes_value(true)
+                            .value_name("USERNAME[:PASSWORD]")
```
ltseeley, line 141: We should probably just add the format to the help text instead of with `value_name`.

adm/src/commands/batch.rs
```diff
@@ -0,0 +1,182 @@
...
+                CliError::EnvironmentError(format!("Failed to create client: {}", err))
+            })?;
+    } else {
+        client = RestApiSawtoothClientBuilder::new()
+            .with_url(&url)
+            .build()
+            .map_err(|err| {
+                CliError::EnvironmentError(format!("Failed to create client: {}", err))
+            })?;
+    }
```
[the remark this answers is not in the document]
isabeltomb: fixed

adm/src/commands/batch.rs
```diff
@@ -0,0 +1,182 @@
...
+            )
+        }
+    } else {
+        let data = parse_batches(batches)?;
+        print_table(data);
+    }
+    Ok(())
+}
+
+fn parse_batches(
```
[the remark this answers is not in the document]
isabeltomb: Changed function name and added doc comment

adm/src/commands/batch.rs
```diff
@@ -0,0 +1,182 @@
...
+        for batch in batches {
+            println!(
+                "{}",
+                serde_json::to_string_pretty(&StructuredOutputBatch::from(&batch.unwrap()))
+                    .map_err(|err| CliError::EnvironmentError(format!(
+                        "Cannot format batches into json: {}",
+                        err
+                    )))?
+            )
+        }
```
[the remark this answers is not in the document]
isabeltomb: Fixed

adm/src/commands/batch.rs
```diff
@@ -0,0 +1,182 @@
...
+    if format == "csv" {
+        let data = parse_batches(batches)?;
+        for row in data {
+            println!("{}", row.join(","))
+        }
+    } else if format == "json" {
+        for batch in batches {
+            println!(
+                "{}",
+                serde_json::to_string_pretty(&StructuredOutputBatch::from(&batch.unwrap()))
```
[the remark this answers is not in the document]
isabeltomb: Fixed

adm/src/main.rs
isabeltomb: Fixed

adm/src/main.rs
isabeltomb: Fixed

ltseeley (approved):

peterschwarz (approved):

PR #2379, opened 2021-06-04
Update protobuf

vaporos:
This updated the usage of protobuf so it no longer causes compiler warnings and specify the minimum version of protobuf to be 2.23.


families/block_info/sawtooth_block_info/src/state.rs
```diff
@@ -211,6 +212,18 @@ impl<'a> From<&'a protos::block_info::BlockInfo> for BlockInfo {
     }
 }
 
+impl From<protos::block_info::BlockInfo> for BlockInfo {
```
agunde406, line 215: Why does this change need to be added?

vaporos: To support the modification in the diff chunk preceding this one, which previously used a reference but does not after this change.

peterschwarz: Should this be in a separate commit?

vaporos: Not sure if that's even possible or makes sense, given it is directly required/attached to the version change of protobuf. (Also, should be non-functional change.)

peterschwarz (approved):

agunde406 (approved):

PR #2419, opened 2022-03-08
fixed outdated deployment configs

jhunter-ngt:
In deploying a Sawtooth single node with the latest minikube release, encountered issues involving outdated API version and missing deployment selector labels.


rbuysse:
Thanks for the PR! The team talked this over and it sounds like we won't be supporting the Kubernetes documentation anymore so I'm going to close this.

PR #2431, opened 2022-07-27
Fix dependencies to build on ARM

CLEckhardt:
Updates many dependencies to enable sawtooth validator to build and
run on ARM. Fixes a memory leak that occurs when using the currently
published image on ARM.

Some packages that were pulled in via apt install (ex.
python3-pyformance, python3-grpcio, etc.) aren't published for ARM, so
we need to install them via pip. To install them via pip, we need to
lock versions for some of the packages to continue to use python3.6 and
be compatible with Grid. Some packages need to be up-to-date and not
version-locked. Installing packages via pip is done in steps, as some
packages need to be finished installing before calling `pip install ...`
for others.

This also removes the dependency declarations in `setup.py` and
`stdeb.cfg` files, as these mostly refer to packages without
compatible ARM versions published. They are also silent failures in the
build process.

I suspect that using an updated version of lmdb resolved the memory
leak.

Signed-off-by: Chris Eckhardt <eckhardt@bitwise.io>

### DIFF
--- a/cli/setup.py
+++ b/cli/setup.py
@@ -33,10 +33,7 @@ setup(
     author='Hyperledger Sawtooth',
     url='https://github.com/hyperledger/sawtooth-core',
     packages=find_packages(),
-    install_requires=[
-        'colorlog', 'protobuf', 'sawtooth-sdk', 'toml', 'PyYAML',
-        'requests'
-    ],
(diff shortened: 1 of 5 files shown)


CLEckhardt:
Moving these updates to the 1-3 branch

PR #2434, opened 2022-09-08
Update sawtooth-default-poet.yaml

bmoore06:
Signed-off-by: Benjamin Moore <74261880+bmoore06@users.noreply.github.com>


bmoore06:
N/A

scealiontach (commented):
These are mostly just formatting/style changes.  I'd suggest that if we are going to update it then we should make it significantly easier to read and manage.  What we've got now is a gnarly mess of escaping and quotes.  We could improve it by changing the commands to follow a form along the lines of 

```
service-0:
  entrypoint:
    - bash
    - -ec
  command: |
    echo "Now we don't need to escape or use && as much";
    tail -f /dev/null
```

PR #2463, opened 2023-10-12
Fix PartialOrd implementation for prune candidate

suchapalaver:
In the previous code, the line `Some(Ordering::reverse(self.0.cmp(&other.0)))` was incorrect because it would return the opposite of what you intended, potentially leading to incorrect ordering of elements.

Signed-off-by: Joseph Livesey [jlivesey@gmail.com](mailto:jlivesey@gmail.com)

### DIFF
--- a/validator/src/state/state_pruning_manager.rs
+++ b/validator/src/state/state_pruning_manager.rs
@@ -49,7 +49,7 @@ impl Ord for PruneCandidate {
 
 impl PartialOrd for PruneCandidate {
     fn partial_cmp(&self, other: &PruneCandidate) -> Option<Ordering> {
-        Some(Ordering::reverse(self.0.cmp(&other.0)))
+        Some(self.cmp(other))
     }
 }
 


vaporos:
Can you elaborate in the commit message on the reasoning on why this is correct vs. the previous code?

suchapalaver:
> Can you elaborate in the commit message on the reasoning on why this is correct vs. the previous code?

@vaporos, thanks. I've updated the commit message to read: 

>`Fix 'PruneCandidate' 'PartialOrd' impl to return result of 'self.cmp(other)', not its reverse`

peterschwarz:
> > Can you elaborate in the commit message on the reasoning on why this is correct vs. the previous code?
> 
> @vaporos, thanks. I've updated the commit message to read:
> 
> > `Fix 'PruneCandidate' 'PartialOrd' impl to return result of 'self.cmp(other)', not its reverse`

I would still think of that comment as the “what” and not the “why”

suchapalaver:
> > > Can you elaborate in the commit message on the reasoning on why this is correct vs. the previous code?
> 
> > 
> 
> > @vaporos, thanks. I've updated the commit message to read:
> 
> > 
> 
> > > `Fix 'PruneCandidate' 'PartialOrd' impl to return result of 'self.cmp(other)', not its reverse`
> 
> 
> 
> I would still think of that comment as the “what” and not the “why”

Any suggestions?

suchapalaver:
This, btw， fixes the build issues on [PR 2458](https://github.com/hyperledger/sawtooth-core/pull/2458).

vaporos (approved):

